PR #1, opened 2022-11-16
Work in progress xilem prototype

raphlinus:
This is a snapshot of the current state of the xilem prototype built on new infrastructure: glazier for window creation, piet-gpu for drawing, and parley for text.

As of this snapshot, only the text widget has been ported, and there are a number of rough edges. It should be considered a proof of concept. That said, it should be possible to build more from here.


leontepe:
Text is blurry.

Screenshot 2022-11-17 at 13 47 04

Is high DPI detection enabled/implemented?

dzhou121:
I guess this question is more for piet-gpu, but how are we going to manage the text rendering on different platforms? Like you'll need sub-pixel rendering for Windows, but not necessary for Macs.

leontepe:
@dzhou121 don't know if that's useful, but we had a small convo about text rendering here: https://xi.zulipchat.com/#narrow/stream/296640-font-tech/topic/Pathfinder.20approach.20vs.2E.20defer.20to.20native

dzhou121:
> https://xi.zulipchat.com/#narrow/stream/296640-font-tech/topic/Pathfinder.20approach.20vs.2E.20defer.20to.20native

Thanks. I can see it's a discussion of the mechanisms of text rendering. My question is more user level, i.e. does text look the same or different across different OS, or Xilem will expose some API for you to choose.

leontepe:
Hm. I suppose we can always expose a config to, say, enable font dilation on Windows or sub-pixel rendering on Mac, but generally I'd say it defaults to the platform native way of rendering fonts.

Psykopear:
Trying this on Linux with Gnome/Wayland.

The `main.rs` runs, I see the button with the gradient background, but if I click on it I get no message on stdout.
I can't close the window normally (`x` button or shortcut to close window) I have to ctrl+c from the terminal.

raphlinus:
Thanks for trying it! I haven't tested on anything but mac yet. The lack of mouse sensitivity sounds like a problem, we'll need to investigate it.

I'm not sure what exactly is wrong with quitting, that appears to be a glazier issue (the code in xilem is adding the quit item to the menu).

The panic on Wayland looks similar to [this observation](https://xi.zulipchat.com/#narrow/stream/351333-glazier/topic/Running.20callbacks.20when.20building.20windows.2E/near/310707714). Again I think the best place to fix this is glazier.

Psykopear:
Right, I also get no menu, so maybe that's related, but I see no logs in the console that could help:

image

raphlinus:
I've tested the glazier-shello example on Windows, but not the integration with piet-gpu. Unfortunately, I'm at the office right now without easy access to a Windows machine. I can take another look when I get home.

raphlinus:
I do repro the "render one frame then hang" behavior with Windows, and also note that it's the same with glazier-hello. This is very likely a glazier problem and I can probably track it down.

raphlinus:
The Windows hang is fixed by https://github.com/linebender/piet-gpu/pull/207. The reason the shello example worked is that the branch predated the robust memory work. So this was not in fact a glazier problem.

jplatte:
On actual wayland (`glazier/wayland` feature), I got a panic (and transparent window):

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/jplatte/code/xilem/src/app.rs:205:47
[wayland-client error] A handler for wl_pointer panicked.
```

With the `gtk` feature, glazier didn't compile at all.

raphlinus:
No worries, it's all good. Very likely it'll be helpful for fixing and validating the panic on Wayland, which I intend to do later today.

src/render.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    pub fn pre_render(&mut self) -> Option<Vec<f64>> {
+        let frame_idx = self.frame_index();
+        if let Some(submitted) = self.submitted[frame_idx].take() {
+            self.cmd_bufs[frame_idx] = submitted.wait().unwrap();
+            Some(unsafe {
+                self.session
+                    .fetch_query_pool(&self.query_pools[frame_idx])
+                    .unwrap()
+            })
```
Psykopear, line 81: I tried adding some prints around, and it seems to hang here on X11

raphlinus: Did you update to latest piet-gpu, it sounds like https://github.com/linebender/piet-gpu/pull/207

DJMcNab: To do so, run `cargo update` before you run it. That solved the hang for me.

raphlinus:
We should now be able to handle window events before the first paint. Also I snapped Cargo.lock to newer github versions. There's lots to do still (including some work on dpi that I'm seeing), but I'm wondering if this might now be "good enough" to merge, and future work on top of it.

flxzt:
I have an eGPU ( Nvidia GTX 1080 ) I canconnect to my laptop (Intel HD graphics) so I am able to test a couple of different configurations:

- with X11 and nvidia: It runs, but panics as soon as I try to resize the window: 
   ```thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ERROR_OUT_OF_DATE_KHR', src/render.rs:118:18```
- with X11 / xwayland and integrated graphics: the window stays black and I can only quit with Ctrl+C
- with the wayland features and integrated graphics: Transparent window..

hope this helps. on windows it runs now for me!

raphlinus:
Thanks for the reports. We're going to switch to wgpu soonish, so it doesn't make sense to go too deep into compatibility issues just now, though it would make sense to fix obvious breakage.

Psykopear:
Ok, now the x11 backend works. It takes some time (~1 second) to get from window creation to the button appearing, but then both the button and closing the window work.

With the `wayland` backend, I get an invisible window and no logs in stdout. But, if I hit the shortcut to close the window, it quits correctly.

Same behaviour with intel integrated and rtx3050

dfrg (approved):
Excited to land this!

PR #5, opened 2022-11-23
Update for piet-scene changes

dfrg:
This should have the appropriate fixes to match the changes in linebender/piet-gpu#208 once that lands.

The ergonomics in these changes look significantly better to me.


raphlinus (approved):
Agree, ergonomics are much nicer. I think the failing CI is just that the piet-gpu PR hasn't been merged. Thanks!

PR #6, opened 2022-11-23
Introduce ChangeFlags and use it instead of bools to track changes in views

Zoxc:
This approach has a number of advantages.

- It makes it clear that `set_label` and `set_text` do not themselves set the flags and that these functions are the source of these flags.
- The type is marked as `#[must_use]` so flags are not accidentally missed, which I have done previously when writing views.
- It seems to make the `update` phase unnecessary.

`ChangeFlags` or `Changes` could be some alternative names. The current name could easily be confused with the `update` phase.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -244,16 +244,14 @@ where
             let state =
                 if let Some(element) = self.root_pod.as_mut().and_then(|pod| pod.downcast_mut()) {
                     let mut state = response.state.unwrap();
-                    let changed = response.view.rebuild(
+                    let changes = response.view.rebuild(
                         &mut self.cx,
                         response.prev.as_ref().unwrap(),
                         self.id.as_mut().unwrap(),
                         &mut state,
                         element,
                     );
-                    if changed {
-                        self.root_pod.as_mut().unwrap().request_update();
-                    }
+                    self.root_pod.as_mut().unwrap().mark(changes);
                     assert!(self.cx.is_empty(), "id path imbalance on rebuild");
                     state
                 } else {
--- a/src/view.rs
+++ b/src/view.rs
@@ -35,7 +35,7 @@ use futures_task::{ArcWake, Waker};
 use crate::{
     event::EventResult,
     id::{Id, IdPath},
-    widget::{UpdateFlags, Widget},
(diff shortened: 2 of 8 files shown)


raphlinus:
Thanks for this! I'm thinking it through, but while I'm doing that let me explain a bit why it is the way it is, and a potential downside.

My initial concept of the `View` trait was to be very generic and generally agnostic about the details of the Element tree implementation that sits under it. So this change feels like a (minor) layering violation, that a detail of the widget tree is now present in the View trait. However, the `View` trait has now evolved to be less generic and has an explicit `Widget` bound on `Element`, so that's not actually a layering violation.

That said, this change puts a (small) additional burden on other potential consumers of the widget tree, they have to set these flags correctly. Not a showstopper, but I think something to be considered.

The real question for me is whether we can make the `update` method go away entirely. I don't much like the current state where all it does is set a flag, it feels like boilerplate. But I haven't thought through the implications of that.

I like `ChangeFlags` as the name.

Zoxc:
I view the current state as having a 1-bit flag, with this PR extending it to n-bit flags. So you could consider the current `View` trait to assume that the `Widget` trait uses a 1-bit flag, which gets changed to n-bit flags, both being "layering violations". Code in widget consumers using a 1-bit flag should be identical to code using n-bit flags, so I'm not sure where the additional burden comes from. We're changing the width of the type, not the structure of the code itself.

> The real question for me is whether we can make the `update` method go away entirely. I don't much like the current state where all it does is set a flag, it feels like boilerplate. But I haven't thought through the implications of that.

I would follow up this PR with the removal of the `update` method, unless we find a reason for it to exist.

raphlinus (approved):
Yeah, let's go ahead and do this. We can back it out later if it turns out to be a problem. Thanks!

Regarding `update`, that decision is tracked in #7, as the `Widget` trait is going to get massively reworked anyway. It might make sense to apply it now so we can experiment with the method missing, but what lands now in `Widget` won't be staying that way for long.

raphlinus:
Ah, there are merge conflicts now. This is approved, so feel free to resolve those and merge it.

PR #8, opened 2022-11-23
Apply scale in MainState::render()

dfrg:
Small patch that applies the current window scale to the fragment when rendering. Makes the button almost visible on high-dpi Windows :) 

Still needs a fix in glazier to support this on macOS.

### DIFF
--- a/src/app_main.rs
+++ b/src/app_main.rs
@@ -15,8 +15,9 @@
 use std::any::Any;
 
 use glazier::{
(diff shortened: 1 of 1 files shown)


raphlinus (approved):
Heh, I was going to do this, you beat me to it :)

Thanks!

dfrg:
I'll take a look at that one. Incoming PR for macOS scale too.

PR #9, opened 2022-11-24
Assign view id paths by parents

Zoxc:
This changes `Id` allocation from using a global unique id per view to letting containers pick unique `Id` segments for their children. The `IdPath` when the `build` method on a view is called uniquely identifies an view. This results in slightly cleaner code. Views like `VStack` can process event dispatch with a jump table by hardcoding children ids instead of using an if/else tree.

A further optimization would be to change `IdPath` to `Vec<u8>` and let containers only use the number of bytes required, reducing `IdPath` sizes overall.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -34,8 +34,7 @@ use crate::{
 pub struct App<T, V: View<T>> {
     req_chan: tokio::sync::mpsc::Sender<AppReq>,
     response_chan: tokio::sync::mpsc::Receiver<RenderResponse<V, V::State>>,
-    return_chan: tokio::sync::mpsc::Sender<(V, V::State, HashSet<Id>)>,
-    id: Option<Id>,
+    return_chan: tokio::sync::mpsc::Sender<(V, V::State, HashSet<IdPath>)>,
     events: Vec<Event>,
     window_handle: WindowHandle,
     root_state: WidgetState,
@@ -53,14 +52,14 @@ const RENDER_DELAY: Duration = Duration::from_millis(5);
 struct AppTask<T, V: View<T>, F: FnMut(&mut T) -> V> {
     req_chan: tokio::sync::mpsc::Receiver<AppReq>,
     response_chan: tokio::sync::mpsc::Sender<RenderResponse<V, V::State>>,
-    return_chan: tokio::sync::mpsc::Receiver<(V, V::State, HashSet<Id>)>,
+    return_chan: tokio::sync::mpsc::Receiver<(V, V::State, HashSet<IdPath>)>,
 
     data: T,
     app_logic: F,
     view: Option<V>,
     state: Option<V::State>,
     idle_handle: Option<IdleHandle>,
-    pending_async: HashSet<Id>,
+    pending_async: HashSet<IdPath>,
     ui_state: UiState,
 }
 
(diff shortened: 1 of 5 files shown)


raphlinus:
I'm still wrapping my head around this one, and am not yet really seeing the advantage, but I want to fully understand it before making a decision.

One observation: a reason to make the id mutable in `rebuild` was to make sure a change of type in an `AnyView` got a fresh id, that seems to no longer be the case. Whether that seriously breaks anything I still haven't figured out.

In any case, thanks for these. I do want to explore the neighborhood of the original Xilem draft and iterate toward things that are actually improvements.

Zoxc:
`AnyView` must be changed to create a unique id for each view it builds. Just increasing a local counter is sufficient.

dfrg:
I appreciate some of the reduced boilerplate but this seems to trade a simple, common `Id` allocation strategy with one that places a burden on containers. For simple cases like fixed `VStack`s this seems nice, but those are also likely to be small and cheap to search for a specific child. In the case of larger, dynamic containers, it seems like adding or removing children would negate any advantage of linear local id allocation scheme. We'd need to resort to a hashmap or linear search anyway if we want to retain stability. But I may be missing something.

infogulch:
It seems like the use cases for ids have goals that pull in conflicting directions: you want ids to be static for the life of each component, and also support dynamic allocation under heavy component construction-destruction load. Maybe something like a generational index [^1] [^2] would be useful here? [^3] [^4]

[^1]: https://docs.rs/slotmap/1.0.6/slotmap/index.html
[^2]: https://docs.rs/bevy_ecs/0.9.0/bevy_ecs/entity/struct.Entity.html
[^3]: https://lucassardois.medium.com/generational-indices-guide-8e3c5f7fd594
[^4]: https://news.ycombinator.com/item?id=17977906

Zoxc:
@dfrg Dynamic containers could just use the existing strategy of a global `Id` allocation for its children, so I don't think an additional burden is required.

It's possible a generational index would be an optimization that dynamic containers could use for better event dispatch.

raphlinus:
I've got my head around this - thanks again for the proposal and discussion. It is a viable alternative way to organize id paths. However, I'm going to close it, as I think it's a sideways move.

Yes, the signatures of build and rebuild become smaller, and the dispatch of events by a tuple to a child is more direct, but I'm also seeing the downsides, which to me outweigh the benefits: async resolution is by the `IdPath` rather than `Id` (and I think this may be a hot path in some cases), and id paths can be twice as long in the `AnyView` case, which may be common when binding to a scripting language.

Just a suggestion, but a blog post on the design neighborhood of the Xilem architecture could make a good read - this proposal as well as the one that's generic over widget trait so can support both native and web from the same app logic.

Zoxc:
> async resolution is by the `IdPath` rather than `Id` (and I think this may be a hot path in some cases)

A major motivation is to optimize event dispatch, which is the mechanism used by async resolution. It relies on `IdPath` before and after this change, so I don't follow your logic here. The `Id` to `IdPath` change in this PR is just for the outstanding async counter, which should get a more optimized representation anyway.

> id paths can be twice as long in the `AnyView` case, which may be common when binding to a scripting language.

I'd expect `IdPath` to be shorter overall if we make it use bytes and let containers represent things more efficiently. Also note that simple view like `Button` would not be longer when stored in a `AnyView` since the `Id` is just shifted to the parent and the button itself no longer needs an `Id`. `AnyView` can be optimized to only add a single byte to the path for the initial view mounted too.

PR #10, opened 2022-11-24
Pass mutable scene builder reference to paint

raphlinus:
Instead of the `paint` method returning a `Rendered` which was just a newtype around `SceneFragment`, pass in `&mut SceneBuilder`. Also store a scene fragment in each `Pod`, which is the storage for these builders.

This change has a number of consequences. Less boilerplate in the paint method, and also less allocation, as the storage can get reused. More importantly, though we don't have the logic for this yet, it sets the stage for retaining scene fragments when widgets don't change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,9 +37,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "ash"
-version = "0.37.0+1.3.209"
+version = "0.37.1+1.3.235"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "006ca68e0f2b03f22d6fa9f2860f85aed430d257fec20f8879b2145e7c7ae1a6"
+checksum = "911015c962d56e2e4052f40182ca5462ba60a3d2ff04e827c365a0ab3d65726d"
 dependencies = [
  "libloading",
 ]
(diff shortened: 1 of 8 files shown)


xarvic:
Could we make the `scene_builder` argument a member of `PaintCx` like it was in druid?

And i am trying to understand how widgets render their children. The paint method of `Pod` has no `scene_builder` parameter and there is no method to retrieve the `SceneFragment` of the Pod. I would be in favor of adding the `scene_builder` parameter and let the `Pod` apply transform and write its `SceneFragment` into that of its parent.

raphlinus:
Is this not addressed in #10?

I haven't implemented containers yet (soon), so it's not clear from the code. `Pod` will have a method to return a scene fragment reference, and the parent will do `append()` with that fragment (also apply a transform based on layout / scrolling / etc). The motivation for this is retaining the rendered scene fragments, so you only call into `paint` when they're invalidated. The idea is that `append` is cheap - currently it is a handful of memcpy operations, but shouldn't allocate in the steady case, and if the memcpy operations do turn out to be expensive, we could go to non-contiguous storage.

raphlinus:
Hopefully that commit makes what I'm trying to do a little clearer.

xarvic:
Yes, thanks :)

From what i understand painting a child widget will look like this most of the time:
```rust
child.paint(ctx);
builder.push_layer(..., Affine::translate(child.origin().to_vec2()), ...);
builder.append(child.fragment());
builder.pop_layer();
```

What i wanted to say is that is could be done by the `Pod` of child if its paint method receives the parent `SceneBuilder`. This was also my rationale for proposing to include the `scene_builder` argument into the `PaintCx`

raphlinus:
Ah, now I understand. I was confusing the paint method of `Pod` with the paint method on the `Widget` trait.

In the general case we will not need `push_layer()`, only when doing clips and blends. Also note that the affine transform has changed from stateful (PostScript/canvas-like) to stateless, so it needs to be given to the `append` method. Thus, though we could certainly add a convenience method to `Pod` to append its fragment, I'm not sure it helps that much.

xarvic:
Ah! I over read the transform parameter in `append`. Now it makes sense. Thanks for the explanation :)

dfrg (approved):
This looks good! I spent some time with it but wanted to give others a chance to offer approval.

PR #12, opened 2022-11-26
Move to piet-wgsl for rendering

dfrg:
Removes the piet-gpu and piet-gpu-hal dependencies and updates the rendering code to use piet-wgsl.


src/app_main.rs
```diff
@@ -189,38 +201,47 @@ where
...
-            self.pgpu_state = Some(
-                crate::render::PgpuState::new(
-                    handle,
-                    handle,
-                    size.width as usize,
-                    size.height as usize,
-                )
-                .unwrap(),
-            );
+            self.surface = Some(self.render_cx.create_surface(handle, width, height));
```
DJMcNab, line 214: `create_surface` laundering the unsafety of `RawWindowHandle` isn't ideal.

Really the ecosystem needs to get their act together and find a safe way to access swapchains from wgpu.
I don't really want to block on it, because afaik the safety comment on literally every call to `instance.create_surface` is along the lines of "hope this works :)"

DJMcNab (approved):
Works on my machine, and looks reasonable.

I like that there's now no unsafe in this repository, although I don't like that this is because of `piet-wgsl` being 'unsound'-ish.

src/app_main.rs
dfrg: Yeah, I suppose the assumption is that if you're sound on both ends of `RawWindowHandle` then you're good. Otherwise, there be dragons. Still, it's a pretty nifty solution to this problem that avoids the need to bless a particular windowing library or marshal these types around yourself.

DJMcNab: I think the issue is that the current raw-window-handle doesn't guarantee that the window will be sound to create a swapchain for, whilst wgpu assumes it is.
I don't know when this breaks down, but that's the real concern. It's not just about soundness on each side, it's about the windows having the expected capabilities.

PR #14, opened 2022-11-29
Experimental composable widgets for styling

DJMcNab:
Inspired by https://tonsky.me/blog/clojure-ui/#tweak-and-reuse and [Tailwind CSS](https://tailwindcss.com/), this demonstrates how our current widget model lends itself to view level styling.

For reference, the user-facing part of this is:
```rust
fn app_logic(data: &mut f64) -> impl View<f64> {
    background(
        Color::RED,
        padding(
            *data,
            button(format!("Padding {data}px"), |data| *data += 1.),
        ),
    )
}
```

Which creates the following application.

Initial state:
image

After 37 clicks:
image

Beautiful, I know.

### DIFF
--- a/src/view/style.rs
+++ b/src/view/style.rs
@@ -1,0 +1,150 @@
+use piet_scene::Color;
+
+use crate::{
+    id::Id,
+    widget::{
+        compose_style::{background::BackgroundWidget, padding::PaddingWidget},
+        Pod,
+    },
+    View,
+};
+
+pub fn padding<V: View<T, A>, T, A>(width: f64, view: V) -> impl View<T, A>
+where
+    V::Element: 'static,
+{
+    PaddingView::new(width, view)
+}
+
+pub struct PaddingView<V> {
+    width: f64,
+    view: V,
+}
+
+impl<T, A, V: View<T, A>> View<T, A> for PaddingView<V>
+where
+    V::Element: 'static,
+{
+    type State = (Id, V::State);
+
+    type Element = PaddingWidget;
+
+    fn build(&self, cx: &mut crate::view::Cx) -> (crate::id::Id, Self::State, Self::Element) {
+        let (id, (child_id, state, element)) = cx.with_new_id(|cx| self.view.build(cx));
+        let element = PaddingWidget::new(Pod::new(element), self.width);
+        (id, (child_id, state), element)
+    }
+
+    fn rebuild(
+        &self,
+        cx: &mut crate::view::Cx,
+        prev: &Self,
+        id: &mut crate::id::Id,
+        (child_id, state): &mut Self::State,
+        element: &mut Self::Element,
+    ) -> bool {
+        let mut changed = prev.width != self.width;
+        if changed {
+            element.set_width(self.width);
+        }
+        cx.with_id(*id, |cx| {
+            let child_element = element.widget.downcast_mut().unwrap();
+            let child_changed = self
+                .view
+                .rebuild(cx, &prev.view, child_id, state, child_element);
+            if child_changed {
+                changed = true;
+                element.widget.request_update();
+            }
+        });
+        changed
+    }
+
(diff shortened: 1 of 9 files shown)


src/view/style.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+
+pub fn padding<V: View<T, A>, T, A>(width: f64, view: V) -> impl View<T, A>
+where
+    V::Element: 'static,
+{
+    PaddingView::new(width, view)
+}
+
+pub struct PaddingView<V> {
```
nilsmartel, line 19: Here's an implicit decision about naming convention, that I really like;

It would be grammatically more obvious to name this `PaddedView`.

But I think this way it's way nicer, and we should adopt this as a naming convention.

E.g. `BorderView` instead of `BorderedView` etc.

This way it's a little easier to navigate the stuff xilem has to offer, even if your english is good. But more than anything, we make the library a little more accessible to non native speakers, who struggle with the language barrier

nilsmartel:
If I am not mistaken, using piets clipping abilities, it would be possible to adopt a "BorderView" as well, rounding the borders of an element, right?

I'm just curious what this system is capable of doing :)

DJMcNab:
Borders is as an open question for me

From a conceptual viewpoint it's easy; just draw a line along the parent clipping path, or a rectangle if the parent isn't a clip.

In terms of actual implementation, I'm less sure how that fits into our painting model at the moment.

Speykious:
Hi, I'm currently looking at Xilem because I am fascinated by the project. I found this pull request and decided to experiment with it, so I made the padding view have individual sizes for all 4 sides:
image
image

(not sure if you want me to add that in, just thought it would be cool to show ^^)
Since we're talking about how we could implement a `BorderView`, @DJMcNab what makes you unsure about how it fits into the painting model? I know the project relies on piet-gpu for drawing but I'm not too familiar with its current state, or even with its API at the moment...

rosefromthedead:
I played around with this a little, and ended up implementing a border widget too. I also found that for simple wrapper widgets like these there isn't really a need for a Pod - especially for the background widget, all its implementation needs to do is paint a rectangle before forwarding all of the calls to the inner widget. I'm interested whether people think this Pod-free way of creating wrapper widgets is a good idea, since it probably eliminates a little runtime checking and allocating?

My code is [here](https://github.com/rosefromthedead/xilem/blob/2bbf69e64b0172965e9087a8cee9d8c06be7b2bd/src/widget/compose_style/padding.rs).

DJMcNab:
I wanted to avoid the Pod, but I found issues with widgets inspecting `PaintCx::size`, for example. It seemed to me that the traits are designed such that `Widget`s were intended to only exist with `Pod`s.

`PaintCx` is not externally constructible, so "spoofying" that isn't intended. And `button` reads `cx.size()`  to get the area to draw.

rosefromthedead:
Oh yeah, size is the one thing I hadn't figured out yet, and I assumed it would be fairly straightforward like the mouse events were... oops. For background at least, it feels like that can still be kept podless.

About the border thing, I think I accidentally avoided the problem you mentioned (how to know the parent clipping path) by just making the clip and the border be part of the same widget.. maybe that's the easy way out. I can't think of a reason why a user might need them to be different widgets, and worst case the path could just be specified twice.

DJMcNab:
Yeah, joining clip and border seems reasonable. One could imagine wanting to have different kinds of border (e.g. an image based border), but that could probably be its own copy-pasted widget.

DJMcNab:
So, the [Xilem Vector Graphics talk](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Xilem.20Vector.20Graphics.20talk) has clarified this for me slightly.

Broadly, I still think the styled views are compelling, but the implementation methods using a unique widget for each is not needed

PR #16, opened 2022-12-01
Uncomment AnyView and update it to use ChangeFlags

Zoxc:


raphlinus (approved):
Thanks! This looks quite straightforward.

PR #18, opened 2022-12-03
Make example an iota more expressive

nilsmartel:
I think this is of importance, because it highlights the reactive nature of xilem.


xStrom (approved):
Thanks for this improvement! 🚀

PR #19, opened 2022-12-03
Create Utility functions

nilsmartel:
This PR introduces wrappers around `AppLauncher`.
I think it's nice to have this, because there is one less struct for newcomers to worry about, when they just want to get into creating UIs.

At the same the additional wrapper-code can be expected to remain this simple.
When AppLauncher grows in functionality, people will be pointed to it this way.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -25,6 +25,7 @@ use tokio::runtime::Runtime;
 use crate::event::{AsyncWake, EventResult};
 use crate::id::IdPath;
 use crate::widget::{CxState, EventCx, LayoutCx, PaintCx, Pod, UpdateCx, WidgetState};
+use crate::AppLauncher;
 use crate::{
     event::Event,
     id::Id,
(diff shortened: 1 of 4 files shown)


src/app.rs
```diff
@@ -271,6 +272,16 @@ where
             false
         }
     }
+
+    /// Runs the app inside a window.
+    pub fn run_with_title(self, title: impl Into<String>) {
+        AppLauncher::new(self).title(title).run()
+    }
+
+    /// Runs the app inside a window.
```
derekdreery, line 281: The documentation here should mention that this is a convenience around `AppLauncher`, and point the user in the direction of that struct when they need to do something more complex.

Same for `run_with_title`.

derekdreery (commented):
I don't know if this PR will be approved in the end, but I've made some comments that I personally think would improve the PR.

src/app.rs
nilsmartel: I agree. I think this would make the code more engaging and rewarding to discover for oneself.

nilsmartel: It is addressed in the latest commit

dhyces:
While it is nice, imo it is a conflict of responsibilities

xorgy:
This has been unmergeable for about a year now, I'm going to go ahead and close it for now.

PR #21, opened 2022-12-08
Handle MouseLeft event

### LINKED ISSUE #13: Hover state retained on window exit
https://user-images.githubusercontent.com/27787342/204162203-b5d94f8d-bd49-4453-8703-690667963831.mov

I suspect this is not intended behavior? All hover state should be discarded on cursor exiting the window.

longmathemagician:
Fixes #13, but requires that the active glazier backend listens for mouse leave events. Support under x11 is added in linebender/glazier#45, remaining platforms will be added as time allows.

### DIFF
--- a/src/widget/core.rs
+++ b/src/widget/core.rs
@@ -225,11 +225,12 @@ impl Pod {
                 }
             }
             RawEvent::MouseLeft() => {
-                Pod::set_hot_state(&mut self.widget, &mut self.state, cx.cx_state, rect, None);
-                if had_active || self.state.flags.contains(PodFlags::IS_HOT) {
-                    false
-                } else {
+                let hot_changed =
+                    Pod::set_hot_state(&mut self.widget, &mut self.state, cx.cx_state, rect, None);
+                if had_active || hot_changed {
                     true
+                } else {
+                    false
                 }
             }
         };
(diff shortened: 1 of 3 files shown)


src/widget/core.rs
```diff
@@ -224,6 +224,14 @@ impl Pod {
                     false
                 }
             }
+            RawEvent::MouseLeft() => {
+                Pod::set_hot_state(&mut self.widget, &mut self.state, cx.cx_state, rect, None);
+                if had_active || self.state.flags.contains(PodFlags::IS_HOT) {
+                    false
+                } else {
+                    true
+                }
```
xarvic, line 233: [Here](https://github.com/linebender/druid/blob/ac3815114c65d46fd388431d3013a9412501916b/druid/src/core.rs#L687)  is druids implementation of the `MouseLeave` event. 
We should probably stay close to that version. Every active or previously hot widget should receive the event.

xarvic (changes requested):

xarvic:
Thanks for adding this :)

src/widget/core.rs
longmathemagician: 🤦‍♂️ yes that makes a lot more sense.

longmathemagician:
Event logic should be fixed now, thanks for catching that! Marking this as ready for review, I can't test it under macOS 13 but it looks like mouse_leave is covered under glazier's mac backend so that should fix #13 on every platform that currently functions.

xarvic (approved):
Looks good, thank you :)

Issue #7, opened 2022-11-23
Design new `Widget` trait


raphlinus:
An important step in building out the widget tree is to get the trait right. Further iteration is possible, but getting it basically right at this time will save hassle later.

The high-level approach is to start with the Druid [Widget trait](https://docs.rs/druid/latest/druid/trait.Widget.html) and make some changes.

* The `data` argument goes away (as does being generic on `T`). This is a natural consequence of the Xilem architecture.

* The `env` argument goes away, I think. This is actually a much bigger question, as we rely on env to retrieve style data, so probably deserves some thought about what will replace that.

* The `paint` method evolves to generate a piet-gpu scene fragment rather than drawing into a Piet render context.

* The trait also needs to update the accessibility tree. In the [prototype Druid integration](https://github.com/mwcampbell/druid/tree/accesskit) for AccessKit, there an [`accessibility()`](https://github.com/mwcampbell/druid/blob/89aea1c48a6f1355afea8bac3f69aa0f486d94e9/druid/src/core.rs#L1215-L1236) method that generates the node, but that's based on generating the entire tree every time rather than doing incremental update. This needs to be worked out in detail.

As discussed in [today's office hours](https://docs.google.com/document/d/1bp-I8D1t9xxtCBmQDxJz9o_stHIwp_qszs04frBi1rg/edit#), we will *not* be making significant changes to layout at this time, so those methods will basically be the same as Druid's existing Flutter-like mechanism.

## The `update` method

Right now the `update` method isn't doing much, and if #6 gets adopted, it will do even less. In current Druid, one of its other roles is to respond to `env` changes (if theme or styling changes), but that may well go away as well.

A potential role may be to update the accessibility tree, but likely that should be a separate method, as it should be lazy (only called when a screen reader is connected).

So one of the questions which needs to be decided is whether to keep the `update` method or remove it. An excellent way to help resolve it is to go through the existing use cases in Druid and see if any really motivate keeping it.

## The `lifecycle` method

We can also consider combining `lifecycle` and `event`, as the motivation for having them separate may no longer be compelling. I consider this a lower priority, as I'm sure it's fine either way, but again if we do decide to change this from Druid, it would be less work to do it now than later.

raphlinus:
[...]

PR #26, opened 2022-12-12
First cut at Widget trait

raphlinus:
This is an attempt to define the Widget trait. It compiles and runs but is a bit incomplete:

* The HotChanged propagation has not been ported.
* It is missing accessibility methods.
* No doubt some cleanup is still needed.

This is work toward #7


src/widget.rs
```diff
@@ -30,57 +30,181 @@ use std::ops::{Deref, DerefMut};
...
+    /// Update the widget's appearance in response to a change in the app's
+    /// [`Data`] or [`Env`].
+    ///
+    /// This method is called when requested by the view layer.
+    /// When the appearance of the widget needs to be updated in response to
+    /// these changes, you can call [`request_paint`] or [`request_layout`] on
+    /// the provided [`UpdateCtx`] to schedule calls to [`paint`] and [`layout`]
+    /// as required.
+    ///
+    /// This method may go around.
```
jeffparsons, line 74: "May go around"? Is this saying that the method may go away in future, like the original comment suggested, or something else?

src/widget.rs
```diff
@@ -30,57 +30,181 @@ use std::ops::{Deref, DerefMut};
...
+}
 
-    fn paint(&mut self, cx: &mut PaintCx, builder: &mut SceneBuilder);
+/// An axis in visual space.
+///
+/// Most often used by widgets to describe
+/// the direction in which they grow as their number of children increases.
+/// Has some methods for manipulating geometry with respect to the axis.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Axis {
```
jeffparsons, line 203: Are these axes directional? I.e. does `Vertical` imply an origin at the top (or bottom) of the visual space, or is that direction independent of this?

src/widget.rs
raphlinus: I meant "may go away" and typo'ed it. That said, my current thinking is to retain it, and I am also considering plumbing the AccessKit node update mechanism through the method. That will require some more careful design.

src/widget/box_constraints.rs
```diff
@@ -0,0 +1,429 @@
...
+        max.width = f64::max(dim, self.min().width);
+        BoxConstraints::new(self.min(), max)
+    }
+
+    /// Shrinks the max height to dim.
+    /// Does NOT shrink beyond min height.
+    pub fn shrink_max_height_to(&self, dim: f64) -> Self {
+        let mut max = self.max();
+        max.height = f64::max(dim, self.min().height);
+        BoxConstraints::new(self.min(), max)
```
Imberflur, line 319: Should there also be something like `f64::min(dim, max.height)` in here, to keep from increasing the max?

raphlinus: This (along with `Axis`) is copied without change from Druid. Not that there's no chance of flaws there, but it's not something I'm focusing on at the moment.

src/widget.rs
Zoxc: I was expecting the accessibility stuff to be handled more like painting with a separate `ChangeFlag`, phase and trait method. I have not yet looked at AccessKit's API however.

Using `update` as a kind of environment change event could be a good idea.

src/widget.rs
```diff
@@ -30,57 +30,184 @@ use std::ops::{Deref, DerefMut};
...
-    /// The fact that `cx` is LayoutCx is just laziness, it should have its
-    /// own cx. And the main methods on that cx should be for region-based
-    /// invalidation.
-    #[allow(unused)]
-    fn prepare_paint(&mut self, cx: &mut LayoutCx, visible: Rect) {}
+    /// Intrinsic is a *could-be* value. It's the value a widget *could* have given infinite constraints.
+    /// This does not mean the value returned by layout() would be the same.
+    ///
+    /// This method **must** return a finite value.
+    fn compute_max_intrinsic(
```
xarvic, line 187: I think we should either add a `compute_max_intrinsic` method to `Pod` or remove this one.

src/widget/core.rs
```diff
@@ -66,54 +70,49 @@ pub struct Pod {
...
+    pub(crate) id: Id,
     pub(crate) flags: PodFlags,
+    /// The origin of the child in the parent's coordinate space.
     pub(crate) origin: Point,
+    /// The origin of the parent in the window coordinate space.
+    pub(crate) parent_window_origin: Point,
     /// The minimum intrinsic size of the widget.
     pub(crate) min_size: Size,
     /// The maximum intrinsic size of the widget.
     pub(crate) max_size: Size,
```
xarvic, line 84: The values of `min_size` and `max_size` are never changed. Are they remains of the exploration of `SwiftUI` layout?

src/widget/contexts.rs
```diff
@@ -146,40 +155,51 @@ impl<'a, 'b> LayoutCx<'a, 'b> {
         }
     }
 
-    pub fn add_event(&mut self, event: Event) {
+    pub fn add_event(&mut self, event: Message) {
```
xarvic, line 158: We should probably rename this and similar methods.

xarvic (approved):
There are some things we could change but i think it works the way it is now. Therefore i am in favor of merging it as soon as possible and create followup PRs with small changes instead of making this PR perfect.

src/widget.rs
raphlinus: I agree, and I also think that should happen in a followup PR. The scope of this one is focused on the `Widget` trait itself, and it's known that there's a bunch more wiring, especially in `Pod`, that needs to be done.

src/widget/core.rs
raphlinus: I will remove them, thanks for the heads-up.

src/widget/contexts.rs
raphlinus: Will do. However, looking at this now, I'm concerned I've introduced a layering violation; `Message` is a Xilem type, and that might make it harder to integrate other reactive layers. I'll file an issue on this.

raphlinus:
Thanks for the review! Apparently the way we've got the repo set up, it needs another approval after a new commit, which honestly feels a bit restrictive. I'll self-approve and merge if there's no activity soon.

DJMcNab (approved):
I've not reviewed; signing off since @xarvic 's review isn't sufficient for GitHub's approval

derekdreery:
I had a look at the settings, and it looks like you can push new commits and the approval will remain. Has someone changed the settings recently, perhaps?

DJMcNab:
We worked it out [on Zulip](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Widget.20trait)

PR #27, opened 2022-12-12
Update for the piet-wgsl -> vello name change

dfrg:
Removes the `piet-scene` dependency and changes `piet-wgsl` to `vello`. Updates all imports for the new crate name. This leaves the git reference as linebender/piet-gpu for now. We can change that to vello in this PR (after the repo is renamed) or in a subsequent one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1595,30 +1595,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


raphlinus (approved):
Looks good, thanks, and I verified that it works (felt worth it to check given how much stuff is in flight).

PR #33, opened 2023-01-09
Styling experiment with `ButtonStyles` struct

matthunz:
This is my idea for how styling can be done in the future. Buttons seem like a good start because they have different internal states that I think are best represented as 3 different stylings.

I really like the idea for strongly-typed styling instead of a `HashMap` or other alternatives. Each component has a specific set of styles it can use and I think it's best to create structs for each.
```rust
pub struct ButtonStyleState {
    border_radius: f64,
    stroke: Option<Stroke>,
    background_color: Option<Color>,
}

pub struct ButtonStyle {
    default: ButtonStyleState,
    hot: ButtonStyleState,
    active: ButtonStyleState,
}
```

### DIFF
--- a/src/view/button.rs
+++ b/src/view/button.rs
@@ -14,7 +14,7 @@
 
 use std::any::Any;
 
-use crate::{event::EventResult, id::Id, widget::ChangeFlags};
+use crate::{event::EventResult, id::Id, widget::{ChangeFlags, button::ButtonStyle}};
 
 use super::{Cx, View};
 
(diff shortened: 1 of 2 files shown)


jaredoconnell:
I do typically like statically typing everything, but I'm not sure if this is the right way to do it.
My concern with this method is how tightly coupled the styling is with the widgets. The consequences of the tight coupling include more boilerplate code when building the widget tree, and it being harder to switch themes at runtime. In addition, things that share similar colors and strokes would need to have those attributes propagated to them somehow.

I think this project would benefit from some level of separation of concerns between behavior and style. If there are states defined by the button, it's possible, depending on how it's designed, to have another section of code take care of the style.

PR #36, opened 2023-01-14
Enable AccessKit and update dependencies

raphlinus:
Turn on the AccessKit dependency and also update to more recent glazier and Vello versions.

Also follow new policy of pinning git dependencies to avoid future breakage.


raphlinus:
Yup, quite a few changes are similar. I don't have strong feelings about the order, if we want to land this I can rebase #26 (which has #36 in its commit change). This one pins to a more recent version which has some changes to gradient encoding, so we definitely want that change but on the other hand it isn't urgent.

DJMcNab (approved):
These changes look good; it compiles and runs for me locally.

Minimising panics, and neither of the menu items work properly. But I'm fairly sure those are both pre-existing

PR #38, opened 2023-01-17
Update for vello changes and pin dependencies

dfrg:
Updates the code for recent changes in vello including gradient API changes and the new strategy for surface creation on dual GPU devices.

Also pins all dependencies to current revs and deletes Cargo.lock.


DJMcNab:
It's not clear to me the relationship this PR has with #36. Some of the dependency updates appear to tread common ground.

dfrg:
Oops, didn’t realize that existed. My changes on top of that one are small, so I’ll just add them to that branch to avoid conflicts.

PR #45, opened 2023-02-28
Update parley

dfrg:
Parley now pulls in @jneem's version of fount directly so the patch is no longer needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ rand = "0.7.3"
 roxmltree = "0.13"
 swash = "0.1.4"
 bytemuck = { version = "1.7.2", features = ["derive"] }
-parley = { git = "https://github.com/dfrg/parley", rev = "e4276b4d1001a050c07121f95ca255c771e9a641" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,7 +21,7 @@ rand = "0.7.3"
 roxmltree = "0.13"
 swash = "0.1.4"
 bytemuck = { version = "1.7.2", features = ["derive"] }
-parley = { git = "https://github.com/dfrg/parley", rev = "e4276b4d1001a050c07121f95ca255c771e9a641" }
+parley = { git = "https://github.com/dfrg/parley", rev = "a885c3c" }
```
DJMcNab, line 24: I think I'd prefer us to use the full commit sha. I don't think there's any risk of collision, so I'm happy to approve anyway

DJMcNab (approved):
I'm not sure how this all fits together, but if we can minimise disruption that's fine.

Cargo.toml
dfrg: Happy to use the full sha. Parley will see further updates soon so I’ll make sure to do this on the next round.

PR #49, opened 2023-03-07
Dynamic ViewSequence

xarvic:
This PR adds a `ViewSequence` implementation for `View`, `Option<ViewSequence>`, tupels of ViewSequences and `List`.

To create no conflicts whith the orphan rules, this PR adds the `ViewMarker` trait. It lets the author decide, whether they need to implement `ViewSequence` for their type.
Currently it is not possible to implement a wrapper types like `Adapt` for `ViewSequence`. This would also be possible by adding a second helper trait. Like this:
```rust

trait ViewMarker {}

trait ViewImpl<T, A> {}

trait View<T, A>: ViewImpl<T, A> { ... }

impl<T, A, V> ViewImpl<T, A> for V where V: View<T, A> {}

impl<T, A, V> ViewImpl<T, A> for V where V: ViewSequence<T, A> {}

impl<T, A, V> ViewSequence<T, A> for V where V: View<T, A> + ViewMarker { ... }
```
This way it is not possible to impl View for a Type without it also having a ViewSequence implementation, while having the flexibility to provide both.

But i am not sure if this is the right way forward. Considering that we probably want to be generic over the element type, we could also define View as:
```rust 
trait View<T, A>: ViewSequence<T, A> {}
```
and have a seperate `TypedView` trait which looks like the current `View` trait. 
This way View only asserts that this sequence has only one element.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -356,7 +356,7 @@ where
                         .downcast_mut()
                         .expect("the root widget changed its type, this should never happen!"),
                 );
-                self.root_pod.as_mut().unwrap().mark(changes);
+                let _ = self.root_pod.as_mut().unwrap().mark(changes);
                 assert!(self.cx.is_empty(), "id path imbalance on rebuild");
                 state
             } else {
--- a/src/app.rs
+++ b/src/app.rs
@@ -408,13 +408,18 @@ impl<T, V: View<T>, F: FnMut(&mut T) -> V> AppTask<T, V, F> {
                         }
                     }
                     AppReq::Wake(id_path) => {
-                        let result = self.view.as_ref().unwrap().message(
-                            &id_path[1..],
-                            self.state.as_mut().unwrap(),
-                            Box::new(AsyncWake),
-                            &mut self.data,
-                        );
-                        if matches!(result, MessageResult::RequestRebuild) {
+                        let needs_rebuild;
+                        {
+                            let result = self.view.as_ref().unwrap().message(
+                                &id_path[1..],
+                                self.state.as_mut().unwrap(),
+                                Box::new(AsyncWake),
+                                &mut self.data,
+                            );
+                            needs_rebuild = matches!(result, MessageResult::RequestRebuild);
+                        }
+
+                        if needs_rebuild {
                             // request re-render from UI thread
                             if self.ui_state == UiState::Start {
                                 if let Some(handle) = self.idle_handle.as_mut() {
--- a/src/event.rs
+++ b/src/event.rs
@@ -45,7 +45,9 @@ impl<A> MessageResult<A> {
     pub fn map<B>(self, f: impl FnOnce(A) -> B) -> MessageResult<B> {
         match self {
             MessageResult::Action(a) => MessageResult::Action(f(a)),
-            _ => self
+            MessageResult::RequestRebuild => MessageResult::RequestRebuild,
+            MessageResult::Stale(event) => MessageResult::Stale(event),
+            MessageResult::Nop => MessageResult::Nop,
         }
     }
 
--- a/src/event.rs
+++ b/src/event.rs
@@ -35,7 +35,7 @@ pub enum MessageResult<A> {
     ///
     /// This is a normal outcome for async operation when the tree is changing
     /// dynamically, but otherwise indicates a logic error.
-    Stale,
+    Stale(Box<dyn Any>),
 }
 
 pub struct AsyncWake;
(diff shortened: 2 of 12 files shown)


raphlinus:
I want to get our container trait story worked out soon, and I have some idea what this PR is about, but I'm not yet sold. A very useful piece for me would be explaining exactly what problem this is solving and why that's worth the extra messiness of the `ViewMarker`.

To me, the most important next step on containers is `impl ViewSequence for Vec<V: View>`. I've been prototyping that a bit (actually in xilemsvg but these traits are pretty much the same), and my take so far is that actually implementing the trait is pretty straightforward, but making the container view have the correct behavior in all the cases (insert, delete, change in place, reorder) will take care.

Regarding `Option<V: View>`, I see a lot of value in making that impl View, where it would make the content conditional in a fairly straightforward way (much as [buildIf](https://developer.apple.com/documentation/swiftui/viewbuilder/buildif(_:)) in SwiftUI), but less in making it a ViewSequence. In fact, I can see some problems with the latter. If we allow ViewSequence to nest (as very much seems to be the case for this PR), then `(A, Option<B>, C)` then it's easy to see how C might lose its stable identity as the vec of widgets changes length between 2 and 3. That seems very undesirable to me.

xarvic:
> Regarding Option<V: View>, I see a lot of value in making that impl View, where it would make the content conditional in a fairly straightforward way (much as [buildIf](https://developer.apple.com/documentation/swiftui/viewbuilder/buildif(_:)) in SwiftUI), but less in making it a ViewSequence. 

I think the main distinction between `View` and `ViewSequence` is that `View` has a specific layout, while in the case of `ViewSequence` the view wrapping it has to take care of that.
For an example if we have a `Cells` layout, which creates a cell for every view, changing a view from `Some` to `None` does tell the `Cells` view to hide the specific cell. This is only possible when `Option` is a `ViewSequence`.

> In fact, I can see some problems with the latter. If we allow ViewSequence to nest (as very much seems to be the case for this PR), then (A, Option\<B>, C) then it's easy to see how C might lose its stable identity as the vec of widgets changes length between 2 and 3. That seems very undesirable to me.

I assume stable identity means in this case that a view is rebuild with the same widget and state even if a view prior in the sequence is added or removed. 
With the current implementation of `Option` and tupels this is the case. In your example we have a sequence 
`Seq` = `(A, Some(B), C)`, the view_state `State` = `(state_A, Some(state_B), state_C)` and the widgets: `Widgets` = `[widget_A, widget_B, widget_C]`. Calling `(A_new, None, C_new).rebuild(Seq, State, Widget)` will first call `A_new.rebuild(A, state_A, widget_A)` then delete `widget_B` in the implementation of `B::rebuild`, call `C_new.rebuild(C, state_C, widget_C)` and finally return the state: `(state_A, None, state_C)` and the widgets `[widget_A, widget_C]`

> I want to get our container trait story worked out soon, and I have some idea what this PR is about, but I'm not yet sold. A very useful piece for me would be explaining exactly what problem this is solving and why that's worth the extra messiness of the ViewMarker.

I agree, that the current design of ViewSequence is messy. My solution for this would be treating `View`s like lists with only one element. Like this: `trait View<T, A>: ViewSequence<T, A> {}`. This would remove the need for the `ViewMarker` trait. The current `View` trait would become `TypedView`.

> To me, the most important next step on containers is impl ViewSequence for Vec<V: View>. I've been prototyping that a bit (actually in xilemsvg but these traits are pretty much the same), and my take so far is that actually implementing the trait is pretty straightforward, but making the container view have the correct behavior in all the cases (insert, delete, change in place, reorder) will take care.

Interesting. I haven't heard about it. Is xilemsvg public and how does the `ViewSequence` implementation of `Vec<V: View>` know the stable identity of the views?

dhardy:
> stable identity

I would assume this to mean that if a widget (such as `widget_B` above) is removed and later re-added, it will have the same `Id`, and further that the `Id` of `widget_C` does not depend on whether `widget_B` is built. I updated #51 with some notes on stable identity which could also apply here.

xarvic:
Thanks for the clarification! With the current design the id of `C` does not change after adding or removing `B`. We could also preserve the `Id` and even `state_B` and `widget_B` and the last `B` in `<Option as ViewSequence>::State` druing the time that `Option` is `None`. I think this should be enough to guarantee stability.

Though for long lists this is probably not an option. Therefore we probably don't get around assigning new Ids when reloading the content.

raphlinus:
This needs a deeper answer, but a few quick clarifications. I do *not* mean that the state of B is preserved when Option<B> toggles between None and Some states. SwiftUI does not do that.

I was referring to the identity of C, but now see your point. My thinking was confused and I need to revisit this.

xilemsvg is not public (I haven't had time to push it). A Vec of views does not preserve their stable identity when their position in the vec is changed.

raphlinus (approved):
I'll go ahead and approve this, as I'm on board with having a trait that produces zero or more widgets, and my concerns about identity have been addressed. I'm interested to see whether the ergonomics around `ViewMarker` can be approved. I will also want to follow up with either an issue or a PR on how the container can properly send WidgetAdded events and otherwise maintain lifecycle invariants when the structure changes, which was possible before (with AnyView) but is even easier to trigger now.

Thanks!

PR #52, opened 2023-03-18
Fix accessibility propagation

raphlinus:
This patch makes a number of changes to change flags propagation, which should at least get the example working again.

This implements the two-flags approach to REQUEST_ACCESSIBILITY and DESCENDANT_REQUESTED_ACESSIBILITY. The framework converts the former flag into the latter as it propagates up the tree, both in `merge_up` and in ChangeFlags propagation.

ChangeFlags propagation only propagates ChangeFlags, it does not move flags from the pod into the rebuild flow. It does have descendant logic as mentioned above.

Layout now triggers an accessibility request. This reduces the burden on custom widgets to explicity request accessibility, and also will be more precise if and when layout becomes more incremental, we'll want to request accessibility only when the layout has actually changed.


src/widget/core.rs
```diff
@@ -130,6 +132,9 @@ impl WidgetState {
 
     fn merge_up(&mut self, child_state: &mut WidgetState) {
         self.flags |= child_state.upwards_flags();
+        if child_state.flags.contains(PodFlags::REQUEST_ACCESSIBILITY) {
+            self.flags |= PodFlags::DESCENDANT_REQUESTED_ACCESSIBILITY;
+        }
```
xarvic, line 137: I think we should do this in `WidgetState::upwards_flags()` to avoid duplication.

src/widget/core.rs
```diff
@@ -172,7 +177,12 @@ impl Pod {
     pub fn mark(&mut self, flags: ChangeFlags) -> ChangeFlags {
         self.state
             .request(PodFlags::from_bits_truncate(flags.bits as _));
-        ChangeFlags::from_bits_truncate(self.state.upwards_flags().bits as _)
+        let mut flags_up =
+            flags & ChangeFlags::from_bits_truncate(PodFlags::UPWARD_FLAGS.bits as _);
+        if flags.contains(ChangeFlags::ACCESSIBILITY) {
+            flags_up |= ChangeFlags::DESCENDANT_REQUESTED_ACCESSIBILITY;
+        }
+        flags_up
```
xarvic, line 185: Same as `merge_up`

raphlinus: The first one I agree with, but this one doesn't have quite the same logic. The flags propagated up the `rebuild` traversal aren't the ones from the `Pod` - I think that was actually a bug in the previous implementation that this PR fixes. And `ChangeFlags` is a different type than `PodFlags`. That said, I could look into making methods to improve DRY.

raphlinus:
Hopefully the commit comment in the followup commit (c4a3732c429c1e8e5dbafe46e66df2a1a71fab85) explains what my concern is here. The next PR I'd like to do (or would be fine if somebody else does if the direction is clear) is to add a new `ChangeFlags::STRUCTURE` which is set whenever parent/child relationships change for any reason: `AnyView` changes type, `Option` toggles between None and Some, vector changes length. Now that I think about it, I *think* there should be a corresponding `PodFlags::STRUCTURE_CHANGED` (with the same bit set), in which case my warning comment should be removed.

I haven't worked out the exact propagation rules; I think they may be a bit subtle. They probably need two flags (one for "this view/widget", one for descendants) like accessibility. On the widget side, when containers receive this flag, they diff their old list of children against the new one, send `WidgetAdded` lifecycle events to the new ones (I think this happens in `update` but this is probably not the only valid choice), and handle deletion as well. I am leaning toward having the framework track the parent/child tree, which would also replace the Bloom filters.

I'm willing to do that as a PR, but am also open to discussion; it's probably most important right now that we get some kind of understanding and agreement about where all this is going.

xarvic:
> The next PR I'd like to do (or would be fine if somebody else does if the direction is clear) is to add a new ChangeFlags::STRUCTURE which is set whenever parent/child relationships change for any reason: AnyView changes type, Option toggles between None and Some, vector changes length.

What would be the difference between `STRUCTURE_CHANGED` and `TREE_CHANGED`?

> Now that I think about it, I think there should be a corresponding PodFlags::STRUCTURE_CHANGED (with the same bit set), in which case my warning comment should be removed.

I guess that `ChangeFlags` will stay a superset of `PodFlags`. But that's just a feeling.

> ChangeFlags propagation only propagates ChangeFlags, it does not move flags from the pod into the rebuild flow. It does have descendant logic as mentioned above.

In druid we relied on `merge_up` being idempotent. I thought that this would also apply here?

xarvic (approved):
Anyways, i think we should merge this as it is.

raphlinus:
Thanks! Somehow I missed the existence of TREE, it is the same idea. We can keep the name. I'm looking over the propagation and see some problems, but given my history of being confused by this stuff probably the best thing is for me to open an issue and discussion can happen there.

PR #55, opened 2023-03-25
Prototype of generic view trait

raphlinus:
Use type system hackery to represent a generic bound over traits.

Note: this patch *just* makes the view itself generic, and doesn't address any of the supporting machinery (ViewSequence, AnyView, etc).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use xilem::view::{button, h_stack, v_stack};
+use xilem::view::{button, h_stack, v_stack, AnyView, WidgetBound};
 use xilem::{view::View, App, AppLauncher};
 
 fn app_logic(data: &mut i32) -> impl View<i32> {
@@ -8,13 +8,14 @@ fn app_logic(data: &mut i32) -> impl View<i32> {
     } else {
         format!("clicked {data} times")
     };
+    let widget: Box<dyn AnyView<i32, WidgetBound> + Send> = Box::new(button(label, |data| {
+        println!("clicked");
+        *data += 1;
+    }));
 
     // The actual UI Code starts here
     v_stack((
-        button(label, |data| {
-            println!("clicked");
-            *data += 1;
-        }),
+        widget,
         h_stack((
             button("decrease", |data| {
                 println!("clicked decrease");
--- a/src/view/sequence.rs
+++ b/src/view/sequence.rs
@@ -1,7 +1,7 @@
 use crate::event::MessageResult;
 use crate::id::Id;
 use crate::view::{Cx, TraitBound, ViewMarker};
-use crate::widget::{ChangeFlags, Pod, Widget};
+use crate::widget::{ChangeFlags, Pod};
 use crate::VecSplice;
 use std::any::Any;
 
@@ -77,13 +77,13 @@ pub trait ViewSequence<T, W: ?Sized + TraitPod, A = ()>: Send {
 // ViewMarker is already a dependency of View but Rusts orphan rules dont work if we remove it here.
 impl<T, A, V: GenericView<T, W, A> + ViewMarker, W: ?Sized + TraitPod> ViewSequence<T, W, A> for V
 where
-    V::Element: Widget + 'static,
+    V::Element: TraitBound<W> + 'static,
 {
     type State = (V::State, Id);
 
     fn build(&self, cx: &mut Cx, elements: &mut Vec<W::Pod>) -> Self::State {
-        let (id, state, element) = <V as GenericView<T, W, A>>::build(self, cx);
-        elements.push(<W as TraitPod>::make_pod(element.boxed()));
+        let (id, state, element) = V::build(self, cx);
+        elements.push(TraitPod::make_pod(element.boxed()));
         (state, id)
     }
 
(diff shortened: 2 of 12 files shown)


src/view/sequence.rs
```diff
@@ -1,10 +1,32 @@
...
+pub trait Element {
+    fn downcast_mut<T: 'static>(&mut self) -> Option<&mut T>;
+
+    fn mark(&mut self, flags: ChangeFlags) -> ChangeFlags;
+}
+
+impl TraitPod for WidgetBound {
+    type Pod = crate::widget::Pod;
+
+    fn make_pod(w: Box<Self>) -> Self::Pod {
```
xarvic, line 25: As i said on Zulip, i would like to avoid the explicit mention of `Box` in in `TraitPod `. Xilem#53 did'nt need it so maybe we can avoid it here too? 
But this is probably something we could change later without touching much code, as it only affects the element types.

xarvic (approved):
With the `TraitPod` trait and the updated version of `ViewSequence` i am in favor of this PR.
I added a comment, but i think we should not block on this.

src/view/view.rs
```diff
@@ -41,14 +59,14 @@ use crate::{
...
+// We depend on ViewMarker to disallow any View implementations which don't implement ViewMarker.
 // Doing that would lead to pretty strange bugs.
-pub trait View<T, A = ()>: ViewMarker + Send {
+pub trait GenericView<T, W: ?Sized, A = ()>: ViewMarker + Send {
     /// Associated state for the view.
     type State: Send;
 
     /// The associated widget for the view.
-    type Element: Widget;
+    type Element: TraitBound<W>;
```
nicoburns, line 69: I'd like to see some better documentation on the generic parameters in this trait (this PR seems to introduce a new one without explaining it, and `A` is never mentioned by name)

src/view/view.rs
```diff
@@ -78,10 +96,14 @@ pub trait View<T, A = ()>: ViewMarker + Send {
     ) -> MessageResult<A>;
 }
 
+pub trait View<T, A = ()>: GenericView<T, WidgetBound, A> {}
+
+impl<T, A, V: GenericView<T, WidgetBound, A>> View<T, A> for V {}
+
```
nicoburns, line 102: Is the idea here that `View` is vello-renderer (=xilem widgets) specific, and `GenericView` could apply to other renderers such as one targetting web, svg, etc? (and that we might therefore get an `SvgView` trait similar to `View` as part of such a backend?)

If so, I'd be tempted to call these `View` and `NativeView` rather than `GenericView` and `View`. But that's probably not important (and could always be changed later). Perhaps just a little note on `View` explaining the difference between it and `GenericView` would be good for now?

src/main.rs
```diff
@@ -8,13 +8,14 @@ fn app_logic(data: &mut i32) -> impl View<i32> {
...
+        *data += 1;
+    }));
 
     // The actual UI Code starts here
     v_stack((
-        button(label, |data| {
-            println!("clicked");
-            *data += 1;
-        }),
+        widget,
```
nicoburns, line 18: This seems like a bit of a regression in ergonomics:

- Will end users be expected to write out types like this?
- Will children always be required to be boxed like this?

src/widget/core.rs
```diff
@@ -560,3 +562,17 @@ impl Pod {
         self.state.flags.contains(PodFlags::REQUEST_LAYOUT)
     }
 }
+
+impl Element for Pod {
```
nicoburns, line 566: Gah: Pods, Elements, Widgets, Views. I'm still struggling to work out / keep track of what each of these is and how they relate to each other. Some kind of glossary / explainer of the key nouns / concepts in Xilem would be incredibly helpful.

nicoburns (commented):
I feel like I've not really been able to provide a substantive review on the key trait changes here, because I feel like I still don't really understand how it all fits together. But I suspect I'm not the only one in that position, so I'm hoping that my questions and requests for documentation will be helpful in their own right.

It's probably a task for after this PR is merged, but I'd *really* like to see a slightly more involved example app built out. Difficult pieces (e.g. text editing) could be avoided, but something with multiple pieces of state and nested views (what do I do once I no longer want to have my entire app in one function? how do I build my own view/widget?)

src/widget/core.rs
jaredoconnell: I agree that extra documentation for this would be beneficial.
Xilem's architecture won't be immediately intuitive to all first learning it, so the more documentation, the better.

jaredoconnell (commented):
While I don't have the insights necessary to foresee all pros and cons of this vs #53, this does look promising.
I don't like that it adds more boilerplate, but the fact that it could lead to entirely different backend implementations could make Xilem very versatile in the future.

I agree with Nico that some extra documentation would be helpful. There are a lot of common terms that have specific meanings in Xilem. Documentation is very helpful in creating a maintainable project.

raphlinus:
I am closing this because I think a macro based approach (#61) is the way forward. In particular, a showstopper is that the trait cannot be made generic over both `Send` and `!Send`, and that is required to support both efficient (multithreaded) native and Web cases.

PR #57, opened 2023-04-12
Draft for async loading

xarvic:
This is a sketch for async loading of view sequences. 

`ViewSequence` currently acts as the interface between views which do layout and strategies for loading views. This PR tries to add a way for layout views like `LinearLayout` to request the parts of a `ViewSequence`which would be visible while keeping the the identity for a specific view which should stay in focus.

It adds a `set_focus` method to the `ViewSequence` trait. This method can be used by sequences with a variable length to load only the elements which would be visible. 
The rebuild method is changed to return a `SequencePosition` this marks the index of the element marked in `set_focus` in the rebuild sequence. Using this the container could call a method like druids `scroll_area_to_view()` to keep the the element in view.

This is a idea i had for some time, which i would like to discuss in the office hours. The implementation for the backend is missing and the design for `SequencePosition` also needs some work.

### DIFF
--- a/src/view/linear_layout.rs
+++ b/src/view/linear_layout.rs
@@ -17,7 +17,7 @@ use std::{any::Any, marker::PhantomData};
 use crate::event::MessageResult;
 use crate::geometry::Axis;
 use crate::id::Id;
-use crate::view::sequence::ViewSequence;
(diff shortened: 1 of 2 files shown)


PoignardAzur:
This is superseded by newer xilem_core elements.

PR #58, opened 2023-04-12
Update linux deps

jneem:
As discussed on the glazier PR, this removes the `x86_64` suffix for the Fedora packages. It also adds some dependencies that I think were missed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,9 @@ Building and running Xilem on Linux and BSD  requires `pkg-config`, `clang`,
 and the development packages of `libxkbcommon`, `libxcb`, and `vulkan-loader`,
 to be installed.
 
-On Fedora this can be achieved using:
-
-To install these packages on Fedora, run
+Most distributions have `pkg-config` installed by default. To install the remaining packages on Fedora, run
 ```
-sudo dnf install clang libxkbcommon-x11-devel libxcb-devel pkg-config vulkan-loader-devel
+sudo dnf install clang libxkbcommon-x11-devel libxcb-devel vulkan-loader-devel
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,19 +76,19 @@ The type erasure of View nodes is not an easy trick, as the trait has two associ
...
-
-Debian:
-
-```sh
-sudo apt-get install libxkbcommon-x11-dev
+sudo dnf install clang libxkbcommon-x11-devel libxcb-devel pkg-config vulkan-loader-devel
+```
+To install them on Debian or Ubuntu, run
+```
+sudo apt-get install clang libxkbcommon-x11-dev pkg-config libvulkan-dev
```
longmathemagician, line 91: I ran this again on a fresh Fedora 37 install and pkg-config was already there, other than that (sorry!) we're good to go for F37 and Ubuntu 22.10.

longmathemagician (approved):

README.md
```diff
@@ -76,19 +76,19 @@ The type erasure of View nodes is not an easy trick, as the trait has two associ
...
+and the development packages of `libxkbcommon`, `libxcb`, and `vulkan-loader`,
+to be installed.
 
-On Fedora this can be installed using:
+On Fedora this can be achieved using:
 
-```sh
-sudo dnf install libxkbcommon-x11-devel.x86_64
-# i686 is a less common architecure, but if you have it, use libxkbcommon-x11-devel.i686 instead
+To install these packages on Fedora, run
```
DJMcNab, line 85: I don't think we need both of +83 and +85?

PR #60, opened 2023-04-24
Factor out xilem_core

raphlinus:
This was a reasonably straightforward refactor, but there are a few rough edges.

The blanket implementation of `TraitBound` failed due to coherence (E0210). There may be a way to bring it back. In the meantime, it seems we can implement it for each concrete type using a macro.

The `ChangeFlags` type is really more a concern of the widget layer than the generic view layer. This PR just brings it into the view, but arguably a better approach would be to make it an associated type. Note that the `upwards()` method is not needed inside xilem_core, only by clients.

There are now two separate `Id` concepts, one for views, one for widgets. Only the latter has AccessKit conversion methods. Having two separate types may in fact be cleaner.


raphlinus:
Closing for the same reason mentioned in #55.

PR #61, opened 2023-04-25
Macro based implementation of xilem_core

raphlinus:
Experimenting with macros to make the View traits more generic.


raphlinus:
I marked this as ready for review because I'm now confident it's the direction we should go. It has one flaw: the new Vec implementation of ViewSequence doesn't handle cases where the number of elements changes. I expect that to be fairly straightforward to fix; I also think it's possible to review without blocking on that.

dfrg (approved):
I can’t claim to completely understand the direction here in full detail but I like the idea of xilem core and the code LGTM.

PR #62, opened 2023-04-26
Add xilemsvg example

raphlinus:
This is a proof of concept that uses the common xilem_core to generate vector graphics content rendered as SVG DOM nodes in the browser.

This could plausibly be made into an actual library, and it's likely that an evolution of this will be used to create interactive illustrations, for example in the raphlinus blog.


crates/xilemsvg/dist/index.html
DJMcNab, line 1: This dist folder should be `gitignore`d, I think

nicoburns: And basic instructions for building/running should be included in it's place (FYI, with the current code you must run a web server in the dist directory for the links to work properly).

raphlinus:
Marking this as ready for review even though it contains a number of rough edges, to maintain forward momentum.

derekdreery:
Ahh sorry I merged the xilem_html PR first. You'll be listed as a co-author. :P

xorgy:
‘Let's just close it’ — @raphlinus

PR #63, opened 2023-05-09
Port FlexLayout from Druid

RagibHasin:


RagibHasin:
Events are not threading through `Sizeable` container yet.

RagibHasin:
> Events are not threading through `Sizeable` container yet.

It was not `Sizeable`. `FlexLayout` was actually at fault.

nicoburns:
This is outdated now (against the old version of Xilem). And the new masonry based version already has a Flex widget.

PR #64, opened 2023-05-10
Update vello to latest

dfrg:
Updates dependencies so we're on the latest version of vello.


DJMcNab (approved):
Seems reasonable to me, have not ran it though

dfrg:
Thanks for taking a look. This will hopefully address the misaligned read issues for good.

PR #68, opened 2023-06-07
fix: Update swash

### LINKED ISSUE #67: The example causes a panic in swash
I tried `cargo run` on Linux (NixOS) and this leads to this panic:

```sh
thread 'main' panicked at 'misaligned pointer dereference: address must be a multiple of 0x2 but is 0x556ae00cc02b', /home/mk/.cargo/git/checkouts/swash-cee884c7ebde3542/af0c735/src/internal/parse.rs:452:13
```

Commit: 4b403d3f53b94fde1dd99712af652b2dec0e43b9

flosse:
Fixes #67

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2347,8 +2347,8 @@ checksum = "8fb1df15f412ee2e9dfc1c504260fa695c1c3f10fe9f4a6ee2d2184d7d6450e2"
 
 [[package]]
 name = "swash"
-version = "0.1.6"
-source = "git+https://github.com/dfrg/swash#af0c735d64ea3ebb7dbeb770f044e4485329037e"
+version = "0.1.8"
+source = "git+https://github.com/dfrg/swash#56ea551ac49e51580c466f03d6c75252a8c8e700"
 dependencies = [
  "yazi",
  "zeno",
@@ -3226,9 +3226,9 @@ checksum = "4d25c75bf9ea12c4040a97f829154768bbbce366287e2dc044af160cd79a13fd"
 
 [[package]]
 name = "yazi"
-version = "0.1.4"
+version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c03b3e19c937b5b9bd8e52b1c88f30cce5c0d33d676cf174866175bb794ff658"
+checksum = "c94451ac9513335b5e23d7a8a2b61a7102398b8cca5160829d313e84c9d98be1"
 
 [[package]]
 name = "zbus"


DJMcNab:
Thanks for the PR!

Should we actually have a `Cargo.lock` in this repository at all?
I wasn't aware we had it not ignored.

Also, this change is surprising - we shouldn't be able to change the rev just in the Cargo.lock - we should be depending on a specific `rev` instead.
I'll look into this more later.

So the first point isn't actionable, but feel free to investigate the second

flosse:
> Should we actually have a `Cargo.lock` in this repository at all? I wasn't aware we had it not ignored.

Yes, a  `Cargo.lock` is very helpful to reproduce a behavior for an example/binary but I'd not use it for a library.
So it might be an idea to move the content of `src/` into a separate example folder so that it's clear.

 
> Also, this change is surprising - we shouldn't be able to change the rev just in the Cargo.lock - we should be depending on a specific `rev` instead.

I agree!

flosse:
I moved `main.rs` now into an example folder. Fell free to do with this PR whatever you want, It's just a feedback from a random curious developer :wink:

xStrom:
> Should we actually have a `Cargo.lock` in this repository at all?

Generally no. However right now we have it configured as an executable binary not just a library, and for an executable the `Cargo.lock` inclusion is common. For libraries (and indeed those who use the current `xilem` as a library) it is ignored anyway.

---

As for migrating from a binary to example instead, I would prefer if we didn't have so many files and nested folders. Which is to say, I don't think we need a `Cargo.toml` for the example, nor a `examples/basic/src/main.rs`. A simple `examples/hello.rs` should work fine.

xStrom:
> Fell free to do with this PR whatever you want, It's just a feedback from a random curious developer 😉

We do appreciate the feedback. To be more precise of my own feedback to your feedback, it divides into two:
* To fix the actual swash problem, I think just updating `Cargo.lock` for now if it solves the issue for you, is a fine solution. I would approve & merge. (If modifying `Cargo.lock` solves it then the same problem doesn't propagate to people using `xilem` as a library.)
* We also want to move from bin to lib+examples, however I think this PR as it is right now needs modifications. Perhaps it's better to split that work into a different PR, or I can just submit a PR of my own as well. In any case, I don't want this to be a blocker for the original swash problem.

flosse:
> * To fix the actual swash problem, I think just updating `Cargo.lock` for now if it solves the issue for you, is a fine solution.

:+1: 
 
> * We also want to move from bin to lib+examples, however I think this PR as it is right now needs modifications. Perhaps it's better to split that work into a different PR

:+1: 

I removed the last commit, so it's now only the update in `Cargo.lock` again.

xStrom (approved):
Thanks!

PR #69, opened 2023-06-26
Add wasm DOM implementation for xilem.

derekdreery:
This PR adds a html DOM backend for xilem, and makes some modest adjustments to `xilem_core` (including adding an `Adapt` view type.

I've been doing all the work in a single commit so it's easier to rebase. Below is a list of the individual changes I did

 - Squash prev commits
 - Remove top-level action handler (use Adapt instead).
 - Type events in the same way as elements (could also do attributes the same)
 - Allow users to avoid compiling the typed html/attributes/events
 - Use more specific types for mouse events from web_sys
 - change "click" from PointerEvent to MouseEvent (PointerEvent caused a panic on `dyn_into().unwrap()`)


crates/xilem_core/src/view.rs
```diff
@@ -55,5 +67,84 @@ macro_rules! generate_view_trait {
                 app_state: &mut T,
             ) -> $crate::MessageResult<A>;
         }
+
+        pub struct Adapt<OutData, OutMsg, InData, InMsg, F: Fn(&mut OutData, AdaptThunk<InData, InMsg, V>) -> $crate::MessageResult<OutMsg>, V: View<InData, InMsg>> {
```
derekdreery, line 71: Perhaps `OutMsg` -> `OutA`, and `InMsg` -> `InA`?

crates/xilem_html/src/app.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    pub fn run(self, root: &web_sys::HtmlElement) {
+        self.0.borrow_mut().ensure_app(root);
+        // Latter may not be necessary, we have an rc loop.
+        std::mem::forget(self)
+    }
+}
+
+impl<T, V: View<T>, F: FnMut(&mut T) -> V> AppInner<T, V, F> {
```
derekdreery, line 52: The top-level app expects the Action type to be `()`. It is expected that if you have a view with a different action type, you'll wrap it in an `Adapt` and handle the action somehow, e.g. ignoring it.

crates/xilem_html/README.md
```diff
@@ -0,0 +1,7 @@
+# Xilemsvg prototype
+
+This is a proof of concept showing how to use `xilem_core` to render interactive vector graphics into SVG DOM nodes, running in a browser. A next step would be to factor it into a library so that applications can depend on it, but at the moment the test scene is baked in.
+
+The easiest way to run it is to use [Trunk]. Run `trunk serve`, then navigate the browser to the link provided (usually `http://localhost:8080`).
+
+[Trunk]: https://trunkrs.dev/
```
jaredoconnell, line 7: Can you expand the documentation for using Xilem SVG and HTML? It's not obvious to someone like me that hasn't used trunk before. Are we supposed to run `trunk serve` in the directory inside web examples? I did that and I got a lot of compile time errors. Are there any prerequisite steps?

jaredoconnell (changes requested):

crates/xilem_html/web_examples/todomvc/src/state.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Filter {
+    All,
+    Active,
+    Completed,
+}
+
+impl Default for Filter {
```
raphlinus, line 83: There's now a `#[default]` attribute you can set on an enum variant.

crates/xilem_html/web_examples/todomvc/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    let footer = (!state.todos.is_empty()).then(|| footer_view(state));
+    el::div((
+        el::header((
+            el::h1(text("TODOs")),
+            el::input(())
+                .attr("class", "new-todo")
+                .attr("placeholder", "What needs to be done?")
+                .attr("value", state.new_todo.clone())
+                .attr("autofocus", "true")
+                .on_keydown(|state: &mut AppState, evt| {
```
raphlinus, line 186: My feeling is that these callbacks should be returning an `A` rather than a raw `MessageResult`, as the latter potentially exposes too many internal details. But this is open for discussion.

I don't think we need an explicit RequestRebuild (the intent of that is to propagate changes when an async future completes, and in particular blasts through memoize or other lazy mechanisms). The re-render of the view tree should happen naturally after a message is delivered.

crates/xilem_html/web_examples/todomvc/index.html
```diff
@@ -0,0 +1,537 @@
+<!DOCTYPE html>
```
raphlinus, line 1: This looks like it came from somewhere. Does it need attribution?

crates/xilem_core/src/message.rs
```diff
@@ -3,21 +3,60 @@
 
 use std::any::Any;
 
-use crate::id::IdPath;
+#[macro_export]
```
raphlinus, line 6: Why does this need to be generic? My understanding was that `Message` and related types do not really depend on the types of the element tree (in particular, they shouldn't need a bound on the trait). But it's entirely possible I'm missing something.

raphlinus (commented):
Wow, really impressive. I haven't examined all the details, and might want to go a little deeper into some aspects. I did leave a couple comments for things that might be discussed. I think this should go in soon, even as we plan to iterate further, so we have something to build on.

Another thing I'm perhaps missing. This PR doesn't have an `adapt` view ported over, but perhaps that's needed to get the message return types lined up? Again, happy to discuss.

crates/xilem_html/README.md
derekdreery: I think because xilem is moving, I'd like to land this PR before addressing your issue. If it continues for you after the PR is merged, could you create a ticket with the errors you get? I'm keen to fix your problem, but the longer I wait the more this PR bitrots.

crates/xilem_html/web_examples/todomvc/index.html
derekdreery: I've added a note to the example's README.

crates/xilem_html/web_examples/todomvc/src/lib.rs
derekdreery: I've changed the callbacks to use one of `()`, `A` or `Option<A>`. Unfortunately this does require a trait bound on `A` (I made a marker trait called Action). An alternative would be to force the callback to return an `Option<A>`, in which case that restriction could be relaxed. The reason for needing the bound is to ensure the implementatinos don't overlap, in particular `Action` is not implemented for `()`.

derekdreery: I haven't changed it on the `Adapt` view because I didn't work out how to do it in a backend-agnostic way.

jaredoconnell (approved):
I don't see any glaring issues. The documentation needs to be improved soon. I didn't successfully run this code.
I'm excited for this to progress in Xilem!

This approval alone isn't enough, I'd say.

crates/xilem_html/src/view_ext.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn on_dblclick<
+        OA: OptionalAction<A>,
+        F: Fn(&mut T, &Event<web_sys::MouseEvent, Self::Element>) -> OA,
+    >(
+        self,
+        f: F,
+    ) -> e::OnDblClick<T, A, Self, F, OA> {
+        e::on_dblclick(self, f)
+    }
+    fn on_input<
```
Philipp-M, line 68: I think at some point in the future this should be limited only the HTML elements that support it (as well as the other extension methods)

crates/xilem_html/src/view.rs
```diff
@@ -0,0 +1,219 @@
...
+
+impl ViewMarker for &'static str {}
+impl<T, A> View<T, A> for &'static str {
+    type State = ();
+    type Element = web_sys::Text;
+
+    fn build(&self, _cx: &mut Cx) -> (Id, Self::State, Self::Element) {
+        let el = new_text(self);
+        let id = Id::next();
+        (id, (), el.into())
```
Philipp-M, line 111: All the `into`s aren't necessary anymore (clippy bleats)

crates/xilem_html/src/view.rs
```diff
@@ -0,0 +1,219 @@
...
+/// just a `web_sys::Element`), but does allow element types that contain other
+/// data, if needed.
+pub struct Pod(pub Box<dyn AnyNode>);
+
+impl Pod {
+    fn new(node: impl DomNode) -> Self {
+        node.into_pod()
+    }
+
+    fn downcast_mut<'a, T: 'static>(&'a mut self) -> Option<&'a mut T> {
```
Philipp-M, line 90: lifetime `'a` can be omitted (I think this is somewhere on main as well though)

Philipp-M (commented):
Just some (not so relevant) nits. I might take another look later

raphlinus:
I think the file structure might be a little messed up, there's no `index.html` at `crates/xilem_html`, and I don't think the dist subdir should be committed. I was planning to play with it a bit, but I'll desk-check the code for now.

raphlinus (approved):
I have some questions, but agree this is good enough to land as-is, and further improvements can be made on top. My biggest question is whether we should commit to the typed approach. If so, I'm inclined to try to make dynamic changes to the element name go away, and have some ideas how to do that (and might experiment with it).

Some parts I haven't reviewed as carefully, especially the event wrappers, and also the details of the todomvc example - I just tried to get a feel for the flow. I'm interested in getting more of a feel how that compares to other reactive layers, Dioxus and leptos in particular.

I'm ok with this pulling the #62 branch in, but think it might be cleaner history to land it separately. (Actually I don't care about clean history, I just want credit in git statistics. Had you fooled though!)

crates/xilem_html/src/element/elements.rs
```diff
@@ -0,0 +1,232 @@
...
+        /// A view representing a
+        #[doc = concat!("`", $name, "`")]
+        /// element.
+        pub struct $ty_name<ViewSeq>(crate::Element<$web_sys_ty, ViewSeq>);
+
+        /// Builder function for a
+        #[doc = concat!("`", $name, "`")]
+        /// view.
+        pub fn $builder_name<ViewSeq>(children: ViewSeq) -> $ty_name<ViewSeq> {
+            $ty_name(crate::element($name, children))
```
raphlinus, line 22: Incidentally, it's possible to use `stringify!($builder_name)` instead of `$name` here, which might save a parameter and reduce duplication, if these names will always be identical.

crates/xilem_core/src/view.rs
```diff
@@ -55,5 +67,137 @@ macro_rules! generate_view_trait {
                 app_state: &mut T,
             ) -> $crate::MessageResult<A>;
         }
+
+        /// A view that wraps a child view and modifies the state that callbacks have access to.
```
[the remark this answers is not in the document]
derekdreery: Perhaps it should be a separate `impl_adapt_view` macro (along with `impl_memo_view` and the rest).

crates/xilem_svg/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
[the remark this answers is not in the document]
derekdreery: Yep landing the svg branch first sounds sensible. IMO xilem is still early/experimental enough that landing stuff that isn't perfect (with a plan for how it will be improved) is fine.

crates/xilem_html/src/view.rs
derekdreery: Good find. I did a `rg "'a"` and this is the only one I found.

crates/xilem_html/src/element/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+            remove_attribute(element, name);
+            changed |= ChangeFlags::OTHER_CHANGE;
+        }
+        while let Some((name, value)) = self_attrs.next() {
+            set_attribute(element, name, value);
+            changed |= ChangeFlags::OTHER_CHANGE;
+        }
+
+        // update children
+        // TODO avoid reallocation every render?
```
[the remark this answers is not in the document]
derekdreery: I've put the scratch in local_storage, so it won't reallocate.

PR #71, opened 2023-07-03
Add a type alias for Box<dyn View> in the anyview macro

Philipp-M:
This should make writing boxed Anyviews as type a little bit less convoluted.

Motivation for this was a bigger refactor in [trui](https://github.com/Philipp-M/trui), where the View additionally needed the super trait `Send`. I had a lot of unidentifiable error messages because the `Box<dyn View>` should've been `Box<dyn View + Send>` somewhere in the view-tree (generally a topic that needs some research, to avoid hair-pulling, I think...).

I think adding a type alias that automatically has all these extra trait bounds might be generally helpful.

### DIFF
--- a/crates/xilem_core/src/any_view.rs
+++ b/crates/xilem_core/src/any_view.rs
@@ -3,7 +3,7 @@
 
 #[macro_export]
 macro_rules! generate_anyview_trait {
-    ($anyview:ident, $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty, $anywidget:ident $($ss:tt)*) => {
+    ($anyview:ident, $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty, $anywidget:ident, $boxedview:ident; $($ss:tt)*) => {
         /// A trait enabling type erasure of views.
         pub trait $anyview<T, A = ()> {
             fn as_any(&self) -> &dyn std::any::Any;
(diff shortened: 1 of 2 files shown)


raphlinus (approved):
Looks good to me, thanks, and I can see how it would be convenient.

One question, are we going to need a `Send` bound in the multithreaded world?

Philipp-M:
Good question, I have basically "copy/pasted" a lot of the code of the xilem repo as I agreed with most of it and didn't want to reinvent the wheel (e.g. async deadline timeout).
I guess it depends, whether we want to send the view(s) between the `AppTask` and the `App`, I can see why though (which is why I've adopted the code).

raphlinus:
I just checked, and the [xilem_tokio branch](https://github.com/linebender/druid/blob/c978a8060e34434e0862c6c222788fd5d830bd29/xilem/src/view/any_view.rs#L114) does have a `Send` bound in its boxed AnyView. I think that should be added, and that this PR is a good place to do that.

That's also a major motivation to have a type alias!

Philipp-M:
I'm not sure if I understand right, but this should already be the case shouldn't it? So I guess I'll go ahead and merge this then?

raphlinus:
Yes, sorry, it looks like it adds the bound. I had misread the original version and missed the `$($ss)*`. Please go ahead and merge, and thanks again for the change.

PR #72, opened 2023-07-05
Port the Memoize view into xilem_core

Philipp-M:
This ports the memoize view into xilem_core as macro.

I haven't added documentation yet (this could probably be its own PR I think...), it's basically just a copy of the original prototype, adjusted to be compilable again.  
I've added two functions, one for a static view and one that just wraps the `Memoize` view (see it in action e.g. [here](https://github.com/Philipp-M/trui/blob/cee0d3d6d6f4b47a56e6365b9bab793f92caa3e6/examples/blocks.rs#L8) or [here](https://github.com/Philipp-M/trui/blob/cee0d3d6d6f4b47a56e6365b9bab793f92caa3e6/examples/rainbow_blocks.rs#L92))

I guess these extra functions (including their names) are likely a point to discuss still. Especially the static view function could probably be a macro (but I think the syntactic sugar benefits are marginal, by avoiding the extra closure)


derekdreery (approved):
LGTM

We should standardize naming/location in a follow-up PR. I think the `generate_` prefix is the better name, but putting the macro in a `view` submodule is tidier.

PR #73, opened 2023-07-05
xilem_html: OnEvent: don't handle children's messages

rosefromthedead:
I tried to leave this as a review last night but unfortunately [this](https://github.com/orgs/community/discussions/10369) happened.


Philipp-M (approved):
Good catch, had this issue myself already, thanks.

derekdreery:
Nice! I didn't run into this bug because I only attached events at elements without children.

derekdreery:
I'm gonna merge if that's OK @rosefromthedead since there is no way to argue that having this PR is better than not. Hope you don't mind (normally people merge their own PRs).

PR #76, opened 2023-07-05
Unify names/locations of memoize and adapt.

derekdreery:
Also use adapt macro to generate adapt view for xilem.


Philipp-M (approved):
Looks good.

I think we should probably move everything view related then into the view mod shouldn't we? (Like anyview and *maybe* sequence)

derekdreery:
@Philipp-M I did think about moving the others, (anyview and sequence) but I thought that since they are specifically *not* views (i.e. their own traits) they shouldn't be in the view subtree. But it's totally an aesthetic decision so really IDC if others think they would be better in the view module.

PR #77, opened 2023-07-07
Remove Cargo.lock

waywardmonkeys:
This isn't typically included for library projects and adds noise to commits that are updating dependencies.


xStrom:
It's there because Xilem currently isn't just a library but also a binary (see `main/src/main.rs`). However the plan has always been to move that `main.rs` into a basic example instead. I guess we could merge this as a first step in that process.

raphlinus:
I'm ok with doing that. The binary is not super important, it's for experimentation.

I've heard some debate about Cargo.lock in libraries, for example @jonhoo advocates for it ([Tweet](https://twitter.com/jonhoo/status/1571290371124260865)). I don't have strong feelings either way, I think we should go with community consensus and above all be consistent.

xStrom:
There is a movement to start including `Cargo.lock` even for libraries indeed. [cargo#8728](https://github.com/rust-lang/cargo/issues/8728) is very topical.

As both the tweet and people in that cargo issues argue, the main benefit is stability for contributors.

When a library user depends on a library, the library's `Cargo.lock` is ignored. So library contributors / CI not testing in a way that matches actual library users is a real ostrich approach to problems. However the proponents of `Cargo.lock` argue that it's not worth adding friction to unrelated PRs that reveal new issues with sub-dependancies. A counter-argument would be that you want to be informed of sub-dependancy issues early so that those issues could be reported and fixed. Waiting around just increases the time window where library users experience a broken build.

This tradeoff space reminds me of the debate we had back in March on pinning a specific Rust toolchain version vs. using `stable` in CI. With that we decided to try pinning with regular explicit updates. (*Glazier-only for now.*) The main benefit being exactly the same - reducing contributor friction.

Perhaps the move here is to apply similar reasoning also to `Cargo.lock`. To utilize `Cargo.lock` for contributor comfort, but also regularly update it to not hide potential issues for too long.

The more I think about it, the more I like the triple-testing advocated by @jonhoo in that tweet. Unfortunately minimal version testing is in its infancy in Rustland, not even possible with stable toolchains, so that's more of a wishlist thing I think. But having another `rm Cargo.lock` job in CI that can fail for PRs has value. It would show problems, but not add direct friction. I still think that `Cargo.lock` should be regularly updated though, because there are plenty of bugs that our minimal test suite would not catch but actual human usage would reveal. The exact cadence is a question, perhaps in lockstep with Rust releases. Perhaps more frequently.

waywardmonkeys:
I understand various sides of this argument (and that this is common practice in other languages).

I wish that some of the proposed things would actually work on a stable `cargo` rather than still requiring a nightly.

I think that doubling (or tripling) CI in the way suggested in some of the discussion is environmentally immoral. That said, having a `Cargo.lock` file in the code would allow for the caching of crates to work better and reduce some of the CI overhead for the regular build.

I'm going to close this for now, hope that no one doubles up on CI tasks in the future, and submit some changes to update various dependencies that are currently out of date.

PR #78, opened 2023-07-08
Improve CI

waywardmonkeys:


.github/workflows/rust.yml
```diff
@@ -11,15 +11,24 @@ env:
...
       run: |
         sudo apt update
         sudo apt install libx11-dev libpango1.0-dev libxkbcommon-dev libxkbcommon-x11-dev
       if: runner.os == 'Linux'
     - uses: actions/checkout@v3
-    - uses: dtolnay/rust-toolchain@stable
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: stable
+        components: rustfmt
```
Philipp-M, line 28: Just a small optimization (to avoid wasting cloud compute resources), I think it's enough to format on just one system (likely linux).

Philipp-M (approved):
Apart from the one comment LGTM. Thanks.

.github/workflows/rust.yml
waywardmonkeys: Can with a conditional here... I didn't want to do it as a separate job as that means a separate spin up, clone, install of rust, etc and that's not a worthwhile use of resources.

waywardmonkeys:
I've got other things I want to add to this ... but having it as a draft lets it get early review (like now) as well as make sure the pieces work as they get written.

.github/workflows/rust.yml
Philipp-M: Yeah that's exactly what I've thought of as well.

Philipp-M:
Yeah I get this, but I think generally it's better/easier (to review) to just have a small "atomic" change if possible in a PR (sometimes it's not, or it's overhead for everyone involved...), so I would suggest splitting changes in separate PRs

.github/workflows/rust.yml
Philipp-M: > I didn't want to do it as a separate job as that means a separate spin up, clone, install of rust, etc and that's not a worthwhile use of resources.

Btw. isn't that exactly what happens with `matrix` (just that it's some syntactic sugar that abstracts it away as the workflow is similar enough)?

Philipp-M: One thing I think would be cool is something like (I'm not sure if this is working, it's untested):

```suggestion
        components: ${{ runner.os == 'Linux' && 'rustfmt' || '' }}
```

derekdreery:
Thanks for working on this - it would be good if we could find a way for the CI to catch problems like #82 , which I think should be possible. Once things are a bit more settled I'd like to write some tests.

Philipp-M:
> it would be good if we could find a way for the CI to catch problems like #82

I doubt that this is possible without a few tests, as it's a runtime error.

Setting up good E2E tests would probably be the best for finding such issues (using something like [playwright](https://playwright.dev/)) but this requires quite some work I think.

waywardmonkeys:
Let's say this is ready for review now and further work on improving CI can happen.

This can't be merged at the moment as it will fail until another PR is landed to fix a formatting issue recently introduced.

Philipp-M:
Have you tried what I've commented? (Don't have to, I will open a follow-up PR otherwise and check, if that's working)

waywardmonkeys:
I did not yet. That's part of why I haven't just gone and merged this. :)

.github/workflows/rust.yml
waywardmonkeys: @Philipp-M That did work. Thanks.

waywardmonkeys:
Ah, you're right. They are both included in the `minimal` profile. Ah well. Maybe I'll revisit it once we can enable `clippy` checks...

PR #79, opened 2023-07-08
svg: clippy: Fix useless format warnings.

waywardmonkeys:
Since `to_svg` returns a `String`, we don't need to format it to a string.


waywardmonkeys:
(The current CI isn't actually testing this... but it will once we're able to land PR #78 and possible follow-ups.)

Philipp-M (approved):
Thanks! (I'll take the liberty to merge this, as it's an obvious improvement)

PR #80, opened 2023-07-08
svg: Rename `xilemsvg` to `xilem_svg`.

waywardmonkeys:
There are 2 other crates (so far) with similar naming, but are using the underscore: `xilem_core` and `xilem_html`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3314,7 +3314,7 @@ dependencies = [
 ]
 
 [[package]]
-name = "xilemsvg"
+name = "xilem_svg"
 version = "0.1.0"
 dependencies = [
  "bitflags 1.3.2",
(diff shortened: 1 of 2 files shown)


Philipp-M (approved):
Yeah I think that makes sense, since the other crates are named like that as well.

Though we should probably generally discuss (probably on zulip) (crate) naming conventions (or just keep the current with underscore).

waywardmonkeys:
I mentioned this on Zulip: https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Crate.20naming

xStrom (approved):
As discussed in zulip, there seems to be consensus to standardize new Linebender crates with `snake_case`.

So merging this makes sense. :rocket:

PR #82, opened 2023-07-08
Fix re-entrancy bug. TodoMVC should work now.

derekdreery:


jplatte (approved):
Fixes a `BorrowMutError` I got when playing around with this!

PR #83, opened 2023-07-09
Implement ViewSequence for Either

jplatte:
I don't have a deep understanding of these traits, but this achieves:

- `Either<impl ViewSequence, impl ViewSequence>` implementing `ViewSequence`
- `Either<impl View + ViewMarker, impl View + ViewMarker>` implementing `View` and thus `ViewSequence`

Regarding the new implementation of `ViewSequence`, I have no idea if it's correct, in particular the `rebuild` method. I just implemented it in as closely to the existing `View::rebuild`, and such that the types line up. Please check if I did it right 😅

### DIFF
--- a/crates/xilem_html/src/view.rs
+++ b/crates/xilem_html/src/view.rs
@@ -126,8 +126,8 @@ where
 
 impl<T, A, V1, V2> View<T, A> for Either<V1, V2>
 where
-    V1: View<T, A>,
-    V2: View<T, A>,
+    V1: View<T, A> + ViewMarker,
+    V2: View<T, A> + ViewMarker,
     V1::Element: AsRef<web_sys::Node> + 'static,
     V2::Element: AsRef<web_sys::Node> + 'static,
 {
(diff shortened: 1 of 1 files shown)


crates/xilem_html/src/view.rs
```diff
@@ -213,6 +211,94 @@ where
...
+
+    fn rebuild(
+        &self,
+        cx: &mut Cx,
+        prev: &Self,
+        state: &mut Self::State,
+        element: &mut xilem_core::VecSplice<Pod>,
+    ) -> ChangeFlags {
+        match (prev, self) {
+            (Either::Left(_), Either::Right(view)) => {
```
Philipp-M, line 236:
```suggestion
            (Either::Left(_), Either::Right(view_sequence)) => {
```

As well for the other `view`s in the `impl`.

Philipp-M (commented):
Apart from the variable names, this looks good to me.

I guess something like this is general enough (including the Either View) that it could probably be integrated into xilem_core as macro.

Philipp-M (approved):
Another day, another way to find issues, but I haven't found any so far :)

One thing (maybe @raphlinus wants to chime in and explain the potential purpose/use-case of that):

Currently the blanket implementation of `ViewSequence` for `View` creates an extra id for the sequence, while the other implementations (e.g. `Option<impl View>` or `(<tuples of views>)`) are not.

I think it's not necessary (here) so this PR is good to merge, but an explanation why something like that makes sense would be nice (and if it doesn't maybe we should slim down the blanket implementation a little bit in xilem_core).

Another thing, if this should/will be integrated into xilem_core:

E.g. the `ViewSequence` implementation of e.g. `Option<impl ViewSequence>` uses
```rust
<$changeflags>::tree_structure()
```
in `ViewSequence::rebuild()`, this is effectively the same as it's currently done in here.

I think we should probably discuss (on zulip) and "standardize" a few changeflags for xilem_core, `empty`, `default` and `tree_structure` seems to be used currently.

crates/xilem_html/src/view.rs
```diff
@@ -155,38 +155,36 @@ where
...
         element: &mut Self::Element,
     ) -> ChangeFlags {
-        let mut change_flags = ChangeFlags::empty();
         match (prev, self) {
             (Either::Left(_), Either::Right(view)) => {
                 let (new_id, new_state, new_element) = view.build(cx);
                 *id = new_id;
                 *state = Either::Right(new_state);
                 *element = Either::Right(new_element);
-                change_flags |= ChangeFlags::STRUCTURE;
```
derekdreery, line 165: I got into the habit of 'or-ing' change flags to make sure I always added rather than replaced flags in cases where there were multiple changes, but here it's fine to just return the changeflag directloy since it's equivalent.

derekdreery (approved):
I don't understand 100% why having `Either` implement `ViewSequence` is useful (since it is always exactly len 1) but I'm sure it will become apparent to me in the fullness of time 🙂 . I also think that we should examine whether this widget could be moved into xilem_core in the future, but there's no rush for this (and certainly no need for it to be in this PR).

@jplatte feel free to merge this yourself. If you need write access just ask for it on the Zulip and one of the admins will give it to you.

Philipp-M:
>  (since it is always exactly len 1)

It should be "either" the length of the left or the right view sequence.

I think it makes sense when having differently sized and typed `ViewSequence`s that are not dynamic (like a `Vec`).
For example a statically typed DOM node that has conditionally different typed children (e.g. `div(Either::Left("1"))` and `div(Either::Right((p("1"), p("2")))`)

jplatte:
> @jplatte feel free to merge this yourself. If you need write access just ask for it on the Zulip and one of the admins will give it to you.

I have write access, just waiting for Raph who wanted to also take a look AFAIU.

Philipp-M:
> Currently the blanket implementation of ViewSequence for View creates an extra id for the sequence

Ah, I get it now, the `Id` of the `View` would be lost otherwise (it's stored in `<impl ViewSequence>::State`), since it's not propagated up the view tree like in `View::build`, the other `ViewSequence` impls don't need that since they act directly on other `ViewSequence`s.

With that in mind it's good to merge IMHO.

raphlinus:
I haven't gotten a chance to look deeply (am trying to keep focus on GPU stuff today), but from a skim of the comments it looks like it's been reviewed adequately so I'm happy for it to be merged. And if there's a problem we can fix it later, this stuff is still fluid.

PR #85, opened 2023-07-09
Implement more of the TodoMVC functionality.

derekdreery:
This patch allows all tasks to be toggled complete/incomplete.

There is a bug where when the filter changes, the 'completed' check gets out of sync with its task. I haven't nailed down the exact cause yet.

### DIFF
--- a/crates/xilem_html/web_examples/todomvc/src/state.rs
+++ b/crates/xilem_html/web_examples/todomvc/src/state.rs
@@ -1,13 +1,7 @@
 use serde::{Deserialize, Serialize};
-use std::sync::atomic::{AtomicU64, Ordering};
 use wasm_bindgen::UnwrapThrowExt;
 
 const KEY: &str = "todomvc_persist";
-
-fn next_id() -> u64 {
-    static ID_GEN: AtomicU64 = AtomicU64::new(1);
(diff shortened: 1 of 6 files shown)


crates/xilem_html/web_examples/todomvc/src/state.rs
```diff
@@ -48,21 +78,42 @@ impl AppState {
...
+            Err(e) => {
+                tracing::error!("couldn't load existing todos: {e}");
+                Default::default()
+            }
+        }
+    }
+
+    /// Save the current state to local_storage
+    pub fn save(&self) {
+        let raw = serde_json::to_string(self).unwrap_throw();
```
Philipp-M, line 98: I'm not sure if `serde_json` isn't a little bit too heavy for this. I'm thinking about something like `serde_wasm_bindgen`, but I think we can optimize that later.

Philipp-M (approved):
LGTM, thanks.

Philipp-M:
Btw. I couldn't trigger the bug (in firefox), do you have a way to consistently reproduce this?

derekdreery:
> Btw. I couldn't trigger the bug (in firefox), do you have a way to consistently reproduce this?

It might have been fixed. I'll ignore it for now and report it again if I find a reliable repro.

PR #90, opened 2023-07-12
Import kurbo via vello, not glazier.

waywardmonkeys:
This makes this consistent in terms of where kurbo things are imported from.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -17,11 +17,10 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use accesskit::TreeUpdate;
-use glazier::kurbo::Size;
 use glazier::{IdleHandle, IdleToken, WindowHandle};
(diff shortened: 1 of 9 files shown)


jaredoconnell (approved):
This seems like a logical change to me. People will just need to keep an eye on the auto-imports if they use them.

waywardmonkeys:
I had previously opened a discussion in Zulip about this: https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Importing.20from.20kurbo

Philipp-M (approved):
Makes sense. Vello is semantically closer to kurbo than glazier and staying consistent here is a good idea. Thanks!

xStrom (approved):
Unifying imports is a good step towards reducing contributor confusion.

The crate sync issue can be handled in other ways and in a different PR, as also discussed in zulip.

PR #96, opened 2023-07-13
deps: Update bitflags from 1.x to most recent 2.x.

waywardmonkeys:


Philipp-M (approved):
Done this already in my framework, so LGTM (as it looks pretty much like in trui)

PR #98, opened 2023-07-15
Fix and use module UseState

Philipp-M:
This fixes the `UseState` module to be compilable and usable again with the current xilem architecture.

A port to `xilem_core` should be straightforward (and could be done in a follow-up PR).

### DIFF
--- a/src/view/use_state.rs
+++ b/src/view/use_state.rs
@@ -109,6 +109,9 @@ where
     }
 }
 
+/// "Injects" additional local state into the provided view (which is returned by `f`)
+/// Requires that the parent state is of `Arc<T>`.
+/// The initial state is provided by `f_init` (when this view is first built)
 pub fn use_state<T, A, S, V, FInit, F>(f_init: FInit, f: F) -> UseState<T, A, S, V, FInit, F>
 where
     V: View<(Arc<T>, S), A>,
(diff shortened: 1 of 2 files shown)


src/view/use_state.rs
```diff
@@ -103,3 +108,12 @@ where
         a
     }
 }
+
+pub fn use_state<T, A, S, V, FInit, F>(f_init: FInit, f: F) -> UseState<T, A, S, V, FInit, F>
```
waywardmonkeys, line 112: Can you add a doc comment here?

Philipp-M: If I would understand exactly what `UseState` does, I'd do ^^ (see zulip thread). I think probably someone else should document the code (maybe in another PR), this function just runs `UseState::new` obviously, I added the trait bounds, so that it's easier to see in rust-analyzer what is required to get this into a `View` (as it otherwise doesn't make much sense I think).

Philipp-M: Added a comment, could probably be better, but I guess better than nothing for now.

DJMcNab:
I think we can close this for now. `use_state` will likely eventually be needed, but we're trying to avoid that for as long as possible...

Philipp-M:
Yeah lets close this, I don't think we should rely on the app state to be `Arc<State>` anyway, as this severely restricts the use-cases.
I do hope we find ways to make `adapt/map_state/lens` more flexible (e.g. modularice distinct state items, as described [here](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Adapting.20non-trivial.20State)), so that we can keep the all the user-state in `AppState`

PR #99, opened 2023-07-15
xilem_html: Add OneOf3..=OneOf8

jplatte:
Screenshot_2023-07-27_173237

Note that the amount of macro-generated code increases linearly w.r.t. the amount of type variables. Thus even `Either8` or an either type with even more generics might be viable w.r.t. compile times / code size.


Philipp-M:
Thanks, switching "dynamically" between views conditionally is certainly an interesting topic and should definitely be researched thoroughly (think about structural identity like in SwiftUI for example, or as discussed on zulip: caching of views).

Since it's wrapped in a macro anyways already and it's so general, I think it makes sense to directly put this into xilem_core. As xilem is still really experimental it makes sense IMHO to increase the experimentation surface for this (say xilem itself (the vello version) or my [TUI framework](https://github.com/Philipp-M/trui) which I use individually for experimentation in a slightly different context)

Checkout e.g. the `Adapt` or `memoize` views for reference (and for the changeflags similar as in the `ViewSequence` macros `<$changeflags>::tree_structure()`).

jplatte:
> Since it's wrapped in a macro anyways already and it's so general, I think it makes sense to directly put this into xilem_core.

It would make the macro a bit more complicated though. We'd need a place to put `Element` type bounds in the macro invocation, and the `throw_str` thing is of course also going to be different for non-web xilem crates. I'd suggest adding `Either` to the regular `xilem` crate via another macro there first, before deciding whether a more general macro for both in `xilem_core` really makes sense.

crates/xilem_html/src/either.rs
```diff
@@ -0,0 +1,225 @@
...
+    };
+}
+
+either! {
+    /// This view container can switch between two views.
+    Either { A, B }
+}
+either! {
+    /// This view container can switch between three views.
+    Either3 { A, B, C }
```
Philipp-M, line 199: Multiples > 2 are usually called `OneOf` so I would suggest using this term instead

crates/xilem_html/src/lib.rs
```diff
@@ -21,13 +22,14 @@ pub use xilem_core::MessageResult;
 pub use app::App;
 pub use class::class;
 pub use context::{ChangeFlags, Cx};
+pub use either::{Either, Either3, Either4, Either5, Either6, Either7, Either8};
```
Philipp-M, line 25: I think a wildcard export is good enough.

Philipp-M:
> We'd need a place to put Element type bounds in the macro invocation

I see... just checked it in more detail, the Element bound for the `View` implementation of `Either` certainly complexifies things.

In that case keeping it in xilem_html for now makes sense.

Btw. doesn't `panic!` lead to a similar result (in this context) as `throw_str`?

crates/xilem_html/src/either.rs
jplatte: I've actually seen `EitherN` more times in Rust context, but I agree that `OneOf` is a bit more natural. Wdyt should happen to `Either` in that case? Should it stay as-is, become `OneOf`, or maybe `OneOf2`?

jplatte:
> Btw. doesn't `panic!` lead to a similar result (in this context) as `throw_str`?

I don't know, I just copied what was there before. I'm *guessing* that `throw_str` has less code size overhead (because no formatting machinery is involved).

crates/xilem_html/src/lib.rs
jplatte: I strongly dislike wildcard imports / re-exports. This list is easy enough to read as-is

crates/xilem_html/src/either.rs
Philipp-M: I think `Either` is fine for two elements, feels more natural than `OneOf2` or `OneOf` IMHO (and is used as well more often in functional programming)

crates/xilem_html/src/lib.rs
Philipp-M: Ok, fair point, I have no strong opinion on this (and generally agree with you)

crates/xilem_html/src/either.rs
jplatte: Hm, then I'm also wondering if it still makes sense to have `Either`s variants keep the `Left` / `Right` variant names.

But thinking more about it, I think I'd really prefer `Either` to be `OneOf2` for consistency.

Maybe somebody else wants to chime in as well to give their opinion on the color of the bikeshed? 😄

jplatte:
Was this discussed at the last office hours? I've now rebased and renamed to `OneOf` since that's what @Philipp-M prefers and I don't mind the change.

jplatte:
Rebased again to fix conflicts with my other PRs ^^

PR #100, opened 2023-07-15
Rename druid_app to glazier_app.

waywardmonkeys:


Philipp-M (approved):
Makes sense, I guess it could be just named `app` too.

But `glazier_app` is certainly more descriptive.

waywardmonkeys:
I had actually renamed it to `app` first and then found it more confusing. :)

PR #101, opened 2023-07-15
Added `cargo clippy` to the CI workflows and fixed/disabled a few clippy lints

Philipp-M:
Currently there are two warnings left, that I think need discussion, like the following:

```
error: module has the same name as its containing module
  --> src/view/mod.rs:24:1
   |
24 | mod view;
```

I think these are good points, and I have named these `core` in my TUI framework for each of these modules (widget and view), and would suggest doing this as well here.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,15 +22,17 @@ jobs:
         sudo apt install libx11-dev libpango1.0-dev libxkbcommon-dev libxkbcommon-x11-dev
       if: runner.os == 'Linux'
     - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


src/test_scenes.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused)]
```
waywardmonkeys, line 1: This file shouldn’t be part of the library build. Should be moved to an example, along with the min.rs

Philipp-M: I agree, but this is out of the scope of this PR.

xStrom (approved):
Thanks!

PR #102, opened 2023-07-15
xilem_html: Add support for non-passive event handlers

jplatte:
Allows adding `.passive(false)` after `.on_click(/* closure */)`. Maybe not the most intuitive but since optional args aren't a thing, I don't have another idea for now.

### DIFF
--- a/crates/xilem_html/src/event/events.rs
+++ b/crates/xilem_html/src/event/events.rs
@@ -26,6 +26,24 @@ macro_rules! event {
             optional_action: std::marker::PhantomData<OA>,
         }
 
+        impl<T, A, V, F, OA> $ty_name<T, A, V, F, OA>
+        where
+            V: crate::view::View<T, A>,
+            F: Fn(&mut T, &$crate::Event<$web_sys_ty, V::Element>) -> OA,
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
Todomvc should call this, or did you want to do that separately?

jplatte:
Right, didn't think to update the example (actually needed this elsewhere).

waywardmonkeys (approved):
I think this looks good, but @Philipp-M or someone else should probably review as well.

Philipp-M (approved):
LGTM.

The design of changing attributes of event listeners (and other "pseudo" views) is also still an open question I think (see also [here](https://github.com/linebender/xilem/pull/103#discussion_r1265698733)).

For now the builder pattern per fine-grained attribute is fine, but I have the feeling that it can get a little bit confusing when overusing the builder pattern (one idea I have in mind is something like [this](https://github.com/Philipp-M/trui/blob/74c46ca985558f6f3129f4a3dd773e00fa5e50ff/src/view/block.rs#L136)). But I guess we should just experiment a little bit to get a feeling about that...

PR #103, opened 2023-07-16
xilem_html: Add ViewExt::{adapt, adapt_state}

jplatte:
More convenient than `{Adapt, AdaptState}::new`.

### DIFF
--- a/crates/xilem_html/src/view_ext.rs
+++ b/crates/xilem_html/src/view_ext.rs
@@ -3,7 +3,9 @@
 
 use std::borrow::Cow;
 
-use crate::{class::Class, event::OptionalAction, events, view::View, Adapt, AdaptThunk, Event};
+use crate::{
+    class::Class, event::OptionalAction, events, view::View, Adapt, AdaptState, AdaptThunk, Event,
+};
 
 /// A trait that makes it possible to attach event listeners and more to views
 /// in the continuation style.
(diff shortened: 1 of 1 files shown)


crates/xilem_html/src/view_ext.rs
```diff
@@ -15,88 +17,71 @@ pub trait ViewExt<T, A>: View<T, A> + Sized {
...
-    }
-    fn on_dblclick<
-        OA: OptionalAction<A>,
-        F: Fn(&mut T, &Event<web_sys::MouseEvent, Self::Element>) -> OA,
-    >(
-        self,
-        f: F,
-    ) -> e::OnDblClick<T, A, Self, F, OA> {
-        e::on_dblclick(self, f)
+    fn adapt<ParentT, ParentA, F>(self, f: F) -> Adapt<ParentT, ParentA, T, A, Self, F>
```
Philipp-M, line 67: I'm not entirely sure if we should use the xilem_core nodes via the builder "attribute" pattern for every HTML element.
As previously discussed on zulip, IMO at some point we want to generate (almost) all possible HTML element attributes strongly typed in a similar fashion as wasm-bindgen via "WebIDL" files.

So I think that only real HTML attributes should be used in a builder "attribute" pattern fashion.

But I think this whole topic needs to be discussed in more detail (by more people) on zulip as it's quite foundational.

Philipp-M: Or maybe directly in a github issue, so that it has more potential surface for external people.

Philipp-M (approved):
I guess we should just experiment/iterate a little bit right now.

My feeling is that we're rewriting/refactoring a lot of the stuff in the future anyways, so I think it makes sense to land this (doesn't add a lot of code as well), and if it turns out not to be useful or is confusing we can still remove it at some time later.

So otherwise LGTM, thanks.

PR #104, opened 2023-07-16
Add a default for generic parameter F of Adapt

jplatte:
I'm not sure I'll actually need this but I think there shouldn't be any serious downsides, either way.

### DIFF
--- a/crates/xilem_core/src/view/adapt.rs
+++ b/crates/xilem_core/src/view/adapt.rs
@@ -59,8 +59,8 @@ macro_rules! generate_adapt_view {
             ParentA,
             ChildT,
             ChildA,
-            F: Fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA>,
             V: $viewtrait<ChildT, ChildA>,
+            F: Fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA>,
         > {
             f: F,
             child: V,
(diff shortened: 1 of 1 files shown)


crates/xilem_core/src/view/adapt.rs
```diff
@@ -59,8 +59,9 @@ macro_rules! generate_adapt_view {
             ParentA,
             ChildT,
             ChildA,
-            F: Fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA>,
             V: $viewtrait<ChildT, ChildA>,
+            F: Fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA> =
```
Philipp-M, line 63: I'm not opposed to this, but what benefits does this offer (right now I only see additional boilerplate IMHO)?

Shouldn't all these generic parameters be inferenced anyways?

jplatte: As per the full commit message of the second commit: This allows spelling out the Adapt type easily if F is a regular function or closure that does not capture. (i.e. you can have a function return `Adapt<…>` explicitly, instead of `impl View<…> + ViewMarker`). This can help to get better compiler errors, and could even be used to have an `Adapt`-typed struct field (though I don't think it's all that useful).

Philipp-M: I'm sold for aiming for better error messages :)
Thanks for the explanation, it shouldn't indeed hurt much.
So if I'm getting it right you could (roughly) do something like this:

```rust
fn my_view() -> Adapt<ParentT, ParentA, ChildT, ChildA, impl View> {...}
```
?

jplatte: Exactly, yes. Before, you would have had to also write out `impl Fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V> -> MessageResult<ParentA>` as a generic argument before the `impl View`. Also, `impl View` could be replaced by a concrete type or `Box<dyn AnyView>`, such that you can name `Adapt` in places where `impl Trait` is not supported.

Philipp-M (approved):

PR #105, opened 2023-07-16
Add AdaptState view

jplatte:


jplatte:
Huh, so this compiles for `xilem_html` but not `xilem`. Need to investigate how `AdaptState` is different from `Adapt` such that the former works without `Send` bounds.

Philipp-M (approved):
Thanks, I'll find it less mentally straining than the current `Adapt` view and I think it's probably good enough for 90% of the use-cases, where otherwise an `Adapt` view would be used.

I was so bold and pushed a few small fixes directly on your branch, hope you don't mind...

If you checkout the other core views, you'll see that they all have a `Send` bound for function type bounds, at least, if it not configurable like the `AnyView` macro (or the `View` macro itself).

I think a follow-up PR should make all the extra bounds + identifiers configurable via macro parameters (for all views, including `Adapt`, `Memoize`) The `generate_anyview_trait` macro is in a good state which can be seen as reference for that IMHO.
Not that identifiers are likely changed, but if we're already using macros for the core views, we can as well make the names configurable.

Another question I'm asking myself, is if we should use one view per file (though `Adapt` and `AdaptState` are really similar on the other hand...)

PR #106, opened 2023-07-16
src/test_scenes.rs: Remove.

waywardmonkeys:
This code isn't used / called and isn't really a xilem test so much as Vello.


DJMcNab (approved):
Seems reasonable to me

PR #107, opened 2023-07-19
Remove `gtk` feature.

waywardmonkeys:
Since this is going away in the underlying glazier, we can remove it now to make sure that everyone is aware. The default is `x11`, so this should have little impact.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,30 +219,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Another relevant detail is that we (should) panic on startup when the gtk feature is enabled, so it's not like it's currently useful (although that same argument is not sufficient to remove the Wayland feature)

PR #108, opened 2023-07-20
xilem_core: Add support for generic type parameters in the `View` trait

Philipp-M:
This can be useful for example if the context is itself generic.

Also makes the trait bounds of the `Memoize` and the `Adapt` views configurable (mostly to mirror the trait bounds of the `View`).

I've implemented a [dummy](https://github.com/Philipp-M/xilem_dummy) implementation of xilem_core to test the (compilation of the) macros, I guess something like this could be useful for testing xilem_core in general, as some usecases might not be covered by the crates of this repo itself.

I needed this for experimentation with an application specific App context that can be used inside View implementations.

### DIFF
--- a/crates/xilem_core/src/sequence.rs
+++ b/crates/xilem_core/src/sequence.rs
@@ -4,8 +4,8 @@
 #[doc(hidden)]
 #[macro_export]
 macro_rules! impl_view_tuple {
-    ( $viewseq:ident, $pod:ty, $cx:ty, $changeflags:ty, $( $t:ident),* ; $( $i:tt ),* ) => {
-        impl<T, A, $( $t: $viewseq<T, A> ),* > $viewseq<T, A> for ( $( $t, )* ) {
+    ( $viewseq:ident <$($tp:ident,)*>, $pod:ty, $cx:ty, $changeflags:ty, $( $t:ident <$($tp_:ident,)*>),* ; $( $i:tt ),* ) => {
+        impl<T, $( $tp, )* A, $( $t: $viewseq<T, $( $tp_, )* A> ),* > $viewseq<T, $( $tp, )* A> for ( $( $t, )* ) {
             type State = ( $( $t::State, )*);
 
             fn build(&self, cx: &mut $cx, elements: &mut Vec<$pod>) -> Self::State {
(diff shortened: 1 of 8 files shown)


crates/xilem_core/src/sequence.rs
```diff
@@ -295,26 +295,36 @@ macro_rules! generate_viewsequence_trait {
...
+            V0<$($tp,)*>, V1<$($tp,)*>, V2<$($tp,)*>, V3<$($tp,)*>, V4<$($tp,)*>, V5<$($tp,)*>, V6<$($tp,)*>;
+            0, 1, 2, 3, 4, 5, 6);
+        $crate::impl_view_tuple!($viewseq <$($tp,)*>, $pod, $cx, $changeflags,
+            V0<$($tp,)*>, V1<$($tp,)*>, V2<$($tp,)*>, V3<$($tp,)*>, V4<$($tp,)*>, V5<$($tp,)*>, V6<$($tp,)*>, V7<$($tp,)*>;
+            0, 1, 2, 3, 4, 5, 6, 7);
+        $crate::impl_view_tuple!($viewseq <$($tp,)*>, $pod, $cx, $changeflags,
+            V0<$($tp,)*>, V1<$($tp,)*>, V2<$($tp,)*>, V3<$($tp,)*>, V4<$($tp,)*>, V5<$($tp,)*>, V6<$($tp,)*>, V7<$($tp,)*>, V8<$($tp,)*>;
+            0, 1, 2, 3, 4, 5, 6, 7, 8);
+        $crate::impl_view_tuple!($viewseq <$($tp,)*>, $pod, $cx, $changeflags,
+            V0<$($tp,)*>, V1<$($tp,)*>, V2<$($tp,)*>, V3<$($tp,)*>, V4<$($tp,)*>, V5<$($tp,)*>, V6<$($tp,)*>, V7<$($tp,)*>, V8<$($tp,)*>, V9<$($tp,)*>;
```
Philipp-M, line 327: I'm not sure if there's a less boilerplate heavy way to do this, but nesting repetitions is not possible unfortunately.

Philipp-M:
Another idea I had is to to be able to add defaults to the type parameters (such that they're more convenient to use in a specific application), but I guess this can be added on top of this in a different PR.

Philipp-M:
I'm not sure whether I want to proceed with this PR, as it introduced quite a bit of generic accidental complexity with my experiments (and may increase maintenance effort for future core views) . I chose a different direction for now (less application context in the views, a more declarative "external" approach). But I'll leave it open, in case someone sees benefit in it (or I'll change my mind).

Philipp-M:
Closing this, I think it mostly increases complexity for no good reason (we have many generic parameters already, don't need to pump it up further...).

PR #109, opened 2023-07-23
xilem_core: Move bounds to where clauses in macros

jplatte:
More readable in documentation, and improves rustfmt's output.

Before:

Screenshot 2023-07-23 at 12-36-45 AdaptState in xilem_html - Rust

After:

Screenshot 2023-07-23 at 12-37-24 AdaptState in xilem_html - Rust

### DIFF
--- a/crates/xilem_core/src/view/adapt.rs
+++ b/crates/xilem_core/src/view/adapt.rs
@@ -59,9 +59,8 @@ macro_rules! generate_adapt_view {
             ParentA,
             ChildT,
             ChildA,
-            V: $viewtrait<ChildT, ChildA>,
-            F: Fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA> =
-                fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA>,
+            V,
+            F = fn(&mut ParentT, AdaptThunk<ChildT, ChildA, V>) -> $crate::MessageResult<ParentA>,
         > {
             f: F,
             child: V,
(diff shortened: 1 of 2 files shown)


crates/xilem_core/src/view/adapt.rs
```diff
@@ -189,13 +175,10 @@ macro_rules! generate_adapt_state_view {
             }
         }
 
-        impl<
-                ParentT,
-                ChildT,
-                A,
-                V: $viewtrait<ChildT, A>,
-                F: Fn(&mut ParentT) -> &mut ChildT + Send,
-            > $viewtrait<ParentT, A> for AdaptState<ParentT, ChildT, V, F>
```
jplatte, line 198: This is what I mean about the rustfmt output, for some reason it was previously indenting generics too much.

jplatte:
Oh, also worth noting that I removed all generic type bounds on the declarations of `Adapt` and `AdaptState`, which is useful for a followup PR / experiment I'm working on. It's a good practice to only put bounds where they are strictly required. (unless one wants to potentially make use of a redundant bound in an upcoming non-breaking release)

Philipp-M (approved):
Haven't thought about docs, that's a good reason to use the where clauses indeed (apart from general readability when trait bounds get too complex).

Btw. I have made the Send bounds configurable in https://github.com/linebender/xilem/pull/108 (`Send` isn't necessary for single threaded environments such as xilem_html). I can rebase on top of these changes as I will split them out of that PR anyways, since I'm not sure anymore if it's a good idea to make `View` even more generic (has lead to a lot of generic accidental complexity with my experiments of an application specific context in views, so I'm currently exploring this from another direction)

PR #112, opened 2023-07-23
xilem_html: Re-export memoize, s functions from crate root

jplatte:
More macro-generated symbols that were not accessible.


Philipp-M (approved):
Thanks

PR #113, opened 2023-07-28
xilem_core: Remove redundant macro match arm in message macro

Philipp-M:
The match arm is already covered in the arm below (the removed one).


Philipp-M:
I've seen that the only difference is a `+ ` in the trait bounds but that shouldn't really matter, all dependent packages (previously using the first match arm) compile.

jplatte (approved):
Makes sense. A trailing `+` will be emitted in two places, but that's not problem.

Issue #54, opened 2023-03-22
Next steps for tree structure tracking


raphlinus:
This is an outline for how I think we should track tree structure in the widget tree.

A number of operations on the view side can change the widget tree structure. The widget tree doesn't change structure by itself, but only in response to mutations from the view side. Those operations are, roughly: an `AnyView` changing view type, and `Option<V>` switching between `None` and `Some`, and a variable length implementor of `ViewSequence` (such as `Vec`) changing length. The goal is for those view operations to mark the tree precisely, then for a traversal of the tree to update accessibility and internal invariants.

## Tree marking

The desired result of tree marking is as follows: every container that has had its list of children changed is marked with `PodFlags::TREE_CHANGED`, and all of its ancestors are marked with the corresponding upward flag; we could create a new one, or use `UPDATE` for this purpose.

Any `ViewSequence` implementation that changes its count should signal `ChangeFlags::TREE`, as should `AnyView` on type change. The tree marking is currently done by the `rebuild` method in the blanket impl of `ViewSequence` for `V: View`, though this plumbing is in a state of flux. We should not propagate `TREE_CHANGED` or `TREE` upward (it is currently in `UPWARD_FLAGS`) but should instead propagate the corresponding upward flag.

## Accessibility

A container generally needs to update its accessibility node in several cases: its geometry changed or its children changed. *Probably* the best way to do this is to have the framework set `REQUEST_ACCESSIBILITY` (and the corresponding upward flag) in both these cases, so widgets don't have to have this logic themselves.

Open question: should that be done during tree marking, or perhaps in a separate traversal? The `update` traversal would be a natural place, but it's not obvious we're going to keep that.

## WidgetAdded

Druid sends a `WidgetAdded` lifecycle event to widgets newly added to the tree. This is propagated from several different traversals. I think it's likely we'll want to send such a lifecycle event also (though I'm not 100% sure it's still needed). We need to figure out which traversal it is. Druid uses the presence of the data field; we'll either need a flag in `PodFlags` indicating whether the widget has been initialized, or make use of structure tracking.

Druid also adds children to the Bloom filter at the same time. The way Bloom filters work, widgets can't be removed (I believe this is a bug in the existing xilem code, but haven't validated it carefully).

## Structure tracking

[...]

PR #115, opened 2023-07-28
xilem: add `ChangeFlag::DESCENDANT_TREE_CHANGED` for finer-grained tree-structure change-detection

Philipp-M:
If I understood https://github.com/linebender/xilem/issues/54#issuecomment-1656103313 right, I think this should be enough to fix the point "Tree marking" in https://github.com/linebender/xilem/issues/54 for xilem (vello) at least.

The behavior itself hasn't changed yet, but this should enable the possibility to have finer-grained tree-structure tracking, as `TREE_CHANGED` is now changed to `DESCENDANT_TREE_CHANGED` in `PodFlags::upwards`.

I think the naming of the flags are likely still open for discussion (as e.g. `TREE_CHANGED` kind of implies that only a descendant tree might have changed).

I feel like that we should have a separate flag for that, for more fine-grained control instead of reusing the `UPDATE` flag, but this is just a gut feeling...


DJMcNab:
This PR is targetting Xilem classic, which is no longer in this tree.

Sorry this never got reviewed - I never really had the expertise in this Widget set to do so.

PR #116, opened 2023-07-29
Convert the main binary target into an example.

xStrom:
Xilem is a library not an application so it makes sense to structure it as a library with examples. This has always been planned and now this PR does it.

Also I removed `.` from workspace members because [that seems to be automatic](https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package) anyway.


Philipp-M (approved):
Thanks!

PR #117, opened 2023-07-30
Fill out and clean up all the `Cargo.toml` files.

xStrom:
This PR adds a bunch of `Cargo.toml` properties so that they better represent their respective packages. I also utilized the workspace variables for commonly reusable properties.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,11 +8,28 @@ members = [
     "crates/xilem_html/web_examples/todomvc",
 ]
 
+[workspace.package]
+edition = "2021"
+homepage = "https://xilem.dev/"
+repository = "https://github.com/linebender/xilem"
+license = "Apache-2.0"
+
+[workspace.dependencies]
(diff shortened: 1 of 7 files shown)


waywardmonkeys (approved):

jplatte (approved):

Cargo.toml
```diff
@@ -8,11 +8,28 @@ members = [
     "crates/xilem_html/web_examples/todomvc",
 ]
 
+[workspace.package]
+edition = "2021"
+homepage = "https://xilem.dev/"
```
DJMcNab, line 13: This isn't really right - that page isn't useful. It's a shame this is now linked from `Cargo.toml`

xStrom: The page can be updated. If you think another page is more useful for now, you could just set up a simple redirect.

DJMcNab: The conclusion we reached for vello is to just not set the homepage field.
I think that's also right in this case, as we have not made a "real" homepage for Xilem - that page doesn't even seem to link to the repository!

PR #119, opened 2023-08-03
xilem_html: Experiment with Strong typed attributes/event-listeners owned by the concrete element type

Philipp-M:
Prototype of how strong typing attributes of HTML elements could look like, when all kinds of attributes and event listeners are directly owned by an HTML element type, instead of using composition types (see #131 for a different approach).

For more information see the discussion in the [relevant zulip thread](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/xilem_html.3A.20strong.20typing.20.28attributes.20and.20elements.29).

This will very likely never be merged (main reason is necessary `'static` bounds for `T` and `A` for `View<T, A>`), but it may be interesting anyway.


Philipp-M:
Closing in favor of #141

PR #123, opened 2023-08-07
Bump vello version for CFF fixes

dfrg:


Philipp-M (approved):
Still works on Linux, thanks!

PR #124, opened 2023-08-08
xilem_html: Add a `DiffMapIterator` that computes a diff between two ordered maps

Philipp-M:
It is used in the diffing of HTML elements in View::rebuild

This is a slightly refactored/generalized version (for usage with different ordered maps) that was introduced [here](https://github.com/linebender/xilem/pull/89)

### DIFF
--- a/crates/xilem_html/src/diff.rs
+++ b/crates/xilem_html/src/diff.rs
@@ -1,0 +1,61 @@
+use std::{cmp::Ordering, collections::BTreeMap, iter::Peekable};
+
+pub fn diff_tree_maps<'a, K: Ord, V: PartialEq>(
+    prev: &'a BTreeMap<K, V>,
+    next: &'a BTreeMap<K, V>,
+) -> impl Iterator<Item = Diff<&'a K, &'a V>> + 'a {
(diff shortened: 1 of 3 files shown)


derekdreery (approved):
Refactoring this out is cleaner and so I'm majorly in favor. Thanks for the PR! :)

crates/xilem_html/src/diff.rs
```diff
@@ -0,0 +1,61 @@
+use std::{cmp::Ordering, collections::BTreeMap, iter::Peekable};
+
+pub fn diff_tree_maps<'a, K: Ord, V: PartialEq>(
```
derekdreery, line 3: The library doesn't have much (any?) test coverage ATM, but this would be a good candidate for some tests checking the diff sequence. Not a blocker tho - we can add them later.

Philipp-M: Yeah, probably a good idea, I can add them here.

derekdreery:
Nice one on the tests! Feel free to merge yourself - let me know if you don't have rights to.

PR #125, opened 2023-08-08
xilem_html: Add a VecMap which is an ordered Map similar as BTreeMap for small `n`

Philipp-M:
A few benchmarks have showed, that for small n (roughly `n < ~20`, with small types < 50 bytes) a Vec is faster than a BTreeMap.
This seems to be the average case for html attributes.

Since it was a low hanging fruit to boldly copy the docs from the `std::collections::BTreeMap`, I have done so, it also implicitly adds doc tests for this (I guess it makes sense to add `cargo test` to the CI).

The VecMap is now used instead of the BTreeMap for the element attributes

(This is currently a draft *only* because it's built on top of #124 and I can't target-branch this on top of that as then the PR is opened on my repo, otherwise it's fine to merge, so the other PR has to be merged first)

### DIFF
--- a/crates/xilem_html/src/diff.rs
+++ b/crates/xilem_html/src/diff.rs
@@ -1,0 +1,61 @@
+use std::{cmp::Ordering, collections::BTreeMap, iter::Peekable};
+
+pub fn diff_tree_maps<'a, K: Ord, V: PartialEq>(
+    prev: &'a BTreeMap<K, V>,
+    next: &'a BTreeMap<K, V>,
+) -> impl Iterator<Item = Diff<&'a K, &'a V>> + 'a {
+    DiffMapIterator {
+        prev: prev.iter().peekable(),
+        next: next.iter().peekable(),
+    }
+}
+
+/// An iterator that compares two ordered maps (like a `BTreeMap`) and outputs a `Diff` for each added, removed or changed key/value pair)
+struct DiffMapIterator<'a, K: 'a, V: 'a, I: Iterator<Item = (&'a K, &'a V)>> {
+    prev: Peekable<I>,
+    next: Peekable<I>,
+}
+
+impl<'a, K: Ord + 'a, V: PartialEq, I: Iterator<Item = (&'a K, &'a V)>> Iterator
+    for DiffMapIterator<'a, K, V, I>
+{
+    type Item = Diff<&'a K, &'a V>;
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            match (self.prev.peek(), self.next.peek()) {
+                (Some(&(prev_k, prev_v)), Some(&(next_k, next_v))) => match prev_k.cmp(next_k) {
+                    Ordering::Less => {
+                        self.prev.next();
+                        return Some(Diff::Remove(prev_k));
+                    }
+                    Ordering::Greater => {
+                        self.next.next();
+                        return Some(Diff::Add(next_k, next_v));
+                    }
+                    Ordering::Equal => {
+                        self.prev.next();
+                        self.next.next();
+                        if prev_v != next_v {
+                            return Some(Diff::Change(next_k, next_v));
+                        }
+                    }
+                },
+                (Some(&(prev_k, _)), None) => {
+                    self.prev.next();
+                    return Some(Diff::Remove(prev_k));
+                }
+                (None, Some(&(next_k, next_v))) => {
+                    self.next.next();
+                    return Some(Diff::Add(next_k, next_v));
+                }
+                (None, None) => return None,
+            }
+        }
+    }
+}
+
+pub enum Diff<K, V> {
+    Add(K, V),
+    Remove(K),
+    Change(K, V),
+}
--- a/crates/xilem_html/src/diff.rs
+++ b/crates/xilem_html/src/diff.rs
@@ -1,5 +1,6 @@
 use std::{cmp::Ordering, collections::BTreeMap, iter::Peekable};
 
+#[allow(unused)]
 pub fn diff_tree_maps<'a, K: Ord, V: PartialEq>(
     prev: &'a BTreeMap<K, V>,
     next: &'a BTreeMap<K, V>,
@@ -11,9 +12,9 @@ pub fn diff_tree_maps<'a, K: Ord, V: PartialEq>(
 }
 
 /// An iterator that compares two ordered maps (like a `BTreeMap`) and outputs a `Diff` for each added, removed or changed key/value pair)
-struct DiffMapIterator<'a, K: 'a, V: 'a, I: Iterator<Item = (&'a K, &'a V)>> {
-    prev: Peekable<I>,
-    next: Peekable<I>,
+pub(crate) struct DiffMapIterator<'a, K: 'a, V: 'a, I: Iterator<Item = (&'a K, &'a V)>> {
+    pub(crate) prev: Peekable<I>,
+    pub(crate) next: Peekable<I>,
 }
 
 impl<'a, K: Ord + 'a, V: PartialEq, I: Iterator<Item = (&'a K, &'a V)>> Iterator
--- a/crates/xilem_html/src/element/mod.rs
+++ b/crates/xilem_html/src/element/mod.rs
@@ -4,10 +4,11 @@
 //! `use xilem_html::elements as el` or similar to the top of your file.
 use crate::{
     context::{ChangeFlags, Cx},
+    diff::{diff_tree_maps, Diff},
     view::{DomElement, Pod, View, ViewMarker, ViewSequence},
 };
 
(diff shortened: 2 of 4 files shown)


Philipp-M:
I have rebased this, it's ready for review/merge.

One thing that I'm not super happy with, is that doc tests force `pub` for the used types in the crate (currently `xilem_html::vecmap::VecMap` and `xilem_html::diff::Diff`). I don't think this is too bad for the necessary exposed types (and they might even be helpful in other crates).

crates/xilem_html/src/diff.rs
```diff
@@ -1,5 +1,6 @@
 use std::{cmp::Ordering, collections::BTreeMap, iter::Peekable};
 
+#[allow(unused)]
```
waywardmonkeys, line 3: I assume this is maybe unused now because it is no longer used by the other files and only by tests?

Not sure what we should be keeping, exposing, etc.

crates/xilem_html/src/element/mod.rs
```diff
@@ -32,7 +33,7 @@ impl<El, ViewSeq> Element<El, ViewSeq> {
         impl<'a, El, VS> fmt::Debug for DebugFmt<'a, El, VS> {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 write!(f, "<{}", self.0.name)?;
-                for (name, value) in &self.0.attributes {
+                for (name, value) in self.0.attributes.iter() {
```
waywardmonkeys, line 36: Would it be nicer to avoid this particular change and have an `IntoIterator` implement for `VecMap`?

crates/xilem_html/src/element/mod.rs
```diff
@@ -123,7 +124,7 @@ where
 
     fn build(&self, cx: &mut Cx) -> (Id, Self::State, Self::Element) {
         let el = cx.create_html_element(&self.name);
-        for (name, value) in &self.attributes {
+        for (name, value) in self.attributes.iter() {
```
waywardmonkeys, line 127: As above...

crates/xilem_html/src/lib.rs
```diff
@@ -10,10 +10,11 @@ use wasm_bindgen::JsCast;
 mod app;
 mod class;
 mod context;
-mod diff;
+pub mod diff;
 mod element;
 mod event;
 mod one_of;
+pub mod vecmap;
```
waywardmonkeys, line 17: What if you just moved the doc tests for this into a regular set of tests and then we wouldn't expose this API?

waywardmonkeys (commented):
It seems like there's almost an easier thing where the diffing is just operating on peekable iterators, so something that takes 2 parameters that are `IntoIterator` would be possible.

waywardmonkeys:
What if you rename `diff_tree_maps` to `diff_kv_iterables` and then make that function look like this:

```
pub fn diff_tree_maps<'a, II, K, V>(
    prev: II,
    next: II,
) -> impl Iterator<Item = Diff<&'a K, &'a V>> + 'a
where
    K: Ord + 'a,
    V: PartialEq + 'a,
    II: IntoIterator<Item = (&'a K, &'a V)> + 'a,
{
    DiffMapIterator {
        prev: prev.into_iter().peekable(),
        next: next.into_iter().peekable(),
    }
}
```

And then impl `IntoIterator` for `VecMap`?

crates/xilem_html/src/diff.rs
Philipp-M: Yes it is/was, I have refactored it into your suggestion below (`diff_kv_iterables`) and used it in the `VecMap::diff` method.

crates/xilem_html/src/element/mod.rs
Philipp-M: I have thought about it, but since `impl` for associated types are not stabilized yet it's not completely trivial as there's a closure in `VecMap::iter` (if at all possible? Does someone know how that could be done with closures?). I avoided more complex code in the `VecMap::iter()` just to make `IntoIterator` possible (as `vecmap.iter()` is a minor drawback IMO compared to `&vecmap`).

crates/xilem_html/src/element/mod.rs
Philipp-M: See above.

crates/xilem_html/src/lib.rs
Philipp-M: Hmm yes I thought about it, but IMO I think that having the examples directly in the rust function docs, with tests verifying that they work is nice, without possibly dedicated duplicated(?) tests as the exposed functionality isn't really "critical" internals (rather stuff that may have use for other crates as well) IMO (there are ways btw. to [avoid](https://users.rust-lang.org/t/doctest-of-private-functions/60805/5) having to expose this by default with a feature flag specifically for testing, might be worth exploring). But I can of course avoid exposing and extract all this into dedicated tests, anyone else agrees here?

Philipp-M:
> What if you rename `diff_tree_maps` to `diff_kv_iterables` ... And then impl `IntoIterator` for `VecMap`?

I have used the possibility to add a member method to `VecMap` directly (since it's not possible without extension traits for `BTreeMap`). But as you say it certainly makes sense to generalize this function and reuse it for the member method (which is done by the latest commit).

waywardmonkeys:
> I have used the possibility to add a member method to VecMap directly (since it's not possible without extension traits for BTreeMap). But as you say it certainly makes sense to generalize this function and reuse it for the member method (which is done by the latest commit).

I've got a headache and don't feel great, so I might be misunderstanding things ... but my function works with `BTreeMap` being passed in with no extension stuff needed. To pass in a `VecMap`, the only extra thing to implement would be `IntoIterator` for it which should be possible without much pain.

But either way ... I was just looking at how pluggable / extensible it could be so that changing types in the future or using the diff code elsewhere would be easier.

Philipp-M:
With a function pointer (`fn(&'a (K, V)) -> (&'a K, &'a V)`) the `IntoIterator` for `&'a VecMap` is possible (as the closure doesn't catch any other vars), not sure whether the optimizer is smart enough to optimize it similarly as with the `iter()` method (as this is a coercion, but I guess it is).

I've extracted the examples as tests, and removed the pub modifiers, this adds (unfortunately) another 100 LOC (but easy to review, since they're just the doc tests).

waywardmonkeys:
Will review within 8 hours, hopefully.

crates/xilem_html/src/element/mod.rs
```diff
@@ -181,7 +182,7 @@ where
         let element = element.as_element_ref();
 
         // update attributes
-        for itm in diff_tree_maps(&prev.attributes, &self.attributes) {
```
waywardmonkeys, line 184: With everything else, I still really think this change should go away and not have a `diff` method and calling `diff_kv_iterables` here since that makes the implementation much easier to swap and experiment.

crates/xilem_html/src/vecmap.rs
```diff
@@ -0,0 +1,382 @@
...
+    #[inline]
+    fn index(&self, key: &Q) -> &V {
+        self.get(key).expect("no entry found for key")
+    }
+}
+
+impl<'a, K, V> IntoIterator for &'a VecMap<K, V> {
+    type Item = (&'a K, &'a V);
+
+    type IntoIter = std::iter::Map<std::slice::Iter<'a, (K, V)>, fn(&'a (K, V)) -> (&'a K, &'a V)>;
```
waywardmonkeys, line 275: I think this can be nicer if you have define a struct `Iter` that holds a `slice::Iter` or something like that and then impl the right traits for that `Iter` ... `Iterator`, `DoubleEndedIterator`, `ExactSizeIterator` ...

But maybe this isn't needed for now since it is all an experiment and can be improved upon later if it works out ...

waywardmonkeys (approved):
I'm marking it as approved even though I think it might be worth iterating a bit further, but as I say in one of the comments ... this is all experimental code anyway, so it needn't be perfect yet.

crates/xilem_html/src/element/mod.rs
Philipp-M: Yeah I *mostly* agree, I have updated it (removed the extra method for the `VecMap`), makes the codesize a slightly shorter too. I've put the doc comment of the previous `VecMap::diff` method above the `diff_kv_iterables` instead.

crates/xilem_html/src/vecmap.rs
Philipp-M: Yeah, sure this could certainly be improved (implement more traits etc.). If this really ends up being used more, it certainly makes sense to make this a little bit more advanced/add features (as you say with a wrapper type). But I think we should do that when we actually really need it (this rather simple addition already introduces quite a bit of LOCs).

PR #127, opened 2023-08-08
xilem_html: Add support for different types in el.attr("attr", <value>) 

Philipp-M:
Where \<value\> can be numbers, booleans and strings and options of values.
This is achieved via an enum that serializes to a string and the From trait.
if the attribute is either set to `None` or `false`, the attribute won't be serialized at all (or added to the attributes map at that).

This is slightly less ambitious (and controversial?) than #119 and is used as basis for all attributes there (slightly different implementation, this PR is an improved version without the `BTreeSet` that is used there for classes). 

(This is also ready to merge/review but is built on top of #124 and #125, so those have to be merged/reviewed first).

### DIFF
--- a/crates/xilem_html/src/element/attributes.rs
+++ b/crates/xilem_html/src/element/attributes.rs
@@ -6,7 +6,6 @@ type CowStr = Cow<'static, str>;
 
 #[derive(PartialEq, Debug)]
 pub enum AttributeValue {
-    Null,
     True, // for the boolean true, this serializes to an empty string (e.g. for <input checked>)
     I32(i32),
     U32(u32),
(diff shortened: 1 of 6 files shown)


crates/xilem_html/src/element/attributes.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl From<&'static str> for AttributeValue {
+    fn from(value: &'static str) -> Self {
+        AttributeValue::String(value.into())
+    }
+}
+
+#[derive(Default)]
+pub struct Attributes(VecMap<CowStr, AttributeValue>);
```
jplatte, line 99: If `Null` is never held here, does it really make sense to be part of the enum? I think it makes sense for the invariant to be enforced properly, though this probably means having to use a custom `ToAttributeValue` trait if we want `insert` to be able to unset an attribute value they we it can now.

crates/xilem_html/src/element/mod.rs
```diff
@@ -34,7 +38,7 @@ impl<El, ViewSeq> Element<El, ViewSeq> {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 write!(f, "<{}", self.0.name)?;
                 for (name, value) in &self.0.attributes {
-                    write!(f, " {name}=\"{value}\"")?;
+                    write!(f, " {name}=\"{}\"", value.serialize())?;
```
jplatte, line 41: This piece of code relies on the invariant that `self.0.attributes` never contains `Null`.

crates/xilem_html/src/element/attributes.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.0.into_iter()
+    }
+}
+
+impl Attributes {
+    // TODO return the previous attribute as an Option?
+    pub fn insert(&mut self, name: impl Into<CowStr>, value: impl IntoAttributeValue) {
+        let value = value.into_attribute_value();
+        // This is a simple optimization in case this is the first attribute inserted to the map (saves an allocation for the Vec)
+        if let Some(value) = value.into_attribute_value() {
```
jplatte, line 116: Double `.into_attribute_value()`. Also, is the comment still accurate?

Finally, how about providing `.remove(name)` as a more obvious alternative to `.insert(name, false)` (which I think makes a lot of sense to work, though).

Philipp-M: Whoops, I guess I was too fast pushing and overlooked that, thanks for catching this...

Yeah without the `Null`, a remove makes more sense probably, and handling that a layer above.
I think the whole wrapper type (`Attributes`) doesn't make much sense anymore (it was a wrapper around `VecMap` precisely because of guarding the potential `Null` value that would've been inserted).

I removed the wrapper type and renamed the file to `attribute_value.rs` as well, as it seems more appropriate in that case.

waywardmonkeys (approved):
This looks like a good improvement to me!

crates/xilem_html/src/element/mod.rs
```diff
@@ -89,12 +87,13 @@ impl<El, ViewSeq> Element<El, ViewSeq> {
     ///
     /// If the name contains characters that are not valid in an attribute name,
     /// then the `View::build`/`View::rebuild` functions will panic for this view.
-    pub fn set_attr(
-        &mut self,
-        name: impl Into<Cow<'static, str>>,
-        value: impl Into<Cow<'static, str>>,
-    ) {
-        self.attributes.insert(name.into(), value.into());
+    pub fn set_attr(&mut self, name: impl Into<CowStr>, value: impl IntoAttributeValue) {
```
jplatte, line 90: Right, so given the new structure of the code, here an `unset_attr` or `remove_attr` could make sense. It doesn't seem terribly important though.

jplatte (approved):

crates/xilem_html/src/element/mod.rs
Philipp-M: The automatic removal is intentional when e.g. doing something like el.attr("checked", false) (or the switch from `true` to `false`), so that the `"checked"` attribute will be removed properly from the element.
Though explicit attribute removal makes sense as well I guess (to avoid something like `el.set_attr(name, false)` which doesn't immediately tell that it's just for removal of the attribute).

I've added that with the latest commit.

Thanks for the review!

PR #128, opened 2023-08-11
Add 'cache-all-crates' to avoid cache miss in CI

Philipp-M:
Currently it seems that the web based crates (I think `xilem_svg`, `xilem_html` and their examples) aren't cached in CI, I think it's because they are not referenced in the main crate, I have added `cache-all-crates: true` which seems to fix this issue.


Philipp-M:
Ok, I have no idea what was previously going wrong, but caching seems to work now without this change (I don't even think this PR would've fixed the issue, but it worked on my fork, so I thought it did). I'm going to close this, as I don't see additional value of this right now.

PR #130, opened 2023-08-16
Add use_state View to xilem-core

matthunz:
Add  the generic `use_state` view to xilem-core.

```rust
fn use_state<F, S, G, V>(
    make_state: F,
    make_view: G
) -> UseState<F, S, G, V>
```

This would generalize and replace the current `UseState` view in the main xilem crate. The main benefit here is no `Arc` to hold the current state. This is still a WIP and is being ported from the [`remember`](https://concoct-rs.github.io/concoct/view/struct.Remember.html) view in my UI crate.

I think this has huge benefits as it would allow us to compose state at no runtime cost. For example, global app state could be replaced by small sections of state created with this view.

### DIFF
--- a/crates/xilem_core/src/view/use_state.rs
+++ b/crates/xilem_core/src/view/use_state.rs
@@ -1,0 +1,62 @@
+#[macro_export]
+macro_rules! generate_use_state_view {
+    ($viewtrait:ident, $cx:ty, $changeflags:ty; $($ss:tt)*) => {
+        pub fn use_state<F, S, G, V>(make_state: F, make_view: G) -> UseState<F, S, G, V> {
+            UseState::new(make_state, make_view)
+        }
+
+        pub struct UseState<F, S, G, V> {
+            make_state: F,
+            state: Option<S>,
+            make_view: G,
+            view: Option<V>,
+        }
+
+        impl<F, S, G, V> UseState<F, S, G, V> {
+            pub fn new(make_state: F, make_view: G) -> Self {
+                Self {
+                    make_state,
+                    state: None,
+                    make_view,
+                    view: None,
+                }
+            }
+        }
+
+        impl<T, A, F, S, G, V> View<T, A> for UseState<F, S, G, V>
+        where
+            S: Send,
+            F: FnMut() -> S + Send,
+            G: FnMut(&mut S) -> V + Send,
+            V: $viewtrait<S, A>,
+        {
+            type State = V::State;
+            type Element = V::Element;
+
+            fn build(&self, cx: &mut $cx) -> (Id, Self::State, Self::Element) {
+                todo!()
+            }
+
+            fn rebuild(
+                &self,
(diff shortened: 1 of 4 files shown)


crates/xilem_core/src/view/use_state.rs
```diff
@@ -0,0 +1,62 @@
...
+            S: Send,
+            F: FnMut() -> S + Send,
+            G: FnMut(&mut S) -> V + Send,
+            V: $viewtrait<S, A>,
+        {
+            type State = V::State;
+            type Element = V::Element;
+
+            fn build(&self, cx: &mut $cx) -> (Id, Self::State, Self::Element) {
+                todo!()
```
raphlinus, line 37: This seems rather important.

raphlinus (commented):
I've taken a quick look. I'm not 100% sure I understand what's going on here, partly because it's incomplete.

The original use_state implementation uses Arc for a reason: it does smart change propagation using the same techniques as Druid. In particular, in event (message) propagation, it offers the child view exclusive mutable access to a copy of the app state in T, then only updates the parent state if the child has changed it. In other words, it preserves pointer equality of the state, which is a primary mechanism to drive incremental computation.

It's not clear whether this approach would support that. It's also not 100% clear that this approach to use_state is what we want to do. It was added to the Xilem prototype speculatively, largely because people were asking after it, and because it's very important in the React way of doing things. Yet, in the Elm way of doing things, people seem to get by just fine without it.

matthunz:
Huh sorry for the confusion!  I wasn't sure why the other implementation required `Arc` but that makes sense for the pointer eq.

The `remember` view in my crate just served the purpose of adding state in views so I didn't need global state. Xilem's `use_state` seems really interesting for change propagation though so I can see why we'd want it.

For this PR then I feel like I could either try to bring the old version of `use_state` into core or rename this `remember` and port the version from my crate. I do still think having remembering state, without the added change propagation, would be useful

crates/xilem_core/src/view/use_state.rs
matthunz: Lol agree with ya there... Just wanted to make sure I was heading in the right direction. Definitely looks like there are a few directions to go here.

If you're curious this is basically the method implementation I had in mind: [remember](https://github.com/concoct-rs/concoct/blob/900f505718d8551ef277d984d7d0b21e320bdde7/src/view/remember.rs#L33)
Here I use the second function argument to create a view with the state from `Remember`. The first function would create the state only once

Philipp-M:
Just that it's not overlooked, I have fixed the original `UseState` some time ago: https://github.com/linebender/xilem/pull/98

It's not yet ported to core in that PR, but should be easy to do so, either in that PR, or IMHO preferably (because of the currently smaller diff) in a new PR. The PR is basically mergeable I think, just needs a rubber stamp review I guess.

matthunz:
Closing in favor of #98

PR #132, opened 2023-08-24
Switch Widget Port

giannissc:


src/view/switch.rs
```diff
@@ -0,0 +1,74 @@
...
+use crate::view::ViewMarker;
+use crate::{view::Id, widget::ChangeFlags, MessageResult};
+
+use super::{Cx, View};
+
+pub struct Switch {
+    is_on: bool,
+}
+
+pub fn switch(is_on: bool) -> Switch {
```
Philipp-M, line 26: I think this should probably have a callback parameter as well, for when the value is changed, similar as in the button example, but since there's a payload (`bool`), it should probably be an `impl Fn(&mut T, bool) -> A` or something like that.

So that it looks as something like this:

```rust
fn my_view(state: &mut AppState) -> impl View<AppState> {
    switch(state.my_switch_value, |state, value| { state.my_switch_value = value; })
}
```

But it would be interesting to "sugar" this callback boilerplate away somehow,
as I think something like the callback above is probably the main use-case, though I think often something additionally should happen when the value is changed as well.

For the simple case above that can probably be done via a macro? But I guess something like that should probably be done on top of this PR in a different PR...
The "UX design" of views is still a relatively big question mark so far (needs more experimentation and discussion to find the most idiomatic style IMO)
(I think SwiftUI is probably the best source of inspiration for most of this stuff)

src/view/switch.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl Switch {
+    pub fn new(is_on: bool) -> Self {
+        Switch { is_on }
+    }
+}
+
+impl ViewMarker for Switch {}
+
+impl<A> View<bool, A> for Switch {
```
Philipp-M, line 38: I think `bool` should actually be a generic parameter `T`, otherwise it's necessary to "`Adapt`" this everywhere, where it's used, unless the central app state is really just a `bool` :)

src/view/switch.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    fn message(
+        &self,
+        _id_path: &[Id],
+        _state: &mut Self::State,
+        _message: Box<dyn Any>,
+        app_state: &mut bool,
+    ) -> MessageResult<A> {
+        *app_state = !*app_state;
+        MessageResult::Nop
```
Philipp-M, line 72: See above, I think this should invocate the callback instead of directly acting on the `app_state` (and then return a `MessageResult::Action`).

Philipp-M (commented):
Left a few comments.

I think something like an additional example showing this in action would be awesome (I think we should generally start with adding more examples, to get a feeling of the UX etc. and examples are good for testing as well)

src/view/switch.rs
giannissc: I figured that the callback was necessary for the button as you don't know what the intention for it is. The function of the toggle switch though is much simpler, it just needs to reverse the previous state. In there a situation that a callback would be useful?

giannissc: Question: When you add the callback to the `MessageResult` does it get executed in `message()` and the resolved value is passed up the tree using the action or do you pass the callback. If the callback is the payload of the action who executes the callback?

src/view/switch.rs
giannissc: Is the purpose of the callback function to be able to access specific fields from the `AppData`?

Philipp-M: Yes kind of.
It's mostly for mutating the actual app state, without restricting this to only accept `bool` as `AppState`. I think it's a slightly nicer form than to wrap it in an `Adapt` node, and "hiding" the mutation inside the `View` impl.

TMU the actual logic of mutating the `AppState` should reside in the `app_logic` function, so that it's obvious what happens. It's also a little bit more flexible, as you can "react" to the event directly in the callback, instead of "guessing" what has changed (in the next update). Not that it would make any realworld sense, but you could do something like this:

```rust
fn my_view(state: &mut AppState) -> impl View<AppState> {
    switch(state.my_switch_value, |state, value| {
        state.count += 1;
        if state.count > 3 {
            state.my_switch_value = value;
        }
    })
}
```

src/view/switch.rs
Philipp-M: The resolved value will just be given in the callback which then actually modifies the app state (the button example could be seen as reference here).

"Optionally" the callback can return `A` (for Action). But I feel that this will likely be used rarely (i.e. `A` will be `()`) in the actual `app_logic` (IMHO more cognitive overhead than just a callback that modifies the app state).

It's executed by passing messages currently in the widget context(s) in e.g. `Widget::event` or `Widget::update`, that in turn propagate to the view via `View::message()` using the id path that then can mutate the app state via the callback.

There are other use cases though for a `MessageResult::Action`, one being async mutations outside of the immediate `app_logic` function (and something I'd like to continue to investigate in more detail, as I feel this will be quite relevant in bigger applications).

If you're interested, I have experimented a little bit here: https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/trui. Though I would approach this currently using the `MessageResult` directly (elm like enum, to be more composable, similar as in the todomvc example).

giannissc:
This was a good exercise! I think I understand a bit more the Xilem model. Thank for the comments they were really helpful :)

It's ready for another review! @Philipp-M

src/view/switch.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn message(
+        &self,
+        _id_path: &[Id],
+        _state: &mut Self::State,
+        _message: Box<dyn Any>,
+        app_state: &mut T,
+    ) -> MessageResult<A> {
+        let bool_state: &mut bool = (self.callback)(app_state);
+        *bool_state = !*bool_state;
+        MessageResult::Nop
```
Philipp-M, line 81: I would just do something like this (like in the button view):

```suggestion
        MessageResult::Action((self.callback)(app_state, !self.is_on))
```

The main reason why I think, modifying the state directly in the callback and returning a potential `A` (instead of `&mut bool`), is that it supports `MessageResult::Action` (checkout the `main_view` function in the todomvc example in xilem_html, specifically the use of `Adapt` there).
`MessageResult::Nop` doesn't seem to be right here for me (There's happening something, i.e. the switch has been turned).
Also it would support (not that it's particularly useful) to inhibit/change the default behavior.

But the idea you're introducing here with selecting/returning the mutable reference of the boolean state inside the callback is certainly interesting though.

Philipp-M (changes requested):
Just tried the example, looks nice :)

giannissc:
Just rebased and force pushed with the new changes @Philipp-M

src/widget/switch.rs
```diff
@@ -0,0 +1,165 @@
...
+    contexts::LifeCycleCx,
+    piet_scene_helpers::{fill_color, stroke},
+    AccessCx, BoxConstraints, ChangeFlags, Event, EventCx, LayoutCx, LifeCycle, PaintCx, UpdateCx,
+    Widget,
+};
+
+pub struct Switch {
+    id_path: IdPath,
+    is_on: bool,
+    is_moved: bool,
```
Philipp-M, line 34: (Slightly nit picky) I would name this something like `is_dragging` as I think it's more appropriate...

src/widget/switch.rs
```diff
@@ -0,0 +1,165 @@
...
+            Event::MouseDown(_) => {
+                cx.set_active(true);
+                cx.request_paint();
+            }
+            Event::MouseUp(_) => {
+                if self.is_moved {
+                    if self.is_on != (self.knob_position.x > SWITCH_WIDTH / 2.0) {
+                        cx.add_message(Message::new(self.id_path.clone(), ()))
+                    }
+                } else if cx.is_active() {
```
Philipp-M, line 78: (Also super nit-picky, because it shouldn't really change the behavior, as a `MouseMove` event should happen before)
```suggestion
                } else if cx.is_active() && cx.is_hot() {
```

Philipp-M (commented):
Sorry for the delay, haven't found the time till now...

Some nit-picks, not really that important.

I think this is good to merge when it's rebased (for the new CI workflows) and the CI is fixed (I think you can just disable clippy for that line), thanks!

I found a bug orthogonal to this PR while testing this, maybe I'll look into this (dragging the knob or pressing a button in the `h_stack` and then moving up to the top button and releasing the cursor)

Philipp-M:
Is this supposed to be closed?

giannissc:
I know what happened here.. It was before I started using git extensively and I had all these into main instead of it's own branch so when synchronized my main to master I discarded my commits for the switch widget..

jplatte:
They are still recoverable though, GitHub keeps everything that's pushed to a PR: https://github.com/linebender/xilem/commit/6275db82c73f14dce5284d7bdfb1507cdbb747f8 (parent commit are also available, and clickable in the web UI)

giannissc:
That's ok it was easy enough to start fresh. But I will keep this in mind next time this happens. Thanks!

Philipp-M:
I'm not sure though, whether it's possible to change the source branch (see https://stackoverflow.com/questions/42381557/changing-source-branch-for-pr-on-github)

So either you'll cherry-pick all the relevant commits onto your main branch and reopen this PR (after merging, you can of course sync with upstream main again), or you have to create a new PR with a different source branch.

src/widget/switch.rs
giannissc: I am replying here even though I will be creating a new PR. The is_hot check I think is wrong here and the only reason it doesn't affect the behaviour is because no actual message is being send here. Had there been any message being sent to the view it would not have if the user moved past the boundaries of the widget. Good UX would be to ensure things function correctly even in light of errors from the user

jplatte:
It's not possible to reopen a PR after the branch that was used to create it is force-pushed while the PR is closed.

src/widget/switch.rs
Philipp-M: Hmm I guess it really depends. A button is not clicked when the user releases it outside the button area. On touchscreens I agree with you though, or generally when the switch is dragged, but this should never be the case here, since `is_dragging` will be set before, so it doesn't really matter...

Philipp-M:
Oh really, interesting, thanks for letting me know.

PR #133, opened 2023-08-28
ProgressBar Widget Port

giannissc:


src/view/progress_bar.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    fn message(
+        &self,
+        _id_path: &[Id],
+        _state: &mut Self::State,
+        _message: Box<dyn Any>,
+        _app_state: &mut T,
+    ) -> MessageResult<A> {
+        MessageResult::Nop
```
Philipp-M, line 75: Although I agree with Raph, that it should be a logic error when a message reaches this, `Stale` may be more appropriate here (it's also used in other views/widgets that don't emit messages/events)? Maybe a new `MessageResult::Error` variant makes sense?
I could also imagine a panic here in debug builds at least.

giannissc:
Should this be merged? I rebased on top of the master @Philipp-M

Philipp-M:
Haven't had yet time to review the progress bar in detail, it looks like this contains the Switch PR (#132) as well?

I think it would be easier to split the Switch into its own PR (any reason why #132 was closed?).

I can quickly approve #132 (or similar) (would be cool to address the remaining review comments, but they're just nit-picks, so not that important).

giannissc:
I am not sure what happened there.. I thought it got submitted and that why I couldn't find it. I will revive it and rebase both and address the comments so that they can both be submitted.

Philipp-M:
I think, since you have used the main branch for that PR, and you have synced with the current linebender main branch, that since there weren't any changes in the PR anymore, that the PR has maybe closed itself? I'm not sure though, it hasn't happened to me yet (but I'm not sure if I ever had such a case).

PoignardAzur:
This PR was made for Xilem classic and is now superseded by #513.

I'm closing this.

PR #135, opened 2023-09-11
Migrate to the generic Linebender CI script.

xStrom:
Having a unified CI script across projects reduces the maintenance burden and allows every project to quickly benefit from lessons learned across all the projects.

The new script in this PR is basically a copy of [Glazier's current CI script](https://github.com/linebender/glazier/blob/19a01ba8a97f15bf499ac64e6ea9690d7390deb5/.github/workflows/ci.yml). It has been designed with the explicit goal of being widely applicable across Linebender projects. It has worked well for Glazier and now is the time to also bring it to Xilem.

Once merged, this will also open up the path to enabling merge queues for Xilem.


waywardmonkeys (changes requested):
Can't look at all of this in detail right now, but you'll have to update the README badge for the build / actions since it is getting renamed here. (I like the renaming.)

waywardmonkeys (changes requested):
Also, `actions/checkout@v4` is out now: https://github.com/actions/checkout/releases/tag/v4.0.0

xStrom:
Fixed `README.md` and upgraded `actions/checkout`. Ready for another look.

waywardmonkeys (approved):

PR #139, opened 2023-10-26
Dust off xilem_svg

raphlinus:
This patch updates the xilem_svg crate to make it suitable for tinkering. The demo app is moved to a separate crate, rather than being hardwired, and the `ViewExt` trait is wired up properly (replacing the previous `KurboShape` hack).

### DIFF
--- a/crates/xilem_svg/src/common_attrs.rs
+++ b/crates/xilem_svg/src/common_attrs.rs
@@ -13,14 +13,12 @@ use crate::{
 
 pub struct Fill<T, V> {
     child: V,
-    // This could reasonably be static Cow also, but keep things simple
     brush: Brush,
     phantom: PhantomData<T>,
 }
 
 pub struct Stroke<T, V> {
     child: V,
-    // This could reasonably be static Cow also, but keep things simple
     brush: Brush,
     style: peniko::kurbo::Stroke,
     phantom: PhantomData<T>,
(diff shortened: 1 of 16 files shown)


DJMcNab:
I think the readme of `xilem_svg` should be updated. `web_examples` probably also needs a readme - probably mostly copying the xilem_svg readme

raphlinus:
Agree on the readme change. In fact, I did a small update but it seems like that got lost (probably a `git restore` when I was struggling with the ViewExt changes). I'll redo that, and also add one in web_examples as well.

crates/xilem_svg/src/common_attrs.rs
```diff
@@ -0,0 +1,165 @@
...
+        id: &mut Id,
+        state: &mut Self::State,
+        element: &mut V::Element,
+    ) -> ChangeFlags {
+        let prev_id = *id;
+        let mut changed = self.child.rebuild(cx, &prev.child, id, state, element);
+        if self.brush != prev.brush || prev_id != *id {
+            element
+                .as_element_ref()
+                .set_attribute("fill", &brush_to_string(&self.brush))
```
Philipp-M, line 143: I think this is supposed to be `"stroke"` isn't it?

crates/xilem_svg/src/common_attrs.rs
```diff
@@ -0,0 +1,165 @@
...
+use xilem_core::{Id, MessageResult};
+
+use crate::{
+    context::{ChangeFlags, Cx},
+    view::{DomElement, View, ViewMarker},
+};
+
+pub struct Fill<T, V> {
+    child: V,
+    // This could reasonably be static Cow also, but keep things simple
```
Philipp-M, line 16: Not sure if this comment makes sense here (did it result from copy/paste from the `Class` view?) (Same for the `Stroke` below)

Philipp-M (approved):
Took another look, looks good.

PR #143, opened 2023-11-08
Enable `Text` widget

nicoburns:
There was already a text widget in the repository. This PR:

- Enables the text view and widget modules
- Fixes them up so they compile with latest Xilem
- Implements compute_max_intrinsic for text

I'm not sure that having `String` for the view makes sense because:
- Strings are not string literals, so the DX isn't that great anyway as you have to wrap in `String::from`, which is longer than some hypothetical `text` function for a custom text type.
- We're almost certainly going to want modifiers on text like font size, color, etc. And we'll need to store those somewhere.

But I've left it as `String` for now.

### DIFF
--- a/src/view/text.rs
+++ b/src/view/text.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Cow;
+
 use crate::view::{Id, ViewMarker};
 use crate::widget::ChangeFlags;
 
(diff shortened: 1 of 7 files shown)


Philipp-M:
> Strings are not string literals, so the DX isn't that great anyway as you have to wrap in String::from, which is longer than some hypothetical text function for a custom text type.

We'll likely need all kinds `View` impls for string based views (&str, Cow\<str> etc.), and I think they may share all the same widget, and maybe even wrap the same text view.

> We're almost certainly going to want modifiers on text like font size, color, etc. And we'll need to store those somewhere.

I think a few composing style views probably make sense for such views (and maybe in general). Something like `StyledText<String>` or `StyledText<&'static str>`.

I'll try to review this later (or in the following few days).

src/widget/text.rs
```diff
@@ -12,44 +12,81 @@
...
     pub fn set_text(&mut self, text: String) -> ChangeFlags {
         self.text = text;
         ChangeFlags::LAYOUT | ChangeFlags::PAINT
     }
+
+    fn get_layout_mut(&mut self, font_cx: &mut FontContext) -> &mut Layout<ParleyBrush> {
+        // Ensure Parley layout is initialised
+        if self.layout.is_none() {
+            let mut lcx = parley::LayoutContext::new();
+            let mut layout_builder = lcx.ranged_builder(font_cx, self.text.trim(), 1.0);
```
Philipp-M, line 48: Not sure if we want to trim strings, having spaces at the beginning or the end of a string may be intended.

Philipp-M (approved):
Apart from the one comment, this looks good to me, thanks, finally support for strings :)

src/view/text.rs
```diff
@@ -43,13 +44,13 @@ impl<T, A> View<T, A> for String {
...
-        _id_path: &[crate::id::Id],
+        _id_path: &[xilem_core::Id],
         _state: &mut Self::State,
-        _event: Box<dyn Any>,
+        _message: Box<dyn std::any::Any>,
         _app_state: &mut T,
-    ) -> MessageResult<A> {
-        MessageResult::Stale
+    ) -> xilem_core::MessageResult<A> {
+        xilem_core::MessageResult::Nop
```
Philipp-M, line 54: One more (nit-picky) thing, this should be `Stale`, as events should not reach this, `Nop` should AFAIK only be used, when this view *expects* an event.

Philipp-M: Oh, just noticed, that I have copied that in my string extension PR as well, so that applies to the other String View impls too I guess...

nicoburns: Not sure I really understand why this is, but I shall take your word for it (+ I can see other code doing this).

PR #146, opened 2023-11-17
Implement `scroll_view` widget

nicoburns:
~Builds on top of https://github.com/linebender/xilem/pull/140 which should be merged first.~ I've now made this PR independent of #140

## Notes

- I'm not very confident about the approach of changing the origin or clipping the painting.
- I had to change the `set_origin` method take `&mut WidgetState` rather than `&mut LayoutCx`. Because otherwise I couldn't call it from `event` and had to request layout every time the scroll position changed.

## Video

https://github.com/linebender/xilem/assets/1007307/dfc921fc-daf7-4010-87e6-81fe3a2e7d7f

### DIFF
--- a/src/view/scroll_view.rs
+++ b/src/view/scroll_view.rs
@@ -14,9 +14,11 @@
 
 use std::{any::Any, marker::PhantomData};
 
-use crate::{event::EventResult, id::Id, View};
+use crate::{view::View, widget::ChangeFlags, MessageResult};
 
-use super::Cx;
+use xilem_core::Id;
+
+use super::{Cx, ViewMarker, ViewSequence};
 
 pub struct ScrollView<T, A, C> {
     child: C,
@@ -36,18 +38,20 @@ impl<T, A, C> ScrollView<T, A, C> {
     }
 }
 
+impl<T, A, VT: ViewSequence<T, A>> ViewMarker for ScrollView<T, A, VT> {}
+
 impl<T, A, C: View<T, A>> View<T, A> for ScrollView<T, A, C>
 where
     C::Element: 'static,
 {
     type State = (Id, C::State);
 
-    type Element = crate::widget::scroll_view::ScrollView;
+    type Element = crate::widget::ScrollView;
 
     fn build(&self, cx: &mut Cx) -> (Id, Self::State, Self::Element) {
         let (id, (child_id, child_state, child_element)) =
             cx.with_new_id(|cx| self.child.build(cx));
-        let element = crate::widget::scroll_view::ScrollView::new(child_element);
+        let element = crate::widget::ScrollView::new(child_element);
         (id, (child_id, child_state), element)
     }
 
@@ -58,27 +62,22 @@ where
         id: &mut Id,
         state: &mut Self::State,
         element: &mut Self::Element,
-    ) -> bool {
+    ) -> ChangeFlags {
         cx.with_id(*id, |cx| {
             let child_element = element.child_mut().downcast_mut().unwrap();
-            let changed =
-                self.child
-                    .rebuild(cx, &prev.child, &mut state.0, &mut state.1, child_element);
-            if changed {
-                element.child_mut().request_update();
-            }
-            changed
+            self.child
+                .rebuild(cx, &prev.child, &mut state.0, &mut state.1, child_element)
         })
     }
 
-    fn event(
+    fn message(
         &self,
         id_path: &[Id],
         state: &mut Self::State,
         event: Box<dyn Any>,
         app_state: &mut T,
-    ) -> EventResult<A> {
+    ) -> MessageResult<A> {
         let tl = &id_path[1..];
-        self.child.event(tl, &mut state.1, event, app_state)
+        self.child.message(tl, &mut state.1, event, app_state)
     }
 }
--- a/src/view/scroll_view.rs
+++ b/src/view/scroll_view.rs
@@ -74,10 +74,15 @@ where
         &self,
         id_path: &[Id],
         state: &mut Self::State,
-        event: Box<dyn Any>,
+        message: Box<dyn Any>,
         app_state: &mut T,
     ) -> MessageResult<A> {
-        let tl = &id_path[1..];
-        self.child.message(tl, &mut state.1, event, app_state)
+        match id_path {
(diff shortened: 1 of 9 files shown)


src/view/scroll_view.rs
```diff
@@ -58,27 +62,22 @@ where
...
+                .rebuild(cx, &prev.child, &mut state.0, &mut state.1, child_element)
         })
     }
 
-    fn event(
+    fn message(
         &self,
         id_path: &[Id],
         state: &mut Self::State,
         event: Box<dyn Any>,
```
Philipp-M, line 77:
```suggestion
        message: Box<dyn Any>,
```

src/view/scroll_view.rs
```diff
@@ -58,27 +62,22 @@ where
...
         app_state: &mut T,
-    ) -> EventResult<A> {
+    ) -> MessageResult<A> {
         let tl = &id_path[1..];
```
Philipp-M, line 80:
```suggestion
```

src/view/scroll_view.rs
```diff
@@ -58,27 +62,22 @@ where
...
-        self.child.event(tl, &mut state.1, event, app_state)
+        self.child.message(tl, &mut state.1, event, app_state)
```
Philipp-M, line 81: I know this is existing code, but I guess it makes sense to make this safer at this point.
(not sure how I can do a multiline suggestion across additions and existing code)

```suggestion
        match id_path {
            [child_id, rest_path @ ..] if *child_id == state.0 => {
                self.child
                    .message(rest_path, &mut state.1, message, app_state)
            }
            _ => MessageResult::Stale(message),
        }
```

src/widget/core.rs
```diff
@@ -94,7 +95,7 @@ pub struct Pod {
 }
 
 #[derive(Debug)]
-pub(crate) struct WidgetState {
+pub struct WidgetState {
```
Philipp-M, line 98: Out of curiosity, why this change?

src/widget/core.rs
```diff
@@ -467,18 +486,23 @@ impl Pod {
...
-            cx.request_paint();
+            parent_state.flags |= PodFlags::REQUEST_PAINT;
+            parent_state.flags |= PodFlags::VIEW_CONTEXT_CHANGED;
 
             self.state.flags.insert(PodFlags::VIEW_CONTEXT_CHANGED);
         }
     }
 
+    /// Get the widgets size (as returned by the layout method)
+    pub fn get_size(&mut self) -> Size {
```
Philipp-M, line 502: I think we should omit `get_` prefixes to stay consistent 
```suggestion
    pub fn size(&mut self) -> Size {
```

src/widget/scroll_view.rs
```diff
@@ -47,41 +44,26 @@ impl ScrollView {
...
-        self.child.event(cx, &child_event);
+    fn event(&mut self, cx: &mut EventCx, event: &Event) {
+        // Pass event through to child
+        self.child.event(cx, event);
+
+        // Handle scroll wheel events
         if !cx.is_handled() {
-            if let RawEvent::MouseWheel(mouse) = event {
-                let new_offset = (self.offset + mouse.wheel_delta.y).max(0.0);
+            if let Event::MouseWheel(mouse) = event {
```
Philipp-M, line 55: Out of a UX perspective it would be interesting (not entirely sure how this is done in web, but I think it's similar as here?), how to handle nested scroll-views.

When I'm getting this right, the nested scroll view handles the scroll event, unless it doesn't scroll (e.g. is at the end).

Just thinking out loud, not suggesting anything.

Philipp-M (commented):
Thanks :)

I think it generally makes sense to add small examples showing the widgets in action (I personally like self-contained examples showing each widgets capabilities on its own, for this maybe a lorem ipsum text in a scroll view), but not a requirement.

I'm not (yet) deep in layouting and vello, so I may be wrong here: Wouldn't it make sense, instead of setting the child origin, to use the vello layer with a transform that has the offset instead?

src/view/scroll_view.rs
nicoburns: Could you explain the difference between "message" and "event" to me? I'm very surprised that there are two functions with identical signatures that seemingly do a similar thing!

src/widget/core.rs
nicoburns: This is to allow `set_origin` to be called with multiple different context types. As `set_origin` is public, and now takes `WidgetState` as a parameter, `WidgetState` also needs to be public.

src/widget/scroll_view.rs
nicoburns: I believe that ideally it would be a little more sophisticated than that, and that the inner view could handle part of a scroll event (some of the pixel delta) and leave the rest for an outer scroll view (perhaps by making the event mutable?). But you suggestion would probably cover the basics.

nicoburns:
> I'm not (yet) deep in layouting and vello, so I may be wrong here: Wouldn't it make sense, instead of setting the child origin, to use the vello layer with a transform that has the offset instead?

I believe that would draw the widget in the correct place, but would not allow input to work correctly. Whereas set_origin applies to both.

src/view/scroll_view.rs
Philipp-M: I think originally (in the druid [xilem_tokio](https://github.com/linebender/druid/tree/xilem_tokio) branch) it was called event (and was put unmodified into this repo), and was renamed later to message, I'm not entirely sure (Raph probably has a better answer), just a naming thing.

I think it's because `message` kind of implies that it "messages" a general payload, while event doesn't necessarily.

Philipp-M:
> I believe that would draw the widget in the correct place, but would not allow input to work correctly. Whereas set_origin applies to both.

That's a good point, might be interesting to research something like virtualized pointer events or similar that can be modified while traversing down the tree (in `Pod::event`).
Hah just saw in that moment, that was done previously like that^^, what speaks against this approach for now?

derekdreery (approved):
Awesome, thanks.

DJMcNab:
@nicoburns are you expecting to revive this branch? It seems to be on the Xilem repo and not deleted.

nicoburns:
No. #155 was merged instead

DJMcNab:
I'll delete the branch then

PR #148, opened 2023-11-21
Correctly pass the `HAS_ACTIVE` state up

DJMcNab:
Previously, this was being overwritten by each widget as the event went up the tree, so only the root widget could be active

This aligns this code with the druid version (see https://github.com/linebender/druid/blob/e53a5ab72c40191b3f92edef9ebf4da07da254f3/druid/src/core.rs#L812)

### DIFF
--- a/src/widget/core.rs
+++ b/src/widget/core.rs
@@ -285,6 +285,14 @@ impl Pod {
             }
         };
         if recurse {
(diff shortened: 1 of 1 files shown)


Philipp-M (approved):
Nice, this actually fixes the issue I [found](https://github.com/linebender/xilem/pull/132#pullrequestreview-1638559934)

PR #150, opened 2023-11-23
Switch widget visual fixes

giannissc:
I have fixed the visuals of the knob in the hot (bigger and lighter), active (smaller and darker) and idle configuration.

The knob would also jump if the dragging started from the negative area of the widget and it seemed kinda jerky.

I have also disabled some of print statement in other places to cleanup the terminal output

### DIFF
--- a/src/widget/core.rs
+++ b/src/widget/core.rs
@@ -278,7 +278,7 @@ impl Pod {
                 had_active || hot_changed
             }
             Event::TargetedAccessibilityAction(action) => {
-                println!("TODO: {:?}", action);
+                // println!("TODO: {:?}", action);
                 self.state
                     .sub_tree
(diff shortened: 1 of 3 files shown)


giannissc:
I have made another (final) PR for the switch widget with some last minute visual changes to it. I will be pushing the progress bar and slider next! @Philipp-M

Philipp-M:
Hmm for me this regresses the dragging. Previously dragging with the mouse worked as expected, now it's kinda janky (it's much faster than the mouse pointer, and lags behind).

giannissc:
Good catch. I rethought the implementation and it should be must smoother now. Let me know how you like it! @Philipp-M

nicoburns (approved):
Not sure if dragging is even really needed for the switch, but it's all working very smoothly for me with the latest version.

PR #152, opened 2023-11-23
Export crates/modules to facilitate external widget definitions

longmathemagician:
Trivial change to allow programs using xilem to share its components, for custom widgets without using a local patch.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,7 @@ mod app_main;
 mod bloom;
 mod geometry;
 mod id;
-mod text;
+pub mod text;
 pub mod view;
 pub mod widget;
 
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
Accesskit, parley and vello seem uncontroversial to me. The text and piet_scene_helpers seem more like private internals for me. However, I think it's probably better to make things public and unblock people for now rather than worry too much about having a perfect API. We're a long way off having a stable API (we don't even have versioned releases at all yet!), so we can always make things private in future.

longmathemagician:
Agreed, I don't see the text and piet_scene modules being exposed as-is long term but for the moment they're very helpful.

PR #154, opened 2023-11-26
xilem_html: Add more text views (support numbers directly as view)

Philipp-M:
I wanted to use numbers as view directly. I don't see why we shouldn't support that, this adds support for that in xilem_html.
(And it reduces a little bit of the string view boilerplate)


derekdreery (approved):
LGTM. We can still format our numbers in different ways if we want by using the output from `format(..)` rather than the number directly.

Philipp-M:
Yes, that was my thought as well, just a convenience, when the format is correct anyway (e.g. for integers) or it doesn't really matter.

Thanks.

PR #155, opened 2023-12-05
Alternative scrollview implementation (doesn't use "origin" for scroll offset)

nicoburns:


src/widget/core.rs
```diff
@@ -498,6 +498,11 @@ impl Pod {
         }
     }
 
+    /// Get the widgets size (as returned by the layout method)
+    pub fn get_size(&mut self) -> Size {
```
Philipp-M, line 502: I still think we should omit `get_` prefixes to stay consistent :)
```suggestion
    pub fn size(&mut self) -> Size {
```

src/widget/scroll_view.rs
```diff
@@ -95,28 +100,73 @@ impl Widget for ScrollView {
...
+            }
+        }
     }
 
-    fn layout(&mut self, cx: &mut LayoutCx, proposed_size: Size) -> Size {
-        let child_proposed = Size::new(proposed_size.width, 1e9);
-        let child_size = self.child.layout(cx, child_proposed);
-        Size::new(child_size.width, proposed_size.height)
+    fn layout(&mut self, cx: &mut LayoutCx, bc: &BoxConstraints) -> Size {
+        cx.request_paint();
```
Philipp-M, line 129: Is this necessary even, when the layout hasn't changed at all?

(I see this is done in other widgets as well though, maybe a point to keep in mind later for optimization etc.)

src/widget/scroll_view.rs
```diff
@@ -95,28 +100,73 @@ impl Widget for ScrollView {
...
+
+        let cbc = BoxConstraints::new(
+            Size {
+                width: 0.0,
+                height: 0.0,
+            },
+            Size {
+                width: bc.max().width,
+                height: f64::INFINITY,
+            },
```
Philipp-M, line 139:
```suggestion
            Size::new(0.0, 0.0),
            Size::new(bc.max().width, f64::INFINITY),
```

(Nitpick, matter of preference I guess, doesn't really matter though)

src/widget/scroll_view.rs
```diff
@@ -95,28 +100,73 @@ impl Widget for ScrollView {
...
+        );
+        let child_size = self.child.layout(cx, &cbc);
+        let size = Size {
+            width: child_size.width.min(bc.max().width),
+            height: child_size.height.min(bc.max().height),
+        };
```
Philipp-M, line 145:
```suggestion
        let size = Size::new(
            child_size.width.min(bc.max().width),
            child_size.height.min(bc.max().height),
        );
```

(same here)

Philipp-M (approved):
Apart from all the nitpicks, this LGTM, thanks :)

src/widget/core.rs
```diff
@@ -498,6 +498,11 @@ impl Pod {
         }
     }
 
+    /// Get the widgets size (as returned by the layout method)
```
xStrom, line 501:
```suggestion
    /// Get the widget's size (as returned by the layout method)
```

xStrom (commented):
Read through it and played around with the example, looks like a good step forward. 👍

I also like this more than #146, because it's closer to how Druid did it. The power of familiarity.

src/widget/core.rs
nicoburns: I've updated this

src/widget/scroll_view.rs
nicoburns: I'm not sure if this is the best way of doing it. But I was seeing the visual rendering on screen failing to update properly before I put this in.

xorgy (approved):
This is a good start, and doesn't regress anything.

PR #158, opened 2024-01-03
Add an `AnyViewSequence` which is analogous to the AnyView for ViewSequences

Philipp-M:
This allows for type-erased ViewSequences.

Original Motivation was to check whether this helps with reducing wasm binary size in `xilem_html` (which wasn't really that [successful](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/AnyViewSequence), static optimization + compression seems to work quite well at this point). But I think this could be helpful otherwise too, e.g. for recursive ViewSequences (see e.g. [here](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Recursively.20returning.20views)) or cross-language bindings that need type erasure in sequences.

### DIFF
--- a/crates/xilem_core/src/sequence.rs
+++ b/crates/xilem_core/src/sequence.rs
@@ -56,7 +56,7 @@ macro_rules! impl_view_tuple {
 
 #[macro_export]
 macro_rules! generate_viewsequence_trait {
-    ($viewseq:ident, $view:ident, $viewmarker: ident, $bound:ident, $cx:ty, $changeflags:ty, $pod:ty; $( $ss:tt )* ) => {
+    ($viewseq:ident, $anyviewseq:ident, $boxedviewseq:ident, $view:ident, $viewmarker: ident, $bound:ident, $cx:ty, $changeflags:ty, $pod:ty; ($( $ss:tt )*); ($( $ss_bounds:tt )*) ) => {
         /// This trait represents a (possibly empty) sequence of views.
         ///
         /// It is up to the parent view how to lay out and display them.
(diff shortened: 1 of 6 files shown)


Philipp-M:
Keeping as draft, until there's a real reason for this (*maybe* recursive ViewSequences, but I'm not sure whether that is more of an artificial use-case)

casey:
I mentioned this in Zulip, but I also wanted to comment on the issue so it wouldn't get lost. I have a place where this could be useful. I have a branch where depending on a state variable, I want to return different `ViewSequence` types. A `.boxed` combinator would be quite useful:

```rust
if foo {
  (
    h1("Foo"),
    span("hello"),
  ).boxed()
} else {
  (
    h1("Bar"),
    div("goodbye"),
  ).boxed()
}
```

Philipp-M:
Yeah I see use-cases for this, and I think we should indeed land an updated version of this (also just to reflect a similar API for `View` <-> `AnyView` to `ViewSequence` <-> `AnyViewSequence`).

For your example you could use `boxed` for the second element to make it work though.

As a more general workaround for now for a changing number of sequences as mentioned by Daniel on zulip:

You could do either something like this:

```rust
let mut seq = Vec::new();
if foo {
    seq.push(h1("Foo").boxed());
    seq.push(span("hello").boxed());
    seq.push(p("baz").boxed());
} else {
    seq.push(h1("Bar").boxed());
    seq.push(div("goodbye").boxed());
}
```

Or alternatively something like:

```rust
if foo {
    (
        h1("Foo"),
        span("hello").boxed(),
        Some(p("baz")),
    )
} else {
    (
        h1("Bar"),
        div("goodbye").boxed(),
        None
    )
}
```
may work.

(The `Option` version should also work with the return type `impl DomFragment`/`impl ViewSequence`, while the `Vec` does not).

PR #159, opened 2024-01-10
Generalize `VecSplice` that is used in the ViewSequences as `ElementsSplice` trait and optimize tree-structure diffing in `xilem_web` with it

Philipp-M:
This slightly changes the `ViewSequence` trait to take a trait `dyn ElementsSplice` instead of the `VecSplice`. Functionally (in xilem) this should result in the same. But it has the advantage to track diffing more precisely in `ViewSequence`s.
This is used in `xilem_web` to only update elements that actually have changed (instead of removing all elements and adding the rebuilt elements again).


Philipp-M:
Closing as this will be merged with/as #160

PR #160, opened 2024-01-10
Add fine-grained tree-structure tracking in `xilem` and effcient tree-updates in `xilem_web`

Philipp-M:
This is based on top of #159 (which is a requirement for this PR, and the reason why this is a draft PR)

This implements fine-grained tree-structure tracking described in #54 using a `TreeTrackerSplice` (which implements the `ElementsSplice` introduced in #159).
Currently it's only replacing the bloom filter which is only used for accessibility. But I expect this to be more useful later. 

It's inspired by the lasagna branch in `druid`.


src/widget/tree_structure.rs
```diff
@@ -0,0 +1,191 @@
...
+
+use crate::view::ElementsSplice;
+
+/// The pure structure (parent/children relations via ids) of the widget tree.
+#[derive(Debug, Default, Clone)]
+pub struct TreeStructure {
+    parent: HashMap<Id, Option<Id>>,
+    // TODO this is taken from the druid lasagna branch, is the Option here intentional, i.e. to track all parentless ids with None?
+    children: HashMap<Option<Id>, Vec<Id>>,
+}
```
Philipp-M, line 18: @raphlinus Since this is copied from your lasagna branch, do you have an explanation for that TODO?

src/lib.rs
```diff
@@ -2,7 +2,7 @@ extern crate core;
 
 mod app;
 mod app_main;
-mod bloom;
+// mod bloom;
```
Philipp-M, line 5: Should the bloom filter implementation be removed completely, or does it make sense to keep it?

src/widget/linear_layout.rs
```diff
@@ -28,14 +28,21 @@ use super::{contexts::LifeCycleCx, EventCx, LayoutCx, LifeCycle, PaintCx, Pod, U
 /// widget.
 pub struct LinearLayout {
     pub children: Vec<Pod>,
+    pub tree_mutations: Vec<tree_structure::SpliceMutation>,
     pub spacing: f64,
     pub axis: Axis,
 }
```
PoignardAzur, line 34: Overall, I'd would really like if `SpliceMutation` and everything related to it was kept out of the `widget` module. It would match how things will look after the Masonry refactor.

This means a few things:

- Nix most of the changes to `src/widget/*.rs`
- In `src/widget/tree_structure.rs`, keep `TreeStructure` and move `SpliceMutation` and `TreeTrackerSplice` out.
- Move the `&mut TreeStructure` out of `CxState` and into `xilem::view::Cx`.
- Basically handle updating the TreeStructure from the view side.

Does that seem feasible to you?

PoignardAzur (commented):
As I mentioned [in the zulip thread](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Fine-grained.20tree-structure.20tracking), I'm of two minds on this.

On the one hand, I kinda hate VecSplice and I think it's a very confusing abstraction, and I think this PR compounds the things I don't like.

On the other hand, this is good quality code, and it's an incremental improvement.

I'd be a lot more comfortable with this PR if it included unit tests. I'm honestly feeling extremely nervous about the existing VecSplice code, because after reading it, I notice that a lot of its inner complexity isn't currently *used* by our web and widget code. Like, `skip` isn't even called anywhere.

So, given that how VecSplice is *supposed* to work is undocumented, I'm guessing that whether it *actually* works can only be known by those who can peer into the thoughts of Raph Levien.

Anyway, tests and documentation aside, my other sticking point is that I'd like the splicing code to stay out of the backend side. Since that backend is soon going to be replaced by a different crate, your changes would be easier to integrate with the Masonry merge if they interacted with the backend as little as possible. Essentially, `TreeStructure` is the only datastructure I'd want the backend to be aware of.

If the backend-related concerns are addressed, I'd be okay to merge this PR. Consider everything else extra credit.

crates/xilem_core/src/sequence.rs
```diff
@@ -53,10 +53,50 @@ macro_rules! impl_view_tuple {
...
+            fn mutate(&mut self) -> &mut $pod;
+            // TODO this could also track view id changes (old_id, new_id)
+            /// Mark any changes done by `mutate` on the current element
+            fn mark(&mut self, changeflags: $changeflags) -> $changeflags;
+            /// Delete the next n existing elements
+            fn delete(&mut self, n: usize);
+            /// Current length of the elements collection
+            fn len(&self) -> usize;
+            // /// skip the next n existing elements
+            // fn skip(&mut self, n: usize);
```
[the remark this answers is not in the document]
Philipp-M: Well I have put it there, because it makes sense as soon as we have reusable/immutable datastructures/ViewSequences such as ropes, where this will be needed. So I put it there as a reminder.

I don't see why this is a big no-no though (I agree though with having code comments that don't have any other sense than just being a comment/unused code, as they usually don't age well). But if it's really a deal-breaker I can remove it...

crates/xilem_core/src/sequence.rs
```diff
@@ -4,11 +4,11 @@
 #[doc(hidden)]
 #[macro_export]
 macro_rules! impl_view_tuple {
-    ( $viewseq:ident, $pod:ty, $cx:ty, $changeflags:ty, $( $t:ident),* ; $( $i:tt ),* ) => {
+    ( $viewseq:ident, $elements_splice: ident, $pod:ty, $cx:ty, $changeflags:ty, $( $t:ident),* ; $( $i:tt ),* ) => {
         impl<T, A, $( $t: $viewseq<T, A> ),* > $viewseq<T, A> for ( $( $t, )* ) {
             type State = ( $( $t::State, )*);
 
-            fn build(&self, cx: &mut $cx, elements: &mut Vec<$pod>) -> Self::State {
+            fn build(&self, cx: &mut $cx, elements: &mut dyn $elements_splice) -> Self::State {
```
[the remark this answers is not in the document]
Philipp-M: I am open for good alternative suggestions.

crates/xilem_core/src/sequence.rs
```diff
@@ -53,10 +53,50 @@ macro_rules! impl_view_tuple {
...
+            fn push(&mut self, element: $pod);
+            /// Mutate the next existing element, and add it to this collection
+            fn mutate(&mut self) -> &mut $pod;
+            // TODO this could also track view id changes (old_id, new_id)
+            /// Mark any changes done by `mutate` on the current element
+            fn mark(&mut self, changeflags: $changeflags) -> $changeflags;
+            /// Delete the next n existing elements
+            fn delete(&mut self, n: usize);
+            /// Current length of the elements collection
+            fn len(&self) -> usize;
```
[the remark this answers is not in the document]
Philipp-M: Yeah I guess this can be improved. Possibly as well via a general description of what it does.

src/widget/linear_layout.rs
Philipp-M: Sure I guess that makes sense.
One thing though, I think it makes sense to at least have readable access in `CxState` (e.g. to target the right accessibility node as is done in `Event::TargetedAccessibilityAction`).

Philipp-M:
> I kinda hate VecSplice and I think it's a very confusing abstraction
> ...
> So, given that how VecSplice is supposed to work is undocumented, I'm guessing that whether it actually works can only be known by those who can peer into the thoughts of Raph Levien.

I still think it's mostly because it's not really documented that well (and it's not the easiest to digest code code admittedly, but with good documentation it should probably not be necessary to dive that deep into the details). The general idea how it's supposed to work can be extracted from the code though?

I don't want to defend it as I also generally like a more mutation-log based approach as discussed on zulip.
But I think it makes sense in the way the architecture is currently built (every view/view sequence manages their corresponding element(s) and recursively traverses their children). I don't see currently how it could be done significantly better otherwise.

Well I can write some tests as this could also well serve as documentation (but I don't like to do unnecessary work when you think the `VecSplice` or generally this approach should go away any time soon). 

Anyway, should #159 be just merged with/as this PR or does it make sense to do these separately?

crates/xilem_core/src/sequence.rs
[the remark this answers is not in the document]
PoignardAzur: Nah, I've got nothing for now. Maybe add a comment explaining the rationale.

crates/xilem_core/src/sequence.rs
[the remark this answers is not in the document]
PoignardAzur: Right now the Xilem codebase is littered with comment-out code that was supposed to be temporary reminders. Right now the consensus is slowly moving towards "if we haven't implemented it yet, don't pretend like having a commented-out stubs means it's just around the corner".

If you want to add a reminder to your code, a TODO is better, ideally with an associated issue link (for now that can be a link to this PR).

crates/xilem_web/src/elements.rs
```diff
@@ -83,7 +164,7 @@ where
         let state = ElementState {
             children_states,
             child_elements,
-            scratch: vec![],
+            scratch,
             attributes,
```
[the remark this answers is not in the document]
PoignardAzur: I'm skeptical, but okay. If we're keeping it, what you just said should added in a comment.

PoignardAzur:
> Anyway, should https://github.com/linebender/xilem/pull/159 be just merged with/as this PR or does it make sense to do these separately?

I think we can close #159, so the discussion is centralized in this PR.

PoignardAzur:
> Well I can write some tests as this could also well serve as documentation (but I don't like to do unnecessary work when you think the VecSplice or generally this approach should go away any time soon).

Fair enough. It's far outside my current priorities that I don't expect this code to go away for at least 6 months, but I do want to eventually refactor it, so I get your reluctance. I think tests would be better, but Xilem current has virtually none, so it would be unfair to hold your PR to a higher standard. Consider tests optional for getting this PR accepted.

Philipp-M:
I think I remember why I put the code of the tree-structure changes into the widget part of the codebase. I think I once had it within the views, until I recognized that the type of the ids were different between view/widget...

This makes stuff more complicated because of the `Pod` (which holds the widget id) that is currently only constructed by the view sequences (more specifically by `impl ViewSequence for impl View`) (and the root pod within `app.rs`). I need the id of the element/pod though, and I can't access that within `View::rebuild` as it's done by the parent context.

I'm currently brainstorming how to get around that without shuffling a lot of code (or ugly hacks).

Philipp-M:
So one idea I have, is extending the `ElementsSplice` methods with a parameter `cx: &mut Cx` to record the changes (or apply them directly in the view context).

So taking the content of the `ViewSequence::build()` for the view:
```rust
let (id, state, element) = <V as View<T, A>>::build(self, cx);
elements.push(Pod::new(element), cx); // applys children tree-structure changes to the newly created Pod
(state, id)
```

Does that sound feasible?

Philipp-M:
I've done that, I'm not *extremely* happy about it, it introduces some kind of accidental complexity I think, and may be slightly brittle (as it requires a clean 1:1 relation of `Widget` <-> `Pod`, which is planned anyway I think).
At this point I'm also considering rethinking the way the View(Sequence) architecture/traits (towards less inplace mutations). But i think this requires quite a bit of thought to be an actual improvement.

Anyway how I have "solved" this for now is by having a `Cx::mark_children_tree_structure_mutations(&mut self, mutations: Vec<SpliceMutation>)`

which is called by the view that owns the children elements. And then register them with the above function.
`TreeTrackerSplice::push(&mut self, element: Pod, cx: &mut Cx)` looks like this then:

```rust
self.mutations.push(SpliceMutation::Add(element.id()));
cx.apply_children_tree_structure_mutations(element.id());
self.splice.push(element);
```

Maybe there's a better way (open for suggestions)

PoignardAzur:
Can you push the commits where you implemented that? I think I don't see them.

crates/xilem_core/src/sequence.rs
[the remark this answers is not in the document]
Philipp-M: Ok, that sounds reasonable, I've converted it into a `TODO` describing also potential use-cases for it.

crates/xilem_web/src/elements.rs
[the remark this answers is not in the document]
Philipp-M: I've added a few `debug_assert`s and a comment.

crates/xilem_core/src/sequence.rs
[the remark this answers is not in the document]
Philipp-M: Is the current version better?

PoignardAzur:
> Anyway how I have "solved" this for now is by having a `Cx::mark_children_tree_structure_mutations(&mut self, mutations: Vec<SpliceMutation>)`
> 
> which is called by the view that owns the children elements. And then register them with the above function. `TreeTrackerSplice::push(&mut self, element: Pod, cx: &mut Cx)` looks like this then:
> 
> ```rust
> self.mutations.push(SpliceMutation::Add(element.id()));
> cx.apply_children_tree_structure_mutations(element.id());
> self.splice.push(element);
> ```

Maybe I'm missing something obvious, but is `SpliceMutation` even necessary a this point? The `TreeStructureTrackingSplice::push()` method takes a Cx and a WidgetPod; `TreeStructure::append_child()` takes a parent id and a child id. It seems like you could easily call one from the other and skip all lot of intermediary steps.

Right now, it looks like you're using `Cx::mark_children_tree_structure_mutations` and `Cx::apply_children_tree_structure_mutations` as a workaround to the fact that Cx doesn't have the parent id. I think your code could get a lot simpler if you just added

```rust
struct Cx {
    // ...
    current_parent: Option<Id>,
}
```

or something in that vein.

(Also, quick question, have you tried running some Xilem apps, eg the Taffy example, and checked that the child tracking actually worked?)

src/widget/tree_structure.rs
PoignardAzur: I'm pretty sure `None` as a key to the children map means "return the root widget". I think you can skip it and have `HashMap<Id, Vec<Id>>` instead.

crates/xilem_core/src/sequence.rs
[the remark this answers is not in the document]
PoignardAzur: Definitely!

src/view/linear_layout.rs
```diff
@@ -84,12 +89,17 @@ impl<T, A, VT: ViewSequence<T, A>> View<T, A> for LinearLayout<T, A, VT> {
         element: &mut Self::Element,
     ) -> ChangeFlags {
         let mut scratch = vec![];
-        let mut splice = VecSplice::new(&mut element.children, &mut scratch);
-
+        let mut tree_mutations = vec![]; // TODO(#160) could save some allocations by using View::State (scratch too)
```
PoignardAzur, line 92: Having an issue/PR number associated with TODOs is good, assuming the linked issue/PR includes enough info for an outsider contributor reading it to know how to contribute.

Since it's awkward to create issues for a PR that hasn't been merged yet, I think a good compromise would be to edit the top-level post of this PR with:

- A list of all the TODOs it adds.
- Enough info to operationalize resolving those TODOS (eg a short description of the problem and how to solve it). I know this is a lot to ask, so it's okay if this info is light, as long as it's immediately visible.

src/view/tree_structure_tracking.rs
```diff
@@ -0,0 +1,91 @@
...
+    Delete(usize),
+    Skip(usize),
+}
+
+/// An ElementsSplice that monitors the tree structure by maintaining a mutation log
+pub struct TreeStructureTrackingSplice<'a, 'b, 'c> {
+    current_child_id: Option<Id>,
+    splice: VecSplice<'a, 'b, Pod>,
+    mutations: &'c mut Vec<SpliceMutation>,
+}
```
PoignardAzur, line 23: The name feels a bit long. Maybe rename to `TreeStructureSplice`? (Not very important.)

src/view/tree_structure_tracking.rs
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+}
+
+impl<'a, 'b, 'c> ElementsSplice for TreeStructureTrackingSplice<'a, 'b, 'c> {
+    fn push(&mut self, element: Pod, cx: &mut Cx) {
+        self.mutations.push(SpliceMutation::Add(element.id()));
+        cx.apply_children_tree_structure_mutations(element.id());
+        self.splice.push(element);
+    }
```
PoignardAzur, line 45:
```suggestion
    fn push(&mut self, element: Pod, cx: &mut Cx) {
        if let Some(parent_id) = cx.id_path().last() {
          cx.tree_structure.append_child(parent_id, element.id);
        }
        self.splice.push(element);
    }
```

See my longer comment.

src/view/view.rs
```diff
@@ -110,4 +116,40 @@ impl Cx {
...
+    /// `cx.apply_children_tree_structure_mutations(pod.id())`
+    /// One needs to be careful with this, when recursively traversing the tree to avoid overwriting other mutations
+    pub fn mark_children_tree_structure_mutations(&mut self, mutations: &mut Vec<SpliceMutation>) {
+        self.current_children_tree_structure_mutations.clear();
+        self.current_children_tree_structure_mutations
+            .append(mutations);
+    }
+
+    // TODO(#160): Currently if there are no children added/mutated the tree will not be mutated.
+    //             Should the parent be inserted if there aren't any children? Should this happen elsewhere?
```
PoignardAzur, line 131: I'm not sure that's a problem?

If I understand the current design, the only widget whose parent info isn't kept up to date is the root widget, which should be fine.

src/widget/tree_structure.rs
```diff
@@ -0,0 +1,155 @@
...
+            if parent == ancestor {
+                return true;
+            }
+            id = parent;
+        }
+        false
+    }
+
+    // TODO: Should any of the following methods return a Result or panic?
+    pub(crate) fn append_child(&mut self, parent_id: Id, id: Id) {
```
PoignardAzur, line 32: Panic for now.

Code that fails silently can stay under the radar for years. Code that panics gets bug reports fast, and panics are easier to find in unit tests and fuzz tests. Returning Result doesn't make sense, because any id-related error here will be because of an faulty implementation in Xilem, not user code or bad IO.

(Maybe mention the panics in doc comments, though.)

src/widget/tree_structure.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        // append children
+        tree_structure.append_child(parent, child1);
+        tree_structure.append_child(parent, child2);
+        tree_structure.append_child(parent, child3);
+        let children = tree_structure.children(Some(parent)).unwrap();
+        assert_eq!(children.len(), 3);
+        assert_eq!(children[0], child1);
+        assert_eq!(children[1], child2);
+        assert_eq!(children[2], child3);
```
PoignardAzur, line 103: Could probably be a single `assert_eq!` with a Vec?

Yeah, by this point I'm nitpicking.

src/widget/tree_structure.rs
```diff
@@ -0,0 +1,155 @@
...
+        assert_eq!(tree_structure.parent(child2), None);
+        assert_eq!(tree_structure.parent(child2_new), Some(Some(parent)));
+        assert_eq!(tree_structure.parent(child3), Some(Some(parent)));
+
+        // delete children
+        tree_structure.delete_children(parent, 0..2);
+        let children = tree_structure.children(Some(parent)).unwrap();
+        assert_eq!(children.len(), 1);
+        assert_eq!(children[0], child3);
+        assert_eq!(tree_structure.parent(child1), None);
```
PoignardAzur, line 126: I feel like maybe all of these should panic. Maybe add an `is_in_tree()` method instead?

PoignardAzur (changes requested):
I feel *much* better about this code now. I definitely like the added unit tests.

If we can resolve the bit with `mark_children_tree_structure_mutations` then we're good to merge.

Philipp-M:
I'm going a bit in more detail over the review later (when I got more time), thanks so far for the detailed review.

One thing though:

> Actually, no, you can get it from the id path.

That's unfortunately not the case, as it's the `ViewId` (we should really rename it (see also #168), I forgot about it when I did the changes again). While we need the `WidgetId`.

This is why it's pretty tricky currently to do this cleanly I think (without reshuffling the whole code-base).

PoignardAzur:
> That's unfortunately not the case, as it's the `ViewId` 

Ah, right, I forgot. Okay, so what I suggest is:

- Add a `parent_widget: Option<widget::Id>` field to Cx, with a getter method.
- Add a `parent_widget: widget::id` argument to `Cx::with_id()` and `Cx::with_new_id`

This adds some cruft to the codebase (we're basically tracking the same thing twice, if I understand correctly), but we'll deal with that later.

> (we should really rename it (see also #168)

Two renames (Id -> ViewId, Id -> WidgetId) would be welcome, though probably not in this PR.

crates/xilem_core/src/sequence.rs
```diff
@@ -53,19 +53,67 @@ macro_rules! impl_view_tuple {
...
+            fn push(&mut self, element: $pod, cx: &mut $cx);
+            /// Mutate the next existing element, and add it to the resulting collection (and increment the index by 1)
+            fn mutate(&mut self, cx: &mut $cx) -> &mut $pod;
+            // TODO(#160) this could also track view id changes (old_id, new_id)
+            /// Mark any changes done by `mutate` on the current element (this doesn't change the index)
+            fn mark(&mut self, changeflags: $changeflags, cx: &mut $cx) -> $changeflags;
+            /// Delete the next n existing elements (this doesn't change the index)
+            fn delete(&mut self, n: usize, cx: &mut $cx);
+            /// Current length of the elements collection
+            fn len(&self, cx: &mut $cx) -> usize;
```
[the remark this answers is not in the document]
Philipp-M: Yeah added it in case an implementor needs access to it. But you're right, in case it's really needed it can still be added later.

src/view/linear_layout.rs
Philipp-M: Hmm, not sure about that, as I don't think a lot of people will look it up (vs the effort of creating such).
And when the TODO isn't clear in the code context (and/or is generally worth an issue) IMO it should really be an issue with a clear description after the PR has been merged.
After the recent update, I don't think such TODOs have survived (I don't think the TODO comment above is worth an issue TBH, I'll probably add a simple PR directly targeting that after this PR, to not further bloat this one).

src/view/view.rs
Philipp-M: > I'm not sure that's a problem?

Yeah it's no problem, I just like to write out thoughts while coding (before I forget them). And clean them up later. A good time is probably now.

> the only widget whose parent info isn't kept up to date is the root widget, which should be fine.

It should not even be necessary to update the root widget, since the root `Pod` is only created once, and never changes its id.
The children of that root widget are updated though in the tree structure.

src/widget/tree_structure.rs
Philipp-M: I've did it via a slice, and yeah that's better (I'm fine with nitpicking btw.)

Philipp-M:
Ok, I've updated this and applied all the review comments.

> Maybe I'm missing something obvious, but is SpliceMutation even necessary a this point?
> ...
> Anyway, if think your code would get a lot more robust if you skipped the mutation list and applied the mutations directly as they come.

Yeah you're right, having access to `Cx` in the splice methods makes all of this quite a bit simpler indeed.
The idea to use something like `with_id` is certainly cleaner, made the code more compact.

I've implemented that though similarly as the `id_path` in `Cx` (via a stack). Just having an option is not enough, as children setting this, overwrite it for other children. So I've added new context methods named `with_pod` and `with_new_pod` so that access to the element id is always valid within the views (via `Cx::element_id`). This is also necessary (as described below) since there's no direct `ViewId` <-> `WidgetId` relation (there are more `ViewId`s than `WidgetId`s).

> Also, quick question, have you tried running some Xilem apps, eg the Taffy example, and checked that the child tracking actually worked?

Sure, I don't don't want to merge untested code (did it via the good ol' `println` debugging).

> Two renames (Id -> ViewId, Id -> WidgetId) would be welcome, though probably not in this PR.

Yeah agree, we should keep PRs small and focused.

> we're basically tracking the same thing twice, if I understand correctly

No that's not necessarily the case. Views can e.g. have ids while having the same `View::Element` of its child view. `xilem_web` has this e.g. for events. Generally this is true for views that take events/messages, without extra widget implementation.

src/widget/tree_structure.rs
Philipp-M: Yeah there's no `Option` necessary for both of these, since it's only the relations. Updated that in the latest commits.

PoignardAzur (approved):
LGTM. We can merge this after a rebase.

PR #164, opened 2024-01-22
Add support for `impl View for Arc/Rc<impl View>`

Philipp-M:


crates/xilem_core/src/view/rc.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2024 the Xilem Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#[macro_export]
+macro_rules! generate_rc_view {
+    (($($rc_ty:tt)*), $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty; $($ss:tt)*) => {
```
jplatte, line 6:
```suggestion
    ($rc_ty:path, $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty; $($ss:tt)*) => {
```

crates/xilem_web/src/view.rs
```diff
@@ -90,6 +90,8 @@ xilem_core::generate_anyview_trait! {AnyView, View, ViewMarker, Cx, ChangeFlags,
 xilem_core::generate_memoize_view! {Memoize, MemoizeState, View, ViewMarker, Cx, ChangeFlags, static_view, memoize;}
 xilem_core::generate_adapt_view! {View, Cx, ChangeFlags;}
 xilem_core::generate_adapt_state_view! {View, Cx, ChangeFlags;}
+xilem_core::generate_rc_view!((std::rc::Rc), View, ViewMarker, Cx, ChangeFlags;);
+xilem_core::generate_rc_view!((std::sync::Arc), View, ViewMarker, Cx, ChangeFlags; + Sync);
```
jplatte, line 94: Seems unnecessarily inconsistent that this macro is invoked using round parentheses while the others are invoked using curly braces.

jaredoconnell:
Now, or later when the examples are improved, this would definitely benefit from a brief example.

crates/xilem_core/src/view/rc.rs
Philipp-M: Just tried that locally, but it has issues with `$rc_ty<V>` unfortunately (I have tried `ty` previously, same issue):
```
expected one of `!`, `+`, `where`, or `{`
```

Not sure if there's another way to have `<V>` after a path-like that is not `tt`.

crates/xilem_web/src/view.rs
Philipp-M: Whoops you're right, was a little bit too fast there.

Philipp-M:
> Now, or later when the examples are improved, this would definitely benefit from a brief example.

You're right, I could add one here, that shows a use-case for this (it's an alternative to `Memoize`)

Philipp-M:
Good idea to mention the example, I've implemented it via an example for memoization and quite a few views were not yet exported... It also revealed a bomb:

`View` needs to be `Sync` to support `Arc<impl View>`.

I think if we want to store shared views in general in the app state this requirement will come sooner or later (e.g. by supporting a rope of views as a `ViewSequence`).

crates/xilem_core/src/view/rc.rs
jplatte: For `ty` it makes sense to me, since there can't be generics after a type in general (any generics would be part of the type, not separate), for `path` I don't understand it but oh well. You could use `$($rc_ty:ident)::*` or `$($tc_ty:tt)::*` to get rid of the surrounding parentheses though, right?

crates/xilem_core/src/view/rc.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2024 the Xilem Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+#[macro_export]
+macro_rules! generate_rc_view {
+    ($($rc_ty:ident)::*, $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty; $($ss:tt)*) => {
+        impl<V> $viewmarker for $(::$rc_ty)*<V> {}
```
jplatte, line 7: Prefixing the paths with `::` in the output seems wrong, semantically. Instead, I would mirror the input expression exactly.

```suggestion
        impl<V> $viewmarker for $($rc_ty)::*<V> {}
```

crates/xilem_core/src/view/rc.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        impl<T, A, V: $viewtrait<T, A> $( $ss )*> $viewtrait<T, A> for $(::$rc_ty)*<V> {
```
jplatte, line 9:
```suggestion
        impl<T, A, V: $viewtrait<T, A> $( $ss )*> $viewtrait<T, A> for $($rc_ty)::*<V> {
```

crates/xilem_core/src/view/rc.rs
```diff
@@ -0,0 +1,44 @@
...
+
+            fn rebuild(
+                &self,
+                cx: &mut $cx,
+                prev: &Self,
+                id: &mut $crate::Id,
+                state: &mut Self::State,
+                element: &mut Self::Element,
+            ) -> ChangeFlags {
+                if !$(::$rc_ty)*::ptr_eq(self, prev) {
```
jplatte, line 26:
```suggestion
                if !$($rc_ty)::*::ptr_eq(self, prev) {
```

crates/xilem_core/src/view/rc.rs
Philipp-M: Well you're probably right, thanks

crates/xilem_core/src/any_view.rs
```diff
@@ -3,22 +3,22 @@
 
 #[macro_export]
 macro_rules! generate_anyview_trait {
-    ($anyview:ident, $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty, $anywidget:ident, $boxedview:ident; $($ss:tt)*) => {
+    ($anyview:ident, $viewtrait:ident, $viewmarker:ty, $cx:ty, $changeflags:ty, $anywidget:ident, $boxedview:ident; ($($super_bounds:tt)*), ($($state_bounds:tt)*)) => {
```
jplatte, line 6: I was going to suggest changing the input to avoid the parentheses here too, but unfortunately that runs into https://github.com/rust-lang/rust/issues/18700 (last comment has a workaround, but probably not worth using that here).

jplatte: Maybe make the last two parameters optional using `$()?` though? The invocation with `(), ()` looks weird.

crates/xilem_core/src/any_view.rs
```diff
@@ -3,22 +3,22 @@
...
         /// A trait enabling type erasure of views.
         pub trait $anyview<T, A = ()> {
             fn as_any(&self) -> &dyn std::any::Any;
 
             fn dyn_build(
                 &self,
                 cx: &mut $cx,
-            ) -> ($crate::Id, Box<dyn std::any::Any $( $ss )* >, Box<dyn $anywidget>);
+            ) -> ($crate::Id, Box<dyn std::any::Any $( $state_bounds )* >, Box<dyn $anywidget>);
```
jplatte, line 14: You could add the `+` here and the following uses of `$state_bounds` and remove it from `generate_anyview_trait!` invocations so they look less weird. A trailing `+` in a list of trait bounds is not a syntax error.

crates/xilem_core/src/view/mod.rs
```diff
@@ -36,9 +37,9 @@ macro_rules! generate_view_trait {
         /// and also a type for actions which are passed up the tree in message
         /// propagation. During message handling, mutable access to the app state is
         /// given to view nodes, which in turn can expose it to callbacks.
-        pub trait $viewtrait<T, A = ()> $( $ss )* {
+        pub trait $viewtrait<T, A = ()> $( $super_bounds )* {
```
jplatte, line 40: Same here with `:`, I'm pretty sure `trait Foo<T, A = ()>: {}` is a valid trait definition with an empty list of supertraits.

crates/xilem_core/src/view/mod.rs
```diff
@@ -36,9 +37,9 @@ macro_rules! generate_view_trait {
...
             /// Associated state for the view.
-            type State $( $ss )*;
+            type State $( $state_bounds )*;
```
jplatte, line 42: And here too :)

crates/xilem_core/src/any_view.rs
Philipp-M: Well that's interesting (same with the `:`), new knowledge acquired...

crates/xilem_core/src/any_view.rs
Philipp-M: I've used optionals as you said

crates/xilem_core/src/any_view.rs
jplatte: You can also have an empty `where` clause, or empty generic parameter list (`<>`) behind non-generic types *including primitives* 😄

Though if you use rustfmt, it will remove all of these things.

Philipp-M:
I've added the super bounds of `View` to `AnyView` as well, to avoid `Box<dyn AnyView + Send + Sync>`

Philipp-M:
Whoops, I was a little bit too eager with deleting old branches...

DJMcNab:
I think this can be closed - this is included in #310

Philipp-M:
Yes it can be closed

PR #165, opened 2024-01-25
Update vello and parley deps...

dfrg:
... and everything else, apparently. This pushes the new font code changes through the ecosystem.

Brings everything into sync and should put us in a good place to start releasing.

This works as is but depends on a vello commit that is still under review: linebender/vello#425 should land first and this should be updated with the new git rev before merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,24 +4,24 @@ version = 3
 
 [[package]]
 name = "accesskit"
-version = "0.11.2"
+version = "0.12.2"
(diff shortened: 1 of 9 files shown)


Philipp-M (approved):
Well, I just started doing exactly that, and then shortly after, this PR popped up haha, thanks again, LGTM

I try to review the other PR as well

waywardmonkeys (approved):

PR #166, opened 2024-02-06
Remove unnecessary Glazier dependencies.

xorgy:


src/id.rs
```diff
@@ -21,9 +21,12 @@ pub struct Id(NonZeroU64);
 impl Id {
     /// Allocate a new, unique `Id`.
     pub fn next() -> Id {
```
DJMcNab, line 23: `Id` is kind of an awful name for this type, although this PR isn't really the place to fix it

DJMcNab (approved):

src/id.rs
xorgy: Yeah, I think the mechanism is maybe not ideal, and personally think the type should maybe be `Option<NonZeroUsize>` (but that's technically complicated by AccessKit, which wants `u64`), and this patch could introduce a needless runtime check when coercing to NonZero (which is just `unsafe` in Glazier) depending on how aware the compiler is of the static initalizer, but it's fine for now I think.

nicoburns: I've turned this into an issue so as to avoid blocking this PR https://github.com/linebender/xilem/issues/168

nicoburns (approved):
Regarding the `kurbo` imports, I don't love importing from a reexport like this. There are a number of alternative strategies we could use:
- We could re-export `vello::kurbo` as `xilem::kurbo` and then have everything import from `crate::kurbo`? The re-export could be private (`pub (crate)`) if we don't want to publicly re-export.
- We could depend directly on `kurbo` in `xilem`
- If `kurbo` use in `xilem` is mainly for interaction with `vello`, we could re-export individual types from `kurbo` in the top-level `vello` namespace (so that we can do `use vello::Circle`)

But that can be resolved in a future PR. This one solves the immediate issue with glazier imports :)

xorgy:
Yeah, regarding the reexports: we already directly import kurbo, and I will be using it for now in the bits of functionality I'm dragging in from Glazier for the winit port. I think it's not crazy to use the xilem public reexport for now, since xilem is tightly coupled to vello anyhow, but this all could be more elegant some day I'm sure.

PR #170, opened 2024-02-13
Upgrade Vello and ditch SceneBuilder/SceneFragment.

xorgy:


xorgy:
This is actually busted, I'll try to fix it.

dfrg:
oops, I did the same PR :)

src/app_main.rs
```diff
@@ -218,8 +217,7 @@ where
             } else {
                 None
             };
-            let mut builder = SceneBuilder::for_scene(&mut self.scene);
-            builder.append(fragment, transform);
+            self.scene.append(fragment, transform);
```
dfrg, line 220: `SceneBuilder::for_scene` implicity reset the scene. We'll have to do this manually now: `self.scene.reset()`

src/widget/core.rs
```diff
@@ -446,28 +448,26 @@ impl Pod {
         };
 
         if needs_paint {
-            let mut builder = SceneBuilder::for_fragment(&mut self.fragment);
-            self.widget.paint(&mut inner_cx, &mut builder);
+            self.widget.paint(&mut inner_cx, &mut self.fragment);
```
dfrg, line 451: ditto, `self.fragment.reset()`

xorgy: Yeah I noticed quickly. Even after closing the examples my GPU driver crashed my system about 15 seconds later... I missed it in core and made the same mistake. :grimacing: 
Anyway, resolved.

PR #171, opened 2024-02-13
Update vello rev

dfrg:
Captures a fix for appending fragments with variable font glyph runs in vello. This also requires adapting to the changes in the scene API.


xorgy:
Great minds think alike.

PR #175, opened 2024-02-16
Support arrays (`[impl ViewSequence; N]`) as `ViewSequence`s

Philipp-M:
Adds the blanket impl
`impl<const N: usize, VT: ViewSequence> ViewSequence for [VT; N]`

Allows for example something like this:

```rust
fn my_tab(impl View) -> impl View {..}
h_stack(["Tab 1", "Tab 2", "Tab 3"].map(my_tab))
```

### DIFF
--- a/xilem_core/src/sequence.rs
+++ b/xilem_core/src/sequence.rs
@@ -552,9 +552,11 @@ where
         ctx: &mut Context,
         elements: &mut impl ElementSplice<Element>,
     ) {
-        for (gen, ((seq, prev_seq), state)) in self.iter().zip(prev).zip(seq_state).enumerate() {
+        for (idx, ((seq, prev_seq), state)) in self.iter().zip(prev).zip(seq_state).enumerate() {
             ctx.with_id(
-                ViewId::new(gen.try_into().expect("usize should fit in u64")),
+                ViewId::new(idx.try_into().expect(
+                    "ViewSequence arrays with more than u64::MAX + 1 elements not supported",
+                )),
                 |ctx| {
                     seq.seq_rebuild(prev_seq, state, ctx, elements);
                 },
@@ -574,11 +576,8 @@ where
             .split_first()
             .expect("Id path has elements for [ViewSequence; N]");
 
-        let index: usize = start
-            .routing_id()
-            .try_into()
-            .expect("u64 should fit in usize");
-        // Panics if index is out of bounds, but we know it isn't because this is the same generation
+        let index: usize = start.routing_id().try_into().unwrap();
+        // We know the index is in bounds because it was created from an index into a value of Self
         let inner_state = &mut seq_state[index];
         self[index].seq_message(inner_state, rest, message, app_state)
     }
@@ -590,9 +589,11 @@ where
         ctx: &mut Context,
         elements: &mut impl ElementSplice<Element>,
     ) {
-        for (gen, (seq, state)) in self.iter().zip(seq_state).enumerate() {
+        for (idx, (seq, state)) in self.iter().zip(seq_state).enumerate() {
             ctx.with_id(
-                ViewId::new(gen.try_into().expect("usize should fit in u64")),
+                ViewId::new(idx.try_into().expect(
+                    "ViewSequence arrays with more than u64::MAX + 1 elements not supported",
+                )),
                 |ctx| {
                     seq.seq_teardown(state, ctx, elements);
                 },
--- a/xilem_core/src/sequence.rs
+++ b/xilem_core/src/sequence.rs
@@ -511,7 +511,7 @@ where
     ) -> MessageResult<Action, Message> {
         let (start, rest) = id_path
             .split_first()
-            .expect("Id path has elements for Option<ViewSequence>");
+            .expect("Id path has elements for Vec<ViewSequence>");
         let (index, generation) = view_id_to_index_generation(*start);
         let stored_generation = &seq_state.generations[index];
         if *stored_generation != generation {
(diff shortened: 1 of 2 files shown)


Philipp-M:
Well I guess I'm getting too used to using unstable Rust (`[].each_ref()` is unstable in 1.76). But fortunately as I've just seen, the corresponding feature was just stabilized 3 weeks ago. So we can just wait for the next Rust version (1.77) and then the feature would be supported (when we don't want to try to keep MSRV low anyhow).

Otherwise this needs I think a more dynamic solution (`State = Vec<VT::State>`).

zoechi:
Your comment made me curious and I looked into `[].each_ref()`, but I wasn't able to figure out what the benefit over for example `[].iter().map(..m)` is, especially in this case where the created array of references isn't kept around afterwards.
What do you think is the benefit?

Philipp-M:
`each_ref` allows to create a different array with the same size (`[T; N].each_ref() -> [&T; N].map() -> [S; N]`).
The difference would be that `Vec` needs an allocation, while `[T; N]` does not. I don't expect much of a practical difference, and it could well be done with a `Vec` (via `iter().map().collect()`) instead (but I think it's a little bit cleaner with a static array).

zoechi:
I see. Because the resulting array is not used (as far as I see), there also would be no `Vec` and `collect()` necessary.  
My approach would be `[].iter().for_each(|x| ...)`. 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2ad60536d7f403943d34758c29a01083
It's a bit more verbose though. I wasn't aware of `each_ref()`.

Philipp-M:
Note the function signature of `ViewSequence::build(&self, cx: &mut Cx, elements: &mut Vec<Pod>) -> Self::State` Which returns the view state. 
So you need to create and return an owned container, it's not enough to just iterate over the child view sequences and invocate `build()`.

It is used in every rebuild and message invocation, where it is handed by the parent/owning view.

zoechi:
Now I get it. I thought I checked 3 times if the result is returned and I didn't see it🤦🏼‍♂️

Philipp-M:
The relevant feature is now in stable rust. So this is ready. We could increase the Rust version in a separate PR as well...

DJMcNab:
Sorry, this probably has suffered the same fate at the hands of #310 as the other PRs

Philipp-M:
I've updated this to the new xilem_core. Hmm I hope this doesn't suffer the same fate again though...
I copied some of the tests from the tuple sequence, since they're seemingly equivalent from what needs to be tested.

xilem_core/src/sequence.rs
```diff
@@ -524,6 +524,83 @@ where
...
+    fn seq_rebuild(
+        &self,
+        prev: &Self,
+        seq_state: &mut Self::SeqState,
+        ctx: &mut Context,
+        elements: &mut impl ElementSplice<Element>,
+    ) {
+        for (gen, ((seq, prev_seq), state)) in self.iter().zip(prev).zip(seq_state).enumerate() {
+            ctx.with_id(
+                ViewId::new(gen.try_into().expect("usize should fit in u64")),
```
DJMcNab, line 557: I think generally we've preferred for these error messages to discuss "why" it failed, not how it failed.
E.g. `ViewSequence arrays with more than u64::MAX + 1 elements not supported` (fencepost at 0)

xilem_core/src/sequence.rs
```diff
@@ -524,6 +524,83 @@ where
...
+
+    #[doc(hidden)]
+    fn seq_rebuild(
+        &self,
+        prev: &Self,
+        seq_state: &mut Self::SeqState,
+        ctx: &mut Context,
+        elements: &mut impl ElementSplice<Element>,
+    ) {
+        for (gen, ((seq, prev_seq), state)) in self.iter().zip(prev).zip(seq_state).enumerate() {
```
DJMcNab, line 555:
```suggestion
        for (idx, ((seq, prev_seq), state)) in self.iter().zip(prev).zip(seq_state).enumerate() {
```

xilem_core/src/sequence.rs
```diff
@@ -524,6 +524,83 @@ where
...
+        app_state: &mut State,
+    ) -> MessageResult<Action, Message> {
+        let (start, rest) = id_path
+            .split_first()
+            .expect("Id path has elements for [ViewSequence; N]");
+
+        let index: usize = start
+            .routing_id()
+            .try_into()
+            .expect("u64 should fit in usize");
```
DJMcNab, line 580: This can just be an unwrap, as it's actually impossible (the index was previously created from a usize)

xilem_core/src/sequence.rs
```diff
@@ -524,6 +524,83 @@ where
...
+        // Panics if index is out of bounds, but we know it isn't because this is the same generation
```
DJMcNab, line 581:
```suggestion
        // We know the index is in bounds because it was created from an index into a value of Self
```

xilem_core/src/sequence.rs
```diff
@@ -524,6 +524,83 @@ where
...
+    }
+
+    #[doc(hidden)]
+    fn seq_teardown(
+        &self,
+        seq_state: &mut Self::SeqState,
+        ctx: &mut Context,
+        elements: &mut impl ElementSplice<Element>,
+    ) {
+        for (gen, (seq, state)) in self.iter().zip(seq_state).enumerate() {
```
DJMcNab, line 593:
```suggestion
        for (idx, (seq, state)) in self.iter().zip(seq_state).enumerate() {
```

DJMcNab (approved):
A few "prose" nits, but otherwise this looks good.

I'm not really sure how *useful* this is, but it also doesn't cost us anything to support.

Philipp-M:
Thanks for the quick review, I'm merging.

PR #176, opened 2024-02-17
basic winit port

xorgy:


dfrg:
This is much cleaner than I would have expected. Overall LGTM and I’m comfortable merging now and making further changes incrementally.

xorgy:
> This is much cleaner than I would have expected. Overall LGTM and I’m comfortable merging now and making further changes incrementally.

It's not really clean, because there are a bunch of hanging references in `App` which rely on the window, which can't really access it because of the ownership boundary between theme.

src/widget/raw_event.rs
```diff
@@ -87,3 +83,116 @@ impl ViewContext {
...
+        self.e.buttons.insert(button);
+        self.e.count = self.counter.count_for_click(self.e.pos);
+        self.e.button = button;
+        self.e.clone()
+    }
+
+    pub fn released(&mut self, button: PointerButton) -> MouseEvent {
+        self.e.wheel_delta = Vec2::ZERO;
+        self.e.buttons.remove(button);
+        self.e.count = self.counter.count_for_click(self.e.pos);
```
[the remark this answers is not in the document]
xorgy: Yeah this was a copy error in this last revision, good catch.

Philipp-M:
> not really any clean way to restore it without changing the App/AppLauncher API

What do you mean exactly?

xorgy:
@Philipp-M 
> What do you mean exactly?

I mean that the `AppLauncher` has to own the window, but the `App` accesses it directly to implement accessibility integration, and there is no clean way to let both of them access it. Right now there is no mechanism for opening subwindows anyway, which was the purpose of separating `App` and `AppLauncher`, so collapsing these interfaces together seems to be the quickest way to get it working.

raphlinus:
I skimmed the PR and tested it. It works for me (tested on mac), and further work blocks on it. My preference is to merge soon, and document all the regressions and rough edges to those can be separate PRs. Thanks for the work!

Philipp-M:
I'm running into a weird issue at this line (so far I could pinpoint it at least):

https://github.com/linebender/xilem/blob/623f18a34f0359e4d449aafda4c6e98d2f2c97ff/src/app_main.rs#L231

This seems to deadlock for me with this PR, after a few frames painted (slightly under 10 frames).
I'm not sure it's directly related to this PR though, as I have sometimes a crash when resizing on current main (at the same line with an `failed to acquire next swapchain texture: Outdated` error)  but it doesn't crash or deadlock at this line otherwise.

I'm not sure where the exact issue could be. I'm testing on Hyprland (wayland + nvidia beta driver) currently. I can try on X11 if it works there for me.

I just tried the wgpu raytracing triangle example which works for me (maybe relevant).

Philipp-M:
I just tested it on X11, there it works. I can still provoke the Outdated swapchain texture issue though when resizing...

Philipp-M:
Ok, I pretty sure it's a wgpu issue, I could reproduce the issue with the examples on wgpu 0.18 (which is what xilem still uses). But *not* on wgpu 0.19

Philipp-M:
Ok, no unfortunately it wasn't as easy as that...
I have quickly hacked in wgpu 0.19 and updated to current vello (I guess I can upstream that on your/this branch, when it's already done anyway?).

Same issue though, deadlock at that line after some frames...

src/widget/raw_event.rs
```diff
@@ -87,3 +82,115 @@ impl ViewContext {
...
+/// Click-count is incremented if both the duration and distance between a pair
+/// of clicks are below some threshold.
+#[derive(Debug, Clone)]
+struct ClickCounter {
+    max_interval: Cell<Duration>,
+    max_distance: Cell<f64>,
+    last_click: Cell<Instant>,
+    last_pos: Cell<Point>,
+    click_count: Cell<u8>,
+}
```
Philipp-M, line 153: Out of curiosity why a `Cell` everywhere? Easy access within non-mutable contexts? (it's not needed currently though)

Philipp-M (approved):
I could "fix" it by commenting out these two lines: https://github.com/gfx-rs/wgpu/blob/b8a8ff6e44162a41b8d54184a88613794dd3d2e4/wgpu-hal/src/vulkan/instance.rs#L967

And since these lines don't exist on the `trunk` branch on wgpu, it does work indeed with
```
[patch.crates-io]
wgpu = { git = 'https://github.com/gfx-rs/wgpu.git' }
```
in `Cargo.toml` (after updating/adjusting wgpu and vello in xilem).

When no one else has these issues (seems to happen on wayland + vulkan + maybe nvidia), I'd be ok to merge with this regression, as it otherwise looks good to me.

Maybe it also makes sense to add the lines above to `Cargo.toml` (but I guess this can be its own PR on top of this, including the update to wgpu and vello).

Philipp-M:
FWIW, here's my changes to get that working for me with wayland: https://github.com/Philipp-M/xilem/commit/2f069ce1237256f73148141884a94deae4dd8958

src/widget/raw_event.rs
xorgy: I just copied this from Glazier as-is. Seems needless in actual practice (since this will always be tracked at the window level, which is naturally synchronous).

xorgy:
> FWIW, here's my changes to get that working for me with wayland: [Philipp-M@9ab5c83](https://github.com/Philipp-M/xilem/commit/9ab5c832106b4f94253f1979f5273b3f5284b74a)

When this is merged, please put that up as a PR. :+ )

PR #177, opened 2024-02-23
Updated dependencies (notably wgpu and vello)

Philipp-M:
While reviewing #176, I had a deadlock in https://github.com/linebender/xilem/blob/21771ef71bd891afa045e08503022bf6b8b5e7fc/src/app_main.rs#L231

(see comments there)

This updates vello and wgpu (to current git HEAD where this deadlock cannot occur, since the relevant lines in wgpu are gone).

Since the surface is lifetime bounded now, it needs a little bit different initialization (the reason why it's already done in `MainState::new`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,9 +35,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.7"
+version = "0.8.9"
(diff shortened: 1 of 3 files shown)


xorgy:
LGTM. I bumped the rest of the things in `Cargo.lock` and popped that on the branch.

xorgy (approved):

Philipp-M:
> that requires bumping the CI Rust toolchain to 1.75.

I don't think we try to keep MSRV low currently (I think at least probably until TAIT is stabilized). So why not.

PR #178, opened 2024-02-29
Handle scroll events and use winit Modifiers and MouseButton

xorgy:


src/app_main.rs
```diff
@@ -75,72 +79,31 @@ impl<T: Send + 'static, V: View<T> + 'static> AppLauncher<T, V> {
...
-                    }
-                    if ms.contains(winit::keyboard::ModifiersState::ALT) {
-                        m |= Modifiers::ALT;
-                    }
-                    main_state.mods(m);
-                }
-                winit::event::Event::WindowEvent {
-                    event:
+            .run(move |event, elwt| {
+                if let winit::event::Event::WindowEvent { event: e, .. } = event {
```
Philipp-M, line 83: Just a thing/tip I like to do to, to avoid intense nesting for cases like this:

```suggestion
                let winit::event::Event::WindowEvent { event: e, .. } = event else {
                    return;
                };
```

Philipp-M (approved):
LGTM. I like how it got more compact/less verbose

src/app_main.rs
xorgy: The reason it's like this is because clippy was complaining about match with a single pattern. There will be cases where we need to handle things other than `WindowEvent` and this will return to being a `match`

PR #179, opened 2024-03-02
Bump vello and wgpu.

xorgy:


Philipp-M:
Hmm just tried it out, it freezes for me (same deadlock issue as before? (Haven't investigated that yet) Seems weird to me, since the release of wgpu 0.19.3 was just 19 hours ago?)

xStrom:
0.19.3, although recent, isn't equal to their `trunk` branch. So if we needed a fix from `trunk` to prevent the deadlock, we probably still need that.

xorgy:
> 0.19.3, although recent, isn't equal to their `trunk` branch. So if we needed a fix from `trunk` to prevent the deadlock, we probably still need that.

When I keep the override/patch in, cargo complains that it's not using it... which is interesting.

Philipp-M:
>  we probably still need that.

Yep, likely.

I haven't dug deeply, but it's freezing at the same place (`get_current_texture()`) as before...

xorgy:
> > we probably still need that.
> 
> Yep, likely.
> 
> I haven't dug deeply, but it's freezing at the same place (`get_current_texture()`) as before...

What platform are you on again? Do you have a thread for this in the Zulip?

Philipp-M:
> Do you have a thread for this in the Zulip?

No, relevant discussion is in #176

(including platform, which is Hyprland (wayland) + nvidia beta driver (550))

xorgy:
This is maybe an NVIDIA Vulkan WSI bug that we are running into because of the way we're using these APIs now... anyway @Philipp-M try the branch with this latest commit on the end. If a hot loop works for now then that's fine.

Philipp-M:
Hmm that unfortunately doesn't help (makes sense to me since the deadlock occurs inside wgpu already, as seen in the link on #176 I've posted).

xorgy:
@Philipp-M Yeah I mistook it for something else that I encountered with a (very outdated) Vulkan software renderer's WSI.

xorgy:
@Philipp-M Does this occur with non-beta/stable NV drivers?

xStrom:
The relevant deadlock issue is [wgpu#4967](https://github.com/gfx-rs/wgpu/pull/4967), which is a merged PR that fixes `wgpu`. Unfortunately it is a breaking change for `wgpu-hal` and so it won't be released as a patch for the v0.19 branch. It will arrive with `wgpu` v0.20 which will be released at approximately 2024-04-20.

I think we can merge this PR right now as-is, knowing that it will break Linux.

Then we will release Vello 0.1 any day now. After that is done, we could change Vello `main` so that it depends on the latest `wgpu` git rev, and then bump the Vello dep again here.

Having this fixed in Vello instead, with a git rev dep, is a more robust solution because `[patch]` only applies to the package that defines it. So if we use `[patch]` here for `xilem` then that won't help other Linux users of Vello, nor will it help Linux users who have `xilem` as a dependency. It will basically only help the Xilem examples being run from the Xilem workspace.

xStrom (approved):

Philipp-M:
Thanks for having looked that up. Yeah sounds like a plan. (If we don't want to create immediate releases for vello in the meantime, but I guess can switch wgpu versions back then)

xorgy:
> I think we can merge this PR right now as-is, knowing that it will break Linux.

To be clear, it works fine on Mesa on Linux, both on X11 and Wayland. It is only broken on NVIDIA's Vulkan stack, either due to a bug in their stack, or a coincidental safeguard in Mesa.

xorgy:
I went ahead and switched the target wgpu revision to a hash on trunk that incorporates some other interesting fixes, and which is compatible with Vello as-is.

Philipp-M (approved):
LGTM, and works for me.

PR #180, opened 2024-03-02
Fix tree-structure tracking in `AnyView` and `ScrollView`

Philipp-M:
I just noticed, that the scroll-view wasn't tracking the tree-structure, this should fix that.

Additionally I've removed the new view-id creation, as it isn' t necessary (yet).

~~I think there's a few gotchas here, we have to find more robust solutions in the future, regarding tree-structure tracking (and other lifecycle stuff).
E.g. with this, the relation is leaked when the scroll-view is removed (e.g. because it's encapsulated in an `Box<dyn AnyView>` and there was a type change).
I think we need some kind of lifecycle for cleaning up views/widgets.~~

Within the ViewSequences this is done externally to the view via `ViewSequence::rebuild`.

Edit: This should now be handled within the `AnyView`.

So I think now all cases are handled correctly. 

This actually does a bit more now than just fixing the `ScrollView`: Fixing `AnyView` tree-structure tracking.

### DIFF
--- a/src/widget/tree_structure.rs
+++ b/src/widget/tree_structure.rs
@@ -162,6 +162,7 @@ mod tests {
         assert_eq!(children, [child1]);
         assert_eq!(tree_structure.parent(child1), Some(parent));
         assert_eq!(tree_structure.parent.len(), 1);
+        assert_eq!(tree_structure.children.len(), 1);
     }
 
     #[test]
--- a/src/widget/tree_structure.rs
+++ b/src/widget/tree_structure.rs
@@ -1,5 +1,5 @@
 use crate::id::Id;
-use std::collections::HashMap;
+use std::{collections::HashMap, ops::RangeBounds};
 
 /// The pure structure (parent/children relations via ids) of the widget tree.
 #[derive(Debug, Default, Clone)]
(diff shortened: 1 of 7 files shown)


src/widget/tree_structure.rs
```diff
@@ -62,15 +62,32 @@ impl TreeStructure {
...
 
-    /// # Panics
-    ///
-    /// When the `parent_id` doesn't exist in the structure or `range` is out of bounds this will panic
-    pub(crate) fn delete_children(&mut self, parent_id: Id, range: std::ops::Range<usize>) {
-        let children = &self.children[&parent_id][range.clone()];
-        for child in children {
-            self.parent.remove(child);
+    /// Deletes all children of `parent_id`, but not their descendants
+    pub(crate) fn delete_children(&mut self, parent_id: Id, range: impl RangeBounds<usize>) {
```
Philipp-M, line 66: This is not needed anymore and I'm not sure if there's a case where this is relevant anymore (it's slightly more efficient than `delete_descendants` as it doesn't require an allocation, but this is only relevant if there are no further descendants, otherwise it could be a pitfall)

Should I rename `delete_descendants` to `delete_children` then (when `delete_children` is removed)?

Philipp-M: Answering my own question: I have removed that for now, as it's well not used, and I don't see an immediate need for it (and CI also wants it to be removed...)

Philipp-M:
I guess this is a little bit out of date and not really mergeable anymore since it targets the old xilem(_core).

PR #182, opened 2024-03-04
invert scroll direction on macos

ijsnow:
The current scroll direction is the opposite of what mac users expect, so this PR inverts it when `target_os = "macos"`.

### DIFF
--- a/src/app_main.rs
+++ b/src/app_main.rs
@@ -171,11 +171,11 @@ where
         self.app
             .window_event(Event::MouseWheel(self.main_pointer.wheel(match delta {
                 MouseScrollDelta::LineDelta(x, y) => {
-                    ScrollDelta::Lines(x.trunc() as isize, y.trunc() as isize)
+                    ScrollDelta::Lines(-x.trunc() as isize, -y.trunc() as isize)
                 }
-                MouseScrollDelta::PixelDelta(PhysicalPosition { x, y }) => {
-                    ScrollDelta::Precise(Vec2::new(x, y) * (1.0 / self.window.scale_factor()))
-                }
+                MouseScrollDelta::PixelDelta(PhysicalPosition { x, y }) => ScrollDelta::Precise(
+                    Vec2::new(-x, -y) * (1.0 / self.window.scale_factor()),
+                ),
             })));
         self.window.request_redraw();
     }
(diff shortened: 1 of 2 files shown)


DJMcNab:
Have you confirmed that this is still as expected when you toggle the "natural" scrolling setting?

xStrom:
This is configurable in the OS, right? Do the events we receive come with a flipped sign then, or do we need to find the value for this configuration setting ourselves?

ijsnow:
Oooooh no I wasn't aware that was configurable

ijsnow:
I'm looking around to see how we can make sure we get the right info from the windowing library or if its inadvertently changed in `xilem` right now. It looks like `floem` gets the right info

ijsnow:
Ok so perhaps this is a discussion outside of a PR but it looks like just inverting the pixels is what both [floem](https://github.com/lapce/floem/blob/main/src/window_handle.rs#L433-L437) and [iced](https://github.com/iced-rs/iced/blob/master/widget/src/scrollable.rs#L1260) do. I've tested it on my mac and in a ubuntu virtual box and it feels correct based on testing against other applications. This also allows the device to dictate the direction of the delta based on the natural scroll system setting on mac. I don't know how this affects windows scrolling, but I imagine the developers of floem (with the lapce editor that has windows support) and iced have come across windows support already.

DJMcNab:
Can you fix the title and description?

To confirm, is this change now because winit uses a positive y is up coordinate system, whereas we use a positive y is down one? (Or the other way around)

I'm not sure why the x is being inverted in that case

xStrom:
I tested this on Windows and as the PR stands right now it incorrectly inverts scrolling.

xStrom:
I also tested `iced`. It works correctly on both Windows and macOS, and respects the macOS natural scrolling setting. So the correct solution should be discoverable by inspecting the `iced` code.

Unfortunately my macOS setup can't currently run `vello` so I can't personally confirm that this is even broken in Xilem on macOS.

I will also note that one somewhat relevant historic thread is [winit#1695](https://github.com/rust-windowing/winit/issues/1695).

ijsnow:
I'll just close this and open an issue. It would be better to start fresh. But yea I'm a little confused at why its not working when I was looking around it seemed like it would be handled with winit, or at least the directions would be different based on the different platforms thanks to winit.

PR #184, opened 2024-03-09
clippy: Fix semicolon_if_nothing_returned lints.

waywardmonkeys:
This doesn't enable the lint as I have a subsequent follow-up PR coming for that.


Philipp-M:
> This doesn't enable the lint as I have a subsequent follow-up PR coming for that.

Why not within this PR?

waywardmonkeys:
> Why not within this PR?

Because it is a new approach and deserves discussion on its own!

xStrom (approved):

PR #185, opened 2024-03-09
clippy: Configure via `lints` table in `Cargo.toml`

waywardmonkeys:
As of Rust 1.74, lints can be configured within the `Cargo.toml` which allows us to not have to configure them in the source code as well as simplifying having a single configuration across an entire workspace.

This is documented at: https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-lints-section

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,9 @@ license = "Apache-2.0"
 xilem_core = { version = "0.1.0", path = "crates/xilem_core" }
 kurbo = "0.11.0"
 
+[workspace.lints]
+clippy.semicolon_if_nothing_returned = "warn"
+
 [package]
 name = "xilem"
 version = "0.1.0"
@@ -38,6 +41,9 @@ rustdoc-args = ["--cfg", "docsrs"]
 default-target = "x86_64-pc-windows-msvc"
 # rustdoc-scrape-examples tracking issue https://github.com/rust-lang/rust/issues/88791
 cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
+
+[lints]
+workspace = true
 
 [features]
 default = ["x11", "taffy"]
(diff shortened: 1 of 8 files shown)


xStrom (approved):
Looks reasonable.

Cargo.toml
```diff
@@ -19,6 +19,9 @@ license = "Apache-2.0"
 xilem_core = { version = "0.1.0", path = "crates/xilem_core" }
 kurbo = "0.11.0"
 
+[workspace.lints]
+clippy.semicolon_if_nothing_returned = "warn"
```
DJMcNab, line 23: I would expect this to use:
```toml
[workspace.lints.clippy]
semicolon_if_nothing_returned = "warn"
```
As this is how it is always documented. Is there a specific reason we chose not to follow this example here?

DJMcNab (commented):
This is incredibly cool, thanks. I don't know how I missed this feature!

I guess this isn't backwards compatible with Rust versions? I think this is a good enough reason to bump MSRVs, though

Cargo.toml
xStrom: From my review perspective I figured that we can change it later when we have more lints, as right now it doesn't make much of a difference.

waywardmonkeys: I've seen it used both ways in the wild. It isn't just clippy that can be controlled here, but rustc warnings as well. I don't care though. I'd written it the way first and changed it before submitting.

waywardmonkeys:
> I guess this isn't backwards compatible with Rust versions? I think this is a good enough reason to bump MSRVs, though

No need for an MSRV bump! Vello's MSRV has this in it and since Xilem doesn't define an MSRV yet, I figured since it typically uses Vello, it would follow along...

DJMcNab:
Yeah - I mean for applying this in e.g. Kurbo and Peniko, for example.

Cargo.toml
DJMcNab: I think following the documentation makes sense here, but it's not that important either way

PR #186, opened 2024-03-10
Mark pod correctly for view sequence

IsseW:
From what I can tell only the last pod in a view sequence was correctly marked when `ChangeFlags::Paint` was requested. This should mark the correct pod and not just the last one.

### DIFF
--- a/crates/xilem_core/src/sequence.rs
+++ b/crates/xilem_core/src/sequence.rs
@@ -173,7 +173,7 @@ macro_rules! generate_viewsequence_trait {
                         el,
                     )
                 });
-                elements.mark(flags, cx)
+                pod.mark(flags)
             }
 
             fn message(


Philipp-M:
Have you noticed any issues?

The `elements.mark()` is necessary, otherwise no tree-structure updates go through.

This code should be run per view, and `elements.mark` takes the *current* last `Pod` to mark changes (which should already be `pod` in that code).
Everytime `mutate` (a few lines earlier) is run, a new pod (current of the view) is added to the elements.

IsseW:
> Have you noticed any issues?

The issue I noticed was for rebuild, where the current last pod is always the last pod. If it's mplemented like it is for `LinearLayout`. Of course I could've missed something though, still pretty new to learning about how xilem works.

IsseW:
> The `elements.mark()` is necessary, otherwise no tree-structure updates go through.
I can see this now with what `TreeStructureSplice` is doing. Only considered what `VecSplice` did earlier... But `TreeStructureSplice` still calls `VecSplice` which will only mark the last element.

Philipp-M:
Can you check if #187 fixes the issue?

IsseW:
Closing in favor of https://github.com/linebender/xilem/pull/186. And moving discussion there

PR #187, opened 2024-03-10
Fix peek in `VecSplice`

Philipp-M:
Fixes #186 

`v` in `VecSplice` can be the original vector, while `ix` is still pointing to the first element, this should fix that. (`peek` and `peek_mut` should point to the current `ix` element)

### DIFF
--- a/crates/xilem_core/src/vec_splice.rs
+++ b/crates/xilem_core/src/vec_splice.rs
@@ -51,11 +51,11 @@ impl<'a, 'b, T> VecSplice<'a, 'b, T> {
     }
 
     pub fn peek(&self) -> Option<&T> {
-        self.v.last()
+        self.v.get(self.ix)
     }
 
     pub fn peek_mut(&mut self) -> Option<&mut T> {
-        self.v.last_mut()
+        self.v.get_mut(self.ix)
     }
 
     pub fn len(&self) -> usize {
--- a/crates/xilem_core/src/vec_splice.rs
+++ b/crates/xilem_core/src/vec_splice.rs
@@ -50,12 +50,20 @@ impl<'a, 'b, T> VecSplice<'a, 'b, T> {
         &mut self.v[ix]
     }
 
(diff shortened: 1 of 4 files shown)


crates/xilem_core/src/vec_splice.rs
```diff
@@ -51,11 +51,11 @@ impl<'a, 'b, T> VecSplice<'a, 'b, T> {
     }
 
     pub fn peek(&self) -> Option<&T> {
-        self.v.last()
+        self.v.get(self.ix)
```
IsseW, line 54: Ah, just found this locally too :grinning:  thanks!

I think this should be `ix - 1` though, since we add 1 to `ix` in mutate

IsseW: Maybe peek isn't the best name for this method either, as what it's used for is look at the last element

Philipp-M: > I think this should be `ix - 1` though, since we add 1 to `ix` in mutate

You're right, just checked it again, `ix` is incremented after accessing it.

> Maybe peek isn't the best name for this method either

You're probably right. Maybe call it `current` instead? I mean `peek` is not totally inaccurate, as it should return the last modified element (if not yet clear, `VecSplice` sweeps through the elements). But I see that it may confuse (additionally to the already quite complex code of `VecSplice`)

IsseW: `current` feels better, but maybe something like `last_mutated` would be the most descriptive.

Philipp-M: Yes makes sense, I've renamed it to `last_mutated` and fixed the index access.

IsseW (approved):
Looks good! This works for the case I had locally too

DJMcNab (approved):
I've not dug into this, but I'll formalise @IsseW's approval

PR #191, opened 2024-03-15
Uninvert the scroll direction

### LINKED ISSUE #183: Scrolling is inverted
Scrolling appears to be inverted for me in the current examples. This is potentially complicated by the fact that I am using macOS with "natural scrolling" enabled (so drag down on touchpad to scroll up and vice versa). But I believe this should be transparent to the application (and even Winit).

DJMcNab:
Fixes #183

Continues #182

The terminology around this is unclear, but I've tried to add hinting comments as well

### DIFF
--- a/src/app_main.rs
+++ b/src/app_main.rs
@@ -171,11 +171,11 @@ where
         self.app
             .window_event(Event::MouseWheel(self.main_pointer.wheel(match delta {
                 MouseScrollDelta::LineDelta(x, y) => {
-                    ScrollDelta::Lines(x.trunc() as isize, y.trunc() as isize)
+                    ScrollDelta::Lines(-x.trunc() as isize, -y.trunc() as isize)
                 }
-                MouseScrollDelta::PixelDelta(PhysicalPosition { x, y }) => {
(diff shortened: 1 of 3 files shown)


xStrom:
I can confirm that vertical scrolling still works as expected on Windows. The `taffy` example doesn't seem to provide horizontal scrolling right now, so no idea about that.

ratmice:
I'm pretty confused as I appear to get the same (correct) behavior both with & without this patch
running the taffy example under sway/wayland.  I would have expected the `pointer_wheel()` changes to have had some effect.

Don't want to complain about things working correctly, just mysterious.

xStrom:
I also had correct behavior before this patch, and still do with it applied. The important question is if this patch has an impact on people who previously experienced inverted scrolling.

DJMcNab:
Ah, that will be because your OS is sending a line delta, of course. That is, you have a coarse scroll mouse wheel.

rictic:
Can confirm that this fixes scrolling on a macbook's built-in trackpad. Without this change, the taffy example's scrolling behavior is inverted, and with this change it matches the behavior of other scrollables. Tested with Natural Scrolling enabled and diabled (the OS-level preference).

xStrom (approved):
Approving it based on the success reports.

PR #193, opened 2024-03-24
add class and style interfaces

derekdreery:
This PR adds more typed interfaces for adding classes and styles, using the DOM APIs under the hood.

I'm opening the PR for discussion. If it's accepted, I'd like to do some tidying up and adding tests before merging.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ members = [
     "crates/xilem_web/web_examples/counter_custom_element",
     "crates/xilem_web/web_examples/todomvc",
     "crates/xilem_web/web_examples/mathml_svg",
-    "crates/xilem_web/web_examples/svgtoy", 
+    "crates/xilem_web/web_examples/svgtoy",
 ]
 
 [workspace.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,6 @@
 members = [
     "crates/xilem_core",
     "crates/xilem_web",
-    "crates/xilem_web/xilem_web_spec",
     "crates/xilem_web/web_examples/counter",
     "crates/xilem_web/web_examples/counter_custom_element",
     "crates/xilem_web/web_examples/todomvc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,11 +2,12 @@
 members = [
     "crates/xilem_core",
     "crates/xilem_web",
+    "crates/xilem_web/xilem_web_spec",
     "crates/xilem_web/web_examples/counter",
     "crates/xilem_web/web_examples/counter_custom_element",
     "crates/xilem_web/web_examples/todomvc",
     "crates/xilem_web/web_examples/mathml_svg",
-    "crates/xilem_web/web_examples/svgtoy",
+    "crates/xilem_web/web_examples/svgtoy", 
 ]
 
 [workspace.package]
--- a/crates/xilem_web/src/class.rs
+++ b/crates/xilem_web/src/class.rs
@@ -24,6 +24,12 @@ impl IntoClasses for &'static str {
     }
 }
 
+impl IntoClasses for Cow<'static, str> {
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        classes.push(self);
+    }
+}
+
 impl<T> IntoClasses for Option<T>
 where
     T: IntoClasses,
--- a/crates/xilem_web/src/class.rs
+++ b/crates/xilem_web/src/class.rs
@@ -1,0 +1,52 @@
+use std::{borrow::Cow, marker::PhantomData};
+
+use xilem_core::{Id, MessageResult};
+
+use crate::{
+    interfaces::{sealed::Sealed, Element},
+    ChangeFlags, Cx, View, ViewMarker,
+};
+
+/// Applies a class to the underlying element.
+pub struct Class<E, T, A> {
+    pub(crate) element: E,
+    pub(crate) class_name: Cow<'static, str>,
+    pub(crate) phantom: PhantomData<fn() -> (T, A)>,
+}
+
+impl<E, T, A> ViewMarker for Class<E, T, A> {}
+impl<E, T, A> Sealed for Class<E, T, A> {}
+
+impl<E: Element<T, A>, T, A> View<T, A> for Class<E, T, A> {
+    type State = E::State;
+    type Element = E::Element;
+
+    fn build(&self, cx: &mut Cx) -> (Id, Self::State, Self::Element) {
+        cx.add_class_to_element(&self.class_name);
+        self.element.build(cx)
+    }
+
+    fn rebuild(
+        &self,
+        cx: &mut Cx,
+        prev: &Self,
+        id: &mut Id,
+        state: &mut Self::State,
+        element: &mut Self::Element,
+    ) -> ChangeFlags {
+        cx.add_class_to_element(&self.class_name);
+        self.element.rebuild(cx, &prev.element, id, state, element)
+    }
+
+    fn message(
+        &self,
+        id_path: &[Id],
+        state: &mut Self::State,
+        message: Box<dyn std::any::Any>,
+        app_state: &mut T,
+    ) -> MessageResult<A> {
+        self.element.message(id_path, state, message, app_state)
+    }
+}
+
+crate::interfaces::impl_dom_interfaces_for_ty!(Element, Class);
--- a/crates/xilem_web/src/class.rs
+++ b/crates/xilem_web/src/class.rs
@@ -7,10 +7,72 @@ use crate::{
     ChangeFlags, Cx, View, ViewMarker,
 };
 
+/// A trait to make the class adding functions generic over collection type
+pub trait IntoClasses {
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>);
+}
+
+impl IntoClasses for String {
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        classes.push(self.into());
+    }
+}
+
+impl IntoClasses for &'static str {
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        classes.push(self.into())
+    }
+}
+
+impl<T> IntoClasses for Option<T>
+where
+    T: IntoClasses,
+{
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        if let Some(t) = self {
+            t.into_classes(classes)
+        }
+    }
+}
+
+impl<T> IntoClasses for Vec<T>
+where
+    T: IntoClasses,
+{
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        for itm in self {
+            itm.into_classes(classes);
+        }
+    }
+}
+
+macro_rules! impl_tuple_intoclasses {
+    ($($name:ident : $type:ident),* $(,)?) => {
+        impl<$($type),*> IntoClasses for ($($type,)*)
+        where
+            $($type: IntoClasses),*
+        {
+            #[allow(unused_variables)]
+            fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+                let ($($name,)*) = self;
+                $(
+                    $name.into_classes(classes);
+                )*
+            }
+        }
+    };
+}
+
+impl_tuple_intoclasses!();
+impl_tuple_intoclasses!(t1: T1);
+impl_tuple_intoclasses!(t1: T1, t2: T2);
+impl_tuple_intoclasses!(t1: T1, t2: T2, t3: T3);
+impl_tuple_intoclasses!(t1: T1, t2: T2, t3: T3, t4: T4);
+
 /// Applies a class to the underlying element.
 pub struct Class<E, T, A> {
     pub(crate) element: E,
-    pub(crate) class_name: Option<Cow<'static, str>>,
+    pub(crate) class_names: Vec<Cow<'static, str>>,
     pub(crate) phantom: PhantomData<fn() -> (T, A)>,
 }
 
@@ -22,7 +84,7 @@ impl<E: Element<T, A>, T, A> View<T, A> for Class<E, T, A> {
     type Element = E::Element;
 
     fn build(&self, cx: &mut Cx) -> (Id, Self::State, Self::Element) {
-        if let Some(class_name) = &self.class_name {
+        for class_name in &self.class_names {
             cx.add_class_to_element(class_name);
         }
         self.element.build(cx)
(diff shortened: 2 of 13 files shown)


crates/xilem_web/src/context.rs
```diff
@@ -41,12 +41,52 @@ fn remove_attribute(element: &web_sys::Element, name: &str) {
...
+fn set_class(element: &web_sys::Element, class_name: &str) {
+    #[cfg(debug_assertions)]
+    if class_name.is_empty() {
+        panic!("class names cannot be the empty string");
+    }
+    #[cfg(debug_assertions)]
+    if class_name.contains(' ') {
+        panic!("class names cannot contain the ascii space character");
+    }
+    element.class_list().add_1(class_name).unwrap_throw()
```
Philipp-M, line 53: [Something, I have observed](https://www.measurethat.net/Benchmarks/Show/30272/0/classname-vs-classlist-with-existing-classes), is that `className` seems to be (significantly) faster than `classList`. Since we have full control over which classes that element uses, it maybe makes sense to use `className` instead and concatenate all the classes when a change occurs there?

crates/xilem_web/src/context.rs
```diff
@@ -41,12 +41,52 @@ fn remove_attribute(element: &web_sys::Element, name: &str) {
...
+}
+
 // Note: xilem has derive Clone here. Not sure.
 pub struct Cx {
     id_path: IdPath,
     document: Document,
     // TODO There's likely a cleaner more robust way to propagate the attributes to an element
     pub(crate) current_element_attributes: VecMap<CowStr, AttributeValue>,
+    pub(crate) current_element_classes: VecMap<CowStr, ()>,
+    pub(crate) current_element_styles: VecMap<CowStr, CowStr>,
```
Philipp-M, line 89: Not that it is particularly relevant to this PR, more a brain-dump/thoughts about it, I'd like to share.

I'm more and more thinking about getting the temporary state of elements out of the context again, likely via associated type bounds as described [here](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/xilem_html.3A.20strong.20typing.20.28attributes.20and.20elements.29/near/387972736), since they're [stabilized](https://github.com/rust-lang/rust/pull/122055) now. This would allow (safe) implementations downstream (i.e. remove the "need" for the `Sealed` trait in the interfaces), better separation of concerns (especially if there's a lot of logic added for all kinds of dom elements), possibly it would also help stripping down to only relevant code (e.g. when style is not used, it could be stripped away in the resulting binary).
And generally make everything cleaner I think. I think it's probably worth the added compilation times. But this probably needs deeper investigation/experimentation (again).

crates/xilem_web/src/context.rs
```diff
@@ -145,26 +187,77 @@ impl Cx {
...
     ) -> ChangeFlags {
         self.apply_attribute_changes(element, attributes)
+            | self.apply_class_changes(element, classes)
+            | self.apply_style_changes(element, styles)
     }
 
     // TODO Not sure how multiple attribute definitions with the same name should be handled (e.g. `e.attr("class", "a").attr("class", "b")`)
     // Currently the outer most (in the example above "b") defines the attribute (when it isn't `None`, in that case the inner attr defines the value)
     pub(crate) fn add_attr_to_element(&mut self, name: &CowStr, value: &Option<AttributeValue>) {
+        // Special-case class so it works with the `class` method
```
Philipp-M, line 221: Interesting idea, I'm not sure whether we should specially handle attributes that are given here, since that would apply to potentially a *lot* of attributes (when they're implemented somewhat in the future). Such as `circle().fill()` (which would bloat the wasm binary size).

Maybe we should explicitly tell the user, that there's either the possibility via `attr` (which would then be discouraged) or typed methods (such as `el.class()`). And that using both at the same time will result in undefined behavior or overwrites previous set values. But I'm not sure about that. We could also introduce a warning (in debug builds?) or even an error when `attr("class")` is used.

crates/xilem_web/src/context.rs
```diff
@@ -145,26 +187,77 @@ impl Cx {
...
+                        self.current_element_classes
+                            .insert(class_name.to_string().into(), ());
+                    }
+                }
+            }
+            return;
+        }
+
+        // parse styles
+        if name == "style" {
```
Philipp-M, line 238: Same as above

crates/xilem_web/src/interfaces.rs
```diff
@@ -91,11 +91,45 @@ where
...
+    ///
+    /// If multiple classes are added, all will be applied to the element.
+    fn class(self, class: impl Into<Cow<'static, str>>) -> Class<Self, T, A> {
+        self.class_opt(Some(class))
+    }
+
+    /// Add an optional class to the wrapped element.
+    ///
+    /// If multiple classes are added, all will be applied to the element.
+    fn class_opt(self, class: Option<impl Into<Cow<'static, str>>>) -> Class<Self, T, A> {
```
Philipp-M, line 104: Not sure if we should maybe achieve this via a trait like I've experimented [here](https://github.com/Philipp-M/xilem_html-typing-tests/blob/5c831060d21c22e46b4f8ce471f2ec94e0230b60/examples/extension_traits-and-attr-map.rs#L196), which would allow adding multiple classes at once in more structured manner. 

Also as the previous `TODO` comment suggested, it would be nice to be able to remove classes as well (but I think `el.class()` should indeed be additive).

crates/xilem_web/src/interfaces.rs
```diff
@@ -91,11 +91,45 @@ where
...
+        Class {
+            element: self,
+            class_name: class.map(Into::into),
+            phantom: PhantomData,
+        }
+    }
+
+    /// Set a style attribute
+    fn style(
```
Philipp-M, line 113: Same here. I could imagine something like this `el.style([("position", "absolute"), ("background", "black")])` or similar as this method `el.style(("position", "absolute"))`.

But yeah I'm generally not sure about the slightly added complexity that a trait introduces to achieve that syntax sugar. But it would e.g. allow custom structs to implement that style trait.

crates/xilem_web/src/svg/kurbo_shape.rs
```diff
@@ -29,32 +29,36 @@ impl ViewMarker for Line {}
 impl Sealed for Line {}
 
 impl<T, A> View<T, A> for Line {
-    type State = VecMap<Cow<'static, str>, AttributeValue>;
+    type State = (
+        VecMap<Cow<'static, str>, AttributeValue>,
+        VecMap<Cow<'static, str>, ()>,
+        VecMap<Cow<'static, str>, Cow<'static, str>>,
+    );
```
Philipp-M, line 36: Maybe we can create extra types (`ElementState`) for (sub) states of `View`s implementing e.g. the `Element` trait, to avoid doing all this for all kinds of types implementing these interfaces?

Philipp-M (commented):
I've added a few comments.

Since I've generally also experimented on this, [this](https://github.com/Philipp-M/xilem/commit/039888a127723c687143b8a7eea02e0731ea6eac#diff-7f94efe64ff3601d7336879db6879f4b9d749789ba33edd2574df4a8cbc8795e) can maybe a source of inspiration as well.

When I remember right, I stopped experimenting with this, because of the added complexity (and resulting code-size). Where I'd think that a more static way (as described in the comment with the assoc state bounds) is more future-proof and less bloaty/easier to strip/optimize by the compiler while being easier to reason as well probably. But this is speculative.

crates/xilem_web/src/interfaces.rs
derekdreery: Aah I quite like this approach. I will revise the PR to use it.

crates/xilem_web/src/interfaces.rs
derekdreery: I think it's worth it because the alternative is more wrapper structs (`Class<Class<Style<Style<Style<Attr<...<Element>..>>>>>`)

crates/xilem_web/src/svg/kurbo_shape.rs
derekdreery: Yep let's do that.

crates/xilem_web/src/context.rs
derekdreery: I propose that se use `classList` to start with, then do an experiment where we compare the two approaches (and then use the fastest).

derekdreery:
Thanks for the review, and for linking to the great conversations that I missed on zulip. I'll do another draft based on the comments when I get chance.

crates/xilem_web/src/interfaces.rs
derekdreery: So you'd probably have `T`, `Option<T>`, `Vec<T>` perhaps for `T: impl Into<Cow<'static, str>>`

Philipp-M: I could imagine even more, but something along those lines yeah. Btw. you're talking about `class`? As style needs key-values? The array/container types could act directly on the trait like `Vec<T> where T: IntoClass`. But I'm not sure about stuff like `el.style((("position", "absolute"), CustomStyle::new()))` which is probably difficult to achieve via blanket-impls. 

Another idea regarding styling I have, is to use an enum for more structured styling, Something like:

```rust
enum StyleAttr {
    Background(BackgroundStyle),
    Position(PositionStyle),
} 
```
etc.

But this could be quite a lot of boilerplate to maintain (including the relevant syntax-sugar logic via traits), so not sure about that... Maybe we should iterate slowly towards more static/structured ways to achieve it (while always comparing whether it improves the current state)?

derekdreery:
Hi @Philipp-M I've made some of the changes you suggested, and in particular went in ot the trait-based polymorphism approach. Let me know what your thoughts are.

Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
     "crates/xilem_web/web_examples/counter_custom_element",
     "crates/xilem_web/web_examples/todomvc",
     "crates/xilem_web/web_examples/mathml_svg",
-    "crates/xilem_web/web_examples/svgtoy",
+    "crates/xilem_web/web_examples/svgtoy",
```
Philipp-M, line 9:
```suggestion
    "crates/xilem_web/web_examples/svgtoy",
```

crates/xilem_web/src/class.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        classes.push(self.into());
+    }
+}
+
+impl IntoClasses for &'static str {
+    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
+        classes.push(self.into());
+    }
+}
```
Philipp-M, line 25:
```suggestion
}

impl IntoClasses for Cow<'static, str> {
    fn into_classes(self, classes: &mut Vec<Cow<'static, str>>) {
        classes.push(self);
    }
}
```

(Simply because it's a low-hanging fruit)

crates/xilem_web/src/context.rs
```diff
@@ -41,12 +166,50 @@ fn remove_attribute(element: &web_sys::Element, name: &str) {
...
+        "class names cannot contain the ascii space character"
+    );
+    element.class_list().remove_1(class_name).unwrap_throw();
+}
+
+fn set_style(element: &web_sys::Element, name: &str, value: &str) {
+    // styles will be ignored for non-html elements (e.g. SVG)
+    if let Some(el) = element.dyn_ref::<web_sys::HtmlElement>() {
+        el.style().set_property(name, value).unwrap_throw();
+    }
```
Philipp-M, line 197:
```suggestion
    if let Some(el) = element.dyn_ref::<web_sys::HtmlElement>() {
        el.style().set_property(name, value).unwrap_throw();
    } else if let Some(el) = element.dyn_ref::<web_sys::SvgElement>() {
        el.style().set_property(name, value).unwrap_throw();
    }
```

SVGElements do also support style.

Just the base DOM interface (`Element`) doesn't, so to be completely accurate (as e.g. `MathML` AFAIK doesn't support `style`) it would make sense to put `Element::style` in `HtmlElement::style` and `SvgElement::style` (`methods { fn style(..) {..} }` in `dom_interface_macro_and_trait_definitions`)

crates/xilem_web/src/context.rs
```diff
@@ -41,12 +166,50 @@ fn remove_attribute(element: &web_sys::Element, name: &str) {
...
+}
+
+fn remove_style(element: &web_sys::Element, name: &str) {
+    // styles will be ignored for non-html elements (e.g. SVG)
+    if let Some(el) = element.dyn_ref::<web_sys::HtmlElement>() {
+        el.style().remove_property(name).unwrap_throw();
+    }
```
Philipp-M, line 204:
```suggestion
    if let Some(el) = element.dyn_ref::<web_sys::HtmlElement>() {
        el.style().remove_property(name).unwrap_throw();
    } else if let Some(el) = element.dyn_ref::<web_sys::SvgElement>() {
        el.style().remove_property(name).unwrap_throw();
    }
```

Philipp-M (commented):
Looks good, thanks :)
Just a few comments, but nothing fancy.

crates/xilem_web/src/style.rs
```diff
@@ -0,0 +1,152 @@
...
+
+/// A trait to make the class adding functions generic over collection type
+pub trait IntoStyles {
+    fn into_styles(self, styles: &mut Vec<(Cow<'static, str>, Cow<'static, str>)>);
+}
+
+struct StyleTuple<T1, T2>(T1, T2);
+
+/// Create a style from a style name and its value.
+pub fn style<T1, T2>(name: T1, value: T2) -> impl IntoStyles
```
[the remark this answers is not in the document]
derekdreery: The idea of i'm here is to keep the number of types exported to a smaller set.

crates/xilem_web/src/interfaces.rs
```diff
@@ -91,11 +95,30 @@ where
...
+
+    /// Set a style attribute
+    fn style(self, style: impl IntoStyles) -> Style<Self, T, A> {
+        let mut styles = vec![];
+        style.into_styles(&mut styles);
+        Style {
+            element: self,
+            styles,
+            phantom: PhantomData,
+        }
```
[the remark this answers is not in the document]
derekdreery: I think there are opportunities to save allocations, but I suggest we push that to a separate PR.

Philipp-M (approved):
Looks good, thanks. Just this comment (was probably missed):

> ... so to be completely accurate (as e.g. MathML AFAIK doesn't support style) it would make sense to put `Element::style` in `HtmlElement::style` and `SvgElement::style` (`methods { fn style(..) {..} } in dom_interface_macro_and_trait_definitions`)

But I can do that in a follow-up PR too and it's not super important

derekdreery:
> Looks good, thanks. Just this comment (was probably missed):
> 
> > ... so to be completely accurate (as e.g. MathML AFAIK doesn't support style) it would make sense to put `Element::style` in `HtmlElement::style` and `SvgElement::style` (`methods { fn style(..) {..} } in dom_interface_macro_and_trait_definitions`)
> 
> But I can do that in a follow-up PR too and it's not super important

I'm not 100% sure I know exactly what you're asking for. Perhaps you could do this in a follow-up PR? :)

PR #195, opened 2024-03-31
xilem_web: Allow `style` only on `HtmlElement` and `SvgElement`

Philipp-M:
According to the spec, `style` is not available on `Element` itself, but it is on `SVGElement` and `HTMLElement`.


derekdreery:
Thanks for doing this instead of getting me to do it - I think it was more efficient that way. :)

derekdreery (approved):

PR #196, opened 2024-04-02
Update parley dependency to bring in new font fallback support

dfrg:
This pulls in a new version of parley with fount replaced by fontique.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -13,7 +13,6 @@ pub fn render_text(scene: &mut Scene, transform: Affine, layout: &Layout<Brush>)
             let run = glyph_run.run();
             let font = run.font();
             let font_size = run.font_size();
-            let font = vello::peniko::Font::new(font.data().0.clone(), font.index());
             let style = glyph_run.style();
             let coords = run
                 .normalized_coords()
@@ -21,7 +20,7 @@ pub fn render_text(scene: &mut Scene, transform: Affine, layout: &Layout<Brush>)
                 .map(|coord| vello::skrifa::instance::NormalizedCoord::from_bits(*coord))
                 .collect::<Vec<_>>();
             scene
-                .draw_glyphs(&font)
+                .draw_glyphs(font)
                 .brush(&style.brush)
                 .transform(transform)
                 .font_size(font_size)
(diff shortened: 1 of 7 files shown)


src/text.rs
```diff
@@ -13,17 +13,21 @@ pub fn render_text(scene: &mut Scene, transform: Affine, layout: &Layout<Brush>)
             let run = glyph_run.run();
             let font = run.font();
             let font_size = run.font_size();
-            let font = vello::peniko::Font::new(font.data().0.clone(), font.index());
             let style = glyph_run.style();
+            let synthesis = run.synthesis();
+            let glyph_xform = synthesis
+                .skew()
+                .map(|angle| Affine::skew(angle.to_radians().tan() as f64, 0.0));
```
nicoburns, line 20: Is this doing "fake italics"?

src/widget/button.rs
```diff
@@ -89,6 +89,9 @@ impl Widget for Button {
         layout_builder.push_default(&parley::style::StyleProperty::Brush(Brush::Solid(
             Color::rgb8(0xf0, 0xf0, 0xea),
         )));
+        layout_builder.push_default(&parley::style::StyleProperty::FontStack(
+            parley::style::FontStack::Source("system-ui"),
+        ));
```
nicoburns, line 94: We should probably start re-exporting parley types at the top-level. I've found that this really helps when consuming Taffy and when the crate is small like Taffy/Parley are there's little reason not to have everything in the top-level namespace.

And/or making use of `use` when type names are reasonably unique like in this case.

nicoburns (approved):
Looks good to me.

src/text.rs
dfrg: Yeah, this always existed in parley but I never hooked it up in xilem. Dropping fount (and swash as a font parser for the fallback engine) required a new synthesis type and some conversions so I decided to add the integration to xilem to make sure it all still works.

src/widget/button.rs
dfrg: This definitely makes sense for the style types. I'll keep this in mind when we move toward release.

src/text.rs
nicoburns: Awesome. I don't suppose "fake bold" would be as easy? Not a blocker for me, but would definitely be nice to have.

dfrg: Almost as easy. The synthesis object notes when it should be applied but we don’t yet have code in vello to dilate the outlines. Ideally we’d like to do this on the GPU at some point but we can and should probably hack it in on the CPU side for completeness until that is done.

PR #197, opened 2024-04-11
xilem_web: Factor element state out of the view context via associated type bounds

Philipp-M:
Based on [this comment](https://github.com/linebender/xilem/pull/193#discussion_r1537625351) and my rather unsatisfying experiments in https://github.com/Philipp-M/xilem/commit/039888a127723c687143b8a7eea02e0731ea6eac

I have removed the need for intermediate state in the view context via a trait `ElementProps` (and in the future likely traits for other DOM interfaces).
This requires a new Rust feature `associated_type_bounds` which is not yet in stable Rust, but will be in 1.79 (it unfortunately just missed the window for 1.78).
So this will likely stay a draft PR until then as there's no hurry, and the [workaround](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/xilem_html.3A.20strong.20typing.20.28attributes.20and.20elements.29/near/387972736) for it is not worth it in the meantime I think, but I wanted to at least make it public (to avoid possibly merge conflicts etc.).

This PR should be seen as a base for future typed features of DOM elements (such as seen in the [video example](https://github.com/Philipp-M/xilem/blob/039888a127723c687143b8a7eea02e0731ea6eac/crates/xilem_html/web_examples/play_video/src/main.rs)), to avoid putting all the state for all kinds of elements in the context (which will likely blow up WASM blob size).


derekdreery:
Hey this looks cool! Are you planning to take it out of draft at some point?

Philipp-M:
As described, I think it makes sense to wait until Rust 1.79 (where the following feature will be stabilized) since this currently requires unstable Rust:
```rust
#![feature(associated_type_bounds)]
```

Not sure whether it makes sense to make an exception here (as xilem_web is still kinda experimental, and it may otherwise block progress in xilem_web?)

xStrom:
I think we should wait for stable Rust. It's not too far away.

DJMcNab:
Yeah - using nightly features makes testing harder, and requires decisions about that.

I could maybe see requiring beta, where this would be "stable" in about two weeks. That's what I did for [Android Trace](https://github.com/linebender/android_trace), for example.

Philipp-M:
I'm closing this, as I'm currently rewriting xilem_web and found a better/cleaner solution for this.

PR #198, opened 2024-04-11
xilem_web: Support arrays (`impl IntoClasses for [impl IntoClasses; N]`) as classes

Philipp-M:
Should allow `el.class(["class1", "class2"])`, as something like `el.class("class1 class2")` doesn't seem to be allowed (probably worth adding more documentation at some time soon).


derekdreery (approved):
Awesome :)

PR #200, opened 2024-04-14
[deps] Use release of peniko.

waywardmonkeys:
This also lets us stop building a separate version of kurbo for xilem_web.


xStrom (approved):
Excellent, thanks!

PR #201, opened 2024-04-15
Remove `test-log` and `env_logger`.

waywardmonkeys:
`test-log` is used with logging to set up `env_logger` for each test and was only used in the box constraints code. This code doesn't actually use logging, so this wasn't doing much.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,15 +47,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aho-corasick"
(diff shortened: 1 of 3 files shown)


DJMcNab (approved):
We might want something like this again, but that would be once we start to be more principled with logging

waywardmonkeys:
And if so, we'd probably want it with the tracing / tracing_subscriber approach since the code that used to log here now uses tracing.

PR #204, opened 2024-04-23
Port Xilem to Masonry backend

PoignardAzur:
This is very much a work in progress.

Posting this so people can look at the code.

PR #205, opened 2024-04-24
Create a minimal xilem_masonry

DJMcNab:
Alternative to #204

Demonstrates a viable path forward to Xilem. This does not get rid of the widget set in the main Xilem crate, or migrate the main Xilem crate to this

### DIFF
--- a/crates/xilem_masonry/examples/mason.rs
+++ b/crates/xilem_masonry/examples/mason.rs
@@ -1,6 +1,7 @@
 // On Windows platform, don't show a console when opening the app.
 #![windows_subsystem = "windows"]
-use xilem_masonry::view::button;
+use masonry::ArcStr;
+use xilem_masonry::view::{button, flex};
 use xilem_masonry::{MasonryView, Xilem};
 
 fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
@@ -14,9 +15,20 @@ fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
 
     // The actual UI Code starts here
 
-    button(label, |data: &mut AppData| {
+    let mut sequence = vec![
+        the_button(label, 1),
+        the_button("Decrement".to_string(), -1),
+    ];
+    for x in 0..count {
+        sequence.push(the_button(format!("+{}", x), x));
+    }
+    flex(sequence)
+}
+
+fn the_button(label: impl Into<ArcStr>, count: i32) -> impl MasonryView<AppData> {
+    button(label, move |data: &mut AppData| {
         println!("clicked");
-        data.count += 1;
+        data.count += count;
     })
 }
 
--- a/crates/xilem_masonry/examples/mason.rs
+++ b/crates/xilem_masonry/examples/mason.rs
@@ -1,6 +1,6 @@
 // On Windows platform, don't show a console when opening the app.
 #![windows_subsystem = "windows"]
-use masonry::ArcStr;
+
 use xilem_masonry::view::{button, flex};
 use xilem_masonry::{MasonryView, Xilem};
 
@@ -15,21 +15,15 @@ fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
 
     // The actual UI Code starts here
 
-    let mut sequence = vec![
-        the_button(label, 1),
-        the_button("Decrement".to_string(), -1),
-    ];
-    for x in 0..count {
-        sequence.push(the_button(format!("+{}", x), x));
-    }
-    flex(sequence)
-}
-
-fn the_button(label: impl Into<ArcStr>, count: i32) -> impl MasonryView<AppData> {
-    button(label, move |data: &mut AppData| {
-        println!("clicked");
-        data.count += count;
-    })
+    let sequence = (0..count)
+        .map(|x| button(format!("+{x}"), move |data: &mut AppData| data.count += x))
+        .collect::<Vec<_>>();
+    flex((
+        button(label, |data: &mut AppData| data.count += 1),
+        button("Decrement", |data: &mut AppData| data.count -= 1),
+        button("Reset", |data: &mut AppData| data.count = 0),
+        sequence,
+    ))
 }
 
 struct AppData {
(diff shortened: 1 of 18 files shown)


crates/xilem_masonry/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+impl<E: Widget + StoreInWidgetMut> RootWidgetMut<'_, E> {
+    pub fn get_element(&mut self) -> WidgetMut<'_, E> {
+        self.ctx.get_mut(&mut self.widget.pod)
+    }
+}
+
+impl<State, Logic, View> Xilem<State, Logic, View>
```
DJMcNab, line 135: This might be too cute, but I'd like to keep it like this

I wanted to avoid the term `App`, because this is plausibly embeddable.

crates/xilem_masonry/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+    const CHANGED: Self = ChangeFlags { changed: true };
+    const UNCHANGED: Self = ChangeFlags { changed: false };
+}
+
+pub struct ViewCx {
+    /// The map from a widgets id to its position in the View tree.
+    ///
+    /// This includes only the widgets which might send actions
+    /// This is currently never cleaned up
+    widget_map: HashMap<WidgetId, Vec<ViewId>>,
```
DJMcNab, line 225: I believe we expect Masonry to do internal tree structure tracking at some point, which I think would include widget deleted events?

I don't know a *clean* way to tidy this up, otherwise

crates/xilem_masonry/src/sequence.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    /// Returns the current amount of widgets built by this sequence.
+    fn count(&self) -> usize;
+}
+
+/// Workaround for trait ambiguity
+///
+/// These need to be public for type inference
+#[doc(hidden)]
+pub struct WasAView;
```
DJMcNab, line 60: This avoids needing ViewMarker

crates/xilem_masonry/src/sequence.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    fn count(&self) -> usize {
+        match self {
+            Some(this) => this.count(),
+            None => 0,
+        }
+    }
+}
+
+// TODO: We use raw indexing for this value. What would make it invalid?
```
DJMcNab, line 158: The design of `ViewSequence` here is subtly different to previous versions of Xilem.

Currently, each non-trivial `ViewSequence`s add a level to the ViewId path, and they have complete control of their own `ViewId`.

This is a quite different, much sparser model of view ids

crates/xilem_masonry/src/vec_splice.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+impl ElementSplice for VecSplice<'_, '_, masonry::WidgetPod<Box<dyn masonry::Widget>>> {
+    fn push(&mut self, element: masonry::WidgetPod<Box<dyn masonry::Widget>>) {
+        self.push(element);
+    }
+
+    fn mutate(&mut self) -> WidgetMut<Box<dyn masonry::Widget>> {
+        unreachable!("VecSplice can only be used for `build`, not rebuild")
```
DJMcNab, line 101: This is an interesting one. I suspect that `ViewSequence::build` might want to take an append only `Vec`, and so only mutate takes an `ElementSplice`.

crates/xilem_masonry/src/view/flex.rs
```diff
@@ -0,0 +1,120 @@
+use std::marker::PhantomData;
+
+use masonry::{
+    widget::{self, WidgetMut},
+    Widget, WidgetPod,
+};
+
+use crate::{ElementSplice, MasonryView, VecSplice, ViewSequence};
+
+// TODO: Allow configuring flex properties. I think this actually needs its own view trait?
```
DJMcNab, line 10: These kinds of specialised view traits would also be the idiomatic way to create and manage the properties of winit `Window`s, for example (in addition to menus?)

crates/xilem_masonry/examples/mason.rs
DJMcNab, line 1:
```
cargo run --package xilem_masonry --example mason
```

PoignardAzur (approved):
LGTM!

crates/xilem_masonry/src/sequence.rs
```diff
@@ -0,0 +1,363 @@
...
+
+            fn message(
+                &self,
+                id_path: &[ViewId],
+                message: Box<dyn std::any::Any>,
+                app_state: &mut State,
+            ) -> MessageResult<Action> {
+                let (start, rest) = id_path
+                    .split_first()
+                    .expect("Id path has elements for vector");
```
PoignardAzur, line 329: Should be "tuple".

crates/xilem_masonry/src/sequence.rs
```diff
@@ -0,0 +1,363 @@
...
+    ) -> MessageResult<Action> {
+        self.0.message(id_path, message, app_state)
+    }
+
+    fn count(&self) -> usize {
+        self.0.count()
+    }
+}
+
+const BASE_ID: NonZeroU64 = match NonZeroU64::new(1) {
```
PoignardAzur, line 280: Could be unwrap, I think?

crates/xilem_masonry/src/sequence.rs
```diff
@@ -0,0 +1,363 @@
...
+
+            fn count(&self) -> usize {
+                // Is there a way to do this which avoids the `+0`?
+                $(self.$idx.count()+)+ 0
+            }
+        }
+    };
+}
+
+// We implement for tuples of length up to 15
```
PoignardAzur, line 348: Add comment for future improvement (new macro metavariables).

crates/xilem_masonry/src/sequence.rs
DJMcNab: [unwrap](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap) isn't const.

`ViewId` being `NonZero` is a little bit weird at the moment, so it might be worth reconsidering.

PR #206, opened 2024-04-25
Implement `AnyView` in `xilem_masonry`

DJMcNab:
[xilem_masonry_anyview.webm](https://github.com/linebender/xilem/assets/36049421/e9110f6b-0a96-497b-8ca5-a7f7aa3e2a7a)

### DIFF
--- a/crates/xilem_masonry/src/any_view.rs
+++ b/crates/xilem_masonry/src/any_view.rs
@@ -1,0 +1,177 @@
+use std::ops::Deref;
+
+use masonry::{
+    declare_widget,
+    widget::{StoreInWidgetMut, WidgetMut, WidgetRef},
+    BoxConstraints, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx, Point, PointerEvent,
+    Size, StatusChange, TextEvent, Widget, WidgetPod,
+};
+use smallvec::SmallVec;
+use vello::Scene;
+
+use crate::{ChangeFlags, MasonryView, MessageResult, ViewCx, ViewId};
+
+/// A view which can have any underlying view type.
+///
+/// This can be used to return type erased views (such as from a trait),
+/// or used to implement conditional display and switching of views.
+///
+/// Note that `Option` can also be used for conditionally displaying
+/// views when placed in a ViewSequence.
+// TODO: Mention `Either` when we have implemented that?
+pub type BoxedMasonryView<T, A = ()> = Box<dyn AnyMasonryView<T, A>>;
+
+impl<T: 'static, A: 'static> MasonryView<T, A> for BoxedMasonryView<T, A> {
+    type Element = DynWidget;
+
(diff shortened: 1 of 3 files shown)


crates/xilem_masonry/src/any_view.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    fn dyn_rebuild(
+        &self,
+        cx: &mut ViewCx,
+        prev: &dyn AnyMasonryView<T, A>,
+        mut element: WidgetMut<DynWidget>,
+    ) -> ChangeFlags {
+        // TODO: Does this need to have a custom view id to enable events sent
+        // to an outdated view path to be caught and returned?
+        // Should we store this generation in `element`? Seems plausible
```
Philipp-M, line 92: It's not really decoupled from the widgets then, is it? (Which AFAIK is necessary, for views not being wired up directly with widgets but being able to receive events from somewhere else, async list is an example for that (via the waker that stores the current view id path)).
I think it's one of the reasons why `View::build` returns a `ViewId`, and rebuild is able to change it (e.g. in this method, because there's a new view).

crates/xilem_masonry/src/any_view.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use crate::{ChangeFlags, MasonryView, MessageResult, ViewCx, ViewId};
+
+/// A view which can have any underlying view type.
+///
+/// This can be used to return type erased views (such as from a trait),
+/// or used to implement conditional display and switching of views.
+///
+/// Note that `Option` can also be used for conditionally displaying
+/// views when placed in a ViewSequence.
```
waywardmonkeys, line 20: `ViewSequence` here should at least be in back-ticks if not linked.

crates/xilem_masonry/src/any_view.rs
```diff
@@ -0,0 +1,177 @@
...
+        message: Box<dyn std::any::Any>,
+        app_state: &mut T,
+    ) -> MessageResult<A> {
+        self.message(id_path, message, app_state)
+    }
+}
+
+/// A widget whose only child can be dynamically replaced
+///
+/// `WidgetPod<Box<dyn Widget>>` doesn't expose this possibility
```
waywardmonkeys, line 123: Some missing periods in this file, like here.

waywardmonkeys (changes requested):

crates/xilem_masonry/src/any_view.rs
DJMcNab: Yeah, this does need to happen in `View::State`.

This isn't actually problematic because we don't have async wiring yet, but the current state is wrong

Philipp-M (approved):

DJMcNab:
This is in a far from perfect state, and it has confirmed that we need to bring back the `State` field/type on `MasonryView`  (entirely because of our future async plans, but doing so later would cause a lot more pain).

But we're merging now despite that

PR #208, opened 2024-04-25
masonry: Update MSRV to 1.73 from 1.65

waywardmonkeys:
The code currently uses some API that is not available in Rust 1.65:

* `is_ok_and` from 1.70
* `next_multiple_of` from 1.73


DJMcNab (approved):
Vello's MSRV is/will be 1.75 since https://github.com/linebender/vello/pull/478, so I think that's our true minimum

But updating this can't hurt. Thanks.

So I'd recommend changing to 1.75 then landing

waywardmonkeys:
Good call. Updated and will merge when CI passes (or I wake up in the morning).

DJMcNab:
(You should be able to merge when ready, which will add to the queue once CI passes)

PR #209, opened 2024-04-25
Rename xilem and masonry `hello` examples.

waywardmonkeys:
When building with `--all-targets --workspace`, there is a warning (on macOS at least) about multiple targets generating the same files due to having the same name.

This renames both `hello` examples to indicate which project they are from.


waywardmonkeys:
The warning message is:

```
warning: output filename collision.
The example target `hello` in package `xilem v0.1.0 (/.../xilem)` has the same output filename as the example target `hello` in package `masonry v0.1.3 (/.../xilem/crates/masonry)`.
Colliding filename is: /.../xilem/target/debug/examples/hello
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The example target `hello` in package `xilem v0.1.0 (/.../xilem)` has the same output filename as the example target `hello` in package `masonry v0.1.3 (/.../xilem/crates/masonry)`.
Colliding filename is: /...xilem/target/debug/examples/hello.dSYM
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

DJMcNab (approved):
I'd *probably* prefer this to use underscores, but that's not really based on much

jaredoconnell:
Looking around the crates in this repository, I agree that underscores would be more consistent.

waywardmonkeys:
Updated to use underscores.

PR #210, opened 2024-04-25
parley: Use via workspace dependency.

waywardmonkeys:
This will help us keep everyone using the same version. This also makes both `xilem` and `masonry` use the same version now.


DJMcNab (approved):
I've not checked which commits are being used - I presume this is a sufficiently recent commit

waywardmonkeys:
It is recent, but will need updating to pick up the split of `fontique` and such prior to publishing.

PR #212, opened 2024-04-26
masonry: Fix typos.

waywardmonkeys:


DJMcNab (approved):
Thanks. I was getting editor warnings about these, so it'll be good to have them fixed

PR #213, opened 2024-04-26
masonry: Use workspace lint config.

waywardmonkeys:
This replaces the old lint configuration in a cargo config with the workspace lint configuration in the Cargo.toml.

The old lint configuration was for tests, which don't currently trigger those lints, so we don't need to supply this configuration elsewhere. (The library has its own lint configuration that supplements this in `src/lib.rs` which can be improved later.)

### DIFF
--- a/crates/masonry/.cargo/config.toml
+++ b/crates/masonry/.cargo/config.toml
@@ -1,9 +1,0 @@
-[target.'cfg(all())']
-# TODO - Replace with Cargo.toml config
-rustflags = [
-    # Global lints/warnings.
-    # We do this here instead of in the crate root because we want to apply
-    # these to example and test targets as well.
-    "-Aclippy::single_match",
-    "-Aclippy::bool_assert_comparison",
-]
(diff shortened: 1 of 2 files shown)


crates/masonry/.cargo/config.toml
```diff
@@ -1,9 +0,0 @@
-[target.'cfg(all())']
-# TODO - Replace with Cargo.toml config
-rustflags = [
-    # Global lints/warnings.
-    # We do this here instead of in the crate root because we want to apply
-    # these to example and test targets as well.
-    "-Aclippy::single_match",
```
DJMcNab, line 7: Ah, yes. I ran into this one in #205. This would explain why it wasn't an issue before

DJMcNab (approved):

crates/masonry/.cargo/config.toml
waywardmonkeys: There's still an `allow` in `crates/masonry/src/lib.rs` which can be removed, but will have to fix the code as well at that time. Another PR, probably another day.

PR #216, opened 2024-04-26
xilem_web: Fix clippy::multiple_bound_locations lints.

waywardmonkeys:
Define generic bounds once rather than partly in the bound predicate and partly in the `where` clause.

### DIFF
--- a/crates/xilem_web/src/vecmap.rs
+++ b/crates/xilem_web/src/vecmap.rs
@@ -34,10 +34,10 @@ impl<K, V> VecMap<K, V> {
     /// assert_eq!(map.get(&1), Some(&"a"));
     /// assert_eq!(map.get(&2), None);
     /// ```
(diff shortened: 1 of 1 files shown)


Philipp-M (approved):
Not that I have strong feelings about this, but I've put the bounds in the where clause already in #197 ([here](https://github.com/linebender/xilem/blob/b3f1068299d3941e9e4575c489c8ae299a519250/crates/xilem_web/src/vecmap.rs)), I think having them in one place (`where ...`) would be better?

waywardmonkeys:
I tired that! But apparently `?Sized` couldn’t be there … will repro later and make sure I didn’t do something dumb.

waywardmonkeys:
@Philipp-M  So, I'd done something dumb earlier when I got an error. I've updated the PR per your suggestion (and added one case that the lint didn't catch for the `impl` block).

PR #217, opened 2024-04-26
masonry: Remove profile from non-root Cargo.toml.

waywardmonkeys:
This is generating a warning every time you build about it being here and that it will be ignored:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /.../xilem/crates/masonry/Cargo.toml
workspace: /.../xilem/Cargo.toml
```

### DIFF
--- a/crates/masonry/Cargo.toml
+++ b/crates/masonry/Cargo.toml
@@ -11,9 +11,6 @@ repository = "https://github.com/PoignardAzur/masonry-rs"
 rust-version = "1.75"
 version = "0.1.3"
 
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):

DJMcNab:
There might be some discussion in adding that elsewhere, but that doesn't really matter much here

waywardmonkeys:
> There might be some discussion in adding that elsewhere, but that doesn't really matter much here

Agreed.

I just want to get things into a much cleaner state to make iteration better.

PR #218, opened 2024-04-26
Add `View::State` and `IdPath` to `xilem_masonry`

Philipp-M:
Draft, to avoid (further) duplicate work.

Basically adds `View::State` back to `xilem_masonry` and the id path, similar to how it was done before (i.e. distinct ids, which may be relevant for a robust async context).

This is not yet 100% done (issues with changing `AnyView` currently).

Also renames the generic params `State` and `T` to `AppState` for clear distinction to `View::State`


Philipp-M:
Closing in favor of #220

PR #220, opened 2024-04-26
xilem_masonry: Robust generational indexing for `ViewSequence`

DJMcNab:
Alternative to #218

The version of viewsequence added in #205 is robust for all properties supported in that version. However, it would be incompatible with future expansions to the Xilem model, in particular async wiring.

In this PR, I propose a system to resolve this, which uses generational indexes in the view id path (for the items where this is relevant).

Some other historical record: #9

This brings back the `View::State` associated type (renamed to `ViewState` and `SeqState` depending on the trait), in order to perform this wiring.

### DIFF
--- a/crates/xilem_masonry/src/sequence.rs
+++ b/crates/xilem_masonry/src/sequence.rs
@@ -212,14 +212,15 @@ impl<T, A, Marker, VT: ViewSequence<T, A, Marker>> ViewSequence<T, A, (WasASeque
 {
     type SeqState = VecViewState<VT::SeqState>;
     fn build(&self, cx: &mut ViewCx, elements: &mut dyn ElementSplice) -> Self::SeqState {
+        let generation = 0;
         let inner = self.iter().enumerate().map(|(i, child)| {
-            let id: u64 = i.try_into().unwrap();
+            let id = create_vector_view_id(i, generation);
 
             cx.with_id(ViewId::for_type::<VT>(id), |cx| child.build(cx, elements))
         });
-        let inner_with_generations = inner.map(|it| (it, 0)).collect();
+        let inner_with_generations = inner.map(|it| (it, generation)).collect();
         VecViewState {
-            global_generation: 0,
+            global_generation: generation,
             inner_with_generations,
         }
     }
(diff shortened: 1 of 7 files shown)


crates/xilem_masonry/src/sequence.rs
```diff
@@ -158,48 +201,91 @@ impl<State, Action, Marker, VT: ViewSequence<State, Action, Marker>>
...
+                    // The known solution mentioned in the above message is to use a different ViewId for the index and the generation
+                    // We believe this to be superfluous for the default use case, as even with 1000 rebuilds a second, each adding
+                    // to the same array, this would take 50 days of the application running continuously.
+                    // See also https://github.com/bevyengine/bevy/pull/9907, where they warn in their equivalent case
+                    // Note that we have a slightly different strategy to Bevy, where we use a global generation
+                    // This theoretically allows some of the memory in `seq_state` to be reclaimed, at the cost of making overflow
+                    // more likely here. Note that we don't actually reclaim this memory at the moment.
+
+                    // We use 0 to wrap around. It would require extremely unfortunate timing to get an async event
+                    // with the correct generation exactly u32::MAX generations late, so wrapping is the best option
```
Philipp-M, line 274: Great writeup, and I agree, the probability that this happens per accident, is likely in the ballpark of a cryptographic hash collision... Not sure if this *could* be a security issue with a targeted attack, but I don't think we care at this point...

Philipp-M (approved):
Looks pretty good, I think this indeed should be robust for async updates. I like that we got rid of the linear search in `Vec<impl ViewSequence>` (and likely other similar containers in the future).

crates/xilem_masonry/src/sequence.rs
```diff
@@ -254,38 +366,36 @@ impl<T, A> ViewSequence<T, A, ()> for () {
...
-        self.0.message(id_path, message, app_state)
+        self.0.message(seq_state, id_path, message, app_state)
     }
 
     fn count(&self) -> usize {
         self.0.count()
     }
 }
 
-const BASE_ID: NonZeroU64 = match NonZeroU64::new(1) {
```
DJMcNab, line 284: For future reference, TIL about [`NonZeroU64::MIN`](https://doc.rust-lang.org/std/num/type.NonZeroU64.html#associatedconstant.MIN)

DJMcNab:
Merging to avoid merge conflicts over the weekend. I'm pretty happy with this now, and I think the main infrastructure work on `xilem_masonry` is done.
One remaining task is `AnyViewSequence`, which is *extremely* analogous to `AnyView`, but otherwise it's improvements to the set of widgets and to Masonry.

(Also on the agenda is working out how to handle flex items - the idiomatic way is a specific `View`/`ViewSequence` trait, but that requires a huge amount of boilerplate)

PR #221, opened 2024-04-26
deps: Update parley.

waywardmonkeys:
This version of parley has fontique as a separate crate and has updated dependencies (as well as a removal of clap as a transitive dep).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,54 +89,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.6.13"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d96bd03f33fe50a863e394ee9718a706f988b9079b20c3784fb726e7678b62fb"
-dependencies = [
- "anstyle",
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Well, looks like a good thing we tried this before publishing `parley`!

(I haven't looked into this yet.)

cc: @dfrg

waywardmonkeys:
This is because of the update to `roxmltree`!

waywardmonkeys:
Once https://github.com/linebender/parley/pull/43 lands, I'll give this a try again so we can update without the change to `roxmltree`. (I don't have a fontconfig system to experiment with.)

waywardmonkeys:
@DJMcNab This is re-running now with the parley-up-through-memmap2-update tree.

DJMcNab (approved):
Seems like it's worth testing this here, even if we can't use latest main.

waywardmonkeys:
Yes, this gets us slimmer dependencies as well as the variants improvement, so it is good. Just sad that `roxmltree` update broke things (but also good to know!).

PR #222, opened 2024-04-26
Properly `deref` the previous view when using `AnyMasonryView`

DJMcNab:
Previously, this meant that the previous was actually a `Box<dyn AnyView>`, which implemented `View` and so was being passed as its own `&dyn AnyView` to rebuild.

I also moved `rebuild` and `message` into the correct order.

This partially fixes the `mason` example - pressing the `Unlimited Power` button, it now retains its state.
This would be more impactful for e.g. a text input view, where the widget had internal state.

Pressing activate and deactivate is still broken. See [#masonry > Hot state for new widgets](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Hot.20state.20for.20new.20widgets)


jaredoconnell (approved):
Looks good to me. Confirmed working.

PR #223, opened 2024-04-26
masonry: Use Kurbo via Vello.

waywardmonkeys:
Rather than have to keep a dependency on `kurbo` at the correct version, always use it via `vello`.

This makes this match how `peniko` is already used within `masonry`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1878,7 +1878,6 @@ dependencies = [
  "futures-intrusive",
  "image",
  "insta",
- "kurbo",
  "nv-flip",
  "once_cell",
  "parley",
--- a/masonry/Cargo.toml
+++ b/masonry/Cargo.toml
@@ -22,7 +22,6 @@ workspace = true
 [dependencies]
 vello.workspace = true
 wgpu.workspace = true
-kurbo.workspace = true
 parley.workspace = true
 winit.workspace = true
 smallvec.workspace = true
--- a/masonry/examples/custom_widget.rs
+++ b/masonry/examples/custom_widget.rs
@@ -8,9 +8,8 @@
 #![windows_subsystem = "windows"]
 
 use accesskit::Role;
-use kurbo::Stroke;
 use masonry::app_driver::{AppDriver, DriverCtx};
-use masonry::kurbo::BezPath;
+use masonry::kurbo::{BezPath, Stroke};
 use masonry::widget::{FillStrat, RootWidget};
 use masonry::{
     AccessCtx, AccessEvent, Action, Affine, BoxConstraints, Color, EventCtx, LayoutCtx, LifeCycle,
--- a/masonry/src/box_constraints.rs
+++ b/masonry/src/box_constraints.rs
@@ -1,7 +1,7 @@
 // Copyright 2019 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::kurbo::Size;
+use vello::kurbo::Size;
 
 /// Constraints for layout.
 ///
--- a/masonry/src/lib.rs
+++ b/masonry/src/lib.rs
@@ -97,9 +97,9 @@
 
 pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 pub use dpi;
-pub use kurbo;
 pub use parley;
 pub use vello;
+pub use vello::kurbo;
 
 #[macro_use]
 mod util;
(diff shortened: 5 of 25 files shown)


waywardmonkeys:
This does leave some `pub use` of kurbo stuff in `crates/masonry/src/lib.rs` that I think a subsequent PR should follow up and remove so that the top-level exports of `masonry` are cleaner and more of just what's needed.

DJMcNab:
I think we need to re-export `vello` and maybe `winit` from Masonry? I'm not sure what the best API here looks like - there are some unclear abstraction boundaries. At the moment, `xilem_masonry` has to depend directly on vello and winit, which probably isn't correct. See also linebender/vello_svg#5

I would propose postponing this until after RustNL.

The phrasing via "vello or peniko" there is quite telling, imo, that this is still messy.

waywardmonkeys:
The vellol or peniko is only because I included changes for xilem_web which uses peniko but not vello.

DJMcNab:
Ah, apologies. That makes sense.

I still think this isn't straightforwardly a win, and needs more careful thought.
It's not really good DX to have multiple paths to the same item up and down the stack, but it's also not good DX to require a specific neighbour dependency - especially when it's only to refer to the types, as it is with Vello, here.
Maybe we re-export all of Vello except for the Kurbo re-export, and re-export Kurbo seperately at the root? 

And so overall, would still prefer to defer this for 3 weeks or so

djc:
It looks like a win to me. If most usage of Kurbo will happen either by crates that need to depend on Vello or (in the case of xilem_web) via Peniko it makes sense then re-exporting Kurbo in Vello makes sense (and so I would also say that https://github.com/linebender/vello_svg/pull/5 does not make much sense to me). In fact I already noticed that xilem_web has a kurbo dependency that it doesn't end up using so removing that would definitely make sense.

Philipp-M:
> In fact I already noticed that xilem_web has a kurbo dependency that it doesn't end up using so removing that would definitely make sense.

It's used though? In the svgtoy for example?

But I guess we can put kurbo and other things there behind a feature flag.

djc:
If it's used (directly) only for the examples it should be a dev-dependency.

waywardmonkeys:
@djc @Philipp-M See #293, I just split that off.

waywardmonkeys:
I've rebased this PR forward, but left in the part that I did in #293 (I'll rebase again).

This should make things more clear that this patch isn't about how we re-export, but about how we use, and this just brings the usage of `kurbo` within `masonry` in line with how `peniko` is used: Not via a direct dep, but via the `vello` re-export.

This also cleans up a number of places where multiple paths to `kurbo` (which might have pointed at different crate versions) were used within a single file.

I'd previously done (and landed) a version of this in `xilem_classic` in PR #90.

I've edited the commit message and description on the commit in this PR to reflect that it isn't going to be involving anything but `masonry` (once I rebase it forward again after #293 lands hopefully).

crates/masonry/src/widget/checkbox.rs
```diff
@@ -4,13 +4,12 @@
 //! A checkbox widget.
 
 use accesskit::{DefaultActionVerb, Role, Toggled};
-use kurbo::{Affine, Stroke};
 use smallvec::{smallvec, SmallVec};
 use tracing::{trace, trace_span, Span};
+use vello::kurbo::{Affine, BezPath, Cap, Join, Size, Stroke};
```
DJMcNab, line 9: I think I'd prefer these all to use `crate::{...}`

If we need the types, we should export them directly, probably.

DJMcNab (approved):
I find this frustrating, because it is not clear what the right rules are here.

I now realise that this is an improvement - apologies, I hadn't fully understood the actual changes being suggested here. I had assumed that this was only doing `pub use vello::kurbo`

waywardmonkeys:
It is frustrating, I agree! And what's worse is that just naively accepting suggestions from rust-analyzer or other tools to resolve imports may or may not do the "right" thing (whatever that even is).

We had a `pub use kurbo` before, so in my last rebase (moments ago, after #293), I added in the `pub use vello::kurbo;` to keep existing code seeing the same thing.

In that case, we could do:

```
use crate::turbo::Foopy;
```

or 

```
pub use crate::vello::kurbo::Foopy;
```

Or the current:

```
pub use vello::kurbo::Foopy;
```

This brings up another thing which is that the import styles are also not identical across things, so some stuff will do:

```
use vello::kurbo::Line;
use vello::peniko::Image;
use vello::Scene;
```

while others might do:

```
use vello::{kurbo::Line, peniko::Image, Scene};
```

So ... I'm happy to make them all do the same thing, I just need to know what that thing should be.

waywardmonkeys:
I'll do that this week.

waywardmonkeys:
I've resurrected this and rebased it.

Philipp-M (approved):
Looks good to me.

PoignardAzur (approved):
LGTM.

PR #224, opened 2024-04-26
check roxmltree update

dfrg:
Updated the fontique fontconfig roxmltree parsing to allow DTDs. Opening this PR to check masonry tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,6 +89,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.13"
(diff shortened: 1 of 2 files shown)


DJMcNab (changes requested):
Not merging until linebender/parley#44 lands and this is updated to the resultant commit on main

DJMcNab:
I think this needs to be updated to https://github.com/linebender/parley/commit/ddb0c6e92361c9b69dd7e0807274eec9f4856772

waywardmonkeys:
This isn't needed now that `parley` is released and we have #238  to switch us to using it.

PR #225, opened 2024-04-27
xilem_masonry: Add checkbox widget

Philipp-M:
Implements the masonry checkbox as view (and some improvements in `Button::message`)

### DIFF
--- a/crates/xilem_masonry/examples/mason.rs
+++ b/crates/xilem_masonry/examples/mason.rs
@@ -1,7 +1,7 @@
 // On Windows platform, don't show a console when opening the app.
 #![windows_subsystem = "windows"]
 
-use xilem_masonry::view::{button, flex};
+use xilem_masonry::view::{button, checkbox, flex};
 use xilem_masonry::{BoxedMasonryView, MasonryView, Xilem};
 
 fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
@@ -20,6 +20,9 @@ fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
         .collect::<Vec<_>>();
     flex((
         button(label, |data: &mut AppData| data.count += 1),
+        checkbox("Check me", data.active, |data: &mut AppData, checked| {
+            data.active = checked;
+        }),
         toggleable(data),
         button("Decrement", |data: &mut AppData| data.count -= 1),
         button("Reset", |data: &mut AppData| data.count = 0),
--- a/crates/xilem_masonry/src/view/button.rs
+++ b/crates/xilem_masonry/src/view/button.rs
@@ -56,10 +56,19 @@ where
             id_path.is_empty(),
             "id path should be empty in Button::message"
         );
-        if let Some(masonry::Action::ButtonPressed) = message.downcast_ref() {
-            return MessageResult::Action((self.callback)(app_state));
+        match message.downcast::<masonry::Action>() {
+            Ok(action) => {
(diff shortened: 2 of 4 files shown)


crates/xilem_masonry/src/view/button.rs
```diff
@@ -53,4 +44,22 @@ where
...
     }
+
+    fn message(
+        &self,
+        _view_state: &mut Self::ViewState,
+        id_path: &[ViewId],
+        message: Box<dyn std::any::Any>,
+        app_state: &mut State,
+    ) -> crate::MessageResult<Action> {
+        debug_assert!(
```
DJMcNab, line 55: I've been doing a bit of thinking here, and I'm wondering if action views *should* get an entry in the id path, entirely for logging purposes (i.e. when printing out an idpath, we can see what kind of view it is targetting)

But that's out-of-scope for this PR.

crates/xilem_masonry/src/view/button.rs
```diff
@@ -53,4 +44,22 @@ where
...
+            id_path.is_empty(),
+            "id path should be empty in Button::message"
+        );
+        if let Some(masonry::Action::ButtonPressed) = message.downcast_ref() {
```
DJMcNab, line 59: To be consistent, I'd probably use `downcast` here; other views likely will want to take ownership of the action, and this box will be immediately dropped in the case where this suceeds anyway

crates/xilem_masonry/src/view/button.rs
```diff
@@ -53,4 +44,22 @@ where
...
+            return MessageResult::Action((self.callback)(app_state));
+        }
+        tracing::error!("Wrong message type in Button::message");
```
DJMcNab, line 62: Ideally, we'd be using a `Box<dyn Any + Debug>` or something similar, to make this more debuggable. But that's also out-of-scope for this PR

crates/xilem_masonry/examples/mason.rs
```diff
@@ -20,6 +20,9 @@ fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
         .collect::<Vec<_>>();
     flex((
         button(label, |data: &mut AppData| data.count += 1),
+        checkbox("Check me", data.active, |data: &mut AppData, checked| {
```
DJMcNab, line 23: Thanks for not removing the activate/deactivate buttons; they're still useful for debugging [#masonry > Hot state for new widgets](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Hot.20state.20for.20new.20widgets)

DJMcNab (approved):
Not ran this locally, but the code looks right, so CI should solve things

Some thoughts for future work

crates/xilem_masonry/src/view/button.rs
Philipp-M: The reason I'm using `downcast_ref` instead of `downcast`, is that then `MessageResult::Stale(message)` is possible. Otherwise I need to do something like `MessageResult::Stale(Box::new(()))` if that's better?

Philipp-M: Oh, wait, `downcast` returns `Self` in `Err`, ok, then I'm changing it.

crates/xilem_masonry/src/view/button.rs
Philipp-M: Good point, maybe implement that behavior directly in `ViewCx`? I mean generally some debugging helpers.

PR #226, opened 2024-04-27
xilem_masonry: Add Label widget

Philipp-M:
Adds the label widget to `xilem_masonry` with some (but not all) attribute setters, and methods to disable the label in masonry.
Also slightly extends the `Flex` view, by being able to specify the direction (and a small fix in `FlexSplice`).

### DIFF
--- a/crates/masonry/src/widget/flex.rs
+++ b/crates/masonry/src/widget/flex.rs
@@ -246,7 +246,7 @@ impl Flex {
 // --- Mutate live Flex - WidgetMut ---
 
 impl<'a> FlexMut<'a> {
-    /// Set the flex direction (see [`Axis`].
+    /// Set the flex direction (see [`Axis`]).
     ///
     /// [`Axis`]: enum.Axis.html
     pub fn set_direction(&mut self, direction: Axis) {
--- a/crates/masonry/src/widget/flex.rs
+++ b/crates/masonry/src/widget/flex.rs
@@ -246,13 +246,20 @@ impl Flex {
 // --- Mutate live Flex - WidgetMut ---
 
 impl<'a> FlexMut<'a> {
+    /// Set the flex direction (see [`Axis`].
+    ///
+    /// [`Axis`]: enum.Axis.html
+    pub fn set_direction(&mut self, direction: Axis) {
+        self.widget.direction = direction;
+        self.ctx.request_layout();
+    }
+
     /// Set the childrens' [`CrossAxisAlignment`].
     ///
     /// [`CrossAxisAlignment`]: enum.CrossAxisAlignment.html
     pub fn set_cross_axis_alignment(&mut self, alignment: CrossAxisAlignment) {
         self.widget.cross_alignment = alignment;
-        // TODO
-        self.ctx.widget_state.needs_layout = true;
+        self.ctx.request_layout();
     }
 
     /// Set the childrens' [`MainAxisAlignment`].
@@ -260,16 +267,14 @@ impl<'a> FlexMut<'a> {
     /// [`MainAxisAlignment`]: enum.MainAxisAlignment.html
     pub fn set_main_axis_alignment(&mut self, alignment: MainAxisAlignment) {
         self.widget.main_alignment = alignment;
-        // TODO
-        self.ctx.widget_state.needs_layout = true;
+        self.ctx.request_layout();
     }
 
     /// Set whether the container must expand to fill the available space on
     /// its main axis.
     pub fn set_must_fill_main_axis(&mut self, fill: bool) {
         self.widget.fill_major_axis = fill;
-        // TODO
-        self.ctx.widget_state.needs_layout = true;
+        self.ctx.request_layout();
     }
 
     /// Add a non-flex child widget.
@@ -317,9 +322,7 @@ impl<'a> FlexMut<'a> {
             }
         };
         self.widget.children.push(child);
-        // TODO
-        self.ctx.widget_state.children_changed = true;
-        self.ctx.widget_state.needs_layout = true;
+        self.ctx.children_changed();
     }
 
     /// Add a spacer widget with a standard size.
--- a/crates/masonry/src/widget/label.rs
+++ b/crates/masonry/src/widget/label.rs
@@ -109,6 +109,12 @@ impl Label {
         self
     }
 
+    /// Builder-style method to disable this label.
+    pub fn with_disabled(mut self, disabled: bool) -> Self {
+        self.disabled = disabled;
+        self
+    }
+
     /// Return the current value of the label's text.
     pub fn text(&self) -> ArcStr {
         self.current_text.clone()
@@ -178,6 +184,13 @@ impl LabelMut<'_> {
     /// Set the [`Alignment`] for this layout.
     pub fn set_text_alignment(&mut self, alignment: Alignment) {
         self.widget.alignment = alignment;
+        self.ctx.request_layout();
+    }
+
+    /// Set disabled
+    pub fn set_disabled(&mut self, disabled: bool) {
+        self.widget.disabled = disabled;
+        // TODO: only request_paint necessary?
         self.ctx.request_layout();
     }
 }
--- a/crates/xilem_masonry/examples/mason.rs
+++ b/crates/xilem_masonry/examples/mason.rs
@@ -1,13 +1,13 @@
 // On Windows platform, don't show a console when opening the app.
 #![windows_subsystem = "windows"]
 
-use xilem_masonry::view::{button, checkbox, flex};
-use xilem_masonry::{BoxedMasonryView, MasonryView, Xilem};
+use xilem_masonry::view::{button, checkbox, flex, label};
+use xilem_masonry::{Axis, BoxedMasonryView, Color, MasonryView, TextAlignment, Xilem};
 
 fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
     // here's some logic, deriving state for the view from our state
     let count = data.count;
-    let label = if count == 1 {
+    let button_label = if count == 1 {
         "clicked 1 time".to_string()
     } else {
         format!("clicked {count} times")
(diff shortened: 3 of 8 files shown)


crates/xilem_masonry/src/lib.rs
```diff
@@ -8,6 +8,7 @@ use masonry::{
     BoxConstraints, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx, Point, PointerEvent,
     Size, StatusChange, TextEvent, Widget, WidgetId, WidgetPod,
 };
+pub use masonry::{widget::Axis, Color, TextAlignment};
```
Philipp-M, line 11: I'm shamelessly reexporting all these attributes to the root of the library (since I think they're mostly unambiguous), not sure what our policy here is.

crates/xilem_masonry/src/view/flex.rs
```diff
@@ -35,7 +44,7 @@ where
         let mut scratch = Vec::new();
         let mut splice = VecSplice::new(&mut elements, &mut scratch);
         let seq_state = self.sequence.build(cx, &mut splice);
-        let mut view = widget::Flex::column();
+        let mut view = widget::Flex::for_axis(self.axis);
```
DJMcNab, line 47: I didn't expose this intentionally, because Masonry doesn't support changing the axis at runtime.

We do not want to have views which give different final trees in `build` and `rebuild`

crates/xilem_masonry/src/view/flex.rs
```diff
@@ -121,7 +130,6 @@ impl ElementSplice for FlexSplice<'_> {
     }
 
     fn len(&self) -> usize {
-        // This is not correct because of the spacer items. Is `len` actually needed?
-        self.element.len() - self.ix
+        self.ix
```
DJMcNab, line 133: Sure! I suspect that we'll probably just remove `len`, anyway; but this is more correct

crates/xilem_masonry/src/view/label.rs
```diff
@@ -0,0 +1,92 @@
+use masonry::{widget::WidgetMut, ArcStr, WidgetPod};
+
+use crate::{ChangeFlags, Color, MasonryView, MessageResult, TextAlignment, ViewCx, ViewId};
+
+pub fn label(label: impl Into<ArcStr>) -> Label {
+    Label {
+        label: label.into(),
+        text_color: Color::default(),
```
DJMcNab, line 8: Hmm, `Color` implementing `Default` is weird - the default would surely depend on the use case?

crates/xilem_masonry/src/view/label.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    fn message(
+        &self,
+        _view_state: &mut Self::ViewState,
+        _id_path: &[ViewId],
+        message: Box<dyn std::any::Any>,
+        _app_state: &mut State,
+    ) -> crate::MessageResult<Action> {
+        MessageResult::Stale(message)
```
DJMcNab, line 90: I *think* we should `warn!` here?

crates/masonry/src/widget/label.rs
```diff
@@ -180,6 +186,13 @@ impl LabelMut<'_> {
         self.widget.alignment = alignment;
         self.ctx.request_layout();
     }
+
+    /// Set disabled
```
DJMcNab, line 190: What does it even mean for a label to be `disabled`?

crates/xilem_masonry/src/view/flex.rs
Philipp-M: Oh, haven't checked that closely yet. That's unfortunate, and I agree, all attributes that are able to set in the view  should of course change the widget tree.
In that case maybe stick to something like `h_flex` or `h_stack` something a like (open for naming suggestions)?

crates/xilem_masonry/src/view/flex.rs
Philipp-M: Yeah probably makes sense, doesn't seem to be needed anymore (I'll probably look into this more deeply, whether that is indeed the case).

crates/xilem_masonry/src/view/label.rs
Philipp-M: I can instead make it black (which seems to be the default) to be more clear. I agree, looking at it more closely, because default for the alpha value seems to be 0 (which I don't think is a good default). Maybe remove `Default` for `peniko::Color` alltogether (or make it explicitely black with opaque alpha)?

crates/xilem_masonry/src/view/label.rs
Philipp-M: Good point

crates/masonry/src/widget/label.rs
Philipp-M: I guess it makes more sense in combination with a checkbox. But basically, that it's not clickable in HTML jargon at least

crates/xilem_masonry/src/view/flex.rs
Philipp-M: Or, well just implement it in masonry, checking that now...

crates/xilem_masonry/src/view/label.rs
DJMcNab: Yeah, removing that `Default` is worth a PR, at worst to get a reason it should stay

crates/xilem_masonry/src/view/label.rs
DJMcNab: In fact, this is actually an `error` condition, because a parent view has messed up if we're seeing this

crates/xilem_masonry/src/view/flex.rs
Philipp-M: It's used in old `xilem::view::List`, so I *guess* we still need this, in case we don't exactly know the size of a `ViewSequence` in the view logic (e.g. lazy loading of items in a list, not sure currently whether that should happen in the xilem reactive layer?)

DJMcNab: I don't think its use in `List` is necessary, but I haven't remade the `List` ViewSequence yet, so I'm not certain

crates/xilem_masonry/src/view/label.rs
Philipp-M: Yeah agree, in fact I have made this an error already (not yet pushed).

crates/masonry/src/widget/label.rs
DJMcNab: Right, so e.g. the contents can't be clicked and dragged? I guess it also would have consequences for the accessibility metadata

crates/xilem_masonry/src/view/flex.rs
Philipp-M: So I've implemented it in masonry, and checked it via

```rust
let axis = if count % 2 == 0 {
    Axis::Horizontal
} else {
    Axis::Vertical
};

...
flex(()).direction(axis)
```

(but have hardcoded the values in the example)

crates/xilem_masonry/src/view/label.rs
Philipp-M: I've made it `Color::BLACK` for now in this PR.

crates/masonry/src/widget/label.rs
Philipp-M: Yeah I guess that is indeed the case.

crates/xilem_masonry/examples/mason.rs
```diff
@@ -39,14 +39,17 @@ fn app_logic(data: &mut AppData) -> impl MasonryView<AppData> {
...
+        Box::new(
+            flex((
+                button("Deactivate", |data: &mut AppData| {
+                    data.active = false;
+                }),
+                button("Unlimited Power", |data: &mut AppData| {
+                    data.count = -1_000_000;
+                }),
+            ))
+            .direction(Axis::Horizontal),
```
DJMcNab, line 51: Should we have a dynamically toggling axis in the same widget here?

crates/masonry/src/widget/flex.rs
```diff
@@ -246,30 +246,35 @@ impl Flex {
 // --- Mutate live Flex - WidgetMut ---
 
 impl<'a> FlexMut<'a> {
+    /// Set the flex direction (see [`Axis`].
```
DJMcNab, line 249: Unclosed bracket here

DJMcNab (approved):
I still think we need some way to check the flex direction change works as expected.

Are you willing to add a Masonry level test for it? I've not dug into the test suite myself, so I'm not sure what that would involve.
If not, I think we can just merge this for expedience.

We do also need to add some `xilem_masonry` level tests. It would be awesome if you could get that started (although feel free to work on whichever features you want, of course)

crates/xilem_masonry/examples/mason.rs
Philipp-M: Not sure, it jumps around the layout, which I didn't find satisfying, but I added it for the `+n` buttons when `data.active` changes, which is less annoying.

Philipp-M:
> We do also need to add some `xilem_masonry` level tests. It would be awesome if you could get that started (although feel free to work on whichever features you want, of course)

I can look into that some time soon, but I think focus on "bling bling" right now seems more appropriate ?

DJMcNab:
> but I think focus on "bling bling" right now seems more appropriate ?

Probably 😆. It would be nice to have at least some tests we can point to, but I see your point

As I say, merging-as-is is fine, although I've not run this myself

Philipp-M:
> Are you willing to add a Masonry level test for it?

I think I would do that, but for me locally a lot of the tests are failing, because I suspect a different font is chosen for me (new test snapshots are generated, with slightly different looking font), not sure currently how to force one specific font (I haven't looked into it, but I think it makes sense to include a font in the repo, and force it in the tests?)
Tagging @PoignardAzur 

Anyway, it should be easy to add a test for the direction, but I'd rather merge this now and do that in different PR.

PoignardAzur:
Yeah, we can probably postpone tests until we figure out how to get a reliable font stack.

Overall LGTM.

crates/xilem_masonry/src/view/label.rs
ratmice: if we kept `Default`, this could just use functional record update syntax,
since `label` is the only non-default field.  Agree about the alpha value.
 
```
Label {
  label: label.into(),
  ..Default::default()
}
```

DJMcNab: I'm not sure I follow. `Label: Default` would not be a good API, because there's nearly no use for an empty label.

(Plus, I think that would allocate without need, which isn't great.)

ratmice: Of course, I just momentarily forgot functional record update syntax requires the structure itself to be `Default` not just the fields.

PR #227, opened 2024-04-27
deps: Use `raw-window-handle` 0.6.0

waywardmonkeys:
We don't need to force the usage of 0.5.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1418,8 +1418,7 @@ dependencies = [
  "log",
  "ndk-sys",
  "num_enum",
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
If it still compiles, then this seems reasonable. Presumably, the implementation for rwh 0.5 was just not being used?

waywardmonkeys:
It does better than just compile! It also runs!

I'm not sure why this was doing this in the first place, probably migration from glazier which was using `raw-window-handle` 0.5 ...

PR #229, opened 2024-04-28
Update the hot state in `ParentWindowOrigin` as well

DJMcNab:
The exact circumstances in which ParentWindowOrigin need to be sent aren't entirely clear to me so this might need some more careful thought

This does resolve [#masonry > Hot state for new widgets](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Hot.20state.20for.20new.20widgets)
This can be tested in the `mason` example, for example when pressing the `activate` button


PoignardAzur (approved):
Looks good. We can do a more complete layout refactor later.

PR #233, opened 2024-04-29
Run tests in CI, and fix fonts for snapshot testing

DJMcNab:
See also linebender/vello#439

This should open the door to doing much more comprehensive testing at higher level in the nearish future (including e.g. in `xilem_masonry`)


DJMcNab:
Discussion in [#masonry > Snapshot test differences in Masonry CI](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Snapshot.20test.20differences.20in.20Masonry.20CI)

crates/masonry/src/testing/harness.rs
```diff
@@ -175,11 +175,24 @@ impl TestHarness {
...
+                },
+            ),
             mouse_state,
             window_size,
             background_color,
         };
+        const FONT_PATH: &str = concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/resources/fonts/roboto/Roboto-Regular.ttf"
+        );
```
PoignardAzur, line 192: Wouldn't that path be wrong for any dependent crate creating a test harness?

I think you want to use `include_bytes` instead (until we come up with an actual resource loading system).

crates/masonry/src/render_root.rs
```diff
@@ -144,6 +162,20 @@ impl RenderRoot {
         self.root_on_text_event(event)
     }
 
+    // TODO: Is this the right place for this?
+    pub fn add_test_font(
```
PoignardAzur, line 166: Seems reasonable. RenderRoot is the type ultimately responsible for turning semantic data into bitmaps. You can either pass a test font with a constructor argument, a build method, or a setter. I think a setter is fine.

PoignardAzur:
Will try to run this on my machine this afternoon, then review.

DJMcNab:
Ooh, a time limit to get it working? Exciting!

DJMcNab:
The status of this is that there's a lot more work needed on testing, including moving this into Xilem. However, this does get our existing tests working in a cross-platform way.

One concern I have with the size of screenshot tests. I'm planning on raising this in office hours, to see if anyone has any good solutions, so won't merge until them.
Ideally, we could just use git lfs, but they don't want to support

PoignardAzur:
Unit tests pass on my machine! :tada: 

(PopOs 22.04, RTX 2070 with X11, for what it's worth.)

DJMcNab:
I'm going to postpone merging this PR until after I port https://github.com/linebender/vello/pull/647 to this repository, to avoid adding the new snapshots into main as-is.

I think I'll do that in a different PR, because it also has implications on other caching choices, and shouldn't be controversial.

PoignardAzur:
Makes sense, yes. FWIW the developer experience of git-lfs in Vello seems pretty good to me.

masonry/src/widget/tests/safety_rails.rs
```diff
@@ -80,6 +80,7 @@ fn check_forget_to_recurse_widget_added() {
 #[should_panic(expected = "not visited in method layout")]
 #[test]
 fn check_forget_to_recurse_layout() {
+    assert!(cfg!(debug_assertions));
```
PoignardAzur, line 83: This is incorrect: you can run unit tests in release mode.

This should probably be an early exit instead.

PoignardAzur (approved):
LGTM.

masonry/src/widget/tests/safety_rails.rs
DJMcNab: An early exit isn't possible here, because this is a `should_panic` test.

PoignardAzur: Oooh! That's annoying. I guess then we should either:

- `#[cfg]` the test.
- Create a synthetic panic in release mode.
- Use `catch_unwind` instead.

In any case, you don't need to solve this now, but a TODO comment would be nice.

PoignardAzur: Created #477 for this, you can link to it.

masonry/src/testing/harness.rs
```diff
@@ -514,7 +528,7 @@ impl TestHarness {
         test_module_path: &str,
         test_name: &str,
     ) {
-        if option_env!("SKIP_RENDER_SNAPSHOTS").is_some() {
+        if std::env::var("SKIP_RENDER_SNAPSHOTS").is_ok_and(|it| !it.is_empty()) {
             // FIXME - This is a terrible, awful hack.
             // We need a way to skip render snapshots on CI and locally
             // until we can make sure the snapshots render the same on
```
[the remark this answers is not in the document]
DJMcNab: I'll do that when I rebase this past #475

masonry/src/testing/harness.rs
```diff
@@ -526,7 +540,7 @@ impl TestHarness {
             return;
         }
 
-        let new_image = self.render();
+        let new_image: DynamicImage = self.render().into();
```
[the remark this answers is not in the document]
DJMcNab: I'm not sure. I'm wanting to redo a lot of this testing code, but I think that will want to come after linebender/rfcs#7

DJMcNab:
Requested review just to get final sign-off on the latest pushed changed (since ac7549d8ade33d95ef925737c4aab5ed156c2861)

masonry/src/testing/harness.rs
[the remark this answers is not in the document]
PoignardAzur: Wait, no, I misread the code. We already redraw in the early exit, the only thing this skips is actually sending the Vello scene to GPU. So that makes sense.

masonry/src/widget/tests/safety_rails.rs
```diff
@@ -53,6 +53,10 @@ fn check_forget_to_recurse_text_event() {
 
 #[should_panic(expected = "not added in method lifecycle")]
 #[test]
+#[cfg_attr(
+    not(debug_assertions),
+    ignore = "This test doesn't work without debug assertions (i.e. in release mode). See https://github.com/linebender/xilem/issues/477"
+)]
```
PoignardAzur, line 59: I don't think you need to link to the issue. Doing a `cfg_attr` fixes it.

For the message I'd write something like:

"Safety rails are disabled when debug_assertions is false (i.e. in release mode)."

PoignardAzur (approved):
LGTM. Tests still compile on my machine.

PR #234, opened 2024-04-29
re-export parley && vello

ykerit:
I try to use masonry to implement my own application, which requires vello and parley

### DIFF
--- a/crates/masonry/src/lib.rs
+++ b/crates/masonry/src/lib.rs
@@ -89,11 +89,8 @@
 
 // TODO - Add logo
 
-#[doc(inline)]
 pub use kurbo;
-#[doc(inline)]
 pub use parley;
-#[doc(inline)]
 pub use vello;
 
 #[macro_use]
--- a/crates/masonry/src/lib.rs
+++ b/crates/masonry/src/lib.rs
@@ -91,6 +91,10 @@
 
 #[doc(inline)]
 pub use kurbo;
+#[doc(inline)]
+pub use parley;
+#[doc(inline)]
+pub use vello;
 
 #[macro_use]
 mod util;


DJMcNab:
Thanks for your interest!

I do think we'll probably want this, however, the same logic also applies elsewhere down the stack. This leads to there being lots of paths to the same item, which leads to a poor experience when using the crates.

E.g. you end up having `masonry::kurbo::Point` and `masonry::vello::kurbo::Point` and `masonry::parley::kurbo::Point`, and worse, you will likely end up using different paths in different imports.

Ideally there would be an easy way to avoid that, but we haven't worked that out yet. Suggestions welcome

See also #223, where I suggest deferring a similar question until after the upcoming conference.

DJMcNab:
I've opened a [thread in the Zulip](https://xi.zulipchat.com/#narrow/stream/419691-linebender/topic/Re-exports.20policy/near/436130449) as an easier place to discuss solutions

ykerit:
ok, looking forward to a better solution

DJMcNab (approved):
After some discussion in Zulip, I'm relatively happy to land this, recognising that we might rework this in future.

I'm not convinced by the `#[doc(inline)]` attribute; it has some pretty big disadvantages. E.g. it breaks [source links](https://docs.rs/druid/latest/druid/kurbo/struct.QuadBez.html) on docs.rs. Cc @xStrom , as you added the attribute causing this in https://github.com/linebender/druid/pull/733

Please remove the `doc(inline)` attribute from at least the new items, then we can land this. Thanks!

(I think for Vello, I'd prefer to re-export a smaller subset of their API, from the root, for example)

simbleau:
Vello and Xilem should have a prelude for ergonomics, where these types live.

ykerit:
> After some discussion in Zulip, I'm relatively happy to land this, recognising that we might rework this in future.
> 
> I'm not convinced by the `#[doc(inline)]` attribute; it has some pretty big disadvantages. E.g. it breaks [source links](https://docs.rs/druid/latest/druid/kurbo/struct.QuadBez.html) on docs.rs. Cc @xStrom , as you added the attribute causing this in [linebender/druid#733](https://github.com/linebender/druid/pull/733)
> 
> Please remove the `doc(inline)` attribute from at least the new items, then we can land this. Thanks!
> 
> (I think for Vello, I'd prefer to re-export a smaller subset of their API, from the root, for example)
Done, Already removed

DJMcNab:
@ykerit what do you mean by this comment? https://github.com/linebender/xilem/pull/234#issuecomment-2085534725

ykerit:
In the last line I mentioned that I have completed what you suggested  😂
@DJMcNab

PR #235, opened 2024-04-30
`xilem_masonry`: Create a generic `View` trait, and use it to complete the `Flex` view

DJMcNab:
Functionally, this code works, but there's a lot of complexity around the different traits which have been added, which is unfortunate.

That is, the ergonomics are good, but that comes at the cost of some how understandable it is.

### DIFF
--- a/crates/masonry/src/widget/flex.rs
+++ b/crates/masonry/src/widget/flex.rs
@@ -383,7 +383,7 @@ impl<'a> FlexMut<'a> {
         child: impl Widget,
         params: impl Into<FlexParams>,
     ) {
-        self.insert_flex_child_pod(idx, WidgetPod::new(Box::new(child)), params)
+        self.insert_flex_child_pod(idx, WidgetPod::new(Box::new(child)), params);
     }
 
     pub fn insert_flex_child_pod(
--- a/crates/masonry/src/widget/widget_mut.rs
+++ b/crates/masonry/src/widget/widget_mut.rs
@@ -70,7 +70,7 @@ impl<W: StoreInWidgetMut> Drop for WidgetMut<'_, W> {
         #[allow(unsafe_code)]
         unsafe {
             // Safety: We never call this drop whilst self is in an invalid state
-            ManuallyDrop::drop(&mut self.inner)
+            ManuallyDrop::drop(&mut self.inner);
         }
     }
 }
@@ -116,7 +116,7 @@ impl<'a> WidgetMut<'a, Box<dyn Widget>> {
         // Logic: The merge_up is definitely called, either in the None
         // arm below or in the destructor of the returned value
         std::mem::forget(self);
-        let (widget, mut ctx) = Box::<dyn Widget>::into_widget_and_ctx(widget_mut);
+        let (widget, ctx) = Box::<dyn Widget>::into_widget_and_ctx(widget_mut);
         match widget.as_mut_any().downcast_mut() {
             Some(widget) => {
                 let ctx = WidgetCtx {
@@ -130,7 +130,7 @@ impl<'a> WidgetMut<'a, Box<dyn Widget>> {
             }
             None => {
                 if root {
-                    parent_widget_state.merge_up(&mut ctx.widget_state);
+                    parent_widget_state.merge_up(ctx.widget_state);
                 }
                 None
             }
--- a/crates/xilem_masonry/src/lib.rs
+++ b/crates/xilem_masonry/src/lib.rs
@@ -206,8 +206,8 @@ pub trait ViewElement {
     type Mut<'a>;
     type Erased: ViewElement;
     fn erase(self) -> Self::Erased;
-    fn downcast<'m>(erased: <Self::Erased as ViewElement>::Mut<'m>) -> Self::Mut<'m>;
-    fn reborrow<'r, 'm>(reference_mut: &'r mut Self::Mut<'m>) -> Self::Mut<'r>;
+    fn downcast(erased: <Self::Erased as ViewElement>::Mut<'_>) -> Self::Mut<'_>;
+    fn reborrow<'r>(reference_mut: &'r mut Self::Mut<'_>) -> Self::Mut<'r>;
 }
 
 impl<W: Widget + StoreInWidgetMut> ViewElement for WidgetPod<W> {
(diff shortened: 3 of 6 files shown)


DJMcNab:
Marking as draft because the distribution of "responsibilities" between the traits isn't very clear

crates/masonry/src/widget/flex.rs
```diff
@@ -482,12 +462,83 @@ impl<'a> FlexMut<'a> {
...
+        let child_val = std::mem::replace(child, Child::FixedSpacer(0.0, 0.0));
+        let widget = match child_val {
+            Child::Fixed { widget, .. } | Child::Flex { widget, .. } => widget,
+            _ => {
+                panic!("Can't update flex parameters of a spacer element");
+            }
+        };
+        let new_child = new_flex_child(params.into(), widget);
+        *child = new_child;
+        // TODO
```
Philipp-M, line 476: Are these `TODO` comments, a result of copy paste, or is there really something todo? (and what?)

Btw. I think the two lines below this, could be summed up with
```rust
        self.ctx.children_changed();
```
couldn't it?

crates/masonry/src/widget/widget_mut.rs
```diff
@@ -29,14 +30,48 @@ use crate::{Widget, WidgetId, WidgetState};
...
-    pub(crate) inner: W::Mut<'a>,
+    pub(crate) parent_widget_state: ManuallyDrop<&'a mut WidgetState>,
+    pub(crate) inner: ManuallyDrop<W::Mut<'a>>,
+    /// Whether this `WidgetMut` is the "root" `WidgetMut` of the reborrow tree, i.e.
+    /// whether we need to call merge_up in drop, or whether that will happen later
+    root: bool,
+}
+
+impl<'a, W: Widget + StoreInWidgetMut> WidgetMut<'a, W> {
+    pub(crate) fn new(parent_widget_state: &'a mut WidgetState, inner: W::Mut<'a>) -> Self {
```
Philipp-M, line 41: I think all of this unsafe code is only "safe" because it's only possible to construct this type within masonry isn't it?

Not super comfortable with this, because it's... well unsafe as well as (slightly) complex (not the best combination). Not sure whether all of that could be achieved with just some reference counting.

crates/xilem_masonry/src/lib.rs
```diff
@@ -201,18 +201,41 @@ where
         EventLoopRunner::new(self.root_widget, window, event_loop, self.driver).run()
     }
 }
-pub trait MasonryView<State, Action = ()>: Send + 'static {
-    type Element: Widget + StoreInWidgetMut;
+
+pub trait ViewElement {
+    type Mut<'a>;
+    type Erased: ViewElement;
```
Philipp-M, line 207: Would be good to at least add some basic documentation to this trait. E.g. that this is type-erased and not necessarily a deleted element.

crates/xilem_masonry/src/sequence.rs
```diff
@@ -65,13 +68,75 @@ pub struct WasAView;
...
+    /// Convert from the mutable form into the immutable form
+    ///
+    /// This may panic *if* the reference doesn't come from an element returned from `Self`
+    fn access_mut<R>(
+        reference: Self::Mut<'_>,
+        f: impl FnOnce(&mut SourceElement::Mut<'_>) -> R,
+    ) -> R;
+}
+
+impl<SourceElement: ViewElement, SequenceElement: ViewElement> SequenceCompatible<SourceElement>
```
Philipp-M, line 95: Maybe put all the bounds into the `where` clause (for ease of understanding this)? Or is this intended?

crates/xilem_masonry/src/sequence.rs
```diff
@@ -65,13 +68,75 @@ pub struct WasAView;
...
+    fn access_mut<R>(
+        reference: Self::Mut<'_>,
+        f: impl FnOnce(&mut <SourceElement as ViewElement>::Mut<'_>) -> R,
+    ) -> R {
+        let mut into = reference.into();
+        f(&mut into)
+    }
+}
+
+pub struct TrivialWrapper<T>(T);
```
Philipp-M, line 114: I don't see any use of this, is this a remnant from previous experiments?

crates/xilem_masonry/src/sequence.rs
```diff
@@ -113,11 +176,16 @@ pub struct OptionSeqState<InnerState> {
 }
 
 // TODO(DJMcNab): Add the generation to the id path
```
Philipp-M, line 178: Not related to this PR, but I don't think that comment is necessary anymore?

crates/xilem_masonry/src/view/flex.rs
```diff
@@ -1,11 +1,14 @@
...
 };
 
-use crate::{ChangeFlags, ElementSplice, MasonryView, VecSplice, ViewSequence};
+use crate::{
+    sequence::SequenceCompatible, ChangeFlags, ElementSplice, MasonryView, VecSplice, View,
+    ViewElement, ViewSequence,
+};
 
 // TODO: Allow configuring flex properties. I think this actually needs its own view trait?
 pub fn flex<VT, Marker>(sequence: VT) -> Flex<VT, Marker> {
```
Philipp-M, line 14: (not directly related to that PR)

I'm not sure whether we actually should add the type bounds here (generally for these "component" functions).

I think for one it helps a little bit with error messages, and since I think it should only be used inside a `View` context, it also shows, what is actually allowed here as sequence. (But well, unfortunately it adds a lot of generic parameters obviously)

crates/xilem_masonry/src/view/flex.rs
```diff
@@ -16,6 +19,13 @@ pub fn flex<VT, Marker>(sequence: VT) -> Flex<VT, Marker> {
     }
 }
 
+pub fn flex_item<V>(view: V, params: impl Into<FlexParams>) -> FlexWrapper<V> {
```
Philipp-M, line 22: More of a UX thing and not entirely relevant to that PR, we could also achieve this via an extension trait. So that this is possible via the builder pattern (maybe also within `MasonryView`, to avoid extra traits? (see also e.g. [here](https://github.com/Philipp-M/trui/blob/dd24a2921eb0587dfffbbafea4a3e62b92f39c43/src/view.rs#L56)))

Philipp-M (approved):
> but that comes at the cost of some how understandable it is.

Yep, took me some time to dive through this...
Pretty clever stuff, neat trick to do this with the `SequenceCompatible` trait.

What I think though, is that we should document this thoroughly (at some time, not necessarily in this PR), to avoid scaring off new Contributors with that added complexity. But what I like, is that the added complexity is mostly internal at the core, so I don't think it's necessary to have a deep understanding of all of this to just add simple stuff (widgets etc.), unless a completely new `ViewSequence<SpliceElement>` context is necessary (like `FlexElement`).

I guess we need the differentiation between the traits `MasonryView` and `View` because of `FlexElement`?
(I can see some use cases with having a "generic" `View` though anyway, maybe we somehow are able to marry/merge xilem_web with xilem_masonry with that at some point?)

One thing I'm wondering is, how much compilation time this adds (for reasonbly large view logic functions), because of added associated trait bounds (which seem to be the main culprit for added compilation time).

(Btw. [related](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Flex.20parameters), but I prefer this, i.e. strong-typed at the expense of complexity as long as it only stays (mostly) inside of xilems implementation over dynamic typing.)

I'm cautiously approving this, but I think it's a good idea to discuss this tomorrow before merging.

DJMcNab:
This track of work is continuing in #310

PR #237, opened 2024-05-01
Port Xilem and Masonry to winit 0.30.0

xorgy:


Cargo.toml
```diff
@@ -70,4 +70,4 @@ bitflags = "2"
 tracing = "0.1.37"
 fnv = "1.0.7"
 instant = { version = "0.1.6", features = ["wasm-bindgen"] }
-winit = "0.29"
+winit = { version = "0.30", features = ["rwh_05"]}
```
waywardmonkeys, line 73: This `rwh_05` isn't needed.

waywardmonkeys (changes requested):

Cargo.toml
xorgy: Ope, yeah had to add that to figure something out, but not needed now.

waywardmonkeys:
How hard would it be to also do the same for the new Masonry-based stuff?

I'd started looking at it, I think, but got busy with other stuff for work ...

xorgy:
> How hard would it be to also do the same for the new Masonry-based stuff?

I have no idea, that's something to ask Daniel. Right now I'm doing this to simplify the AccessKit integration in classic Xilem.

xorgy:
We could pin masonry winit to `0.29.0` for now if that's helpful.

xorgy:
Looks like there are four examples directly interacting with winit event loops, and also `crates/masonry/src/event_loop_runner.rs`

waywardmonkeys:
I just suspect that a bare-bones port might be simple enough to just do it at all at the same time and not bloat CI times by having to compile multiple versions of all sorts of things ... If it is, then great. If it isn't, then we could just keep it as it is where one thing is using 0.30 and the other is still on 0.29.

xorgy:
I have a quick and dirty port of masonry event_loop_runner, just testing it out now

Cargo.toml
```diff
@@ -70,4 +70,4 @@ bitflags = "2"
 tracing = "0.1.37"
 fnv = "1.0.7"
 instant = { version = "0.1.6", features = ["wasm-bindgen"] }
-winit = "0.29"
+winit = "0.30"
```
waywardmonkeys, line 73: Might as well switch to a workspace dependency now.

crates/masonry/src/event_loop_runner.rs
```diff
@@ -5,123 +5,169 @@ use tracing::warn;
...
+                }
+                render_root::RenderRootSignal::SetCursor(cursor) => {
+                    self.window.set_cursor(cursor);
+                }
+                render_root::RenderRootSignal::SetSize(size) => {
+                    // TODO - Handle return value?
+                    let _ = self.window.request_inner_size(size);
+                }
+                render_root::RenderRootSignal::SetTitle(title) => {
+                    self.window.set_title(&title);
```
waywardmonkeys, line 167: These should probably use user events in winit, but that's for the future, not today.

waywardmonkeys (changes requested):
As long as others are okay with the Masonry specific parts, I'm okay with the rest for a first pass. Just should change the one dependency to be a workspace dep.

PoignardAzur:
LGTM.

crates/masonry/src/event_loop_runner.rs
xorgy: I think Masonry signals serve a different purpose.

DJMcNab (approved):
I have *not* scanned through this, but am just formalizing the two other approvals

waywardmonkeys (approved):

PR #238, opened 2024-05-01
deps: Use workspace vello, release parley.

waywardmonkeys:
`xilem` now uses the same `vello` as others (git for now) and everyone uses the new released version of `parley`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,12 +89,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.82"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
We'll need a new release of `vello` before publishing Xilem stuff. Hopefully, we can land `wgpu` 0.20 prior to that ...

waywardmonkeys:
Hmm, looks like we could use the released 0.1.0 of Vello and things still build ...

waywardmonkeys:
If we use released `vello`, then we end up with multiple versions of `skrifa` and `read-fonts` , but it builds and runs.

But I think we want the wgpu 0.20 update if we can get it for the shader compilation improvements!

xStrom (approved):
Another step towards publishing! 🎉

Using workspace Vello makes sense. Also, yeah, still holding out hope for `wgpu` v0.20.

PR #240, opened 2024-05-01
Remove StoreInWidgetMut and declare_widget

PoignardAzur:
This makes the WidgetMut type much simpler to implement, at the cost of not being able to use it as a receiver for custom widgets.

See this discussion for details:
https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Thoughts.20on.20simplifying.20WidgetMut


crates/masonry/src/widget/widget_mut.rs
```diff
@@ -21,67 +18,34 @@ use crate::{Widget, WidgetId, WidgetState};
...
-/// which is a special type declared with the `declare_widget` macro. Methods to mutate
-/// the widget will be implemented in that `W::Mut` type, which `WidgetMut<W>` derefs to.
+/// ## WidgetMut as a Receiver
 ///
-/// See [`declare_widget`](crate::declare_widget) for details.
-pub struct WidgetMut<'a, W: Widget + StoreInWidgetMut> {
-    pub(crate) parent_widget_state: &'a mut WidgetState,
-    pub(crate) inner: W::Mut<'a>,
+/// Once the Receiver trait is stabilized, WidgetMut will implement it so that custom
+/// widgets in downstream crates can use WidgetMut as the receiver for inherent methods.
```
DJMcNab, line 24: I'd at least mention extension traits here

crates/masonry/src/widget/widget_mut.rs
```diff
@@ -21,67 +18,34 @@ use crate::{Widget, WidgetId, WidgetState};
...
+pub struct WidgetMut<'a, W: Widget> {
+    pub ctx: WidgetCtx<'a>,
+    pub widget: &'a mut W,
```
DJMcNab, line 27: Not being able to access the inner widget mutably from outside was in my mind a strong advantage of the old approach.

It meant that widgets could expose mutable access to their state during construction, and yet not have a chance for these methods to be externally called. It also opens the door to `mem::swap` nonsense

I don't have a suggestion for how to recover this property, unfortunately

DJMcNab: This is especially troubling, because these mutable methods will be at least slightly more ergonomic than the "proper" versions which can access the context properly (although not having `DerefMut` does help slightly there)

crates/masonry/src/widget/widget_mut.rs
PoignardAzur: Right, this is very minimal documentation. I should probably have added a TODO.

crates/masonry/src/widget/widget_mut.rs
PoignardAzur: > Not being able to access the inner widget mutably from outside was in my mind a strong advantage of the old approach.

Nope, you could get a `&mut` to the inner widget with the old method too. It's necessary for the setter methods to be able to access said inner widget.

> It meant that widgets could expose mutable access to their state during construction

I think we should encourage users to only use the builder pattern in those cases.

> It also opens the door to mem::swap nonsense

At that point, if the user wants to shoot themselves in the foot that hard, it's their problem.

DJMcNab: Oh, as in you could access it using the `StoreInWidgetMut` methods

Good point

DJMcNab (approved):
I do think this is probably an improvement in readability etc., but it does have some unfortunate consequences

PR #241, opened 2024-05-02
A foundation for text editing and selection in Masonry

DJMcNab:
This brings in a lot of the old work from Druid and Masonry  (prior to linebender/masonry#56) on text, as well as some types from Glazier.

Needed work:
- [X] Text display using abstract types
- [ ] Text selection with mouse
- [ ] Text input with keyboard
- [ ] IME integration (of the kind winit understands)

Follow up work:
- [ ] Keyboard control of the selection area (hard-coded hotkeys)
- [ ] [Proper placement](https://raphlinus.github.io/text/2020/10/26/text-layout.html#shaping-cluster) of cursor
- [ ] Input methods
- [ ] Proper hotkey handling
- [ ] Copy and/or paste


crates/masonry/src/text2/layout.rs
```diff
@@ -0,0 +1,463 @@
...
+            self.brush = brush;
+            self.invalidate();
+        }
+    }
+
+    /// Set the default brush used for the layout.
+    ///
+    /// See also [`set_text_color`][Self::set_text_color]
+    /// for a convenience method when using a solid color.
+    pub fn set_overrid_brush(&mut self, override_brush: Option<Brush>) {
```
waywardmonkeys: Typo in function name

crates/masonry/src/text2/layout.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// See also [`set_text_color`][Self::set_text_color]
+    /// for a convenience method when using a solid color.
+    pub fn set_brush(&mut self, brush: Brush) {
+        if brush != self.brush {
+            self.brush = brush;
+            self.invalidate();
+        }
+    }
+
+    /// Set the default brush used for the layout.
```
waywardmonkeys, line 129: Shouldn't say `default` here as that's copied from the above function, I think.

crates/masonry/src/text2/layout.rs
```diff
@@ -0,0 +1,463 @@
...
+    pub fn set_text_alignment(&mut self, alignment: Alignment) {
+        if self.alignment != alignment {
+            self.alignment = alignment;
+            self.invalidate();
+        }
+    }
+
+    /// Set the width at which to wrap words.
+    ///
+    /// You may pass `f64::INFINITY` to disable word wrapping
```
waywardmonkeys: What would passing `None` here mean?

crates/masonry/src/text2/layout.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// The nominal size of the layout.
+    pub size: Size,
+    /// The distance from the nominal top of the layout to the first baseline.
+    pub first_baseline: f32,
+    /// The width of the layout, inclusive of trailing whitespace.
+    pub trailing_whitespace_width: f32,
+    //TODO: add inking_rect
+}
+
+impl<T> TextLayout<T> {
```
waywardmonkeys, line 76: Just a general comment that some things are prefixing with `set_` and others with `set_text_` and sometimes `set_color` is used in here and in other places `set_text_color` and it might be nice to make this more consistent. (Probably not immediately before RustNL...)

DJMcNab: Yeah, all of this API design could do with a tiny bit of cleanup; there's quite a lot of redudancy in these methods

xStrom (approved):
Approving in the interest of v0.1 for RustNL, with the understanding that the code introduced here is buggy and is missing functionality that will be addressed in future PRs.

PR #243, opened 2024-05-02
Masonry: Only activate buttons when they are hot

DJMcNab:
This matches the behaviour in Xilem classic

### DIFF
--- a/crates/masonry/src/widget/button.rs
+++ b/crates/masonry/src/widget/button.rs
@@ -82,7 +82,7 @@ impl Widget for Button {
                 }
             }
             PointerEvent::PointerUp(_, _) => {
-                if ctx.is_active() && !ctx.is_disabled() {
+                if ctx.is_active() && ctx.is_hot() && !ctx.is_disabled() {
                     ctx.submit_action(Action::ButtonPressed);
                     ctx.request_paint();
                     trace!("Button {:?} released", ctx.widget_id());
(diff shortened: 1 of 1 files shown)


crates/masonry/src/widget/button.rs
```diff
@@ -82,13 +82,18 @@ impl Widget for Button {
                 }
             }
             PointerEvent::PointerUp(_, _) => {
-                if ctx.is_active() && !ctx.is_disabled() {
+                if ctx.is_active() && ctx.is_hot() && !ctx.is_disabled() {
```
waywardmonkeys, line 85: Just curious when this would be active, but not hot?

crates/masonry/src/widget/button.rs
```diff
@@ -82,13 +82,18 @@ impl Widget for Button {
...
                     ctx.submit_action(Action::ButtonPressed);
                     ctx.request_paint();
                     trace!("Button {:?} released", ctx.widget_id());
                 }
                 ctx.set_active(false);
             }
+            PointerEvent::PointerLeave(_) => {
+                // If the screen was locked whilst holding down the mouse button, we don't get a `PointerUp`
+                // event, but should no longer be active
```
waywardmonkeys, line 94: This doesn't need checking but since I often switch between having 1 or 2 external monitors attached to my laptop, I wonder if `winit` handles a `PointerLeave` in that scenario as well!

crates/masonry/src/widget/button.rs
DJMcNab: When you have started holding down the mouse, moved it away, then raised the mouse.

PoignardAzur (approved):
Huh. It's weird that Masonry didn't already implement that.

PR #244, opened 2024-05-02
Add initial accessibility support

PoignardAzur:
Add AccessKit dependency.
Add accesskit_winit dependency.
Add methods to the Widget trait which create the accessibility tree and react to accessibility events.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,94 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
+name = "accesskit"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6cf780eb737f2d4a49ffbd512324d53ad089070f813f7be7f99dbd5123a7f448"
+
+[[package]]
+name = "accesskit_atspi_common"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3f1db0583df3e2f52501231d49001d17403bbdae4d14e359776cc87f591b618"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "atspi-common",
+ "serde",
+ "thiserror",
+ "zvariant",
+]
+
+[[package]]
+name = "accesskit_consumer"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "329bc2cce90b24356497badc4824c02986298f7b6c5ea208ceed59532448b6f6"
+dependencies = [
+ "accesskit",
+]
+
+[[package]]
+name = "accesskit_macos"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5abdf2e75371122f769baeee90ecfd3987b1b00d6ff596fd1eec5d778795e19a"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "objc2",
+ "objc2-app-kit",
+ "objc2-foundation",
+ "once_cell",
+]
+
+[[package]]
+name = "accesskit_unix"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "64bb5d87ed1f38c0e905de65cfe39b8adea00ec00c842308d020ebeb8c5da7ed"
+dependencies = [
+ "accesskit",
+ "accesskit_atspi_common",
+ "async-channel",
+ "async-executor",
+ "async-task",
+ "atspi",
+ "futures-lite 1.13.0",
+ "futures-util",
+ "serde",
+ "zbus",
+]
+
+[[package]]
+name = "accesskit_windows"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2f80369680515d6929059cd19a6d154800506afc36430aed3ddddb9f2ab153a"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "paste",
+ "static_assertions",
+ "windows 0.54.0",
+]
+
+[[package]]
+name = "accesskit_winit"
+version = "0.20.0"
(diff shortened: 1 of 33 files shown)


PoignardAzur:
Note that accessibility events are still missing. I'm hoping to get them done before office hours today.

mwcampbell:
I suggest you drop the new `app_name` method from `AppDriver` and don't set `app_name` on `accesskit::Tree`. Requiring every application to explicitly provide an app name adds to the boilerplate, and I don't think it's justified. Only AT-SPI has an app name property in the first place, and strictly speaking, it's not needed. I'm having second thoughts anyway about whether AccessKit should have such a property; our AT-SPI backend could set it from the executable name, after all.

mwcampbell:
Are we always going to require every application to directly create its winit window, as I'm seeing in the Masonry examples? If so, that's going to make it inconvenient to implement accessibility correctly, and very easy to get it wrong, as the examples currently do. As documented in `accesskit_winit` and shown in that crate's examples, it's necessary to create the window invisible, then create the AccessKit adapter, then show the window. Creating the window initially visible (the default) and then creating the AccessKit adapter doesn't work reliably on Windows. I haven't tested that case on macOS.

mwcampbell:
Calling `accesskit_winit::Adapter::process_event` for every `WindowEvent`, as shown in the `accesskit_winit` examples, is absolutely necessary on macOS and Linux.

mwcampbell:
I see that `accesskit_winit::Adapter::update_if_active` is being called for the initial tree, but not for subsequent updates. Is that what you meant by accessibility events being missing? I assumed that was referring to action requests.

mwcampbell:
It looks like `root_accessibility` leaves the role of the root node at the default, which is `Role::Unknown`. It should be `Role::Window`.

PoignardAzur:
> I suggest you drop the new app_name method from AppDriver and don't set app_name on accesskit::Tree

Done.

> Are we always going to require every application to directly create its winit window, as I'm seeing in the Masonry examples?

I'd like the default workflow to be "Provide WindowAttributes to Masonry, Masonry creates the window".

> I see that accesskit_winit::Adapter::update_if_active is being called for the initial tree, but not for subsequent updates.

Fixed in last commit.

> It looks like `root_accessibility` leaves the role of the root node at the default, which is `Role::Unknown`. It should be `Role::Window`.

Actually, the one place where `NodeBuilder::default` is called creates kind of a fake "parent of the root" node. That node is never added to accessibility tree.

I can set `Role::Window` in the xilem root widget, though.

PoignardAzur:
> Calling `accesskit_winit::Adapter::process_event` for every `WindowEvent`, as shown in the `accesskit_winit` examples, is absolutely necessary on macOS and Linux.

Done.

> If so, that's going to make it inconvenient to implement accessibility correctly, and very easy to get it wrong, as the examples currently do. As documented in `accesskit_winit` and shown in that crate's examples, it's necessary to create the window invisible, then create the AccessKit adapter, then show the window.

Done.

mwcampbell:
In widget containers or wrappers, it's incorrect for `accessibility_role` to return `Role::GenericContainer` while `accessibility` simply forwards the call to the child/inner widget. That would lead to a single accessibility node with a role of `GenericContainer` but other properties from the child widget. In other words, the correct accessibility role from the child widget wouldn't be set. You have to either forward the call to `accessibility_role` to the child as well, or the parent's `accessibility` method has to push a child node.

mwcampbell:
Please disregard my previous comment. In the cases I had looked at, `self.child` is a `WidgetPod`. So in that case, the `accessibility` implementation on `self.child` does push a child node, and what you're doing is correct.

PoignardAzur:
> Please disregard my previous comment. In the cases I had looked at, `self.child` is a `WidgetPod`. So in that case, the `accessibility` implementation on `self.child` does push a child node, and what you're doing is correct.

Right, in every single case in Masonry when a widget calls another widget's method, it's through WidgetPod. There's actually some code that tries to enforce that.

DJMcNab (approved):
Approval for process purposes

I've not had time to review this code myself. But it's really good we're getting this working

mwcampbell:
It's OK with me if we merge this now and do further fixes as separate PRs, so we don't have to keep rebasing this one.

jaredoconnell:
I've never used accessessiblity features, so I'm not sure exactly what to test, but if I turn on Windows 10's narrator, the app just crashes. Tested on the `mason` and `calc` examples. It crashes if it's already open, and it crashes if it's opened. The error message is not very descriptive.
The `mason` example does not crash on main.
```
 process didn't exit successfully: `C:\Users\Admin\source\repos\xilem\target\debug\examples\mason.exe` (exit code: 101)
```

What environment are you testing this on?

mwcampbell:
Oh, I hadn't tested the mason example yet. Looking at that now.

mwcampbell:
Yes, I get this panic with the mason example:

```
thread 'main' panicked at C:\Users\mattc\.cargo\registry\src\index.crates.io-6f17d22bba15001f\accesskit_consumer-0.19.0\src\tree.rs:142:9:
assertion `left == right` failed
  left: 5
 right: 0
```

It's panicking on this assertion in `src/tree.rs` in the `accesskit_consumer` crate:

```
assert_eq!(pending_nodes.len(), 0);
```

This happens if you include child nodes in the tree update without adding their IDs to the `children` property of any parent node.

mwcampbell:
If I run the calc example with NVDA already running, the bounding rectangles of all the nodes are set to (0, 0, 0, 0). This suggests that the `InitialTreeRequested` event handler is producing an accessibility tree before layout is done. As I said in chat, it's not strictly necessary to produce the initial tree inside that event handler. That event handler is already running asynchronously, due to a limitation of winit, so AccessKit is already having to use a placeholder tree at first. So it might be better to request a repaint, set some flag indicating that a full accessibility tree is required, and do the accessibility tree update as part of the normal lifecycle. We just have to make sure that the initial tree update is pushed soon, but not necessarily immediately inside that event handler.

mwcampbell:
Is Masonry setting the AccessKit bounding rectangles in physical coordinates? If not, it needs to. If I test with NVDA and move the mouse over the buttons in the calc example, what it speaks doesn't match what's actually under the mouse cursor, suggesting that the bounding rectangles are inaccurate.

mwcampbell:
The role of the root node should be `Role::Window`. As it is now, NVDA and Narrator announce, for example, "Simple Calculator grouping" instead of "Simple Calculator window".

PR #245, opened 2024-05-02
Wire up scale factor

raphlinus:
Generally prefer use of LogicalPosition to PhysicalPosition for layout, events, etc. Use scale factor to convert to logical position. Apply scale factor as final transform before Vello rendering.

This PR also makes physical_position available in PointerState.

There is a TODO for handling WindowEvent::Rescale, which remains and is not addressed. It should be at some future point.

### DIFF
--- a/crates/masonry/src/contexts.rs
+++ b/crates/masonry/src/contexts.rs
@@ -631,9 +631,7 @@ impl LayoutCtx<'_> {
         self.widget_state.local_paint_rect =
             self.widget_state.local_paint_rect.union(child.paint_rect());
 
-        let mouse_pos = self
-            .mouse_pos
-            .map(|pos| LogicalPosition::new(pos.x, pos.y));
+        let mouse_pos = self.mouse_pos.map(|pos| LogicalPosition::new(pos.x, pos.y));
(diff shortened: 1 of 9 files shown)


jaredoconnell (commented):
Works well on Mac. I'll test it on Windows and Linux later.
image

jaredoconnell (approved):
Works well on Windows and Mac. My linux VMs are too out of date to test it right now.
Should be ready to go once the issue the CI is flagging is fixed.

Philipp-M (approved):
Looks good. Also works on Linux (X11 (xmonad) and wayland (COSMIC)) as expected.

eeshvardasikcm:
> There is a TODO for handling WindowEvent::Rescale, which remains and is not addressed. It should be at some future point.

I have only been following headlines from an outside view, but I wouldn't think this is a problem for now. This WindowEvent::Rescale class signature looks similar to the way Android Studio Game Activity new project template first allowed a similar remaining to-do the last I remember. I will have to study more about this repository to see how the objectives and timeline for handling this call relate, but the recent trend appears to be that allowing some gap here is functional.

PR #247, opened 2024-05-03
masonry: Fix non-resetting visual active state when releasing active button outside of its area

Philipp-M:
When pressing the button and releasing it outside of its active area, the visual pressed state (background gradient) stays "pressed". This should fix this.

(Not related to that PR) I'm not sure, if there's a case, where the disabled state could change while pressing the button and trigger a similar behavior.

### DIFF
--- a/crates/masonry/src/widget/button.rs
+++ b/crates/masonry/src/widget/button.rs
@@ -86,6 +86,9 @@ impl Widget for Button {
                     ctx.request_paint();
                     trace!("Button {:?} released", ctx.widget_id());
                 }
+                if ctx.is_active() && !ctx.is_hot() && !ctx.is_disabled() {
+                    ctx.request_paint();
+                }
                 ctx.set_active(false);
             }
             PointerEvent::PointerLeave(_) => {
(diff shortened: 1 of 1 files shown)


Philipp-M:
> (Not related to that PR) I'm not sure, if there's a case, where the disabled state could change while pressing the button and trigger a similar behavior.

I think even that should be covered with the latest commit, by not checking the disabled state at all (should not be necessary, because having an active button can not happen without having it previously disabled AFAIK.

crates/masonry/src/widget/button.rs
```diff
@@ -86,6 +86,9 @@ impl Widget for Button {
                     ctx.request_paint();
                     trace!("Button {:?} released", ctx.widget_id());
                 }
+                if ctx.is_active() && !ctx.is_hot() {
```
DJMcNab, line 89: The only scenario where this case would not happen (if the first "happy path" case didn't happen) is if:
1) The mouse button was down (and that didn't activate another widget); then
2) The cursor was dragged into this widget then released

I would probably remove this conditional entirely, and deal with the extra paint cycle in that case. That also lets us remove the `request_paint` from the first condition

DJMcNab (approved):
Thanks for catching this. It seems like these kinds of state management issues are going to be hard to avoid

PR #248, opened 2024-05-03
Revert to `vello` v0.1.0.

xStrom:
As discussed at office hours we're going with Vello v0.1.0 for now, as using the git dependency blocks publishing and Vello 0.2 isn't ready.


waywardmonkeys (approved):
LGTM, but you did a full `cargo update` that isn't reflected in the commit message.

PR #249, opened 2024-05-03
Prepare `Cargo.toml` files for release.

xStrom:
* Rename classic `xilem` to `xilem_classic` and the new `xilem_masonry` to `xilem`.
  No directory structure changes yet to avoid merge conflicts before RustNL.
* Base common dependencies on workspace dependencies.
* Bump `masonry` version to `0.2.0`.
* Remove some legacy docs.rs configuration inherited from Druid/Glazier that is no longer needed.
* Unify ordering of properties.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1285,7 +1285,7 @@ dependencies = [
 
 [[package]]
 name = "masonry"
-version = "0.1.3"
+version = "0.2.0"
 dependencies = [
  "assert_matches",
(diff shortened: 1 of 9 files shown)


jaredoconnell (approved):
Looks good. I do find it a little unfortunate that we're losing the explicit `xilem_masonry`, since it exemplified how modular xilem is, but the alternatives to that renaming are no better.

PR #251, opened 2024-05-03
docs: Start fixing links and formatting.

waywardmonkeys:


waywardmonkeys:
There's a lot left to do, but I have to travel, so figured something was better than nothing.

jaredoconnell (approved):
Looks good. It's definitely nice to have types specified with back-ticks.

Issue #250, opened 2024-05-03
Implement a default logger for Masonry


PoignardAzur:
[From this discussion on Zulip.](https://xi.zulipchat.com/#narrow/stream/317477-masonry)

## Default subscriber

Setting the default subscriber is a very side-effecty action. You can only do it once, later attempts will get errors/panics. I had some problems in my tests because of that.

But I do like the idea of always having a subscriber on by default, eg one that dumps all traces to a log file in `/tmp`. Always-on logging means that if the user gets a crash or a bug, they don't need to restart/recompile their program with the logs on. This is especially important for intermittent bugs.

We should call `set_global_default` in `event_loop_runner::run_with`, and if we get a "default subscriber already set" error, just swallow it. By that point we've already taken control of the main loop; in theory the user could still try to set a default subscriber later and shoot themselves in the foot, but it seems very unlikely to happen. We can include a "default subscriber set" log to make it really clear what's happening.

The default tracer would:

- Dump logs to a non-local temp file (exact path is platform-dependent).
- On wasm, dump to the console, I guess.
- Have a default level of `DEBUG` in dev mode and `INFO` in release.
- Include an env filter, so users can do `LOG_LEVEL=trace cargo run` to change the log level without having to recompile.

## High-quality logs

If we want the logs to always run in the background by default, they must be performance-friendly, so we should probably keep to one log per pass (3-4 logs per frame). That log could carry some basic aggregated statistics like:

- What event started the pass (mouse click, keyboard key, repaint, etc.)
- How many widgets were visited.
- The focused widget's ID.
- Was the event handled?
- What actions were emitted?
- What lifecycle events were run?

I think that keeps the noise level low; but at the same time it's helpful info when you're browsing logs after a bug already occurred and trying to figure out what your program did.

## Skip high-density events

I'd also want some way to skip the traces of high-density events (eg mouse moves), because otherwise you can really easily flood your logs with repetitive garbage.

A simple implementation would be:

- Don't send a log at the beginning of a mouse move event.
- At the end, look if anything changed in the app (eg the hot widget changed).
  - If so, print a trace.
  - If not, skip the trace and increment a "mouse move frames skipped" counter.
- The next time you print a trace, first include a "X frames have been skipped" message.

### Batch high-density events

A more ambitious solution would be to implement a custom "batching" subscriber that would intercept all logs during said high-density events, and add them to a buffer.

When a low-density event is called, the high-density event buffer is flushed and prints an aggregation of all events.

So instead of

```
[...]

PR #252, opened 2024-05-03
Add default subscriber to Masonry

PoignardAzur:
This subscriber uses EnvFilter to change its value depending on the RUST_LOG env variable, so log levels can be changed without recompiling the app.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1303,6 +1303,7 @@ dependencies = [
  "smallvec",
  "tempfile",
  "tracing",
+ "tracing-subscriber",
  "unicode-segmentation",
  "vello",
  "wgpu",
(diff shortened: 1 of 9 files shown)


raphlinus (approved):
Looks good to me. Do you want to add a comment so that it auto-closes #250 on merge?

waywardmonkeys:
This is already merged but I think it is a bad idea for it to be part of the lib and it should be part of the executable targets even if that is somewhat annoying.

That is how this is usually done.

PoignardAzur:
We may revisit the decision at a later date, but for now the goal is maximum debuggability over everything else, and bundling a default subscriber helps with that.

See #250 for details.

PR #254, opened 2024-05-04
Improve formatting of logs

PoignardAzur:


crates/masonry/src/event_loop_runner.rs
```diff
@@ -307,22 +307,35 @@ impl MainState<'_> {
...
         };
+        // Use EnvFilter to allow the user to override the log level without recompiling.
+        // TODO - Print error message if the env var is incorrectly formatted.
         let env_filter = EnvFilter::builder()
             .with_default_directive(default_level.into())
             .with_env_var("RUST_LOG")
             .from_env_lossy();
+        // This format is more concise than even the 'Compact' default:
+        // - We print the time without the date (GUI apps usually run for very short periods).
+        // - We print the time with seconds precision (we really don't need microseconds=.
```
DJMcNab, line 332: Milliseconds would surely be useful?

crates/masonry/src/event_loop_runner.rs
```diff
@@ -307,22 +307,35 @@ impl MainState<'_> {
...
+        // - We don't print the target
```
DJMcNab, line 333: Not sure about this either - it seems useful to have some context about where a log is coming from?

Also, this comment doesn't seem to serve any purpose - it only duplicates the `with_target` line. It should probably explain why we exclude the target

PoignardAzur: My experience with these logs so far is that the target is visual noise. It almost never gives me info I'm looking for.

If you have a log message, you can already grep the log's location. The target only gives you "Which module was this log defined in?", which I've never found useful.

You're right that I should include this explanation.

crates/masonry/src/event_loop_runner.rs
PoignardAzur: In GUI? I don't see it. We don't do multi-threading yet, and we deal very little with network events. In my experience there hasn't been one case where the millisecond count was helpful.

The coarse time at least helps me get some context, eg "Ok, this happened 3 minutes ago when I was doing X, so this is probably Y".

DJMcNab: I was under the impression that we should think about GUI as a soft-realtime system, and so it is important to be following e.g. how long things take, and milliseconds is the right scale for that
I know I found the timestamps very useful in my Vello work.

But maybe our conceptions of what these logs are for are misaligned

DJMcNab (approved):
I would still strongly prefer to have the milliseconds - it should only add 4 more characters to the messages

But otherwise this is good, and I'm not going to block the merge on that

crates/masonry/src/event_loop_runner.rs
Philipp-M: We can still add that later, when it makes sense I guess. I do lean to also finding milliseconds useful though.

PoignardAzur: Part of the reason I'm moving this way, is that I plan to produce multiple logs at once, as outlined in #250. The console logs will be the least detailed ones; the logs written to a temp file will have the full detail level.

PR #255, opened 2024-05-04
Fix accessibility-related crash with Checkbox and Button

PoignardAzur:


Philipp-M (approved):
Not tested, but looks good

PR #256, opened 2024-05-04
Fix bugs in the accessibility tree created by Masonry

PoignardAzur:
Fix the layout values set in the accessibility tree.


Philipp-M (approved):
Haven't tested the accessibility part, but changes look good.

PoignardAzur:
@mwcampbell I think we should be good to merge?

PR #257, opened 2024-05-04
Create TodoMVC example

PoignardAzur:


crates/masonry/src/text2/edit.rs
```diff
@@ -101,38 +101,49 @@ impl<T: EditableText> TextEditor<T> {
...
-                                    selection.min() + c.len(),
-                                    // We have just added this character, so we are "affined" with it
-                                    Affinity::Downstream,
-                                ));
-                                Handled::Yes
-                            } else {
-                                debug_panic!("Got text input event whilst not focused");
-                                Handled::No
-                            }
+                            let selection = self.selection.selection.unwrap_or(Selection {
```
DJMcNab, line 130: Ok, so this unwrap is handling the case where the text gets reset by the driver?

I don't think the caret at index 0 is the right choice there. If the text is being changed, I think that's a sufficient event to drop the focus; yes that's disruptive, but logically the Textbox with a different content is a new text field, even if it holds the same location in the layout.

crates/masonry/src/text2/edit.rs
```diff
@@ -101,38 +101,49 @@ impl<T: EditableText> TextEditor<T> {
...
+                                anchor: 0,
+                                active: 0,
+                                active_affinity: Affinity::Downstream,
+                                h_pos: None,
+                            });
+                            // TODO: We know this is not the fullest model of copy-paste, and that we should work with the inner text
```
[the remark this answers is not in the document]
PoignardAzur: Fixed.

crates/masonry/src/text2/edit.rs
DJMcNab: Perhaps. It still a pattern which we should try and discourage if it all possible.

Can we force the caller of `set_text` to provide the new caret position?

Ok, I can imagine this being right for a textbox like at the top of a spreadsheet? I guess you could also be moving down a carousel? Although logically in that case, you're moving focus to the next item; it just happens to have moved to the same place as the previously focused item

PoignardAzur: You make a good point, but I don't really feel in the mindset to figure out the ideal DX for this. For now I think I'll just add a TODO comment.

crates/masonry/src/text2/edit.rs
```diff
@@ -101,38 +101,45 @@ impl<T: EditableText> TextEditor<T> {
...
+                            });
+                            let c = ' ';
+                            self.text_mut().edit(selection.range(), c);
+                            self.selection.selection = Some(Selection::caret(
+                                selection.min() + c.len_utf8(),
+                                // We have just added this character, so we are "affined" with it
+                                Affinity::Downstream,
+                            ));
+                            let contents = self.text().as_str().to_string();
+                            ctx.submit_action(Action::TextChanged(contents));
```
DJMcNab, line 118: I was trying to avoid passing the context in here (e.g. because its origin is wrong). Not sure what the right interface here is, though.

DJMcNab (commented):
Working this selection logic out properly in a massive pain. E.g. pressing `enter` should be able to reset the contents, but that kind of imperative logic doesn't really match the Xilem model

DJMcNab (approved):
Approving so we can hand this over

crates/xilem_masonry/src/view/textbox.rs
```diff
@@ -61,6 +85,10 @@ impl<State, Action> MasonryView<State, Action> for Textbox {
     ) -> crate::ChangeFlags {
         let mut changeflags = ChangeFlags::UNCHANGED;
 
+        if self.contents != element.text().as_str() {
```
DJMcNab, line 88: This is very unidiomatic to Xilem. We should have a comment here mentioing that, at least

crates/masonry/src/text2/edit.rs
PoignardAzur: I think this code will be due for a refactor after RustNL anyway (not to be dismissive of your work, I just think we haven't decided on a lot of semantics yet).

crates/xilem_masonry/src/view/textbox.rs
PoignardAzur: Fair, it's kind of a band-aid. But it's needed to avoid calling set_text on every new keypress.

PR #258, opened 2024-05-04
xilem_masonry: Fix build of textbox (wire up action_widget correctly)

Philipp-M:
Didn't seem to be wired up correctly, this should fix that


Philipp-M:
Whoops, on main there aren't events yet, (was on todomvc branch), hmm I guess I'll just cherry-pick it directly there... (Maybe we should split out the text-box changes into its own PR?)

PoignardAzur:
It's already fixed in #257. Thanks for the contribution =)

PR #259, opened 2024-05-04
Return the correct accessibility role in DynWidget

mwcampbell:
This fixes the problem where the container widget had the same role as its child, leading to, for example, nested buttons in the mason example.


DJMcNab (approved):
I've not been able to test this, but approving/merging based on the analysis in Zulip

Thanks!

PR #260, opened 2024-05-05
Cleanup the text widgets to not be generic

DJMcNab:
This is fallout from #241. As discussed on Zulip, using different underlying text types here is a very advanced feature, which will not be needed by any real consumers. It also creates some footguns

Probably conflicts with #257, so that should land first


PoignardAzur (approved):
LGTM!

waywardmonkeys (approved):

PR #261, opened 2024-05-05
Add VS Code marker comments to code

PoignardAzur:
[VS Code marker comments](https://code.visualstudio.com/docs/editor/custom-layout#_minimap-and-breadcrumbs) are a [new feature](https://code.visualstudio.com/updates/v1_88#_minimap-section-headers) that supposedly help with browsing the code with a minimap. I don't know if other editors have equivalents.

I haven't tested them before, and I don't know whether they actually improve DX, but they're pretty simple and unobstrusive, so I think it'd be worth adding them experimentally and seeing how it goes.

### DIFF
--- a/masonry/doc/ARCHITECTURE.md
+++ b/masonry/doc/ARCHITECTURE.md
@@ -166,3 +166,20 @@ The GlobalPassCtx types stores two timer handlers: **timers** and **mock_timer_q
 When a widget calls `request_timer` in a normal running app, a normal timer is requested from the platform. When a widget calls `request_timer` from a simulated app inside a TestHarness, mock_timer_queue is used instead.
 
 All this means you can have timer-based tests without *actually* having to sleep for the duration of the timer.
+
+
+## VS Code markers
+
+Masonry uses VS Code markers to help users browse code with the minimap:
+
+https://code.visualstudio.com/docs/getstarted/userinterface#_minimap
+
+These markers look like this:
+
+```rust
+// --- MARK: MARKER NAME ---
+```
+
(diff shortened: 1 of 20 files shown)


PoignardAzur:
This will probably wait until after RustNL.

PoignardAzur:
This feels ready to merge.

jaredoconnell (approved):

DJMcNab:
I would like to test this out before merging.

It seems to me that keeping these up-to-date will be quite a big cognitive burden, without obviously high value for vscode users and no value for developers who don't use vscode.

Are you familiar with the Outline tree/view?

PoignardAzur:
image

> It seems to me that keeping these up-to-date will be quite a big cognitive burden,

I don't think it's much more of a cognitive burden than any regular comments. We already had some comments marking "regions" in large files before, this is just a more advanced version of that.

> Are you familiar with the Outline tree/view?

I am. The main difference is that markers are more of a visual aid, whereas the outline view is more syntaxical.

Basically minimap markers is for when you're scrolling around and you've memorized "okay, the code I'm looking for is roughly after the rendering code and before the trait impls". Outline tree is for when you're memorized "okay, I'm looking for a function called X".

I know that I often have a more visual memory of how the code is laid out, so markers feel like they would be more helpful for me than a tree view.

> I would like to test this out before merging.

Given the low cost (no build-times or runtime cost, low visual noise), I'd like to merge first and try them for a few months. We can take them away, say, in September if we think they're not worth the effort.

To be transparent, I don't know if they're worth it (I don't really like how they're rendered), but the only way to know for sure is to try.

DJMcNab (approved):
I personally think it unlikely that I will maintain these, but I'm not going to block this experiment.

PR #262, opened 2024-05-05
Remove ChangeFlags from xilem_masonry

PoignardAzur:


DJMcNab (approved):
I still feel like this *should* be useful, albeit only for debugging.

But that small advantage is not worth the huge amount of detritus it brings with it

Philipp-M (commented):
Thought about starting a zulip thread about this already a few days ago...

Still not sure, whether we want `ChangeFlags` or not. I think either has its advantages.

`ChangeFlags` being "recursive", has "more control" directly over the child views.

On the other hand the `ViewCx` is used already to track state for widgets (and I think in the future it will be used for rather more than less "view state".

So I'm slightly in favor of removing the `ChangeFlags` and just tracking the whole "meta" state of views in the `ViewCx` directly... I hope all that stuff in `ViewCx` doesn't get too messy then though...

PoignardAzur:
> I still feel like this should be useful, albeit only for debugging.

I still left a print line for debugging.

> On the other hand the ViewCx is used already to track state for widgets (and I think in the future it will be used for rather more than less "view state".

Note that a major reason this is removed is that Widgets currently have track their own state. Initially ChangleFlags kept track of stuff like "needing layout", "needing repaint" that Widgets now handle directly.

Philipp-M:
I'm thinking more about view specific Changeflags (I honestly have no concrete use-case in my head though TBH, but there may be something that only concerns the reactive layer in the future...).

I agree, that all that stuff directly concerning widgets, should probably be tracked in masonry.

PR #263, opened 2024-05-05
Unbreak text rendering

DJMcNab:
Previously, we were rendering the first line `n` times, the second line `n-1` times, etc.

(Because the copy was in the loop)

See [#xilem > ✔ Text rendering when using nested Scenes](https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/.E2.9C.94.20Text.20rendering.20when.20using.20nested.20Scenes)


Philipp-M (approved):
Looks good :)

PR #264, opened 2024-05-05
xilem_masonry: Fix Flex splice (increment ix by 2 because of spacer widget)

Philipp-M:
While adding filters to the TodoMVC example, I produced a crash. Which seems to originate because there was a spacer added, which also increments the underlying element child count.
So in case we want to have that spacer, we need to increment by 2 instead...

(Since `mutate` and `delete` skip spacers, because `child_mut` returns none, and the index is only incremented for concrete children it's not necessary to change the index incrementation code there)

### DIFF
--- a/crates/xilem_masonry/src/view/flex.rs
+++ b/crates/xilem_masonry/src/view/flex.rs
@@ -100,7 +100,7 @@ impl ElementSplice for FlexSplice<'_> {
     fn push(&mut self, element: WidgetPod<Box<dyn masonry::Widget>>) {
         self.element.insert_child_pod(self.ix, element);
         self.element.insert_default_spacer(self.ix);
-        self.ix += 1;
+        self.ix += 2;
     }
 
     fn mutate(&mut self) -> WidgetMut<Box<dyn Widget>> {
@@ -142,6 +142,6 @@ impl ElementSplice for FlexSplice<'_> {
     }
 
     fn len(&self) -> usize {
-        self.ix
+        self.ix / 2
     }
 }


crates/xilem_masonry/src/view/flex.rs
```diff
@@ -142,6 +142,6 @@ impl ElementSplice for FlexSplice<'_> {
     }
 
     fn len(&self) -> usize {
-        self.ix
+        self.ix / 2
```
PoignardAzur, line 145: Oooooh, you had to change the `len()` method as well?

I tried to do `self.ix += 2` at first, and I got crashes. I naively assumed that `self.ix += 1` would still work with spacers. Shame on me for not testing it worked ^^

PoignardAzur (approved):
LGTM.

crates/xilem_masonry/src/view/flex.rs
Philipp-M: I don't think `len` used currently (IIRC it was previously only really used by the `List` ViewSequence, it's not yet clear whether it's really needed), but I think it's probably more correct that way, because the spacers are "invisible" views?

Weird that you got crashes with `self.ix += 2`. Do you know how you've got those crashes?

DJMcNab:
Just gut-checking a policy thing here - I think I would have expected @Philipp-M to activate the merge here, as the PR author.

raphlinus:
Generally yes but I think we can speed things along a bit right now.

PoignardAzur:
Right, I sped this one along because we agreed that bug blocked other work, but maybe I overreached a bit.

PR #265, opened 2024-05-05
xilem_masonry: Add filter ("All", "Active", "Completed") to todomvc example

Philipp-M:
This adds the filter-buttons in the footer of TodoMVC via checkboxes (as we currently don't have better widgets for that...), in the future this should be updated with proper radio buttons.

### DIFF
--- a/xilem/examples/to_do_mvc.rs
+++ b/xilem/examples/to_do_mvc.rs
@@ -61,21 +61,6 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<TaskList> + use<> {
         }),
     ));
 
-    let filter_view = |label, filter| {
-        // TODO: replace with combo-buttons
-        checkbox(
-            label,
-            task_list.filter == filter,
-            move |state: &mut TaskList, _| state.filter = filter,
-        )
-    };
-
-    let footer_view = flex_row((
-        filter_view("All", Filter::All),
-        filter_view("Active", Filter::Active),
-        filter_view("Completed", Filter::Completed),
-    ));
-
     let tasks = task_list
         .tasks
         .iter()
@@ -101,7 +86,24 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<TaskList> + use<> {
         })
         .collect::<Vec<_>>();
 
-    flex_col((first_line, tasks, footer_view)).padding(50.)
+    let filter_tasks = |label, filter| {
+        // TODO: replace with combo-buttons
+        checkbox(
+            label,
+            task_list.filter == filter,
+            move |state: &mut TaskList, _| state.filter = filter,
+        )
+    };
+    let has_tasks = !task_list.tasks.is_empty();
+    let footer = has_tasks.then(|| {
+        flex_row((
+            filter_tasks("All", Filter::All),
+            filter_tasks("Active", Filter::Active),
+            filter_tasks("Completed", Filter::Completed),
+        ))
+    });
+
+    flex_col((first_line, tasks, footer)).padding(50.0)
 }
 
 fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
--- a/xilem/examples/to_do_mvc.rs
+++ b/xilem/examples/to_do_mvc.rs
@@ -16,8 +16,16 @@ struct Task {
     done: bool,
 }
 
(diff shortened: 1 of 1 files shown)


PoignardAzur:
Nicely done, but as you point out, this PR makes these other bugs very visible. I'd rather not merge it until they are fixed, since this is our showcase. ^^

Philipp-M:
Sure, I also think this should only be merged, when it doesn't produce obvious bugs at this point :). But I think it's probably good for debugging these bugs.

Philipp-M:
(Now this weird issue can be debugged, since #264 is merged)

bram209:
I ran in (I think) the same issue that you are describing, put some reproduction steps here: https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/TodoMVC.20example/near/437314215

Philipp-M:
Since #278 is merged, I've tested this again, it works now as expected

jaredoconnell:
I tested this. I think this filter should be placed before the list:
image
This makes it so that it doesn't move around when you're using it.

I do find that it's in need of polish. You're using multi-select checkboxes as radio buttons. If we're going with multi-select, you technically don't want an `All` option, and instead check both options.
I don't know what level of polish we're going for this week.

xStrom:
> I think this filter should be placed before the list:
> This makes it so that it doesn't move around when you're using it.

For what it's worth, it also moves around in the examples on [todomvc.com](https://todomvc.com).

Philipp-M:
Yes checkboxes are not ideal indeed, these should be exchanged with the right widgets, when they're implemented.

I'm not sure how much we're striving towards the original TodoMVC.

PoignardAzur:
I've been looking through our PR backlog, and I'd be interested in merging this, if everything works correctly after a rebase.

I'm fine having checkboxes masquerade as combo buttons for now, as long as we track it in an issue. We can add a big ugly "TODO - replace with combo buttons" comment to the example.

Philipp-M:
I've rebased it (+ added comment), and well it revealed yet another new bug^^ (this example seems to be a quite good test as well)

To reproduce:
Press "Active" and then 
"Buy milk"

DJMcNab:
Hmm, so the behaviour we see is that the checkbox "thinks" it is clicked, but without sending an action.

That's very strange

Philipp-M:
I rebased this again, to check whether that issue resolved itself - not the case.
I've found another issue (that is actually *the* issue for the other one as well):
It's possible to toggle the checkboxes, whereas I'd expect, that these should behave more like radio-buttons, since they should be driven by xilem.

So the issue is related to two-way dataflow and non-keyed widgets/viewsequences. I.e. masonry modifies the checkbox value while it shouldn't do it (at least in this case).
This opens the broader question about two-way/unidirectional dataflow/state again.

I think this is an example, that two-way dataflow can be a pitfall for declarative frameworks on top of masonry. I think it would be great to find a general clean solution for this issue (something like opt-in control of the state in masonry, as in this case).

I've pushed a ["fix"/workaround](https://github.com/linebender/xilem/commit/927dbb79e90bea124895a54e1bf71be399597431), which should make this work as expected.

PoignardAzur:
This seems like a very "brute force" solution to the two-way dataflow problem, buuuuut nobody has a better idea, so we may as well ship it.

Philipp-M:
I've removed the workaround in masonry, as this is now more properly handled in masonry since #1366 

I think this is now not surfacing any further bugs, and should be mergeable.

xilem/examples/to_do_mvc.rs
```diff
@@ -45,39 +53,62 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<TaskList> + use<> {
...
+                let delete_button = button("Delete", move |data: &mut TaskList| {
+                    data.tasks.remove(i);
+                });
+                Some(flex_row((checkbox, delete_button)))
+            }
         })
         .collect::<Vec<_>>();
 
-    flex_col((first_line, tasks)).padding(50.)
+    flex_col((first_line, tasks, footer_view)).padding(50.)
```
DJMcNab: In "upstream" todomvc, the "All", "Active", "Completed" filters seem to be hidden if there are no tasks. I'm not sure if we should emulate that

DJMcNab (approved):
Thanks! This works well for me

PR #270, opened 2024-05-05
Tweak code in post_event_processing

PoignardAzur:
Inline `wants_animation_frame`.
Request new frame for both visual and accessibility updates.

### DIFF
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -548,8 +548,9 @@ impl RenderRoot {
                 .push_back(RenderRootSignal::RequestAnimFrame);
         }
 
-        // If we either need to be repainted of need to update accessibility info,
-        // we request a redraw.
+        // We request a redraw if either the render tree or the accessibility
+        // tree needs to be rebuilt. Usually both happen at the same time.
+        // A redraw will trigger a rebuild of the accessibility tree.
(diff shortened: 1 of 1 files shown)


crates/masonry/src/render_root.rs
```diff
@@ -532,14 +532,15 @@ impl RenderRoot {
...
-        // If we need a new paint pass, make sure winit knows it.
-        if self.wants_animation_frame() {
+        if self.root.state().request_anim {
             self.state
                 .signal_queue
                 .push_back(RenderRootSignal::RequestAnimFrame);
         }
 
-        if self.root.state().needs_paint {
+        // If we either need to be repainted of need to update accessibility info,
```
waywardmonkeys, line 541: ... or need ...

waywardmonkeys (changes requested):
Typo in comment and it really needs a commit comment (especially since it contains an actual functionality change, not just changing the animation frame handling.

crates/masonry/src/render_root.rs
DJMcNab: (I'm not sure that this comment actually contains any information either)

DJMcNab (commented):
Agree with the review from @waywardmonkeys

DJMcNab (approved):
Thanks

PR #271, opened 2024-05-05
Convert Masonry `README.md` to the Linebender standard.

xStrom:
This is a basic skeleton based on the latest Linebender standards.

The installation instructions were stale, based on `druid-shell`. The example code and roadmap were also stale, not reflective of the recent barrage of changes. Thus I removed these stale sections.

I removed `CONTRIBUTING.md` because the new readme has a section about that as standard, and a more thorough `CONTRIBUTING.md` will eventually live in the repo root.

I moved `ARCHITECTURE.md` and `ROADMAP.md` to a new `doc` sub-directory, as is standard for Linebender. These are actually also stale, but won't be immortalized by publishing so changing/deleting them isn't as urgent.


waywardmonkeys (approved):

crates/masonry/README.md
```diff
@@ -1,157 +1,41 @@
...
-    AppLauncher::with_window(main_window)
-        .with_delegate(Delegate {
-            next_task: String::new(),
-        })
-        .log_to_console()
-        .launch()
-        .expect("Failed to launch application");
-}
-```
+Masonry gives you a platform to create windows (using [winit] as a backend) each with a tree of widgets.
```
DJMcNab, line 17: Unless I'm mistaken, I think we might only support one window at the moment

But we are definitely planning to support multi-window, it's just a matter of wiring it up

crates/masonry/README.md
```diff
@@ -1,157 +1,41 @@
...
+It also gives you tools to inspect that widget tree at runtime, write unit tests on it, and generally have an easier time debugging and maintaining your app.
```
DJMcNab, line 19: Do we also want to mention a similar thing of alpha status to Vello (maybe not as a GitHub alert, though)

Stuff like:
- text input being somewhat janky
- snapshot testing not being consistent cross platform

DJMcNab (approved):
I wonder if `/doc/masonry/..` is the right place for these?

Doesn't need to happen imminently. If they're out of date, I suppose hidden in a folder is probably for the best

(GitHub mobile won't let me add this as a comment to the renaming)

Approving for expedience, although I think Olivier might have opinions about this

crates/masonry/README.md
xStrom: Added a note about it being alpha.

crates/masonry/README.md
xStrom: Yeah, might be a new limitation. I haven't investigated.

Changed it to singular for now.

PR #272, opened 2024-05-05
Sync Masonry's `.gitignore` with the repo's `.gitignore`.

xStrom:
The general editor/OS ignore filters are useful across the repo.

Fixees one of the Masonry specific snapshot filters, which was broken due to a copy-paste accident.


waywardmonkeys (changes requested):
I'm mostly against the things moving into the root `.gitignore`.

Some of them should be global ignores for the people who need them, some of them are really old and not useful in the wild (like `.AppleDouble` and `.LSOverride`).

DJMcNab (commented):
Agreed, we should just use the usual standard, i.e.
```rust
/target
.DS_Store
```

xStrom:
Alright, I removed them.

waywardmonkeys (approved):

PR #273, opened 2024-05-05
Implement basic IME interaction, backspace, and delete.

xorgy:


xorgy:
I understand that `Movement` is preferred for some of the cursor movement things, but I was mainly focused on getting something working. An understander of `Movement` can surely complete the selection and cursor movement commands.

crates/masonry/src/text2/edit.rs
```diff
@@ -97,19 +97,55 @@ impl<T: EditableText> TextEditor<T> {
...
+                        }
+                        Key::Named(NamedKey::Delete) => {
+                            if let Some(selection) = self.with.selection {
+                                if !selection.is_caret() {
+                                    self.text_mut().edit(selection.range(), "");
+                                    self.with.selection = Some(Selection::caret(
+                                        selection.min(),
+                                        Affinity::Downstream,
+                                    ));
+                                } else if let Some(o) =
```
DJMcNab, line 125: I don't think `o` is a good name for this variable; we're not rationing characters

crates/masonry/src/text2/edit.rs
```diff
@@ -97,19 +97,55 @@ impl<T: EditableText> TextEditor<T> {
...
                         Key::Named(NamedKey::Backspace) => {
-                            eprintln!("Got backspace, not yet handled");
-                            Handled::No
+                            if let Some(selection) = self.with.selection {
+                                if !selection.is_caret() {
+                                    self.text_mut().edit(selection.range(), "");
+                                    self.with.selection =
+                                        Some(Selection::caret(selection.min(), Affinity::Upstream));
+                                } else if let Some(o) =
+                                    self.text().prev_grapheme_offset(selection.active)
```
DJMcNab, line 106: According to https://github.com/xi-editor/xi-editor/issues/655, deleting entire grapheme clusters is "too aggressive".

We need to at least document why that is no longer true

crates/masonry/src/text2/edit.rs
```diff
@@ -147,17 +183,115 @@ impl<T: EditableText> TextEditor<T> {
                             Handled::No
                         }
                     }
+                } else if mods.control_key() || mods.super_key()
```
DJMcNab, line 186: On macOS, I believe that this case specifically actually uses "Option"

Iced calls this [`is_jump_modifier_pressed`](https://github.com/iced-rs/iced/blob/master/widget/src/text_input.rs#L1287)

Of course, we should just use the events from the platform for macOS, but you know. winit

DJMcNab (commented):
As I mentioned, this IME support doesn't work for me, because we haven't called [`set_ime_allowed`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_ime_allowed)

We should also set at least [cursor area](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_ime_cursor_area) as well

crates/masonry/src/text2/edit.rs
```diff
@@ -97,19 +97,55 @@ impl<T: EditableText> TextEditor<T> {
...
+                                        Some(Selection::caret(selection.min(), Affinity::Upstream));
+                                } else if let Some(o) =
+                                    self.text().prev_grapheme_offset(selection.active)
+                                {
+                                    self.text_mut().edit(o..selection.active, "");
+                                    self.with.selection =
+                                        Some(Selection::caret(o, selection.active_affinity));
+                                }
+                                Handled::Yes
+                            } else {
```
raphlinus, line 113: Something Olivier and I noticed while experimenting: this should submit a `TextChanged` action as does insertion.

raphlinus:
The backspace and delete functionality is working, but I'm not getting any IME behavior on mac, not even the ability to insert characters through the alt key.

crates/masonry/src/event_loop_runner.rs
```diff
@@ -49,6 +49,7 @@ pub fn run(
     let event_loop_proxy = event_loop.create_proxy();
     let adapter = Adapter::with_event_loop_proxy(&window, event_loop_proxy);
     window.set_visible(visible);
+    window.set_ime_allowed(true);
```
raphlinus, line 52: Add a TODO to enable this on a more fine-grained basis - primarily when focusing on a text edit widget.

crates/masonry/src/text2/edit.rs
```diff
@@ -52,27 +52,31 @@ impl EditableText for String {
 
 /// A region of text which can support editing operations
 pub struct TextEditor<T: EditableText> {
-    selection: TextWithSelection<T>,
+    with: TextWithSelection<T>,
```
raphlinus, line 55: Daniel suggests "inner" as a better name for this, and I agree. I realize it's tricky to come up with a perfect name for this.

crates/masonry/src/text2/selection.rs
```diff
@@ -121,8 +121,39 @@ impl<T: Selectable> TextWithSelection<T> {
         match event {
             TextEvent::KeyboardKey(key, mods) if key.state.is_pressed() => {
                 match key.key_without_modifiers() {
-                    winit::keyboard::Key::Named(NamedKey::ArrowLeft) if mods.shift_key() => {
-                        // TODO: Expand selection
+                    winit::keyboard::Key::Named(NamedKey::ArrowLeft) => {
+                        if mods.shift_key() {
+                        } else {
```
raphlinus, line 126: It seems like this case is still a TODO.

raphlinus (commented):
I'd like to see this go in soon. I've added a couple comments, but will let Daniel do the approval after his comments are addressed.

crates/masonry/src/text2/edit.rs
xorgy: Some day... but for today it's TODO

crates/masonry/src/text2/edit.rs
```diff
@@ -52,27 +52,31 @@ impl EditableText for String {
...
         Self {
-            selection: TextWithSelection::new(text, text_size),
-            _preedit_range: None,
+            inner: TextWithSelection::new(text, text_size),
+            preedit_range: None,
         }
     }
 
+    pub fn reset_preedit(&mut self) {
+        self.preedit_range = None;
```
DJMcNab, line 70: I think this should just be part of `set_text`; we might need to "redefine" it directly on the `TextEditor`

DJMcNab (approved):
I can still see a few issues in the logic, but overall looks like it mostly works

And this is much better than not having it

crates/masonry/src/text2/edit.rs
```diff
@@ -97,19 +101,61 @@ impl<T: EditableText> TextEditor<T> {
...
+                                        selection.min(),
+                                        Affinity::Downstream,
+                                    ));
+                                } else if let Some(offset) =
+                                    self.text().next_grapheme_offset(selection.active)
+                                {
+                                    self.text_mut().edit(selection.min()..offset, "");
+                                    self.inner.selection = Some(Selection::caret(
+                                        selection.min(),
+                                        selection.active_affinity,
```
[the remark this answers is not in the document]
xorgy: We will be using Affinity in my next PR  > :+ )

crates/masonry/src/text2/selection.rs
```diff
@@ -121,8 +120,39 @@ impl<T: Selectable> TextWithSelection<T> {
...
+                            let t = self.text();
+                            if let Some(selection) = self.selection {
+                                if mods.control_key() {
+                                    let offset = t.prev_word_offset(selection.active).unwrap_or(0);
+                                    self.selection =
+                                        Some(Selection::caret(offset, Affinity::Downstream));
+                                } else if let Some(o) = t.prev_grapheme_offset(selection.active) {
```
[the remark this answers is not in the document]
xorgy: Might as well, though in practice it is always capable of being 0 because the start of text is a straight grapheme break.

crates/masonry/src/text2/edit.rs
xorgy: Maybe, but I don't see how to do that cleanly right now... possibly out of intellectual laziness. :+ )

PR #275, opened 2024-05-06
Fix the handling of textbox width

DJMcNab:
Previously, filling up the line would overflow the right border of the box, because the `max_advance` was too big.

This improves the margin hack, and also fixes the resultant issue.

### DIFF
--- a/crates/masonry/src/widget/textbox.rs
+++ b/crates/masonry/src/widget/textbox.rs
@@ -23,6 +23,10 @@ use crate::{
 use super::{LineBreaking, WidgetMut, WidgetRef};
 
 const TEXTBOX_PADDING: f64 = 3.0;
+/// HACK: A "margin" which is placed around the outside of all textboxes, ensuring that
+/// they do not fill the entire width of the window
+///
+/// In theory, this should be proper margin/padding the parent widget.
 const TEXTBOX_MARGIN: f64 = 8.0;
 
 /// The textbox widget is a widget which shows text which can be edited by the user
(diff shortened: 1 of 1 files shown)


crates/masonry/src/widget/textbox.rs
```diff
@@ -22,7 +22,12 @@ use crate::{
 
 use super::{LineBreaking, WidgetMut, WidgetRef};
 
-const TEXTBOX_PADDING: f64 = 4.0;
+const TEXTBOX_PADDING: f64 = 3.0;
+/// HACK: A "margin" which is placed around the outside of all textboxes, ensuring that
+/// they do not fill the entire width of the window
+///
+/// In theory, this should be proper margin/padding the parent widget.
```
waywardmonkeys, line 29: Is a word missing before "the parent" ?

crates/masonry/src/widget/textbox.rs
```diff
@@ -249,8 +254,7 @@ impl Widget for Textbox {
         let max_advance = if self.line_break_mode != LineBreaking::WordWrap {
             None
         } else if bc.max().width.is_finite() {
-            // TODO: Does Prose have different needs here?
-            Some(bc.max().width as f32 - 2. * TEXTBOX_PADDING as f32)
+            Some(bc.max().width as f32 - 2. * TEXTBOX_PADDING as f32 - 2. * TEXTBOX_MARGIN as f32)
```
waywardmonkeys, line 257: Why not just do the math at 64 bit and then one cast to `f32`?

`Some((bc.max().width - 2. * TEXTBOX_PADDING - 2. * TEXTBOX_MARGIN) as f32)`

jaredoconnell (approved):
This works as intended for me.

PR #276, opened 2024-05-06
Implement underlining and strikethrough for Masonry

DJMcNab:
This will be needed to follow on from #273 (showing the pre-edit region visually)

underline on Lorem Ipsum.

and

stikethrough on Lorem Ipsum

cc @dfrg for awareness, I'm not certain on the maths (but that doesn't need to block merging)

### DIFF
--- a/crates/masonry/src/text_helpers.rs
+++ b/crates/masonry/src/text_helpers.rs
@@ -148,6 +148,39 @@ pub fn render_text(
                     &line,
                 )
             }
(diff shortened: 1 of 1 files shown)


dfrg (approved):
I haven’t tested this but the math looks correct. Thanks!

DJMcNab:
Fixed image of the underlining:
some lorem ipsum text with an underline.

PR #278, opened 2024-05-06
masonry: Require `ParentWindowOrigin` lifecycle when origin of child has changed

Philipp-M:
Fixes the issue described in #265.

I *think* this is the proper fix, it seems to be similar as in previous xilem (where I think that pass is called `ViewContextChanged`).


DJMcNab (approved):
Approving based on Zulip discussion.
Unblocks more `to_do_mvc` features

PR #283, opened 2024-05-07
Update dependencies.

xStrom:
These are the results of running

```sh
cargo upgrade --ignore-rust-version
cargo update
```

in preparation for the upcoming release.

These are semver compatible updates.


raphlinus (approved):
Thanks!

PR #284, opened 2024-05-07
Implement underlining for IME areas

DJMcNab:
image


xStrom (approved):
Needs more research on Windows behavior, but approving for other platform sakes.

PR #285, opened 2024-05-08
Additional flex view options

keeslinp:
Wired up some more existing masonry flex properties so they can be set on the flex view.

I was messing around with a dummy example to try out the v0.1 and noticed I couldn't easily center something because flex was only exposing direction not alignment/cross-alignment. It looked like all the logic already exists in masonry so it was just a matter of wiring it up. This is my first contribution so sorry if I missed some steps.

Sidenote: I didn't see any examples of testing that a view updates the underlying element so I am guessing that is still tbd. If someone has a suggestion for a good way to do that I am happy to try it out.

I'm also not sure what the etiquette is on adding/editing examples so I just included the example I was using to verify it worked but no stress if you want to remove it.

### DIFF
--- a/crates/xilem_masonry/examples/flex.rs
+++ b/crates/xilem_masonry/examples/flex.rs
@@ -8,26 +8,15 @@ use xilem::{
     MasonryView, Xilem,
 };
 
-struct AppState {
-    count: u32,
-}
-
-impl AppState {
-    fn decrement(&mut self) {
-        if self.count > 0 {
-            self.count -= 1;
-        };
-    }
-    fn increment(&mut self) {
-        self.count += 1;
-    }
-}
-
-fn app_logic(data: &mut AppState) -> impl MasonryView<AppState> {
+fn app_logic(data: &mut i32) -> impl MasonryView<i32> {
     flex((
-        button("-", AppState::decrement),
-        label(format!("count: {}", data.count)),
-        button("+", AppState::increment),
+        button("-", |data| {
+            *data -= 1;
+        }),
+        label(format!("count: {}", data)),
+        button("+", |data| {
+            *data += 1;
+        }),
     ))
     .direction(xilem::Axis::Horizontal)
     .cross_axis_alignment(CrossAxisAlignment::Center)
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Thanks for the PR! This code all looks perfectly idiomatic.

I have not yet had a chance to test out the new example - I think adding a new example is very reasonable, whilst we're still in the early stages. I'll merge once I've run it, (or someone else can do the same)

crates/xilem_masonry/examples/flex.rs
```diff
@@ -0,0 +1,42 @@
...
+    view::{button, flex, label},
+    MasonryView, Xilem,
+};
+
+struct AppState {
+    count: u32,
+}
+
+impl AppState {
+    fn decrement(&mut self) {
```
DJMcNab, line 16: Looking at this code again - I'm not sure I like these two methods; I think these methods would probably just be closures inline in the `app_logic`.

It's nice to keep the examples light on ceremony

crates/xilem_masonry/examples/flex.rs
```diff
@@ -0,0 +1,42 @@
...
+    ))
+    .direction(xilem::Axis::Horizontal)
+    .cross_axis_alignment(CrossAxisAlignment::Center)
+    .main_axis_alignment(MainAxisAlignment::Center)
+}
+
+fn main() -> Result<(), EventLoopError> {
+    let data = AppState { count: 0 };
+    let app = Xilem::new(data, app_logic);
+    app.run_windowed("Demo".into())?;
```
DJMcNab, line 40: The window title here should probably be "centered flex"?

DJMcNab (commented):
Tried the example and it looks good. Just one small comment about code style for examples. For bigger examples we probably want to use this form.

screenshot of the demo.

crates/xilem_masonry/examples/flex.rs
keeslinp: Good call, this started out as my own demo that I copied in. Should I just nuke the counter logic altogether since this is just showing the layout?

keeslinp: This has been slimmed down considerably now

keeslinp:
> Tried the example and it looks good. Just some small comment about code style for examples. (For bigger examples we possibly do want to use this form of data with methods, but this is a very small example)
> 
> screenshot of the demo.


For future reference, should I include a screenshot in the initial PR description?

Philipp-M (approved):
> For future reference, should I include a screenshot in the initial PR description?

Would be great, but since that needs to be tested anyway, not necessary (to me at least).

Otherwise looks good.

(Should we just merge, or give @keeslinp Contributor rights to follow our policy?)

DJMcNab:
I'm going to just merge at this moment

PR #286, opened 2024-05-08
Add crates.io and docs.rs links to README.md

matthunz:
Congrats everyone on v0.1!

Not sure if this is the best styling to go for, but I think it'd be great to link to the new releases

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 **An experimental Rust architecture for reactive UI**
 
+[![Crate](https://img.shields.io/crates/v/xilem?style=flat-square)](https://crates.io/crates/xilem)
+[![Documentation](https://img.shields.io/badge/docs-latest-blue.svg?style=flat-square)](https://docs.rs/xilem)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,8 @@
 
 **An experimental Rust architecture for reactive UI**
 
+[![Crate](https://img.shields.io/crates/v/xilem?style=flat-square)](https://crates.io/crates/xilem)
+[![Documentation](https://img.shields.io/badge/docs-latest-blue.svg?style=flat-square)](https://docs.rs/xilem)
```
DJMcNab, line 8: There are existing badges for these at the bottom of the list of badges, which are commented out.

Could you please change this PR to uncomment those?

DJMcNab (commented):
Hi! Thanks for your interest. I agree that this is really important to have!

We had already done some preparation for this change, but hadn't gotten around to uncommenting it. Can you please change this PR to instead use that?

README.md
matthunz: Oh wow I didn't even notice😅 sounds good I just switched over

DJMcNab (approved):
Thanks!

PR #287, opened 2024-05-08
masonry: docs: Don't `deny(rustdoc::broken_intra_doc_links)`.

waywardmonkeys:
Having this denied means we can't generate docs locally. We can re-enable this once the issues are fixed.


DJMcNab (approved):
We really need to update CI at some point :)

Thanks!

PR #289, opened 2024-05-08
xilem_classic: deps: Require current `futures-task`

waywardmonkeys:
Previously, this depended only on version `0.3`, which had a advisory issued for it.


waywardmonkeys:
This will clear the "maybe insecure" status that deps.rs reports.

DJMcNab (approved):

PR #290, opened 2024-05-08
Remove `xilem_classic`

waywardmonkeys:
This can be accessed in the git history and has been replaced by the version using `masonry`.


waywardmonkeys:
This is exploratory and I'm happy to iterate on it with whoever to get it to where we can land this.

waywardmonkeys:
A logical next step from here would be to do the next step in the multi-crate repository setup:

* `git mv crates/* .`
* `git m xilem_masonry xilem`
*  Edit the `Cargo.toml` accordingly

That could be done as a separate PR though or as part of this one and smashed into one giant commit.

My thought in favor of doing this step first was that it would eliminate some of the confusion that exists now and let people figure out what to do with their PRs that are outstanding before the big re-org of things.

DJMcNab:
The position I'm seeing at RustNL is that we should remove the `xilem_classic` code entirely.

waywardmonkeys:
That could work too and I could modify this PR to just remove it instead of move it.

waywardmonkeys:
@DJMcNab Updated to remove instead. This also removes some entries from `Cargo.lock` which always brings a little joy to my heart.

DJMcNab (commented):
Looks good, although I agree it's worth getting Raph's rubber stamp

raphlinus (approved):
Thanks for the cleanup!

waywardmonkeys:
MERGE AHOY.

PR #292, opened 2024-05-08
masonry: Disable render tests by default

waywardmonkeys:
Set both `ENABLE_RENDER_TESTS=1` and `ENABLE_RENDER_SNAPSHOTS=1` in your environment to run the tests.

### DIFF
--- a/crates/masonry/src/testing/harness.rs
+++ b/crates/masonry/src/testing/harness.rs
@@ -33,6 +33,11 @@ pub const HARNESS_DEFAULT_SIZE: Size = Size::new(400., 400.);
 /// Default background color for tests.
 pub const HARNESS_DEFAULT_BACKGROUND_COLOR: Color = Color::rgb8(0x29, 0x29, 0x29);
 
+fn check_env_enabled(key: &'static str) -> bool {
+    let val = std::env::var_os(key);
+    val.is_some_and(|it| !it.is_empty())
+}
+
 /// A safe headless environment to test widgets in.
 ///
 /// `TestHarness` is a type that simulates an [`AppRoot`](crate::AppRoot)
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
My concerns with this PR:

1) There's no documentation for how to run the tests yet, so nowhere to mention how to re-enable them.
2) I'm not sure we need 2 separate environment variables for now.

crates/masonry/src/testing/harness.rs
```diff
@@ -238,7 +243,7 @@ impl TestHarness {
     /// Create a bitmap (an array of pixels), paint the window and return the bitmap as an 8-bits-per-channel RGB image.
     pub fn render(&mut self) -> RgbaImage {
         let (scene, _tree_update) = self.render_root.redraw();
-        if std::env::var("SKIP_RENDER_TESTS").is_ok_and(|it| !it.is_empty()) {
+        if !check_env_enabled("ENABLE_RENDER_TESTS") {
```
DJMcNab, line 246: So as it happens, the render tests mostly work fine

So I'd probably revert this half of the change

waywardmonkeys:
One other difference is that one part of this was reading the env variable at runtime via `std::env::var` while the other part was reading it at compile time via `option_env!`.

DJMcNab:
So, my expected updates for this PR would be.
1) Keep the change to read at runtime for both
2) Re-invert the `SKIP_RENDER_TESTS` flag, as that is *only* useful for Windows CI

Alternatively, we could close this pending the fixed snapshot tests (i.e. #233), but that is at least another couple of weeks away.
I can make these changes if needed.

PoignardAzur:
I think we can close this now that render tests work?

PR #293, opened 2024-05-08
xilem_web: Remove unused direct dep on `kurbo`.

waywardmonkeys:
This is always used via `peniko`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3859,7 +3859,6 @@ version = "0.1.0"
 dependencies = [
  "bitflags 2.5.0",
  "gloo",
- "kurbo",
  "log",
  "paste",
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This is an offshoot from #223.

waywardmonkeys:
All uses of `kurbo` within `xilem_web`, including the `svgtoy` sample are via `peniko`.

`xilem_web::svg` exports both `peniko` and `peniko::kurbo`, so the `svgtoy` sample is a bit textually unclear about the provenance, but it is via `peniko`.

DJMcNab (approved):
Sure

I now think we probably should do this all up the stack, although I still find the tradeoffs annoying

PR #298, opened 2024-05-09
Handle IME in a way which winit expects

DJMcNab:
Doing some investigation, the issues I'm seeing is that ibus is a bit/a lot broken around compose keys

But this has correct behaviour for me. This matches the documentation of the `winit` [`Ime`](https://docs.rs/winit/latest/winit/event/enum.Ime.html) event.

### DIFF
--- a/crates/masonry/src/text2/edit.rs
+++ b/crates/masonry/src/text2/edit.rs
@@ -271,12 +271,11 @@ impl<T: EditableText> TextEditor<T> {
                     if let Some(preedit) = self.preedit_range.clone() {
                         self.text_mut().edit(preedit.clone(), preedit_string);
                         let np = preedit.start..(preedit.start + preedit_string.len());
-                        if preedit_string.is_empty() {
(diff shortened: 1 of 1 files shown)


crates/masonry/src/text2/edit.rs
```diff
@@ -290,7 +290,11 @@ impl<T: EditableText> TextEditor<T> {
                         let sr = self.selection.map(|x| x.range()).unwrap_or(0..0);
                         self.text_mut().edit(sr.clone(), preedit_string);
                         let np = sr.start..(sr.start + preedit_string.len());
-                        self.preedit_range = Some(np.clone());
+                        if preedit_string.is_empty() {
```
xorgy, line 293: Maybe hoist the assignment out of the if statement.

xorgy (approved):

xorgy (changes requested):
It's broken on X11 for me.

PR #299, opened 2024-05-09
docs: fix deps discrepency between Fedora/Ubuntu

simbleau:
Slight discrepancy between the instructions for Fedora and Ubuntu.

In the Fedora instructions, it doesn't include `pkg-config`, in Ubuntu, it does.


xStrom:
Are any of these still needed? These instructions were for Glazier.

simbleau:
> Are any of these still needed? These instructions were for Glazier.

Not sure, I just figured.

DJMcNab (approved):
Thanks!

PR #300, opened 2024-05-11
`masonry`: Use `cursor_icon` crate directly.

waywardmonkeys:
`cursor_icon` is a crate that is used by `winit` to provide its `CursorIcon` type separately from the `winit` crate.

This re-exports the `CursorIcon` from `masonry` and then uses that rather than via `winit::window::CursorIcon`, allowing the use of the `CursorIcon` without having to independently discover its provenance.

This is also one step towards not requiring `winit` within the `masonry` internals apart from the actual window / event loop management.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1722,6 +1722,7 @@ dependencies = [
  "accesskit",
  "accesskit_winit",
  "assert_matches",
+ "cursor-icon",
  "float-cmp",
  "fnv",
  "futures-intrusive",
--- a/masonry/Cargo.toml
+++ b/masonry/Cargo.toml
@@ -41,6 +41,7 @@ tracing-subscriber = { version = "0.3.18", features = ["env-filter", "time"] }
 accesskit.workspace = true
 accesskit_winit.workspace = true
 time = { version = "0.3.36", features = ["macros", "formatting"] }
+cursor-icon = "1.1.0"
 
 [dev-dependencies]
 float-cmp = { version = "0.9.0", features = ["std"], default-features = false }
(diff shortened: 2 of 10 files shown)


waywardmonkeys:
Apart from the discussion of separating `winit` stuff, I think this is a win because it makes a type that you have to use with `masonry` part of the `masonry` interface (`CursorIcon`) and improves on the current status quo where you have to look at the code to figure out where to get it from.

DJMcNab (approved):
Thanks. IWBN if we could lint against using these paths (`winit::cursor_icon`), but it doesn't appear to be possible in a built-in way.
The closest I can find is [`disallowed_methods`](https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_methods), but that isn't documented to handle paths in general. We could I guess use a janky grep solution, but that isn't good.

The only other potential issue is external access to `cursor_icon::ParseError` (e.g. for creating an error enum). I possibly would also export that as `CursorIconParseError` at the root.

waywardmonkeys:
[disallowed_types](https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_types) might work.

waywardmonkeys:
And good catch on `cursor_icon::ParseError`. I'll update this later today (hopefully) as I'm about to go to dinner.

waywardmonkeys:
Unfortunately, we can't use `disallowed-types` as setting that up in a `clippy.toml` to disallow `winit::window::CursorIcon` realizes that this is just `cursor_icon::CursorIcon` and then bans our usage of it via `masonry` as well:

```
warning: `cursor_icon::CursorIcon` is not allowed according to config
  --> masonry/src/widget/mod.rs:93:36
   |
93 |     pub fn cursor(&self) -> Option<CursorIcon> {
   |                                    ^^^^^^^^^^
   |
   = note: use masonry::CursorIcon (from clippy.toml)
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_types
```

So sad.

waywardmonkeys:
@DJMcNab Going to let you re-review and then I can hopefully land this.

DJMcNab (approved):

PR #303, opened 2024-05-11
Actually use the druid backspace logic

DJMcNab:
This is followup to #273.

This code originally came from https://github.com/xi-editor/xi-editor/pull/837

I've also brought the tests back


jaredoconnell (approved):
Tested and works well for me. The lack of backspace before was very annoying.

I wish functions like backspace_offset had more documentation on the reason behind the special-case logic, but I'm assuming you didn't write the original code, so you're not the right person to ask for that.

DJMcNab:
> The lack of backspace before was very annoying.

@jaredoconnell: To be clear, backspacing was added in #273; this merely makes it more robust when there are emoji. E.g. we would delete all of `👨🏿‍❤️‍👨🏻` instead of doing what vscode does and deleting leading to `👨🏿‍❤️`

PR #304, opened 2024-05-12
fix invalid HTML in examples

alexpyattaev:
For some reason the recent version of Trunk refused to serve some of the examples due to them missing the <html> tags. Adding the tags calms it down and allows it to work again. Trivial change, but would save someone the pain of having to fix examples before running them with trunk serve.

### DIFF
--- a/xilem_web/web_examples/counter/index.html
+++ b/xilem_web/web_examples/counter/index.html
@@ -1,4 +1,6 @@
 <!DOCTYPE html>
+<html>
 <title>Pythagorean theorem</title>
 
 <body></body>
+</html>
--- a/xilem_web/web_examples/counter_custom_element/index.html
+++ b/xilem_web/web_examples/counter_custom_element/index.html
@@ -1,4 +1,5 @@
 <!DOCTYPE html>
+<html>
 <head>
 <style>
     .gray {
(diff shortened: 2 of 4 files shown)


waywardmonkeys:
Without this, you get errors like:

```
2024-05-12T05:58:09.811900Z ERROR ❌ error
error from build pipeline

Caused by:
    0: HTML build pipeline failed (1 errors), showing first
    1: failed to finalize asset pipeline
    2: Document has neither a <link data-trunk rel="rust"/> nor a <body>. Either one must be present.
```

waywardmonkeys:
@DJMcNab Should we merge this directly or ... ?

waywardmonkeys (approved):

DJMcNab (approved):
Thanks for the PR.

Is this due to a recent change in trunk? That error message doesn't look like it's correctly describing the issue, as there is a `body` tag.
It would be good to open an issue there, or link to it from this if there's an existing one.

But this is an improvement for us, so thanks!

alexpyattaev:
I believe it is indeed a bug in trunk, but having valid html here is nice too:) I will make a bug report for trunk as well (assuming they do not have one already)

12.5.2024 9.03.39 Daniel McNab ***@***.***>:

waywardmonkeys:
Perhaps see https://github.com/trunk-rs/trunk/issues/791

PR #305, opened 2024-05-12
Add rustfmt config

PoignardAzur:
These standardized options can make code a little nicer. Right now they don't change anything, though they include an implicit commitment to adopt the # `imports_granularity = "Module"` and `group_imports = "StdExternalCrate"` settings.

While these settings are unstable, we can apply them with rustfmt nightly without actually committing to using nightly in CI. We should progressively move parts of the codebase towards that format in future PRs.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,7 +1,6 @@
 # See https://rust-lang.github.io/rustfmt for more information
 
-# Git can handle translation with Windows newlines, but if you have
-# configured git correctly, which is not the default on windows
+# Ensure lines end with \n even if the git configuration core.autocrlf is not set to true
 newline_style = "Unix"
 
 # `Foobar { foo, bar }` is more readable than `Foo { foo: foo, bar: bar }`
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,14 @@
+# See https://github.com/rust-lang/rustfmt for more information
+
+# `Foobar { foo, bar }` is more readable than `Foo { foo: foo, bar: bar }`
+use_field_init_shorthand = true
+
+# Forces let else blocks to always be their own line(s)
+single_line_let_else_max_width = 0
+
+# Subjectively, this combo improves readability on imports. The full prefix is always
+# visible right after `use`, and different types of imports are grouped together in
+# a predictable way (first std, then other crates, then local imports).
+# (These options are still unstable and thus commented out)
+# imports_granularity = "Module"
+# group_imports = "StdExternalCrate"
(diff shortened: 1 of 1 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,14 @@
+# See https://github.com/rust-lang/rustfmt for more information
+
+# `Foobar { foo, bar }` is more readable than `Foo { foo: foo, bar: bar }`
+use_field_init_shorthand = true
```
xStrom, line 4: This matches our current standard `rustfmt.toml` 👍

rustfmt.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# Forces let else blocks to always be their own line(s)
+single_line_let_else_max_width = 0
```
xStrom, line 7: This is a new introduction. It's quite minor and so I don't personally care much either way.

Maybe someone else has opinions as this is new.

rustfmt.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# Subjectively, this combo improves readability on imports. The full prefix is always
+# visible right after `use`, and different types of imports are grouped together in
+# a predictable way (first std, then other crates, then local imports).
+# (These options are still unstable and thus commented out)
+# imports_granularity = "Module"
+# group_imports = "StdExternalCrate"
```
xStrom, line 14: I agree with both of these values and yes the only reason they haven't been enabled is because they're still unstable. We can have them in the file already though as seen here.

xStrom (commented):
Let's also add

```toml
# Ensure line endings are \n
newline_style = "Unix"
```

which is part of our [current standard `rustfmt.toml`](https://github.com/linebender/vello/blob/6938a2893d6a2ba658709d1d04720f6c6033700f/rustfmt.toml) file.

PoignardAzur:
Does the `newline_style = "Unix"` lint have an impact? Git does the conversion for you anyway, right?

rustfmt.toml
```diff
@@ -0,0 +1,18 @@
+# See https://rust-lang.github.io/rustfmt for more information
+
+# Git can handle translation with Windows newlines, but if you have
+# configured git correctly, which is not the default on windows
```
xStrom, line 4: Let's be more specific about the git config property here.

```suggestion
# Ensure lines end with \n even if the git configuration core.autocrlf is not set to true
```

DJMcNab (approved):
Thanks!

We should probably port this across our other repositories as well

xStrom (approved):

PR #306, opened 2024-05-13
Check for typos in CI

jplatte:
Currently, the only legit typo being found is `seeked` but I'm not sure what to do about that.

Would be nice to check for new typos in CI anyways, right?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,9 +121,11 @@ jobs:
         # We currently skip checking masonry's docs
         run: cargo doc --workspace --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples --exclude masonry
 
+  # If this fails, consider changing your text or adding something to .typos.toml
   typos:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v4
-      - uses: crate-ci/typos@v1.21.0
+
+      - name: check typos
+        uses: crate-ci/typos@v1.21.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,3 +120,10 @@ jobs:
       - name: cargo doc
         # We currently skip checking masonry's docs
         run: cargo doc --workspace --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples --exclude masonry
+
+  typos:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: crate-ci/typos@v1.21.0
--- a/.typos.toml
+++ b/.typos.toml
@@ -3,7 +3,12 @@
 
 [default]
 extend-ignore-re = [
+    # Matches lorem ipsum text.
+    # In general, regexes are only matched until the end of a line by typos,
+    # and the repeated matcher at the end of both of these also ensures that
+    # matching ends at quotes or symbols commonly used to terminate comments.
     "Lorem ipsum [a-zA-Z .,]*",
+    "Phasellus in viverra dolor [a-zA-Z .,]*",
 ]
 
 # Corrections take the form of a key/value pair. The key is the incorrect word
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,13 +1,22 @@
+# See the configuration reference at
+# https://github.com/crate-ci/typos/blob/master/docs/reference.md
+
 [default]
 extend-ignore-re = [
     "Lorem ipsum [a-zA-Z .,]*",
 ]
 
+# Corrections take the form of a key/value pair. The key is the incorrect word
+# and the value is the correct word. If the key and value are the same, the
+# word is treated as always correct. If the value is an empty string, the word
+# is treated as always incorrect.
+
 [default.extend-identifiers]
 FillStrat = "FillStrat" # short for strategy
 
+# Case insensitive
 [default.extend-words]
-seeked = "seeked" # FIXME?
+seeked = "seeked" # Part of the HTML standard
 
 [files]
 extend-exclude = [
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,0 +1,15 @@
+[default]
+extend-ignore-re = [
+    "Lorem ipsum [a-zA-Z .,]*",
+]
+
+[default.extend-identifiers]
+FillStrat = "FillStrat" # short for strategy
+
+[default.extend-words]
+seeked = "seeked" # FIXME?
+
+[files]
+extend-exclude = [
+    "masonry/resources/i18n"
+]


.typos.toml
```diff
@@ -0,0 +1,15 @@
+[default]
+extend-ignore-re = [
+    "Lorem ipsum [a-zA-Z .,]*",
+]
```
DJMcNab, line 4: This regular expression looks like it either:
1) only handles to the end of the first line
2) disables checking for the entire file

I presume that it's option 1, and the perceived typos are only on the first line. But what happens if an update to typos makes something in that second paragraph be treated as a typo?

.typos.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[default.extend-identifiers]
+FillStrat = "FillStrat" # short for strategy
```
DJMcNab, line 7: This probably isn't a great name, but fixing that is not the job of this pr

.typos.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[default.extend-words]
+seeked = "seeked" # FIXME?
```
DJMcNab, line 10: Where is `seeked` used?

.typos.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[files]
+extend-exclude = [
+    "masonry/resources/i18n"
+]
```
DJMcNab, line 15: What does this do?

.github/workflows/ci.yml
```diff
@@ -120,3 +120,10 @@ jobs:
       - name: cargo doc
         # We currently skip checking masonry's docs
         run: cargo doc --workspace --all-features --no-deps --document-private-items -Zunstable-options -Zrustdoc-scrape-examples --exclude masonry
+
+  typos:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: crate-ci/typos@v1.21.0
```
DJMcNab, line 129: How often does this update? Do you think it's likely we'll remember to update it manually?

DJMcNab (commented):
I want to understand why https://github.com/linebender/vello/pull/492 is still draft before committing to this

Cc @waywardmonkeys

.typos.toml
jplatte: Agreed.

.typos.toml
jplatte: The intention was "until the end of the string literal" (notice how that character class repetition at the end doesn't include quotes), but you might be right that it only matches the first line. Could add `\n` to the character class but I kind of doubt that typos uses multiline regx. If an update to typos detects something in the second paragraph as a typo then (a) we'll only see it when explicitly bumping the version used in CI and (b) I'd say we should try with `\n` then and if it doesn't work, just make a similar regex for the second line, I guess?

.typos.toml
jplatte: https://github.com/search?q=repo%3Alinebender%2Fxilem%20seeked&type=code

.typos.toml
jplatte: Exclude translation files from typo detection because there's a lot of non-English content in there.

.github/workflows/ci.yml
jplatte: It updates at least once a month, but there's really no strong need to bump the version used here as often. I am subscribed to release notifications for it and tend to rerun it on the couple of projects I am contributing to that use it when I get an update, then if it finds a new typo I post a PR and include a version bump.

jplatte: It's also possible to use `@main` but I wouldn't recommend it as it can make CI fail somewhat frequently in annoying ways, plus I've seen it a couple of times that running it right after a release has the GitHub action try to get the latest release which hasn't been built + uploaded yet.

.typos.toml
DJMcNab: Oh yeah - my eyes entirely skimmed over the first line

Apologies for the confusion. We definitely should check that it works as expected

.github/workflows/ci.yml
waywardmonkeys: This is a good time to bring up Dependabot perhaps.

waywardmonkeys (approved):
This LGTM, but someone else should also approve.

waywardmonkeys:
> I want to understand why [linebender/vello#492](https://github.com/linebender/vello/pull/492) is still draft before committing to this
> 
> Cc @waywardmonkeys

Because of @xStrom's concerns about needing a `CONTRIBUTING.md`

waywardmonkeys:
Apparently, I'm my own worst enemy and I'm the one that brought up discoverability and `CONTRIBUTING.md` in my PR. Anyway, I've rebased, added a `.typos.toml` and marked it ready as well.

.github/workflows/ci.yml
waywardmonkeys: Also, see my similar PR in vello and I have a format that is slightly different than that here.

xStrom, line 129:
```suggestion

  # If this fails, consider changing your text or adding something to .typos.toml
  typos:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: check typos
        uses: crate-ci/typos@v1.21.0
```

.typos.toml
```diff
@@ -0,0 +1,15 @@
+[default]
+extend-ignore-re = [
+    "Lorem ipsum [a-zA-Z .,]*",
+]
+
+[default.extend-identifiers]
```
xStrom, line 6:
```suggestion
# Corrections take the form of a key/value pair. The key is the incorrect word
# and the value is the correct word. If the key and value are the same, the
# word is treated as always correct. If the value is an empty string, the word
# is treated as always incorrect.

[default.extend-identifiers]
```

.typos.toml
```diff
@@ -0,0 +1,15 @@
...
+FillStrat = "FillStrat" # short for strategy
+
+[default.extend-words]
```
xStrom, line 9:
```suggestion
# Case insensitive
[default.extend-words]
```

.typos.toml
```diff
@@ -0,0 +1,15 @@
+[default]
```
xStrom, line 1: Worth linking to the reference.

```suggestion
# See the configuration reference at
# https://github.com/crate-ci/typos/blob/master/docs/reference.md

[default]
```

.typos.toml
xStrom, line 10:
```suggestion
seeked = "seeked" # Part of the HTML standard
```

xStrom (commented):
Let's align this with [vello#492](https://github.com/linebender/vello/pull/492) so that it can become part of the unified Linebender CI standard.

.typos.toml
xStrom: It's part of the [HTML standard](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/seeked_event) so we should keep the typo to avoid confusion.

.github/workflows/ci.yml
xStrom: Using `@master` for dictionary updates can introduce failures for unrelated PRs, so similarly as how we update the Clippy version manually, updating this manually makes sense.

Maybe some rarely scheduled batched update by Dependabot can make sense in the future, but for now we can just add this to the checklist of our every-six-weeks CI update where we update the stable Rust version.

xStrom (approved):
Thanks!

PR #307, opened 2024-05-16
Ignore synthetic keyinput events for now

xorgy:
They are unused (and rarely useful), and they cause unwanted behavior.

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -120,7 +120,7 @@ impl ApplicationHandler<accesskit_winit::Event> for MainState<'_> {
             WinitWindowEvent::KeyboardInput {
                 device_id: _,
                 event,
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
To clarify, this would be useful e.g. in cases where an app has different behaviour whilst a key is pressed.

The only example I can really think of is a keyboard visualiser which wants to show which keys are currently pressed (e.g. for setting up keyboard shortcuts?).

I think this might need a todo comment, because it is a shame to lose the invariant that each key pressed event has a corresponding key release; it is definitely a footgun that we should document in `TextEvent`

xorgy:
@DJMcNab That's correct. I think there's probably a time and a place to reintroduce synthetic events, but it is a niche case, and right now propagating synthetic key events makes reasonably-programmed widgets behave in unexpected ways.

PR #309, opened 2024-05-20
Make execution on Android work properly

DJMcNab:
```
cargo apk run --example mason_android -p xilem
```

image

There will some followup needed here, of course. But some of that will want to come after xilem#235, when the way to do multiple windows has become clearer


xorgy:
Looks good to me. I will likely approve when I get home.

xorgy (approved):
Looks good to me, and runs. 🚀

PR #310, opened 2024-05-20
Move `xilem` onto a new `xilem_core`, which uses a generic View trait

DJMcNab:
This:
1) Renames the current/old `xilem_core` to `xilem_web_core` and moves it to the `xilem_web/xilem_web_core` folder
2) Creates a new `xilem_core`, which does not use (non-tuple) macros and instead contains a `View` trait which is generic over the `Context` type
3) Ports `xilem` to this `xilem_core`, but with some functionality missing (namely a few of the extra views; I expect these to straightforward to port)
4) Ports the `mason` and `mason_android` examples to this new `xilem`, with less functionality.

This continues ideas first explored in #235 

The advantages of this new View trait are:
1) Improved support for ad-hoc views, such as views with additional attributes.
     This will be very useful for layout algorithms, and will also enable native *good* multi-window (and potentially menus?)
2) A lack of macros, to better enable using go-to-definition and other IDE features on the traits

Possible disadvantages:
1) There are a few more traits to enable the flexibility
2) It can be less clear what `Self::Element::Mut` is in the `rebuild` function, because of how the resolution works
3) When implementing `View`, you need to specify the context (i.e. `impl<State, Action> View<State, Action, [new] ViewCtx> for Button<State, Action>`.


xilem_core2/src/element.rs
```diff
@@ -0,0 +1,100 @@
...
+///
+/// In most cases, there will be a corresponding implementation of [`SuperElement<Self>`] for
+/// some other type.
+/// This will be the generic form of this type, which is used for the implementation of [`AnyView`].
+///
+/// [`AnyView`]: crate::AnyView
+///
+// TODO: Rename so that it doesn't conflict with the type parameter names
+pub trait Element {
+    type Mut<'a>;
```
derekdreery, line 23: Could you comment this associated type - what is it?

DJMcNab: Added - does that makes sense to you?

masonry/src/action.rs
```diff
@@ -18,7 +18,7 @@ pub enum Action {
     TextEntered(String),
     CheckboxChecked(bool),
     // FIXME - This is a huge hack
-    Other(Arc<dyn Any>),
+    Other(Arc<dyn Any + Send + Sync>),
```
DJMcNab, line 21: I made this change because it was already using `Arc` - if it was meant to be single threaded, it should have been `Rc`

xilem/src/any_view.rs
```diff
@@ -22,186 +21,50 @@ use crate::{MasonryView, MessageResult, ViewCx, ViewId};
...
-    for BoxedMasonryView<State, Action>
-{
-    type Element = DynWidget;
-    type ViewState = AnyViewState;
+pub type AnyWidgetView<State, Action = ()> =
+    Box<dyn AnyView<State, Action, ViewCtx, Pod<DynWidget>> + Send + Sync>;
 
-    fn build(&self, cx: &mut ViewCx) -> (masonry::WidgetPod<Self::Element>, Self::ViewState) {
-        self.deref().dyn_build(cx)
+impl<W: Widget> SuperElement<Pod<W>> for Pod<Box<dyn Widget>> {
```
DJMcNab, line 27: This implementation is technically mostly used for sequences, not `any_view`

Maybe there's a better file for it?

xilem/src/view/button.rs
```diff
@@ -20,39 +21,48 @@ pub struct Button<F> {
...
-    type Element = masonry::widget::Button;
+    type Element = Pod<widget::Button>;
     type ViewState = ();
 
-    fn build(&self, cx: &mut ViewCx) -> (WidgetPod<Self::Element>, Self::ViewState) {
-        cx.with_leaf_action_widget(|_| {
-            WidgetPod::new(masonry::widget::Button::new(self.label.clone()))
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        ctx.with_leaf_action_widget(|_| {
+            Pod::from(WidgetPod::new(widget::Button::new(self.label.clone())))
```
DJMcNab, line 33: This should use `.into` - it's a mistake I made due to the old signature of `ViewCtx::with_leaf_action_widget`

xilem/src/view/flex.rs
```diff
@@ -52,127 +53,145 @@ impl<VT, Marker> Flex<VT, Marker> {
...
-            let child = self.element.child_mut(self.ix);
-            if child.is_some() {
-                break;
-            }
-            self.ix += 1;
+    fn with_scratch<R>(&mut self, f: impl FnOnce(&mut AppendVec<Pod<Box<dyn Widget>>>) -> R) -> R {
+        let ret = f(&mut self.scratch);
+        for element in self.scratch.drain() {
+            self.element.insert_child_pod(self.ix, element.inner);
+            self.element.insert_default_spacer(self.ix);
```
DJMcNab, line 157: Ah... That's a mistake.


```suggestion
            self.element.insert_default_spacer(self.ix + 1);
```

(Or any of the equivalent reformulations)

xilem/src/view/flex.rs
```diff
@@ -52,127 +53,145 @@ impl<VT, Marker> Flex<VT, Marker> {
...
+        f: impl FnOnce(<Pod<Box<dyn Widget>> as xilem_core::ViewElement>::Mut<'_>) -> R,
+    ) -> R {
+        let child = self
+            .element
+            .child_mut(self.ix)
+            .expect("ElementSplice::mutate won't overflow");
+        let ret = f(child);
+        self.element.remove_child(self.ix);
+        // Also remove the implicit spacer
+        // TODO: Make the spacers be explicit?
```
DJMcNab, line 188: I wonder if we want two forms of this?

xilem_core/Cargo.toml
```diff
@@ -1,21 +1,16 @@
...
-description = "Common core of the Xilem Rust UI framework."
-keywords = ["xilem", "ui", "reactive", "performance"]
-categories = ["gui"]
-publish = false # Until it's ready
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-homepage.workspace = true
 
-[package.metadata.docs.rs]
```
DJMcNab, line 13: We should recover this metadata

We probably need to have a standard set of this documented (on the wiki?)

xilem_core/examples/filesystem.rs
```diff
@@ -0,0 +1,218 @@
+// Copyright 2024 the Xilem Authors
```
DJMcNab, line 1: This is incomplete - need to add folders

Help welcome.

This is a bit messy, because it's not completely clear what the source of truth with the filesystem is. (We will likely have a similar issue with DOM nodes)

xilem_core/src/element.rs
```diff
@@ -0,0 +1,113 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! The types which can be used as elements in a [`View`](crate::View)
+
+/// A type which can be used as the `Element` associated type for a [`View`](crate::View).
+///
+/// It is expected that most libraries using `xilem_core` will have a generic
+/// implementation of this trait for their widget type.
+/// Additionally, this may also be implemented for additional types.
+/// In Xilem (the user interface library)
```
DJMcNab, line 11: Sentence fragment. I meant to talk about windows and layout properties here

xilem_core/src/element.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// This does require the reference type to be reborrowable.
```
DJMcNab, line 13: At the moment, we've avoided this (and it's non-trivial for `WidgetMut`). I hope this is YAGNI

xilem_core/src/element.rs
```diff
@@ -0,0 +1,113 @@
...
+/// An element which can be used for an [`AnyView`]
+pub trait AnyElement<Child>: SuperElement<Child>
+where
+    Child: ViewElement,
+{
+    /// Replace the inner value of this reference entirely
+    fn replace_inner(this: Self::Mut<'_>, child: Child) -> Self::Mut<'_>;
+}
+// TODO: What do we want to do here? This impl seems nice, but is it necessary?
+// It lets you trivially have sequences of types with a heterogenous element type,
```
DJMcNab, line 107:
```suggestion
// It lets you trivially have sequences of types with a homogenous element type,
```

xilem_core/src/message.rs
```diff
@@ -1,74 +1,99 @@
...
-    /// a value up the tree, rather than changing their internal state.
-    Action(A),
-    /// The event handler received a change request that requests a rebuild.
-    ///
-    /// Note: A rebuild will always occur if there was a state change. This return
-    /// type can be used to indicate that a full rebuild is necessary even if the
-    /// state remained the same. It is expected that this type won't be used very
-    /// often.
-    #[allow(unused)]
+    /// This allows for sub-sections of your app to use an elm-like architecture
```
DJMcNab, line 18: How useful is this really?

xilem_core/src/message.rs
```diff
@@ -1,74 +1,99 @@
...
 
-impl<A> MessageResult<A> {
-    pub fn map<B>(self, f: impl FnOnce(A) -> B) -> MessageResult<B> {
-        match self {
-            MessageResult::Action(a) => MessageResult::Action(f(a)),
-            MessageResult::RequestRebuild => MessageResult::RequestRebuild,
-            MessageResult::Stale(event) => MessageResult::Stale(event),
-            MessageResult::Nop => MessageResult::Nop,
-        }
+impl<T> Message for T
```
DJMcNab, line 54: This has some footguns around e.g. `Box(Message)` also implementing these same methods

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,650 @@
-// Copyright 2023 the Xilem Authors
+// Copyright 2024 the Xilem Authors
```
DJMcNab, line 1: Not sure about these year changes; they're mostly accidental

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,650 @@
...
+///    These can be nested if an ad-hoc sequence of more than 15 sequences is needed.
+///
+pub trait ViewSequence<State, Action, Context: ViewPathTracker, Element: ViewElement, Marker>:
+    'static
+{
+    type SeqState;
+
+    /// Build the associated widgets into `elements` and initialize all states.
+    #[must_use]
+    fn seq_build(&self, ctx: &mut Context, elements: &mut AppendVec<Element>) -> Self::SeqState;
```
DJMcNab, line 68: There are no previous elements (that this sequence should know about), so it's not meaningful for this to use the full splice.

Using AppendVec here simplifies management of this

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,650 @@
...
+
+impl<T> AppendVec<T> {
+    pub fn into_inner(self) -> Vec<T> {
+        self.inner
+    }
+    pub fn push(&mut self, item: T) {
+        self.inner.push(item);
+    }
+    pub fn drain(&mut self) -> impl Iterator<Item = T> + '_ {
+        self.inner.drain(..)
```
DJMcNab, line 23: Users could use std::mem::replace to remove items anyway, so we may as well provide this affordance.

We don't provide direct exclusive access to the vector to limit the potential for incorrectness

DJMcNab (commented):
Some thoughts which might be helpful for reviewers

xilem_core/src/message.rs
Philipp-M: I'm not sure yet about it exactly, but I guess it (also) allows to modularize views, possibly in external libraries. Maybe we don't need it, when we have found a satisfying solution for local/modularized state.

xilem_core/src/sequence.rs
Philipp-M: I mean most of the code here is rewritten, so 2024 makes sense? But maybe we also want to have year-ranges to be more accurate.

xilem_core/src/sequence.rs
Philipp-M: Originally using `ElementSplice` in `build` had the purpose of tracking the tree-structure (and in xilem_web adding children to the parent element in the DOM tree in place), and before that, it was just a `Vec`, but I guess when we're reimplementing tree-structure/xilem_web, we could revisit that. I want to investigate in xilem_web, to also use a separate element-tree (in Rust, not the DOM-tree I mean), so maybe it's not necessary, and we probably want to track the tree-structure directly in masonry anyway.

It was mostly a convenience thing, to reuse the trait, you're right, only elements are added in `build` (so IIRC only `ElementSplice::push` was needed).

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,650 @@
...
+    #[must_use]
+    fn seq_build(&self, ctx: &mut Context, elements: &mut AppendVec<Element>) -> Self::SeqState;
+
+    /// Update the associated widgets.
+    fn seq_rebuild(
+        &self,
+        prev: &Self,
+        seq_state: &mut Self::SeqState,
+        ctx: &mut Context,
+        elements: &mut impl ElementSplice<Element>,
```
Philipp-M, line 76: I suspect this makes the trait "object-unsafe".
Just noting, I see `ElementSplice` is already not object-safe, I remember working around that, to make it possible to use `dyn`, but maybe it's not necessary to have this object-safe.

Philipp-M (commented):
I've added a first round of comments, I'll go deeper into this later.

xilem/src/view/flex.rs
DJMcNab: Lol, fixed now, but I had to refind this bug

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,651 @@
...
+    let view_id = view_id.routing_id();
+    let id_low_ix = view_id as u32;
+    let id_high_gen = (view_id >> 32) as u32;
+    (id_low_ix as usize, id_high_gen)
+}
+
+/// The implementation for an `Option` of a `ViewSequence`.
+///
+/// Will mark messages which were sent to any index as stale if
+/// that index has been unused in the meantime.
```
PoignardAzur, line 345: Copy-paste error.

xilem_core/src/views/memoize.rs
```diff
@@ -42,17 +42,18 @@ It's not possible in Rust currently to check whether the (content of the) callba
...
-    D: PartialEq + Send + Sync + 'static,
-    V: MasonryView<State, Action>,
-    F: Fn(&D) -> V + Send + Sync + 'static,
+    Context: ViewPathTracker,
+    Data: PartialEq + 'static,
+    V: View<State, Action, Context>,
+    ViewFn: Fn(&Data) -> V + 'static,
 {
-    type ViewState = MemoizeState<State, Action, V>;
+    type ViewState = MemoizeState<State, Action, Context, V>;
```
PoignardAzur, line 52: I notice that `ViewState` is either `()` or some variant of `Child::ViewState` for every single implementation except this one. You might want to add a doc comment in the `View` and `ViewSequence` trait indicating that `Memoize` is the "bottom turtle" of both `ViewState` and `SeqState`.

xilem_core/examples/filesystem.rs
PoignardAzur, line 1: It's really nice that you added examples that don't require understanding Masonry, but I think what you actually want is unit tests.

PoignardAzur (approved):
I don't think we've really nailed "React architecture in Rust" yet, but this PR gets us one step closer.

xilem_core/src/views/memoize.rs
DJMcNab: I'm not sure that these docs would be true - the main use case for `ViewState` is to store an inner `SeqState`; the main use case for `SeqState` is actually to store generations to allow proper async wiring.

This is a use case of `ViewState`, but I'm personally inclined to try and find alternative patterns of memoization, because of the hot-reloading use cases.
Although actually doing some more thinking on that; I suppose that the Arc case is actually pretty similar...

xilem_core/examples/filesystem.rs
DJMcNab: Yeah, definitely. These models were mostly to test that things compiled as I was building up this core.

Do you want to block the PR on these being added?

xilem_core/src/views/memoize.rs
PoignardAzur: Yeah, I probably missed some stuff in the rebuild code, then. In any case, I'd like documentation to include a better description of what `ViewState` and `SeqState` do, because on a cursory read the only thing you see is "storing someone else's ViewState/SeqState, which ends up being `()`".

xilem_core/examples/filesystem.rs
PoignardAzur: Honestly, unless this PR is blocking further work (and even then), I'd say yes, until at least one test is added, for good practice.

We don't have to get 100% coverage, but writing tests in the same PR as the code is an habit you have to build, and now is as good a time as any to start. And It's a good precedent for this codebase.

DJMcNab: Great. For some context on this untested state, the final version of this pr was submitted as my final action before going on holiday.

My first task for next week can certainly be adding tests here.

Philipp-M (approved):
Phew... not the easiest to digest code, but indeed better than #235 (when it can cover the functionality introduced there without the reborrowing)
I do think the added complexity is worth it though to avoid the macros and being more flexible with different view contexts.
And it seems, that not every contributor of xilem does need to understand all this code for (non-core) contributions...
I hope, that due to all the added typing, composing views will not let the compiler explode (in compilation time mostly).

Great, well-thought work!

> I don't think we've really nailed "React architecture in Rust" yet

I'm not sure we even want that, having to deal with a lot of React ugliness in my job currently (e.g. `useEffect` or magic (unwanted) reconciliations).
I'm not the biggest fan of distributed "reactive" state as it's done in React (nor the "centralized state" solutions that exist there)...
We still want to find a satisfying and simple solution for modularized views and state anyway, to support convenient library crates for (composed?) views. (but that's orthogonal to this PR)

xilem_core/src/any_view.rs
```diff
@@ -1,137 +1,342 @@
...
+/// This is useful for a view which can be any of several view types, by using
+/// `Box<dyn AnyView<...>>`, which implements [`View`].
+// TODO: Mention `Either` when we have implemented that?
+///
+/// This is also useful for memoization, by storing an `Option<Arc<dyn AnyView<...>>>`,
+/// then [inserting](Option::get_or_insert_with) into that option at view tree construction time.
+///
+/// Libraries using `xilem_core` are expected to have a type alias for their own `AnyView`, which specifies
+/// the `Context` and `Element` types.
+pub trait AnyView<State, Action, Context, Element: crate::ViewElement> {
```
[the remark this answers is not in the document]
DJMcNab: So this lets you use any type which can be converted into this element type.

I.e. `Box<AnyWidgetView>` can internally be a `Pod<Button<...>>` or `Pod<Flex<...>>`

xilem_core/src/element.rs
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    /// If you need to get a return value from the .
+    /// Unfortunately, it isn't possible to abstract over reborrowing without a closure or equivalent.
+    ///
+    /// [#49601](https://github.com/rust-lang/rust/issues/49601)
+    fn with_reborrow(this: Self::Mut<'_>, f: impl FnOnce(Self::Mut<'_>)) -> Self::Mut<'_> {
+        let (this, ()) = Self::with_reborrow_val(this, f);
+        this
+    } */
+}
```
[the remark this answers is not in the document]
DJMcNab: IMO reborrowing isn't unidiomatic.

This API form is needed because it's the only way to represent reborrowing in the type system.
But yes, Masonry makes it complicated

xilem_core/src/element.rs
```diff
@@ -0,0 +1,115 @@
...
+/// A type which can be used as the `Element` associated type for a [`View`](crate::View).
+///
+/// It is expected that most libraries using `xilem_core` will have a generic
+/// implementation of this trait for their widget type.
+/// Additionally, this may also be implemented for other types, depending on the
+/// needs of the specific parent view.
+/// In Xilem (the user interface library), this is also used for types containing the
+/// flex properties of their child views, and window properties.
+///
+// /// This does require the reference type to be reborrowable.
```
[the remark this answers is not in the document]
DJMcNab: These docs are commented out

xilem_core/src/view.rs
```diff
@@ -0,0 +1,200 @@
...
+/// The `View` trait is parameterized by `State`, which is known as the "app state",
+/// and also a type for actions which are passed up the tree in message
+/// propagation.
+/// During message handling, mutable access to the app state is given to view nodes,
+/// which will in turn generally expose it to callbacks.
+///
+/// ## Alloc
+///
+/// In order to support the open-ended [`DynMessage`] type, this trait requires an
+/// allocator to be available.
```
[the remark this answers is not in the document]
DJMcNab: The allocator story isn't great, but a lot of views need it - i.e. sequences and AnyView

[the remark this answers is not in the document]
Philipp-M: I see, in that case we may want to wait for Allocator traits etc. anyways before committing to something like that.
I don't think there are good ways for these cases without allocators?

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,651 @@
...
+}
+
+impl<T> AppendVec<T> {
+    pub fn into_inner(self) -> Vec<T> {
+        self.inner
+    }
+    pub fn push(&mut self, item: T) {
+        self.inner.push(item);
+    }
+    pub fn drain(&mut self) -> impl Iterator<Item = T> + '_ {
```
Philipp-M, line 22:
```suggestion
    pub fn drain(&mut self) -> Drain<'_, T> {
```

Or should this hide the underlying type/implementation details?
With the concrete type it e.g. also has the properties `impl ExactSizeIterator` and `impl DoubleEndedIterator`.

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,651 @@
...
 
-            fn mark(&mut self, changeflags: $changeflags, _cx: &mut $cx) -> $changeflags
-            {
-                self.last_mutated_mut().map(|pod| pod.mark(changeflags)).unwrap_or_default()
-            }
+/// Marker type to workaround trait ambiguity.
+#[doc(hidden)]
+pub struct WasAView;
+
+impl<State, Action, Context, V, Element> ViewSequence<State, Action, Context, Element, WasAView>
```
Philipp-M, line 118:
```suggestion
impl<State, Action, Context, V, Element> ViewSequence<State, Action, Context, Element, fn() -> WasAView>
```

Otherwise it's e.g. not possible to `Clone` a ViewSequence (I just ran into that).
Alternatively we could also do
```rust
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default, Copy)]
pub struct WasAView;
```

xilem_core/src/sequence.rs
```diff
@@ -1,363 +1,651 @@
...
-            fn mutate(&mut self, _cx: &mut $cx) -> &mut $pod
-            {
-                self.mutate()
-            }
+/// A temporary "splice" to add, update and delete in an (ordered) sequence of elements.
+/// It is mainly intended for view sequences.
+pub trait ElementSplice<Element: ViewElement> {
+    fn with_scratch<R>(&mut self, f: impl FnOnce(&mut AppendVec<Element>) -> R) -> R;
+    /// Insert a new element at the current index in the resulting collection.
+    fn push(&mut self, element: Element);
```
Philipp-M, line 105: Do we still need this, when there's `with_scratch` (which I'm not super happy with the naming, maybe it should be called `append` or this should be the new `push` instead)?
If we do keep this, it should either push at the end (this was the behavior before IIRC) or be called `insert` as the doc comment already implies IMO.

xilem_core/src/message.rs
```diff
@@ -1,74 +1,99 @@
...
+/// We are aware of potential backwards-compatible workarounds, but
+/// are not aware of any tangible need for this.
+///
+/// [`View`]: crate::View
+pub type DynMessage = Box<dyn Message>;
+/// Types which can be contained in a [`DynMessage`].
+// The `View` trait could have been made generic over the message type,
+// primarily to enable flexibility around Send/Sync and avoid the need
+// for allocation.
+pub trait Message: 'static + Send {
```
Philipp-M, line 48: I'm currently running into the issue while rewriting xilem_web event listeners, that this requires `Send`. `web_sys::Event` doesn't implement `Send`.

So the `Send` bound should be optional (another reason to make it generic?).

Philipp-M:
Hmmm, I just discovered one major issue because of the orphan rule (in hindsight it's obvious):
The trait can not be implemented for foreign types in downstream crates, and I think this is somewhat a dealbreaker for nice API design, following e.g. is not possible:
```rust
impl<State, Action> View<State, Action, ViewCtx> for &'static str {..}
```

Philipp-M:
> The trait can not be implemented for foreign types in downstream crates

I mean I'm honest, I really don't like it, as it convolutes everything with wrapper types such as `text("text")` or `number(123)`, but I guess it will be difficult to have different implementations in one project anyway (e.g. we have a masonry text node and a web text node and want to use it in the same project (after a potential merger of those crates), e.g. in an embedded web-view).

I'm asking myself, if there's a general way to provide such an implementation in core and have different contexts define which implementation/element is used or something like that... But I'm not sure if that's possible... That makes [this](https://github.com/rust-lang/rust-project-goals/blob/main/src/2024h2/Relaxing-the-Orphan-Rule.md) even more relevant

xilem_core/src/sequence.rs
DJMcNab: Yeah - it would be nice to have an `AnyViewSequence` (and we'll probably need it for hot reloading?), but it should still be a relatively straightforward adaptation.

DJMcNab:
Those general implementations can be added, as shown in DJMcNab/xilem#4

They do lead to *some* jankiness however. In particular, you can't have:
```
fn test()->impl WidgetView<...> {
    "my string"
}
```
because of  the new required marker

xilem_core/src/sequence.rs
DJMcNab: I *think* it's needed to avoid additional e.g. allocations when they're not needed.

It's also a bit simpler to use. Although really, manual implementation of `ViewSequence` seem so unlikely that it wouldn't be unreasonable to remove this

xilem/src/view/flex.rs
DJMcNab: This will probably be a follow-up PR

xilem/src/lib.rs
```diff
@@ -200,36 +159,20 @@ impl ViewCx {
 
     pub fn with_leaf_action_widget<E: Widget>(
         &mut self,
-        f: impl FnOnce(&mut Self) -> WidgetPod<E>,
-    ) -> (WidgetPod<E>, ()) {
+        f: impl FnOnce(&mut Self) -> Pod<E>,
```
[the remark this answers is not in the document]
DJMcNab: I've added a new constructor - `Pod::new`

xilem_core/src/sequence.rs
DJMcNab: Oh, so it's a [perfect derive](https://smallcultfollowing.com/babysteps//blog/2022/04/12/implied-bounds-and-perfect-derive/) issue?
I'd say that you should just implement `Clone` for `Flex` manually; it's a logical mistake that `Flex: Clone` would require `Marker: Clone`, since that implementation is never used - working around it here isn't very principled.

That the same issue arises with `Debug` is unfortunate, and was a hazard I didn't see before. Overall, I'm happy to add the `Debug` derive here, but I'm going to push back on the fixes to make `Clone` work like that here.

xilem_core/src/any_view.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 the Xilem Authors
+// Copyright 2023 the Xilem Authors
```
Philipp-M, line 1: Does it make sense to add ranges (i.e. 2023-2024) here? I think that's more accurate?

xilem_core/src/message.rs
```diff
@@ -17,6 +17,8 @@ pub enum MessageResult<Action> {
     /// This allows for sub-sections of your app to use an elm-like architecture
     Action(Action),
     // TODO: What does this mean?
+    /// This message's handler needs a rebuild to happen.
+    /// The exact semantics of this method haven't been determined.
```
Philipp-M, line 21: Well, it has been, but that was when we had async in xilem, it was used when a future has completed, sent a message and in `View::message` a rebuild was requested (e.g. when a list view entry in the async list has finished loading)

Philipp-M (approved):
Great work, certainly a big improvement in code quality.
Nice job also with all those exhaustive tests (seems like they surfaced some issues as well?)

xilem_core/src/message.rs
DJMcNab: But my question is how is that different from the implied result of an action handler at the root?

The semantics here are clearly not very principled

xilem_core/src/any_view.rs
DJMcNab: I don't think we want to do that

Afaik, the years don't really mean much, and are much more of a formality.

xilem_core/src/message.rs
Philipp-M: You mean via `MessageResult:Action`?

DJMcNab: Yeah, exactly.

PR #311, opened 2024-05-20
masonry: Remove `'static'` bound requirement for `AppDriver`

Philipp-M:
Not sure if the boxing of the `AppDriver` is somehow intended. But in case it's not, we can remove the `'static` bound requirement for that.


DJMcNab (commented):
This will also be incompatible with the web in future - see winit's [`spawn_app`](https://docs.rs/winit/0.30.0/winit/platform/web/trait.EventLoopExtWebSys.html#tymethod.spawn_app)

DJMcNab:
Agreed in office hours [2024-05-23](https://xi.zulipchat.com/#narrow/stream/359642-office-hours/topic/2024-05-23)

Issue #312, opened 2024-05-22
[Masonry] Implement unified PointerEvent


xorgy:
There is some consensus that a PointerEvent-style API (like existed in Glazier) is probably a good way to handle touch and mouse in general. This is a tracking issue for that topic.

raphlinus:
There are a few layers to this. The simplest layer is just to enable mouse-like behavior from touch, so that touch can operate widgets like buttons and sliders, and also selection (and focus) in text input.

Another more sophisticated layer is to support multi-touch well. That has potentially far-reaching changes, as multiple widgets can be active at the same time. A careful design is needed.

And yet another more sophisticated layer is gesture disambiguation and recognition. A touch-drag gesture on a slider widget should move the slider, but on mobile, the same gesture *not* on a suitably sensitive widget should be interpreted as a scroll. There *may* be platform-specific gesture recognition code we want to call into, or it may be that we want a pure Rust cross-platform implementation of all this, tunable to feel native on the specific platform.

There may also be improvements desired on the winit side. I see a bunch of PRs to add pen support (https://github.com/rust-windowing/winit/pull/2647 is the most recent) but they seem to be stuck in review. https://github.com/rust-windowing/winit/issues/99 seems to be the main tracking issue there.

To be clear, I think we should have the simplest layer first, so we can unblock simple UI interactions on mobile, then carefully design what we want to do for the other two.

PoignardAzur:
How much progress have we made on the Android side of this?

xorgy:
@PoignardAzur 
Winit has unified their PointerEvent interface upstream, and 0.31.0 will have this, they have also slated click counts on winit `PointerEvent`s for 0.31.0.
I think at this point it'd be best to just keep what we have, and when we move to winit 0.31.0 we can make things more elegant (and delete a bunch of code).

We still have to work out properly intercepting PointerEvents from touch (and maybe pen) devices. On the web, the UA handles this under the hood. Handling touch gesture recognition properly will still involve retaining the full history of a touch (timestamp, position, and force) between `Start` and `End`/`Cancel`

PoignardAzur:
I think we've implemented all of the above.

DJMcNab:
I'm not sure that we have. Our handling of touch is currently to pretend that they are the mouse, which isn't really right.

xorgy:
> I'm not sure that we have. Our handling of touch is currently to pretend that they are the mouse, which isn't really right.

[...]

PR #313, opened 2024-05-24
Add basic translation of Touch to PointerEvent.

xorgy:
https://github.com/linebender/xilem/assets/1272018/6e9aeefb-f324-4738-9aa2-f54d643631e2

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -284,6 +284,46 @@ impl ApplicationHandler<accesskit_winit::Event> for MainState<'_> {
                         self.pointer_state.clone(),
                     ));
             }
+            WinitWindowEvent::Touch(winit::event::Touch {
+                location, phase, ..
+            }) => {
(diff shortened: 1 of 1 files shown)


masonry/src/event_loop_runner.rs
```diff
@@ -284,6 +284,46 @@ impl ApplicationHandler<accesskit_winit::Event> for MainState<'_> {
                         self.pointer_state.clone(),
                     ));
             }
+            WinitWindowEvent::Touch(winit::event::Touch {
+                location, phase, ..
+            }) => {
+                // FIXME: This is naïve and should be refined for actual use.
```
DJMcNab, line 290: When this lands, we should add a comment in #312 pointing at this section of code (as an example of what not to do lol)

DJMcNab (approved):
Functions for me as expected

masonry/src/event_loop_runner.rs
xorgy: I mean... it is ‘Unified’ :upside_down_face:

PR #314, opened 2024-05-24
Signal TextFieldFocused and ImeMoved to enable and position the IME.

xorgy:


xorgy:
This works, but I can almost guarantee there will be disagreement about the details, so this draft is the minimum working thing.

masonry/src/render_root.rs
```diff
@@ -570,8 +570,9 @@ impl RenderRoot {
             self.state.focused_widget = new;
             self.root_lifecycle(event);
 
-            // TODO - Handle IME
-            // Send TextFieldFocused(focused_widget) signal
+            self.state
+                .signal_queue
+                .push_back(RenderRootSignal::TextFieldFocused);
```
PoignardAzur, line 575: The code shouldn't unconditionally send a TextFieldFocused event. There are cases where `new` is `None`, or refers to a non-textbox widget.

masonry/src/widget/textbox.rs
```diff
@@ -277,6 +280,21 @@ impl Widget for Textbox {
...
+                        x: origin.x,
+                        y: origin.y + size.height,
+                    },
+                    LogicalSize {
+                        width: size.width,
+                        height: size.height,
+                    },
+                ),
+            );
+        }
```
PoignardAzur, line 297: I get the intent, but this code doesn't really work. `layout()` is called *before* a widget is placed. It determines its size, not its position. The way layout currently works, the only point where you know a widget's position for sure is after the entire layout pass has ran.

You can fetch a widget's position from the root using the `find_widget_by_id` methods.

masonry/src/event_loop_runner.rs
```diff
@@ -430,13 +428,13 @@ impl MainState<'_> {
...
                 }
                 render_root::RenderRootSignal::ImeStarted => {
-                    // TODO
+                    window.set_ime_allowed(true);
                 }
-                render_root::RenderRootSignal::ImeMoved => {
-                    // TODO
+                render_root::RenderRootSignal::ImeMoved(position, size) => {
+                    window.set_ime_cursor_area(position, size);
                 }
```
PoignardAzur, line 438: We may want to remove some of these events. They're placeholders, it's not like I had a plan in mind for what IME would look like when I created them.

PoignardAzur (changes requested):
I think we need to hash out what our architecture for IME placement is supposed to be.

masonry/src/event_loop_runner.rs
xorgy: I tried not to mark any of this as YAGNI because I know you have thought a lot about _other_ things in here. It definitely didn't seem like all of these should exist. I'll cut some of them out.

masonry/src/render_root.rs
xorgy: I handle this now, thank you.

masonry/src/event_loop_runner.rs
xorgy: I've gone ahead and cleaned these up a bit. Feel free to bikeshed the names a bit.

masonry/src/widget/textbox.rs
```diff
@@ -170,6 +171,7 @@ impl Widget for Textbox {
                         ctx.request_layout();
                         ctx.request_paint();
                         ctx.request_focus();
+                        ctx.widget_state.needs_window_origin = true;
```
xorgy, line 174: I don't know if this is the right way. I am surprised @PoignardAzur had nothing to say about it so far. :+ )

masonry/src/render_root.rs
```diff
@@ -570,8 +567,11 @@ impl RenderRoot {
             self.state.focused_widget = new;
             self.root_lifecycle(event);
 
-            // TODO - Handle IME
-            // Send TextFieldFocused(focused_widget) signal
+            self.state.signal_queue.push_back(if new.is_some() {
```
DJMcNab, line 570: It's not clear to me how you know that `new` is a textbox.

masonry/src/widget/textbox.rs
PoignardAzur: It's not the right way. I'm not sure I understand the logic here.

masonry/src/render_root.rs
xorgy: No distinction is made between text boxes and other focused widgets, so all we know is that a widget is focused or that no widget is focused. 
Forgot that this was unaddressed.

PoignardAzur (approved):
Overall this code is un-idiomatic, but it can be merged as a first implementation.

masonry/src/widget/textbox.rs
PoignardAzur, line 174: I think that moving the `ImeMoved` signal to `paint()` means you can remove this line.

masonry/src/render_root.rs
```diff
@@ -570,8 +567,11 @@ impl RenderRoot {
...
+                RenderRootSignal::StartIme
+            } else {
+                RenderRootSignal::EndIme
```
PoignardAzur, line 573: I think that part is a bit over-simplified.

What happens if you switch between two textboxes? Shouldn't the IME session be restarted, instead of just moved to a different position?

xorgy: No it should not be restarted, however the surround should be updated (which is part of the Android IME work).

xorgy: Though maybe this is more just exposing an issue with conflation of opening the soft keyboard with starting the IME. The issue with restarting it has nothing to do with resetting the IME (which is desirable) but rather relates to opening the soft keyboard. On Wayland, these two things are not separable, but on Android they are separable.
Winit might be able to work around this by coalescing the two `set_ime_allowed` state changes but idk if that introduces more problems.

masonry/src/render_root.rs
xorgy: I've added a TODO so that this can be addressed in the next iteration of this (probably at the same time as introducing the Android/OSK IME changes).

PR #320, opened 2024-05-29
xilem,masonry: Use `dpi` crate directly

waywardmonkeys:
This removes a class of dependencies from within both `masonry` and `xilem` on the `winit` crate where we can just use `dpi` directly instead.

The `dpi` crate is meant (like `cursor_icon`) to be a shared ecosystem crate rather than only for usage with `winit`.

### DIFF
--- a/masonry/examples/calc.rs
+++ b/masonry/examples/calc.rs
@@ -11,6 +11,7 @@ use std::sync::Arc;
 
 use accesskit::{DefaultActionVerb, Role};
 use masonry::app_driver::{AppDriver, DriverCtx};
+use masonry::dpi::LogicalSize;
 use masonry::widget::{Align, CrossAxisAlignment, Flex, Label, RootWidget, SizedBox, WidgetRef};
 use masonry::{
     AccessCtx, AccessEvent, Action, BoxConstraints, Color, EventCtx, LayoutCtx, LifeCycle,
@@ -20,7 +21,6 @@ use masonry::{
 use smallvec::{smallvec, SmallVec};
 use tracing::{trace, trace_span, Span};
 use vello::Scene;
-use winit::dpi::LogicalSize;
 use winit::window::Window;
 
 #[derive(Clone)]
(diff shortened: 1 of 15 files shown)


masonry/examples/calc.rs
```diff
@@ -10,6 +10,7 @@
 use std::sync::Arc;
 
 use accesskit::{DefaultActionVerb, Role};
+use dpi::LogicalSize;
```
DJMcNab, line 13: If our users need to use this, we should re-export this entire crate

xilem/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 #![allow(clippy::comparison_chain)]
 use std::{any::Any, collections::HashMap};
 
+use dpi::LogicalSize;
```
DJMcNab, line 7: i.e. here and in masonry

```rust
pub use dpi;
```

masonry/src/widget/widget_pod.rs
```diff
@@ -2,9 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use accesskit::{NodeBuilder, NodeId};
+use dpi::LogicalPosition;
```
DJMcNab, line 5: Then e.g. here:

```suggestion
use crate::dpi::LogicalPosition;
```

DJMcNab (approved):
Thoughts on re-exporting, although this current state is also an improvement.

I do also have the same concern about there being multiple paths to the same item and that being a maintenance burden

waywardmonkeys:
Just noting that I agree with all of the feedback (so far( and will address it soon, I hope.

masonry/examples/calc.rs
waywardmonkeys: Done!

waywardmonkeys:
@DJMcNab All done! I think that's a good improvement, thanks for the feedback.

(It is probably worth noting that similar changes should be done for AccessKit and perhaps other things, separately.)

DJMcNab (approved):
Thanks!

PR #322, opened 2024-05-30
Add diagram to document xilem layers

PoignardAzur:
Note that this adds a link to a section added in https://github.com/linebender/parley/pull/69

I removed Taffy from the diagram for simplicity, and because Taffy's current integration is in flux.

Overall I think the Xilem README is in desperate need of a rewrite, and this PR should be considered a stopgap. Further PRs should better integrate the ecosystem descriptions into the README, move some sections around, etc.


README.md
```diff
@@ -44,7 +44,7 @@ These are all just vanilla data structures. The next step is diffing or reconcil
 
 The closures are the interesting part. When they're run, they take a mutable reference to the app data.
 
-## Components
+### Components
```
Philipp-M, line 47: Not sure about these changes, they don't seem to be under a relevant section?

README.md
```diff
@@ -109,6 +109,21 @@ To install them on Debian or Ubuntu, run
 sudo apt-get install pkg-config clang libwayland-dev libxkbcommon-x11-dev libvulkan-dev
 ```
 
+## Project structure
+
+This diagram gives an idea what the Xilem project is built on:
+
+![docs/assets/xilem-layers.svg]
```
Philipp-M, line 116:
```suggestion
![Xilem project layers](docs/assets/xilem-layers.svg)
```

README.md
```diff
@@ -109,6 +109,21 @@ To install them on Debian or Ubuntu, run
...
+
+On a very coarse level, Xilem is built directly on top of xilem_core and Masonry, both of which are crates in this repository.
+
```
Philipp-M, line 119: When introducing this new section, it probably makes sense to link to `ARCHITECTURE.md` (which needs some rework as well) for more details?

docs/assets/xilem-layers.svg
raphlinus, line 1: I approve the inclusion of this file (which I authored) in the repo.

README.md
```diff
@@ -65,29 +87,16 @@ fn app_logic(data: &mut AppData) -> impl View<AppData, (), Element = impl Widget
...
-            data.count += 1
-        })
-    }),
-}
-```
-
-The current code uses a `PartialEq` bound, but in practice I think it might be much more useful to use pointer equality on `Rc` and `Arc`.
-
-The combination of memoization with pointer equality and an adapt node that calls [Rc::make_mut] on the parent type is actually a powerful form of change tracking, similar in scope to Adapton, self-adjusting computation, or the types of binding objects used in SwiftUI. If a piece of data is rendered in two different places, it automatically propagates the change to both of those, without having to do any explicit management of the dependency graph.
+When a subtree is a pure function of some data, as is the case for the button above, it makes sense to *memoize.* The data is compared to the previous version, and only when it's changed is the view tree build. The signature of the memoize node is nearly identical to [Html.lazy] in Elm:hide the underlying type/implementation details?
```
waywardmonkeys, line 94: Is the text from `:hide` on supposed to be here?

README.md
```diff
@@ -65,29 +87,16 @@ fn app_logic(data: &mut AppData) -> impl View<AppData, (), Element = impl Widget
...
+With the concrete type it e.g. also hastype is actually a powerful form of change tracking, similar in scope to Adapton, self-adjusting computation, or the types of binding objects used in SwiftUI. If a piece of data is rendered in two different places, it automatically propagates the change to both of those, without having to do any explicit management of the dependency graph.
```
waywardmonkeys, line 95: I think the first sentence here isn't clear (`hastype`) ...

waywardmonkeys (changes requested):

README.md
PoignardAzur: Oh wow, no idea how that happened. I must have accidentally copy-pasted the wrong snippet somewhere. Good catch!

docs/assets/xilem-layers.svg
PoignardAzur: (To be clear, I took that file from a page of Raph's last talk, made some minor changes, then included it here.)

waywardmonkeys (approved):

PR #324, opened 2024-05-31
masonry: Pass in winit `CursorEntered` events

waywardmonkeys:


PoignardAzur (approved):
Looks good to me. Not sure why it wasn't already the case in the first place.

waywardmonkeys:
Probably because nothing uses it so far. (You care about it leaving, but when it enters the window, you'd start getting moved events...)

xStrom:
It wasn't already in due to the Druid ancestry, as in Druid this would be redundant because of `HotChanged(true)`.

PR #325, opened 2024-05-31
Allow feeding masonry from an externally event loop.

waywardmonkeys:
This allows for running `masonry` with an externally managed event loop for integrating with an application that is already running an event loop.

* Expose `PointerState` to allow creating `PointerEvent`
* Expose `WindowEvent` to allow managing window size and scale factor.
* Make `DriverCtx` usable from outside the crate. While this struct was already `pub`, the sole member of it was not.

### DIFF
--- a/masonry/src/app_driver.rs
+++ b/masonry/src/app_driver.rs
@@ -8,7 +8,12 @@ use crate::{Action, Widget, WidgetId};
 
 pub struct DriverCtx<'a> {
     // TODO
-    pub(crate) main_root_widget: WidgetMut<'a, Box<dyn Widget>>,
+    // This is exposed publicly for now to let people drive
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
I think there are going to be better ways to integrate other event loops in the future, but so far, this is the bare minimum for me to get parts of `masonry` usable from an application that is already running and driving a winit event loop.

DJMcNab (approved):
Can we document that the fields of DriverCtx are only exposed to expediently enable external running, and that this API is not stable/supported?

waywardmonkeys:
I've added a comment and a `#[doc(hidden)]`. Re-approve and I'll land it when I'm awake.

DJMcNab (approved):

PR #328, opened 2024-06-03
masonry: Rename a scale factor variable

waywardmonkeys:
This helps make it more clear when skimming through this code that this is the scale factor being used to transform the scene.


waywardmonkeys:
This may seem silly, but this cost me a lot of time in tracking down why my own integration wasn't working correctly.

I should probably help add some comments / documentation somewhere to help document / describe how window scale factors are used.

DJMcNab (approved):
Seems reasonable. It's a shame that this scale factor application needs a new scene, wrt to memory pressure (both allocation and just movement)

But our scenes will be small enough that this shouldn't have a huge impact

PR #329, opened 2024-06-03
Document masonry

PoignardAzur:
This was initially a supposed to be a small documentation pass, which grew into a few changes:

- Adding back a to-do-list example.
- Fixing the bugs revealed by that example (infinite bounding boxes, wrong accessibility handling in Portal, etc).
- Making sure all widgets return the correct spans instead of the less-useful default one.
- Adding a trace to the layout pass for easier debugging.

Overall I'm pretty happy with this!


masonry/README.md
```diff
@@ -20,11 +20,74 @@ It also gives you tools to inspect that widget tree at runtime, write unit tests
...
+}
+
+impl AppDriver for Driver {
+    fn on_action(&mut self, ctx: &mut DriverCtx<'_>, _widget_id: WidgetId, action: Action) {
+        match action {
+            Action::ButtonPressed => {
+                let mut root: WidgetMut<RootWidget<Portal<Flex>>> = ctx.get_root();
+                let mut root = root.get_element();
+                let mut flex = root.child_mut();
+                flex.add_child(Label::new(self.next_task.clone()));
```
DJMcNab, line 52: Do we also want to clear the text widget here?

masonry/doc/ARCHITECTURE.md
```diff
@@ -20,31 +22,31 @@ Masonry has some opinionated design goals:
...
 Contains the TestHarness type, various helper widgets for writing tests, and the snapshot testing code.
 
+### `src/text2/`
+
+Contains text-handling code, for both displaying and editing text. Has been overhauled during the port to winit, but still in rough shape. Here be dragons.
+
 ### `src/text/`
 
-Contains text-handling code, for both displaying and editing text. Hasn't been maintained in a while, here be dragons.
+Dead code, should probably be cleaned up.
```
DJMcNab, line 35: Yeah - it's provided for reference of future direction, e.g. `movement.rs` which we expect to move to Parley.
At this point, we have most of the functionality elsewhere.

masonry/src/lib.rs
```diff
@@ -3,40 +3,35 @@
...
-//! **(TODO: FIX THIS EXAMPLE)**
-//! The todo-list example looks like this:
+//! The to-do-list example looks like this:
 //!
-//! ```ignore
-//! use masonry::widget::{prelude::*, TextBox};
-//! use masonry::widget::{Button, Flex, Label, Portal, WidgetMut};
-//! use masonry::Action;
-//! use masonry::{AppDelegate, AppLauncher, DelegateCtx, WindowDescription, WindowId};
+//! ```
```
DJMcNab, line 16: I'm not a fan of having the same example in three different places, but since doc examples aren't runnable, at least one of those duplications is inevitable. I expect we'll lose the second when we move to `#[doc = concat!(include!())]`

masonry/src/widget/label.rs
```diff
@@ -261,6 +261,10 @@ impl Widget for Label {
         SmallVec::new()
     }
 
+    fn make_trace_span(&self) -> Span {
+        trace_span!("Label")
```
DJMcNab, line 265: If this is necessary, is it worth just removing the default impl?

DJMcNab (approved):
I've not carefully reviewed the Portal code, but otherwise this looks good to me.

masonry/README.md
PoignardAzur: Nah, I'm deliberately keeping the README example short. We'll clear the textbox once we have a way to make this code more concise.

masonry/src/widget/label.rs
PoignardAzur: It's not really necessary, it just makes better logs. I thought about removing the default impl, but the Widget trait is hard enough to implement as it is.

PR #332, opened 2024-06-04
masonry: Use `dpi` via `crate::dpi`

waywardmonkeys:


waywardmonkeys:
@DJMcNab I know this is what you were afraid would happen. I think in the end, we'll hopefully be able to make `winit` an optional dependency (or a separate integration crate, `masonry_winit`) and then this issue would go away.

DJMcNab (approved):
Thanks.

This is really a tooling failure, although working out what rules in specific should be set is non-trivial.

PR #333, opened 2024-06-04
masonry: Use own `PointerButton` enum

waywardmonkeys:
This brings the `PointerButton` enum from glazier and has all code outside of the winit event loop integration using it.

For now, it has a `todo!()` for the `MouseButton::Other` as it isn't clear what that is for.

### DIFF
--- a/masonry/src/event.rs
+++ b/masonry/src/event.rs
@@ -11,7 +11,7 @@ use crate::WidgetId;
 use std::{collections::HashSet, path::PathBuf};
 
 use accesskit::{Action, ActionData};
-use winit::event::{Ime, KeyEvent, Modifiers, MouseButton};
+use winit::event::{Ime, KeyEvent, Modifiers};
 use winit::keyboard::ModifiersState;
 
 // TODO - Occluded(bool) event
@@ -30,14 +30,34 @@ pub enum WindowEvent {
     RebuildAccessTree,
 }
 
+/// An indicator of which pointer button was pressed.
+#[derive(PartialEq, Eq, Clone, Copy, Debug, Hash)]
+#[repr(u8)]
+pub enum PointerButton {
+    /// No mouse button.
+    None,
+    /// Primary button, commonly the left mouse button, touch contact, pen contact.
+    Primary,
+    /// Secondary button, commonly the right mouse button, pen barrel button.
+    Secondary,
+    /// Auxiliary button, commonly the middle mouse button.
+    Auxiliary,
+    /// X1 (back) Mouse.
+    X1,
+    /// X2 (forward) Mouse.
+    X2,
+    /// Other mouse button. This isn't fleshed out yet.
+    Other,
+}
+
 // TODO - How can RenderRoot express "I started a drag-and-drop op"?
 // TODO - Touchpad, Touch, AxisMotion
 // TODO - How to handle CursorEntered?
 // Note to self: Events like "pointerenter", "pointerleave" are handled differently at the Widget level. But that's weird because WidgetPod can distribute them. Need to think about this again.
 #[derive(Debug, Clone)]
 pub enum PointerEvent {
-    PointerDown(MouseButton, PointerState),
-    PointerUp(MouseButton, PointerState),
+    PointerDown(PointerButton, PointerState),
+    PointerUp(PointerButton, PointerState),
     PointerMove(PointerState),
     PointerEnter(PointerState),
     PointerLeave(PointerState),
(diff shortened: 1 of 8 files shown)


waywardmonkeys:
I think this is one step of many ...

* `buttons: HashSet<PointerButton>` should be a `buttons: PointerButtons` like in glazier, there's no need for a `HashSet`.
* There's a lot of mixed terminology with some things being `Pointer` and others being `Mouse`
* The `PointerButton` (and later `PointerButtons`) and other types for pointers could probably move into their own files to clean things up.
* Some of these types could, in the future, be in an inter-operability crate much like `keyboard-types`. (There is a `pointer-types` crate, but it doesn't seem to have a lot of use...)

masonry/src/event.rs
```diff
@@ -30,14 +30,33 @@ pub enum WindowEvent {
     RebuildAccessTree,
 }
 
+/// An indicator of which pointer button was pressed.
+#[derive(PartialEq, Eq, Clone, Copy, Debug, Hash)]
+#[repr(u8)]
+pub enum PointerButton {
+    /// No mouse button.
+    // MUST BE FIRST (== 0)
```
DJMcNab, line 38: This needs a "why" here - *I* know it's because of the future `PointerButtons` type, but since that doesn't presently exist, this is actually entirely unneeded

masonry/src/event_loop_runner.rs
```diff
@@ -13,16 +13,29 @@ use vello::{peniko::Color, AaSupport, RenderParams, Renderer, RendererOptions, S
...
 
+impl From<WinitMouseButton> for PointerButton {
+    fn from(button: WinitMouseButton) -> Self {
+        match button {
+            WinitMouseButton::Left => PointerButton::Primary,
+            WinitMouseButton::Right => PointerButton::Secondary,
+            WinitMouseButton::Middle => PointerButton::Auxiliary,
+            WinitMouseButton::Back => PointerButton::X1,
+            WinitMouseButton::Forward => PointerButton::X2,
+            WinitMouseButton::Other(_) => todo!(),
```
DJMcNab, line 34: I think this is for mice with more than 5 buttons. For example, my mouse has a DPI up/down button (although these specifically don't actually use this).
It looks like some users have seen this actually be used, e.g. https://github.com/rust-windowing/winit/issues/2562

I'd prefer to use a `tracing::warn!("...")` here (maybe a `tracing::debug!`?) rather than a full `panic`; this isn't an error condition, just one we're not sure how to expose.

DJMcNab (approved):
Overall reasonable. Not sure we should be panicking in the case you identified

masonry/src/event.rs
waywardmonkeys: I've removed this comment for now and can re-add it when `PointerButtons` happens.

masonry/src/event_loop_runner.rs
waywardmonkeys: I've added a `PointerButton::Other` for now (no value attached) and mapped to it here with a `tracing::warn!` emitted.

masonry/src/event.rs
```diff
@@ -30,14 +30,34 @@ pub enum WindowEvent {
...
+    Primary,
+    /// Secondary button, commonly the right mouse button, pen barrel button.
+    Secondary,
+    /// Auxiliary button, commonly the middle mouse button.
+    Auxiliary,
+    /// X1 (back) Mouse.
+    X1,
+    /// X2 (forward) Mouse.
+    X2,
+    /// Other mouse button. This isn't fleshed out yet.
```
DJMcNab, line 49: According to https://www.w3.org/TR/pointerevents3/#the-button-property, there's also the Pen Eraser button? I guess if winit doesn't expose that, it's not really that important

masonry/src/event.rs
```diff
@@ -30,14 +30,34 @@ pub enum WindowEvent {
...
+    /// Primary button, commonly the left mouse button, touch contact, pen contact.
+    Primary,
+    /// Secondary button, commonly the right mouse button, pen barrel button.
+    Secondary,
+    /// Auxiliary button, commonly the middle mouse button.
+    Auxiliary,
+    /// X1 (back) Mouse.
+    X1,
+    /// X2 (forward) Mouse.
+    X2,
```
DJMcNab, line 48: Not a huge fan of these names, but this is what the web spec uses, so 🤷

DJMcNab (approved):
This is ready. Thanks!

There are some aspects that I don't really like, but the choices we have made are also the correct ones given our current constraints.

masonry/src/event.rs
waywardmonkeys: Should we file a bug to follow up on this?

masonry/src/event.rs
waywardmonkeys: Same

masonry/src/event.rs
DJMcNab: Probably...

PR #334, opened 2024-06-04
masonry: More efficient set of PointerButtons.

waywardmonkeys:
This brings code over from glazier for a set of PointerButtons that just uses a single value and bit flags rather than a `HashSet`.

### DIFF
--- a/masonry/src/event.rs
+++ b/masonry/src/event.rs
@@ -8,7 +8,7 @@ use crate::kurbo::Rect;
 // TODO - See issue https://github.com/linebender/xilem/issues/367
 use crate::WidgetId;
 
-use std::{collections::HashSet, path::PathBuf};
+use std::path::PathBuf;
 
 use winit::event::{Force, Ime, KeyEvent, Modifiers};
 use winit::keyboard::ModifiersState;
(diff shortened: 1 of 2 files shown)


masonry/src/event.rs
```diff
@@ -50,6 +50,84 @@ pub enum PointerButton {
...
+    /// Count the number of pressed buttons in the set.
+    #[inline]
+    pub fn count(self) -> u32 {
+        self.0.count_ones()
+    }
+}
+
+impl std::fmt::Debug for PointerButtons {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "PointerButtons({:05b})", self.0)
```
DJMcNab, line 127: Can we display this in full? I.e. `Primary | Secondary | ...`

We can maybe implement [`std::fmt::Binary`](https://doc.rust-lang.org/std/fmt/trait.Binary.html) to use this display?

masonry/src/event.rs
```diff
@@ -50,6 +50,84 @@ pub enum PointerButton {
     Other,
 }
 
+/// A set of [`PointerButton`]s.
+#[derive(PartialEq, Eq, Clone, Copy, Default)]
+pub struct PointerButtons(u8);
```
DJMcNab, line 55: Can we have the same utils as from bitflags (such as `|`, etc.)

I know we don't want to have different public names for these versions of the values (which is right), so we can't really use bitflags directly.

In theory, I guess we could make `PointerButton | PointerButton` yield `PointerButtons`, but I think that's too cute.

DJMcNab, line 55: Is it worth implementing `From<PointerButton> for PointerButtons`?

masonry/src/event.rs
```diff
@@ -50,6 +50,84 @@ pub enum PointerButton {
     Other,
 }
 
+/// A set of [`PointerButton`]s.
```
DJMcNab, line 53: We should point to the web spec for their equivalent

DJMcNab (approved):
Some suggestions for added utilities, and a docs improvement.

masonry/src/event.rs
waywardmonkeys: We barely use `PointerButtons` right now (only on the `PointerState`).

If I did this, I'd want to move it to a separate file as it would start to grow too large to just have it in `event.rs` ... logically, we could do `masonry/src/event/pointer.rs` and then `pub use` the stuff from it into `masonry::event`. It would be easiest then if we left `src/event.rs` rather than renaming it to `src/event/mod.rs` ...

waywardmonkeys: Unsure again since we barely actually. use this.

masonry/src/event.rs
waywardmonkeys: Maybe! There's some interesting differences between what we do for button / buttons and what the mouse event and pointer events specs say.

https://www.w3.org/TR/pointerevents3/#button-states

waywardmonkeys: (I think we should aim to move more towards pointer events and a cleaner model over time.)

masonry/src/event.rs
waywardmonkeys: This also starts to get closer to where I'd rather have a new crate called `input-types` (name not taken yet) or something and go all out there since this isn't specific to Masonry (and I have another version of code like this in my own codebases...)

masonry/src/event.rs
```diff
@@ -49,6 +49,118 @@ pub enum PointerButton {
...
+
+    /// Returns `true` if the set is empty.
+    #[inline]
+    pub fn is_empty(self) -> bool {
+        self.0 == 0
+    }
+
+    /// Returns `true` if all the `buttons` are in the set.
+    #[inline]
+    pub fn is_superset(self, buttons: PointerButtons) -> bool {
```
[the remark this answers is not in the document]
waywardmonkeys: There's already a `contains` for  a single button ... I'll rename this to `contains_all` for now unless you have another suggestion to deal with both.

masonry/src/event.rs
```diff
@@ -49,6 +49,118 @@ pub enum PointerButton {
...
+
+fn button_bit(button: PointerButton) -> u8 {
+    match button {
+        PointerButton::None => 0,
+        PointerButton::Primary => 0b1,
+        PointerButton::Secondary => 0b10,
+        PointerButton::Auxiliary => 0b100,
+        PointerButton::X1 => 0b1000,
+        PointerButton::X2 => 0b10000,
+        // TODO: When we properly do `Other`, this changes
```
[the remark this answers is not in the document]
waywardmonkeys: I don't.

masonry/src/event.rs
waywardmonkeys: This is done now as well.

waywardmonkeys:
This is done now, I think ... I didn't add tests, but probably will in the future. But right now, this code is barely even used! (But definitely should be used more.)

DJMcNab (approved):

PR #381, opened 2024-06-10
deps: Update `wgpu`, `vello`, `peniko`

waywardmonkeys:
This brings us to the 0.2 release of Vello, which brings the 0.20 release of wgpu.


DJMcNab (approved):
Looks good. I'm seeing some slightly slow startup times on desktop, but it's much improved on Android. And it seems like this startup time issue is also present on main.

waywardmonkeys:
I feel like that changed recently. I don't recall the startup white flash on my machine before, but it is definitely there now and without these changes.

PR #382, opened 2024-06-10
deps: Update accesskit to 0.15, accesskit_winit to 0.21

waywardmonkeys:


DJMcNab (approved):
I don't have a setup to actually test accessibility, but I can confirm this still works on my machine and on Android.

waywardmonkeys:
I tested that the accessibility still works on macOS prior to submitting the PR.

PR #383, opened 2024-06-10
Add milliseconds in the tracing formatting

DJMcNab:
In #254, we had an initial log format added.

However, that doesn't include support for milliseconds, which are useful in a GUI system because it allows getting a first approximation of latency. I'm adding this now because of the startup latency issues first observed in #381 ([#masonry > Slight Startup Time Regression](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Slight.20Startup.20Time.20Regression))


waywardmonkeys (approved):

Philipp-M (approved):
Yeah I also prefer more detailed timing information, I don't think seconds is really enough

PoignardAzur:
LGTM.

As long as we're not going to nanosecond level, that's an acceptable level of verbosity.

PR #391, opened 2024-06-12
Consistently use `ctx` over `cx` in Xilem

DJMcNab:
This was brought up in [#xilem > paper cuts](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/paper.20cuts)

The memoize change was a copy-and-paste failure


PoignardAzur (approved):
LGTM.

I assume you did a project-wide grep for Cx?

DJMcNab:
Yeah - I ignored the instances in Masonry, e.g. `fcx`

I think there are a couple in Masonry which would make a reasonable follow-up, but those are potentially less obviously wins.

PR #396, opened 2024-06-13
Implement widgets-in-arenas RFC

PoignardAzur:
See https://github.com/linebender/rfcs/pull/6 and [zulip thread](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/New.20RFC.3A.20.22widgets-in-arena.22).

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -3,8 +3,6 @@
 
 //! The context types that are passed into various widget methods.
 
-#![allow(private_bounds)]
-
 use std::any::Any;
 use std::time::Duration;
 
@@ -119,12 +117,14 @@ pub struct WorkerCtx<'a> {
 
 pub struct WorkerFn(pub Box<dyn FnOnce(WorkerCtx) + Send + 'static>);
 
+// --- MARK: GETTERS ---
+// Methods for all context types
 impl_context_method!(
     WidgetCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
-    PaintCtx<'_>,
     LayoutCtx<'_>,
+    PaintCtx<'_>,
     AccessCtx<'_>,
     {
         /// get the `WidgetId` of the current widget.
@@ -145,6 +145,7 @@ impl_context_method!(
         }
 
         #[allow(dead_code)]
+        /// Helper method to get a direct reference to a child widget from its WidgetPod.
         fn get_child<Child: Widget>(&self, child: &'_ WidgetPod<Child>) -> &'_ Child {
             let (child, _child_token) = self
                 .widget_children
@@ -154,6 +155,7 @@ impl_context_method!(
         }
 
         #[allow(dead_code)]
+        /// Helper method to get a direct reference to a child widget's WidgetState from its WidgetPod.
         fn get_child_state<Child: Widget>(&self, child: &'_ WidgetPod<Child>) -> &'_ WidgetState {
             let (child_state, _child_state_token) = self
                 .widget_state_children
@@ -164,12 +166,17 @@ impl_context_method!(
     }
 );
 
+// Methods for all mutable context types
 impl_context_method!(
     WidgetCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
     {
+        /// Helper method to get a mutable reference to a child widget's WidgetState from its WidgetPod.
+        ///
+        /// This one isn't defined for PaintCtx and AccessCtx because those contexts
+        /// can't mutate WidgetState.
         fn get_child_state_mut<Child: Widget>(
             &mut self,
             child: &'_ mut WidgetPod<Child>,
@@ -183,12 +190,15 @@ impl_context_method!(
     }
 );
 
-// methods on everyone but layoutctx
+// --- MARK: GET LAYOUT ---
+// Methods on all context types except LayoutCtx
+// These methods access layout info calculated during the layout pass.
 impl_context_method!(
     WidgetCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     PaintCtx<'_>,
+    AccessCtx<'_>,
     {
         /// The layout size.
         ///
@@ -219,7 +229,19 @@ impl_context_method!(
         pub fn to_window(&self, widget_point: Point) -> Point {
             self.window_origin() + widget_point.to_vec2()
         }
+    }
+);
 
+// --- MARK: GET STATUS ---
+// Methods on all context types except LayoutCtx
+// Access status information (hot/active/disabled/etc).
+impl_context_method!(
+    WidgetCtx<'_>,
+    EventCtx<'_>,
+    LifeCycleCtx<'_>,
+    PaintCtx<'_>,
+    AccessCtx<'_>,
+    {
         /// The "hot" (aka hover) status of a widget.
         ///
         /// A widget is "hot" when the mouse is hovered over it. Widgets will
@@ -308,6 +330,8 @@ impl_context_method!(
     }
 );
 
+// --- MARK: CURSOR ---
+// Cursor-related impls.
 impl_context_method!(EventCtx<'_>, {
     /// Set the cursor icon.
     ///
@@ -352,8 +376,9 @@ impl_context_method!(EventCtx<'_>, {
     }
 });
 
+// --- MARK: WIDGET_MUT ---
+// Methods to get a child WidgetMut from a parent.
 impl<'a> WidgetCtx<'a> {
-    // FIXME - Assert that child's parent is self
     /// Return a [`WidgetMut`] to a child widget.
     pub fn get_mut<'c, Child: Widget>(
         &'c mut self,
@@ -381,10 +406,11 @@ impl<'a> WidgetCtx<'a> {
     }
 }
 
-// TODO - Remove
+// TODO - It's not clear whether EventCtx should be able to create a WidgetMut.
+// One of the examples currently uses that feature to change a child widget's color
+// in reaction to mouse events, but we might want to address that use-case differently.
 impl<'a> EventCtx<'a> {
     /// Return a [`WidgetMut`] to a child widget.
-    // FIXME - Assert that child's parent is self
     pub fn get_mut<'c, Child: Widget>(
         &'c mut self,
         child: &'c mut WidgetPod<Child>,
@@ -411,10 +437,9 @@ impl<'a> EventCtx<'a> {
     }
 }
 
-// TODO - Remove
+// TODO - It's not clear whether LifeCycleCtx should be able to create a WidgetMut.
 impl<'a> LifeCycleCtx<'a> {
     /// Return a [`WidgetMut`] to a child widget.
-    // FIXME - Assert that child's parent is self
     pub fn get_mut<'c, Child: Widget>(
         &'c mut self,
         child: &'c mut WidgetPod<Child>,
@@ -441,7 +466,7 @@ impl<'a> LifeCycleCtx<'a> {
     }
 }
 
-// methods on event and lifecycle
+// Methods on WidgetCtx, EventCtx, and LifeCycleCtx
 impl_context_method!(WidgetCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
     /// Request a [`paint`](crate::Widget::paint) pass.
     pub fn request_paint(&mut self) {
@@ -539,7 +564,7 @@ impl_context_method!(WidgetCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
     }
 });
 
-// methods on everyone but paintctx
+// Methods on all context types except PaintCtx and AccessCtx
 impl_context_method!(
     WidgetCtx<'_>,
     EventCtx<'_>,
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -18,6 +18,7 @@ use crate::promise::PromiseToken;
 use crate::render_root::{RenderRootSignal, RenderRootState};
 use crate::text2::TextBrush;
 use crate::text_helpers::{ImeChangeSignal, TextFieldRegistration};
+use crate::tree_arena::TreeArenaTokenMut;
 use crate::widget::{CursorChange, WidgetMut, WidgetState};
 use crate::{AllowRawMut, CursorIcon, Insets, Point, Rect, Size, Widget, WidgetId, WidgetPod};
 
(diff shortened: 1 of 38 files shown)


masonry/src/contexts.rs
```diff
@@ -704,3 +706,146 @@ impl AccessCtx<'_> {
...
+            widget: &child.inner,
+        }
+    }
+}
+
+// --- MARK: WRAPPER
+
+pub struct RawWrapper<'a, Ctx, W> {
+    pub(crate) ctx: Ctx,
+    pub(crate) widget: &'a W,
```
DJMcNab, line 799: I don't think these need to be `pub(crate)`, as the only constructors live in this file

masonry/src/contexts.rs
```diff
@@ -704,3 +706,146 @@ impl AccessCtx<'_> {
...
+}
+
+pub struct RawWrapperMut<'a, Ctx: Context, W> {
```
DJMcNab, line 802: The warning about `Context` being private is unfortunate. I think I'd prefer to make `Context` a sealed trait rather than a private one, as the resultant docs of `RawWrapperMut` are quite poor.

masonry/src/contexts.rs
```diff
@@ -704,3 +706,146 @@ impl AccessCtx<'_> {
...
+        self.parent_widget_state
+            .merge_up(self.ctx.get_widget_state());
+    }
+}
+
+trait Context {
+    fn get_widget_state(&mut self) -> &mut WidgetState;
+}
+
+macro_rules! impl_context_trait {
```
DJMcNab, line 839: I don't think this being a macro wins us much

masonry/src/lib.rs
DJMcNab, line 1: We also need to re-export the new contexts

masonry/src/widget/portal.rs
```diff
@@ -341,8 +354,11 @@ impl<W: Widget> Widget for Portal<W> {
             !self.constrain_vertical && portal_size.height < content_size.height;
 
         if self.scrollbar_horizontal_visible {
-            self.scrollbar_horizontal.widget_mut().portal_size = portal_size.width;
-            self.scrollbar_horizontal.widget_mut().content_size = content_size.width;
+            let mut scrollbar = ctx.get_raw_mut(&mut self.scrollbar_horizontal);
+            scrollbar.widget().portal_size = portal_size.width;
+            scrollbar.widget().content_size = content_size.width;
+            std::mem::drop(scrollbar);
```
DJMcNab, line 360: Do we need to request a paint on the scrollbar here?

masonry/src/widget/portal.rs
DJMcNab, line 1: I've not debugged this, but the `Portal` in the `to_do_list` example doesn't seem to be working. However, they also aren't working on main, so this isn't a new regression

masonry/src/contexts.rs
```diff
@@ -704,3 +706,146 @@ impl AccessCtx<'_> {
...
     }
 }
+
+macro_rules! impl_get_raw {
+    ($SomeCtx:tt) => {
+        impl<'s> $SomeCtx<'s> {
+            /// Get a child context and a raw shared reference to a child widget.
+            ///
+            /// The child context can be used to call context methods on behalf of the
+            /// child widget.
```
DJMcNab, line 716: I notice that in the current code, the context in `RawWrapper` is never used. If the child cannot have been mutated, how meaningful even are these context methods?

DJMcNab (commented):
It's really good to see progress towards this RFC being implemented.

This has sparked a few questions which need thinking about, in particular about the role of `RawWrapper`. I think it's good to have some sort of read-only access to the child's state, but what form that takes is not clear to me.

masonry/src/contexts.rs
PoignardAzur: They have a few getters. But yes, I mostly include the context types for symmetry between the RawWrapper and RawWrapperMut.

masonry/src/widget/widget_pod.rs
```diff
@@ -161,50 +161,58 @@ impl<W: Widget> WidgetPod<W> {
...
-        visit: impl FnOnce(&mut Self) -> Ret,
-    ) -> Ret {
-        if cfg!(not(debug_assertions)) {
-            return visit(self);
-        }
+        visit: impl FnOnce(&mut Self) -> bool,
+    ) {
+        let _span = self.inner.make_trace_span().entered();
+
+        // TODO #370 - Re-implement debug logger
```
DJMcNab, line 171: We should use full links to issues, as seen by our experience with Masonry

masonry/src/contexts.rs
PoignardAzur: I've done that in the last version, though I'm not super confident in my code.

masonry/src/widget/portal.rs
PoignardAzur: LayoutCtx currently doesn't have a request_paint method. I'm choosing to procrastinate on that.

masonry/src/render_root.rs
```diff
@@ -234,15 +290,38 @@ impl RenderRoot {
...
-    // --- MARK: POINTER_EVENT  ---
+    pub fn get_widget(&self, id: WidgetId) -> Option<WidgetRef<dyn Widget>> {
+        let (state, state_token) = self.widget_arena.widget_states.find(id.to_raw())?;
+        let (widget, widget_token) = self
+            .widget_arena
+            .widgets
+            .find(id.to_raw())
+            .expect("found state but not widget");
+
+        let widget = widget.as_any().downcast_ref::<Box<dyn Widget>>().unwrap();
```
DJMcNab, line 301: This is confusing to me - this seems to from `&Box<dyn Widget>` returned from the arena. Is the arena storing `Box<Box<dyn Widget>> as Box<dyn Widget>`?

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO - Factor out into a separate crate.
+#![allow(dead_code)]
```
DJMcNab, line 5:
```suggestion
// We intentionally don't make the types in this module part of our public API,
// but implement some useful methods which we don't use yet.
#![allow(dead_code)]
```

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+
+struct TreeNode<Item> {
+    id: u64,
+    item: Item,
+    children: Vec<TreeNode<Item>>,
+}
+
+#[derive(Default)]
+pub struct TreeArena<Item> {
```
DJMcNab, line 14: Every `pub` in this module should be `pub(crate)`

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+        *self
+    }
+}
+
+impl<'a, Item> Copy for TreeArenaToken<'a, Item> {}
+
+pub struct TreeArenaTokenMut<'a, Item> {
+    children: &'a mut Vec<TreeNode<Item>>,
+}
+
```
DJMcNab, line 33: We should have an explanatory comment about this *somewhere*.

i.e. that this is the safe version, and the unsafe version will look like this

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+impl<'a, Item> Clone for TreeArenaToken<'a, Item> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<'a, Item> Copy for TreeArenaToken<'a, Item> {}
+
+pub struct TreeArenaTokenMut<'a, Item> {
+    children: &'a mut Vec<TreeNode<Item>>,
```
DJMcNab, line 31: Do these need to be ordered?

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+                    TreeArenaToken {
+                        children: &child.children,
+                    },
+                ));
+            }
+        }
+        None
+    }
+
+    pub fn iter_children(&self) -> impl Iterator<Item = (&'_ Item, TreeArenaToken<'_, Item>)> {
```
DJMcNab, line 169: This won't be possible in the arena version - every place that uses this should be using `widget.children_ids()`, I think.

masonry/src/widget/mod.rs
```diff
@@ -52,7 +52,7 @@ pub use widget_state::WidgetState;
 
 pub use sized_box::BackgroundBrush;
 #[doc(hidden)]
-pub use widget::{Widget, WidgetId};
+pub use widget::{AllowRawMut, Widget, WidgetId};
```
DJMcNab, line 55: I don't think these should be re-exported here. Instead, we should make `widget::widget` be `pub(crate)`, then re-export from that module in the root.

That is morally equivalent to making this export item `pub(crate)`, although there's an inaccurate diagnostic if you actually try and do that.

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+    }
+
+    pub fn root_token_mut(&mut self) -> TreeArenaTokenMut<'_, Item> {
+        TreeArenaTokenMut {
+            children: &mut self.roots,
+        }
+    }
+
+    // TODO - Use list of parents
```
DJMcNab, line 69: I'm not sure what this is saying

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub fn find(&self, id: u64) -> Option<(&Item, TreeArenaToken<'_, Item>)> {
```
DJMcNab, line 70: This should have a comment saying either:
1) This is `O(n)` where `n` is the total number of items in the subtree; OR
2) This is currently as in (1), but will be `O(1)` in future.

masonry/src/tree_arena.rs
```diff
@@ -0,0 +1,297 @@
...
+impl<'a, Item> Copy for TreeArenaToken<'a, Item> {}
+
+pub struct TreeArenaTokenMut<'a, Item> {
+    children: &'a mut Vec<TreeNode<Item>>,
+}
+
+#[cfg(FALSE)]
+struct TreeArena<Item> {
+    items: HashMap<u64, UnsafeCell<Item>>,
+    parents: HashMap<u64, u64>,
```
DJMcNab, line 37: We is this not inline in `items`?

DJMcNab (commented):
This code is too dense for me to review it extremely carefully.

I've made sure to review the `tree_arena` code somewhat carefully. It shouldn't surprise you that the main issue there is lack of docs.

masonry/src/render_root.rs
PoignardAzur: > Is the arena storing `Box<Box<dyn Widget>> as Box<dyn Widget>`?

It is, and no, that's not ideal. Long term the plan would be to remove a lot of uses of `Box<dyn Widget>` and remove this downcast.

This probably deserves a comment in the code.

masonry/src/tree_arena.rs
PoignardAzur: You mean children? I guess not, especially if we don't allow iteration.

masonry/src/tree_arena.rs
PoignardAzur: Good catch. It might be possible depending on the implementation, but for simpler implementations it wouldn't be.

masonry/src/debug_logger.rs
```diff
@@ -199,7 +199,7 @@ impl DebugLogger {
                 "is_expecting_place_child_call",
                 w_state.is_expecting_place_child_call,
             ),
-            StateTree::new("is_new", w_state.is_new),
+            //StateTree::new("is_new", w_state.is_new),
```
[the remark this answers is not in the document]
PoignardAzur: Honestly, I'm of half a mind to remove the debug_logger code entirely. It's getting more and more vestigial as time goes on. I don't want to do it right now though.

masonry/src/render_root.rs
PoignardAzur: After some testing, I'm realizing I was confused about what some of the downcasts were doing, and which were necessary. I'm adjusting the code to focus less on the box-of-box aspect, but it's still going to be a bit confusing. I'm adding documentation too.

masonry/src/tree_arena.rs
PoignardAzur: For clarity, I guess. It's essentially pseudocode anyway.

masonry/src/tree_arena.rs
PoignardAzur: I get your point, but I think the "this will factored out" aspect is clearer with pub items.

masonry/src/tree_arena.rs
```diff
@@ -10,11 +19,25 @@ struct TreeNode<Item> {
     children: Vec<TreeNode<Item>>,
 }
 
+// TODO - Keep track of parent relationships, and use them to implement
```
DJMcNab, line 22: Ooh, would this be a global map of parents? I see

PoignardAzur:
I think the PR is ready to merge now. It would help me if someone could give it a quick once-over and approved it for merging.

We can probably fix remaining issues in follow-up PRs.

DJMcNab (approved):
My biggest concern is the `O(n)` find. However, since those are currently unused, it's fairly reasonable. We should definitely resolve that *before* pass order lands.

masonry/src/testing/helper_widgets.rs
```diff
@@ -335,8 +335,8 @@ impl Widget for ReplaceChild {
         self.child.accessibility(ctx);
     }
 
-    fn children(&self) -> SmallVec<[WidgetRef<'_, dyn Widget>; 16]> {
-        self.child.widget().children()
+    fn children_ids(&self) -> SmallVec<[WidgetId; 16]> {
+        todo!()
```
Philipp-M, line 339: Did you want to handle that `todo!` in this PR?

masonry/src/widget/widget_pod.rs
```diff
@@ -552,163 +528,198 @@ impl<W: Widget> WidgetPod<W> {
...
+        // Note: this code is the absolute worse and needs to die in a fire.
+        // We're basically implementing a system where RouteWidgetAdded is
+        // propagated to a bunch of widgets, and transformed into WidgetAdded,
+        // which is *also* propagated to children but we want to skip that case.
+        match event {
+            LifeCycle::WidgetAdded => {
+                return true;
+            }
+            _ => (),
+        }
```
Philipp-M, line 561:
```suggestion
        if let LifeCycle::WidgetAdded = event {
            return true;
        }
```

Philipp-M (commented):
I did not find time to dive deeper, so just a few nits.

masonry/src/testing/helper_widgets.rs
PoignardAzur: Not sure why I kept it. Will address in smaller PR.

PoignardAzur:
> My biggest concern is the O(n) find. However, since those are currently unused, it's fairly reasonable. We should definitely resolve that before pass order lands.

I'm not too worried about these kinds of holes with well-defined edges; if nothing else, because they make for good first contributions for newcomers.

PR #397, opened 2024-06-14
Differentiate pointer buttons in `Button` Widget and View

DJMcNab:
In most cases, you want a button which only actuates when the primary mouse button is pressed, so the easy case is still that. This is a short term hack, because e.g. the active state is still based on any button being pressed, not just those we are interested in.

That is, we probably need to represent a set of buttons we are interested in. However, this change minimally unblocks additional work with Xilem. In particular, see [#xilem > Minesweeper converted from Iced to Xilem](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Minesweeper.20converted.20from.20Iced.20to.20Xilem).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ rust.unexpected_cfgs = { level = "warn", check-cfg = [
     'cfg(FALSE)',
     'cfg(tarpaulin_include)',
 ] }
-rust.unreachable_pub = "warn"
 
 [workspace.dependencies]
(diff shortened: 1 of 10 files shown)


waywardmonkeys:
Sounds like this would (in subsequent iterations) benefit from my `PointerButtons` PR (#334)?

Just asking because if so, that's motivational and something I can look at more once I am well again.

DJMcNab:
Yeah, I think that's right.

jaredoconnell (approved):
I can confirm that this works for me on Mac OS.

PR #398, opened 2024-06-14
Create a `View` for `SizedBox`

DJMcNab:
image

I have chosen not to set the background and border properties, as those are moreso styling properties - in our current state, we seem to be much preferring only specifying layout values

### DIFF
--- a/masonry/src/widget/sized_box.rs
+++ b/masonry/src/widget/sized_box.rs
@@ -83,6 +83,18 @@ impl SizedBox {
         }
     }
 
+    /// Construct container with child in a pod, and both width and height not set.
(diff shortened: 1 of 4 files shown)


jaredoconnell (approved):
Tested. It works. Though I noticed a bug that is in both this PR and main: the button's active region is shifted relative to the button. At least it is on my scaled Mac desktop.

jaredoconnell:
> Hmm, is that in the `flex` example? Hopefully I'll be able to reproduce on Monday.

Yes, it is.

DJMcNab:
I've not been able to reproduce this on my machine. I don't have a Mac to test it with

PR #399, opened 2024-06-16
Update stable Rust version to 1.79

waywardmonkeys:


DJMcNab (approved):
I think we also should use this opportunity to update all dependencies - that is, not necessarily in this pr, but we use the rust release as the catalyst for that

waywardmonkeys:
I had my eye on the wgpu update (0.20.1), but wasn't sure there was something compelling to force / require that (unlike winit 0.30.1 which fixes an observed problem).

PR #401, opened 2024-06-16
Port `Adapt` from old `xilem_core`

Philipp-M:
Ports `Adapt` to new xilem_core.
It also adds `MapState` which was previously called `AdaptState` and a new view `MapAction` which is basically about how elm works.
`MapAction` and `Adapt` is shown in action in the added `elm` example, while `MapState` can be seen in action in the added `components` example.

### DIFF
--- a/xilem/examples/elm.rs
+++ b/xilem/examples/elm.rs
@@ -1,6 +1,10 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+//! Xilem supports several patterns for creating modular components.
+//! You can also emulate the elm architecture for a subset of your app.
+//! Though usually it's more idiomatic to update state directly within event callbacks, as seen in the `direct_counter` view.
+
 use masonry::widget::{CrossAxisAlignment, MainAxisAlignment};
 use winit::error::EventLoopError;
 use xilem::{
@@ -17,9 +21,9 @@ struct AppState {
 }
 
 // `map_state()` maps a subset of the state from the parent, such that views can be modularized by state
-fn map_state_view(count: &mut i32) -> impl WidgetView<i32> {
+fn direct_counter(count: &mut i32) -> impl WidgetView<i32> {
     flex((
-        label(format!("adapt state count: {}", count)),
+        label(format!("direct count: {count}")),
         button("+", |count| *count += 1),
         button("-", |count| *count -= 1),
     ))
@@ -32,9 +36,9 @@ enum CountMessage {
 
 // `map_action()` is basically how elm works, i.e. provide a message that the parent view has to handle to update the state.
 // In this case the parent adjusts the count that is given to this view according to the message
-fn map_action_view<T>(count: i32) -> impl WidgetView<T, CountMessage> {
+fn elm_counter<T>(count: i32) -> impl WidgetView<T, CountMessage> {
     flex((
-        label(format!("adapt state count: {}", count)),
+        label(format!("elm count: {count}")),
         button("+", |_| CountMessage::Increment),
         button("-", |_| CountMessage::Decrement),
     ))
@@ -48,10 +52,10 @@ enum AdaptMessage {
 
 // `adapt()` is the most flexible but also most verbose way to modularize the views by state and action,
 // This is basically a combination of the two ways above, but it also allows to change the `MessageResult` for the parent view
-fn adapt_view(count: i32) -> impl WidgetView<i32, AdaptMessage> {
+fn adapt_counter(count: i32) -> impl WidgetView<i32, AdaptMessage> {
     flex((
         flex((
-            label(format!("adapt count: {}", count)),
+            label(format!("adapt count: {count}")),
             button("+", |count| {
                 *count += 1;
                 AdaptMessage::Changed
@@ -74,18 +78,18 @@ fn adapt_view(count: i32) -> impl WidgetView<i32, AdaptMessage> {
 fn app_logic(state: &mut AppState) -> impl WidgetView<AppState> {
     flex((
         map_state(
-            map_state_view(&mut state.map_state_count),
+            direct_counter(&mut state.map_state_count),
             |state: &mut AppState| &mut state.map_state_count,
         ),
         map_action(
-            map_action_view(state.map_action_count),
+            elm_counter(state.map_action_count),
             |state: &mut AppState, message| match message {
                 CountMessage::Increment => state.map_action_count += 1,
                 CountMessage::Decrement => state.map_action_count -= 1,
             },
         ),
         adapt(
-            adapt_view(state.adapt_count),
+            adapt_counter(state.adapt_count),
             |state: &mut AppState, thunk| match thunk.call(&mut state.adapt_count) {
                 MessageResult::Action(AdaptMessage::Reset) => {
                     state.adapt_count = 0;
--- a/xilem/examples/elm.rs
+++ b/xilem/examples/elm.rs
@@ -1,0 +1,110 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use masonry::widget::{CrossAxisAlignment, MainAxisAlignment};
+use winit::error::EventLoopError;
+use xilem::{
+    view::{button, flex, label},
+    EventLoop, WidgetView, Xilem,
+};
+use xilem_core::{adapt, map_action, map_state, MessageResult};
+
+#[derive(Default)]
+struct AppState {
+    adapt_count: i32,
+    map_state_count: i32,
+    map_action_count: i32,
+}
+
+// `map_state()` maps a subset of the state from the parent, such that views can be modularized by state
+fn map_state_view(count: &mut i32) -> impl WidgetView<i32> {
+    flex((
+        label(format!("adapt state count: {}", count)),
+        button("+", |count| *count += 1),
+        button("-", |count| *count -= 1),
+    ))
+}
+
+enum CountMessage {
+    Increment,
+    Decrement,
+}
+
+// `map_action()` is basically how elm works, i.e. provide a message that the parent view has to handle to update the state.
+// In this case the parent adjusts the count that is given to this view according to the message
+fn map_action_view<T>(count: i32) -> impl WidgetView<T, CountMessage> {
+    flex((
+        label(format!("adapt state count: {}", count)),
+        button("+", |_| CountMessage::Increment),
+        button("-", |_| CountMessage::Decrement),
+    ))
+}
+
+enum AdaptMessage {
+    Changed,
+    Reset,
+    Nop,
+}
+
+// `adapt()` is the most flexible but also most verbose way to modularize the views by state and action,
+// This is basically a combination of the two ways above, but it also allows to change the `MessageResult` for the parent view
+fn adapt_view(count: i32) -> impl WidgetView<i32, AdaptMessage> {
+    flex((
+        flex((
+            label(format!("adapt count: {}", count)),
+            button("+", |count| {
+                *count += 1;
+                AdaptMessage::Changed
+            }),
+            button("-", |count| {
+                *count -= 1;
+                AdaptMessage::Changed
+            }),
+        )),
+        flex((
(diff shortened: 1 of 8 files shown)


xilem_core/src/views/adapt.rs
```diff
@@ -0,0 +1,186 @@
...
+///     .enumerate()
+///     .map(|(idx, todo)| {
+///         Adapt::new(
+///             move |data: &mut AppState, thunk| {
+///                 if let MessageResult::Action(action) = thunk.call(&mut data.todos[idx]) {
+///                     match action {
+///                         TodoAction::Delete => data.todos.remove(idx),
+///                     }
+///                 }
+///                 MessageResult::Nop
```
DJMcNab, line 55: I don't *think* `Nop` is right here if the `MessageResult` was `Action`. But, as I say in my comments on those types, the semantics have very much not been nailed down.

I suspect we might need a `map_action` method on `MessageResult` which does this properly?

xilem_core/src/views/adapt.rs
```diff
@@ -0,0 +1,186 @@
...
+use crate::{DynMessage, MessageResult, Mut, View, ViewId, ViewPathTracker};
+
+/// A view that wraps a child view and modifies the state that callbacks have access to.
+///
+/// # Examples
+///
+/// Suppose you have an outer type that looks like
+///
+/// ```ignore
+/// struct State {
```
DJMcNab, line 15: Please don't call this `State`; mixing and matching the name of the generic parameter and its value is a recipe for confusion

xilem_core/src/views/adapt.rs
```diff
@@ -0,0 +1,186 @@
...
+where
+    Context: ViewPathTracker,
+    ChildView: View<ChildState, ChildAction, Context>,
+    ProxyFn: Fn(
+        &mut ParentState,
+        AdaptThunk<ChildState, ChildAction, Context, ChildView>,
+    ) -> MessageResult<ParentAction>,
+{
+    /// Creates a new [`Adapt<ParentState, ParentAction, ChildState, ChildAction, Context, V, F>`].
+    pub fn new(f: ProxyFn, child: ChildView) -> Self {
```
DJMcNab, line 107: I suspect that this would be more idiomatic as `fn adapt()` at the top level

DJMcNab (commented):
This `thunk` concept is so hard to explain 😆

Some concerns on the documentation, but otherwise this looks worth having.

IMO, we need at least one instantion somewhere, whether in an example or in a test.

xilem_core/src/views/adapt.rs
Philipp-M: That resulted from copying the old source, but I agree. I'll rename it to `AppState` to stay consistent. We should probably generally stay consistent with one name for the `AppState` as well (as I've seen `AppData` sometimes).

xilem_core/src/views/adapt.rs
Philipp-M: Same as below, but I agree as well, a `map_action` is probably better and reduces boilerplate as well

Philipp-M:
> IMO, we need at least one instantion somewhere, whether in an example or in a test.

Yeah an example is certainly worth it, also because I don't think it's the easiest thing to grasp for newcomers. I guess I can add one for xilem itself? Or should that be in xilem_core?

xilem_core/src/views/adapt.rs
Philipp-M: > I suspect we might need a map_action method on MessageResult which does this properly?

btw. that already existed as `MessageResult::map` should I just port that?

Philipp-M: > I don't think `Nop` is right here if the `MessageResult` was Action

On a second look, I do think it is? I think anything else should probably be redirected to the parent view, but when the `Adapt` has handled everything in `MessageResult::Action`, it's a `Nop` for the parent view.

But yes semantics need to be nailed down at some point. 
E.g. if we want something like `MessageResult::RequestRebuild` but I think, unless we want to constrain `Action` with a trait bound (in case it has to be mapped and shouldn't just be a `Nop`), we probably want some "hardcoded" variants.

DJMcNab: In my mind, what `Nop` signifies is that this message did nothing; therefore, the View tree does not need to be rebuilt, because `State` wasn't modified

DJMcNab:
I think it would be good to have an example for Xilem itself.

I'd probably call it something like `elm`? Because really that's us emulating the elm architecture

xilem_core/src/views/adapt.rs
DJMcNab: I'd probably prefer these to be idiomatically named something more instructive. Like, maybe `TodoList` or `Tasks` in this case.

For example, in `xilem_minesweeper`, I'd be inclined to call it `Minesweeper` (or indeed `Game`, as it is now)

xilem_core/src/views/adapt.rs
Philipp-M: Ah, I see that bit of logic in xilem(_masonry) now, and can follow your thoughts a little bit more.
So previously in old (async) xilem, every message/event resulted in a full rebuild, only when an async `Wake` event was sent (that was given to the view via `Cx::waker()`), it listened to `MessageResult::RequestRebuild` to check whether the async view needs (another) rebuild.

I'm not sure how safe it is to not rebuild when `MessageResult` is `Nop`, I think we need to be sure that the user had no real access to the `AppState` in that case.

DJMcNab: It seems reasonable that a developer would want to filter down messages they are interested.

The best example for that is probably the `button_any_state` we currently have, where you would send a `Nop` reply for any button presses you aren't interested in (but yes, filtering by `PointerButtons` for this case is *better* anyway)

Philipp-M: Yeah good point, though I think it *can* result in "unexpected" bugs when the user gets more power over the rebuild process (which should not be too expensive anyway).

Probably a good point for discussion in office hours, I think Raph had some thoughts to that as well.

jaredoconnell:
Is it possible for you to add a runnable example at this stage? It'd help with testing.

xilem_core/src/views/adapt.rs
```diff
@@ -0,0 +1,189 @@
...
+    ChildAction: 'static,
+    ParentState: 'static,
+    ParentAction: 'static,
+    Context: ViewPathTracker + 'static,
+    ChildView: View<ChildState, ChildAction, Context>,
+    ProxyFn: Fn(
+            &mut ParentState,
+            AdaptThunk<ChildState, ChildAction, Context, ChildView>,
+        ) -> MessageResult<ParentAction>
+        + 'static,
```
Philipp-M, line 107: All those bounds add up a lot of noise (one reason why I wanted to get rid of the `'static` bound), but I think it's more correct to provide the same bounds here as for the `View` implementation to avoid misuse...

xilem_core/src/views/map_action.rs
```diff
@@ -0,0 +1,118 @@
+use core::marker::PhantomData;
+
+use crate::{DynMessage, Mut, View, ViewId, ViewPathTracker};
+
+/// A view that maps a child [`View<State,ChildAction,_>`] to [`View<State,ParentAction,_>`] while providing mutable access to `State` in the map function.
+///
+/// This is very similar to the Elm architecture, where the parent view can update state based on the action message from the child view.
+pub struct MapAction<
```
Philipp-M, line 8: Previously named `AdaptState` I'm still not sure about the naming, but I *think* `MapState` makes more sense?

xilem_core/src/views/map_state.rs
```diff
@@ -0,0 +1,104 @@
+use core::marker::PhantomData;
+
+use crate::{DynMessage, MessageResult, Mut, View, ViewId, ViewPathTracker};
+
+/// A view that "extracts" state from a [`View<ParentState,_,_>`] to [`View<ChildState,_,_>`].
+/// This allows modularization of views based on their state.
+pub struct MapState<ParentState, ChildState, V, F = fn(&mut ParentState) -> &mut ChildState> {
```
Philipp-M, line 7: Previously named `AdaptState` I'm still not sure about the naming, but I *think* `MapState` makes more sense?

jplatte: I was the one who added `AdaptState`, simply because `Adapt` was already a thing so it seemed like a fitting name. `MapState` seems fine as well.

xilem_core/src/views/map_state.rs
```diff
@@ -0,0 +1,107 @@
...
+///         button("+", |count| *count += 1),
+///         button("-", |count| *count -= 1),
+///     ))
+/// }
+///
+/// fn main() -> Result<(), EventLoopError> {
+///     Xilem::new(AppState::default(), |state| {
+///         map_state(count_view(state.count), |state: &mut AppState|  &mut state.count)
+///     })
+///     .run_windowed(EventLoop::with_user_event(), "Map action example".into())?;
```
DJMcNab, line 42: This API is almost certain to change; I can easily imagine this becoming outdated and not being caught - I'd probably just remove this whole `fn main`, and start from an "app_logic" function

xilem/examples/elm.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(Default)]
+struct AppState {
+    adapt_count: i32,
+    map_state_count: i32,
+    map_action_count: i32,
+}
+
+// `map_state()` maps a subset of the state from the parent, such that views can be modularized by state
+fn map_state_view(count: &mut i32) -> impl WidgetView<i32> {
```
DJMcNab, line 20: This would not be the idiomatic name for this view; maybe something like `direct_counter`?

xilem/examples/elm.rs
```diff
@@ -0,0 +1,110 @@
...
+enum CountMessage {
+    Increment,
+    Decrement,
+}
+
+// `map_action()` is basically how elm works, i.e. provide a message that the parent view has to handle to update the state.
+// In this case the parent adjusts the count that is given to this view according to the message
+fn map_action_view<T>(count: i32) -> impl WidgetView<T, CountMessage> {
+    flex((
+        label(format!("adapt state count: {}", count)),
```
DJMcNab, line 37: This string value is a bit weird. Also, is there a reason this doesn't use the `{count}` shorthand?

DJMcNab (approved):
One thing I would say is that emulating the elm architecture is not necessarily the most idiomatic approach to using Xilem. I think we should explain that, but also explain that it is possible.

On the other hand, I don't think that's a job for this PR.

A few details, then this can land

xilem/examples/elm.rs
Philipp-M: Whoops, you're right, probably an artifact resulted from copy/paste. I changed all the view names and their label (elm/direct/adapt counter). No there's no reason, I changed it.

Philipp-M:
> One thing I would say is that emulating the elm architecture is not necessarily the most idiomatic approach to using Xilem

Certainly not, though I could think that in an async context this could be useful, but that's a different story that needs more research.

> I think we should explain that, but also explain that it is possible.

Probably (and I've added a comment at the top of that example).
But on the other hand I don't think people will stumble over it and say "this is the better approach to handle state/messages in xilem" ^^. You likely need a good use-case for this.

Btw. does the name `elm` for the example still make sense? As there's also e.g. `map_state` in there.

DJMcNab:
I'd consider splitting into two examples: `components` and `elm`.

But I'm worried that `components` might be a very small example. Maybe we can just stick it onto `to_do_mvc`?

Philipp-M:
> Maybe we can just stick it onto `to_do_mvc`

Hmm I'm not sure where it would fit best without complicating things, an `adapt` similar as in the xilem_web todomvc (i.e. send a message to the parent view, that the task can be deleted) would probably make more sense there.
Or do you mean using `adapt` there instead of `map_state`?

Philipp-M:
> But I'm worried that `components` might be a very small example.

I don't think that's a bad thing, focus on one thing. A user can quickly see what it is about. I guess I'll do that `components` example instead, and leave `adapt` and `map_action` in the `elm` example

PR #402, opened 2024-06-17
xilem_core: Provide a way to add `View` implementations for external types

Philipp-M:
Since we're suffering from the orphan rule with the new xilem_core implementation, we cannot use `View` in downstream crates for external types such as a `WidgetView` implementation for `&'static str` in Xilem. 

This PR adds an `OrphanView` trait which the `Context` of the `View` has to implement for the type which has an implementation in xilem_core for this. The `View` impl for these types is basically a proxy for `OrphanView` (which in turn has the same method signature as `View`), so downstream crates can achieve the same as with the `View`.

### DIFF
--- a/xilem_core/src/views/orphan.rs
+++ b/xilem_core/src/views/orphan.rs
@@ -1,0 +1,189 @@
+// TODO document everything, possibly different naming
+#![allow(missing_docs)]
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{DynMessage, MessageResult, Mut, View, ViewElement, ViewId, ViewPathTracker};
+
+// TODO it would be nice to be able to decide between those two ways (mostly to avoid allocations, as otherwise `as_view` would be better I think)
+
+/// A way to implement `View` for foreign types
+pub trait AsOrphanView<T, State, Action>: ViewPathTracker + Sized {
+    type V: View<State, Action, Self>;
+    fn as_view(value: &T) -> Self::V;
+}
+
+pub trait OrphanView<T, State, Action>: ViewPathTracker + Sized {
+    type Element: ViewElement;
+    type ViewState;
+
+    fn build(view: &T, ctx: &mut Self) -> (Self::Element, Self::ViewState);
+    fn rebuild<'el>(
+        new: &T,
+        prev: &T,
+        view_state: &mut Self::ViewState,
+        ctx: &mut Self,
+        element: Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element>;
+
+    fn teardown(
+        view: &T,
+        view_state: &mut Self::ViewState,
+        ctx: &mut Self,
+        element: Mut<'_, Self::Element>,
+    );
+    fn message(
+        view: &T,
+        view_state: &mut Self::ViewState,
+        id_path: &[ViewId],
+        message: DynMessage,
+        app_state: &mut State,
+    ) -> MessageResult<Action>;
+}
+
+macro_rules! impl_orphan_view_for {
+    ($ty: ty) => {
+        impl<State, Action, Context> View<State, Action, Context> for $ty
+        where
+            Context: OrphanView<$ty, State, Action>,
+        {
+            type Element = Context::Element;
+
+            type ViewState = Context::ViewState;
+
+            fn build(&self, ctx: &mut Context) -> (Self::Element, Self::ViewState) {
+                Context::build(self, ctx)
+            }
+
+            fn rebuild<'el>(
+                &self,
+                prev: &Self,
+                view_state: &mut Self::ViewState,
+                ctx: &mut Context,
+                element: Mut<'el, Self::Element>,
+            ) -> Mut<'el, Self::Element> {
+                Context::rebuild(self, prev, view_state, ctx, element)
+            }
+
+            fn teardown(
+                &self,
+                view_state: &mut Self::ViewState,
+                ctx: &mut Context,
+                element: Mut<'_, Self::Element>,
+            ) {
+                Context::teardown(self, view_state, ctx, element);
+            }
+
+            fn message(
+                &self,
+                view_state: &mut Self::ViewState,
+                id_path: &[ViewId],
+                message: DynMessage,
+                app_state: &mut State,
+            ) -> MessageResult<Action> {
+                Context::message(self, view_state, id_path, message, app_state)
+            }
+        }
+    };
+}
+
+macro_rules! impl_as_orphan_view_for {
+    ($ty: ty) => {
+        impl<State, Action, Context> View<State, Action, Context> for $ty
+        where
+            Context: AsOrphanView<$ty, State, Action>,
+        {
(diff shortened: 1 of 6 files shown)


xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,189 @@
...
+                )
+            }
+        }
+    };
+}
+
+// string impls
+impl_as_orphan_view_for!(&'static str);
+#[cfg(feature = "std")]
+impl_orphan_view_for!(String);
```
DJMcNab, line 152:
```suggestion
impl_orphan_view_for!(alloc::string::String);
```

xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,189 @@
...
+#[cfg(feature = "std")]
+impl_as_orphan_view_for!(std::borrow::Cow<'static, str>);
```
DJMcNab, line 154:
```suggestion
impl_as_orphan_view_for!(alloc::borrow::Cow<'static, str>);
```

xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,189 @@
...
+// Why does the following not work, but the `Cow` impl does??
```
DJMcNab, line 155: I presume this would be because we already have an impl for `Arc<T>`, and the compiler (for some reason) can't reason that `str` is not `View`.

DJMcNab (commented):
It's not clear why you've added a std feature for this; none of the things behind it actually need it.

I'm also not convinced it's meaningful to have a `View` implementation for numbers. It's *cute*, but it's a bit weird imo.

xilem_core/src/views/orphan.rs
Philipp-M: Ahh, that makes sense, thanks for the hint.

Philipp-M:
> It's not clear why you've added a std feature for this; none of the things behind it actually need it.

I guess I wasn't thorough enough and thought, these require `std`, even better if that's not necessary, I've removed the `std` feature.

> I'm also not convinced it's meaningful to have a `View` implementation for numbers. It's *cute*, but it's a bit weird imo.

xilem_web actually already had support for numbers as `Text` node, apart from syntax-sugar and being slightly more optimized than using something like `format!()` or `num.to_string()`, I do think it can be useful in a more mathematical focused impl of `xilem_core`. I haven't looked into how viable xilem_core could be e.g. for that animatables experiment that I did in trui, but that's e.g. a use-case (i.e. animatable properties).

> shows two different ways (unfortunately they aren't possible for the same type at the same time)

So at this point I'm leaning towards the more traditional approach with providing an actual `View` impl (in the form of `OrphanView` in the context of this PR), so I'm thinking about removing `AsOrphanView` in favor of `OrphanView` as it allows more control with just little bit more syntactical overhead. Maybe we can get similar behavior like `as_view()` somehow else in a more general way (without extra marker in `View`, because I don't think that gives good DX TBH).

Philipp-M:
I've removed the `AsOrphanView`, cleaned up a little, and added a "test" (more to show how it's used, as the implementation is trivial). When there's no objection that would be ready to merge/review now. The kurbo feature is used in the xilem_web rewrite already.

xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,125 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{DynMessage, MessageResult, Mut, View, ViewElement, ViewId, ViewPathTracker};
+
+/// This trait provides a way to add [`View`] implementations for types that would be restricted otherwise by the orphan rules.
+/// Every type that can be supported with this trait, needs a concrete `View` implementation in xilem_core, possibly feature-gated.
```
DJMcNab, line 7:
```suggestion
/// Every type that can be supported with this trait, needs a concrete `View` implementation in `xilem_core`, possibly feature-gated.
```

xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    };
+}
+
+// string impls
+impl_orphan_view_for!(&'static str);
+impl_orphan_view_for!(alloc::string::String);
+impl_orphan_view_for!(alloc::borrow::Cow<'static, str>);
+
+// number impls
```
DJMcNab, line 95: I'm still not convinced that these are that meaningful, but I'm not going to block on that.

xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,125 @@
...
+                message: Message,
+                app_state: &mut State,
+            ) -> MessageResult<Action, Message> {
+                Context::orphan_message(self, view_state, id_path, message, app_state)
+            }
+        }
+    };
+}
+
+// string impls
```
DJMcNab, line 90:
```suggestion
// string impls - should be used for immutable strings which can be selected within and copied from
```
I.e. we should be very explicit that this maps to `Prose` and `Text`

xilem_core/src/views/orphan.rs
```diff
@@ -0,0 +1,125 @@
...
+impl_orphan_view_for!(i32);
+impl_orphan_view_for!(u32);
+impl_orphan_view_for!(i64);
+impl_orphan_view_for!(u64);
+impl_orphan_view_for!(u128);
+impl_orphan_view_for!(isize);
+impl_orphan_view_for!(usize);
+
+#[cfg(feature = "kurbo")]
+mod kurbo {
```
DJMcNab, line 111: We should explain why these are needed (i.e. for an svg-like sub-tree)

DJMcNab (approved):
The reason I really preferred the "AsView" style is that I think it's important that these views are *only* syntax sugar for an existing view, and are not the only way to access certain functionality. That is mostly because of extensibility, but also in terms of allowing people to make the meaning of their code clearer if needed, by migrating from the syntax sugar.

I do agree that for `String` specifically, that has issues around needing to clone the string for a short time, which is very unfortunate. But IMO, the answer there is to make non-`'static` views be supported.

That being said, this is *probably* a slightly more idiomatic actual implementation than the "AsView" style, and so I think we should have it this way. This is a really good trick to avoid the marker trait awkwardness the previous solution had (djmcnab#4)

xilem_core/src/views/orphan.rs
Philipp-M: Yeah apart from a little bit macro-expanded boilerplate (which I don't think should contribute too much to the compilation) doesn't hurt?
(In case we can also feature gate this, like the kurbo module I guess, would probably be interesting whether this meaningfully contributes to compilation time...)

I think it's cool to have something like (taken from xilem_web) `p(("The answer to life is ", 42))` and it should be cheap as view, since equality checks are trivial.

Philipp-M:
> "AsView" style is that I think it's important that these views are only syntax sugar for an existing view

Yep that's why I originally wanted to have both of these (and the full `View` equivalent to avoid allocs for more optimized views), but these unfortunately intersect...
I really hope we can find ways in general for something like this, because I think it's quite useful to avoid a full `View` impl. especially for end users not being familiar with all the specifics of the `View` trait.

Thanks for the review and the initial impl in https://github.com/DJMcNab/xilem/pull/4 that inspired this.

PR #403, opened 2024-06-18
Rewrite xilem_web to support new xilem_core

### LINKED ISSUE #274: `xilem_web`: Panic using `BoxedView` inside HTML element
When trying to use a `BoxedView` in an `html::div` I get the following panic:

```
panicked at C:\Users\matth\.cargo\git\checkouts\xilem-031d790027f85a11\1e37164\crates\xilem_web\src\context.rs:304:14:
Element type has changed, this should never happen!
```

Reproducible by changing the counter example:
```rust
fn btn(
    label: &'static str,
    click_fn: impl Fn(&mut AppState, web_sys::MouseEvent) + 'static,
) ->  BoxedView<AppState> {
    Box::new(el::button(label).on_click(click_fn))
}

fn app_logic(state: &mut AppState) -> impl View<AppState> {
    el::div((
        btn("+1 click", |state, _| state.increment()),
        state.text.clone(),
    ))
}
```

<details>
  <summary>Stack trace</summary>

  ```
  Stack:
  
  Error
      at imports.wbg.__wbg_new_abda76e883ba8a5f (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9.js:338:21)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.console_error_panic_hook::Error::new::h516a79c7e649e619 (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[2255]:0xa99ca)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.console_error_panic_hook::hook_impl::h5e4d6d7771b2c70d (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[432]:0x55f76)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.console_error_panic_hook::hook::h99fffc916c6dad72 (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[3139]:0xb8210)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.core::ops::function::Fn::call::h97eb8c787f0b3f7f (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[2707]:0xb2178)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.std::panicking::rust_panic_with_hook::h0833ff9dc94df5f1 (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[1002]:0x7eb17)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.std::panicking::begin_panic_handler::{{closure}}::hc3329d7ae0363c6f (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[1373]:0x90226)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.std::sys_common::backtrace::__rust_end_short_backtrace::h4268105044501476 (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[3596]:0xbc275)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.rust_begin_unwind (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[2830]:0xb4058)
      at lookbook_xilem_example-f6a974525f70f53f.wasm.core::panicking::panic_fmt::h765acaeb0182f5af (http://localhost:8080/lookbook-xilem-example-57aa3af25ad35cc9_bg.wasm:wasm-function[2970]:0xb5f83)
  
  
  imports.wbg.__wbg_error_f851667af71bcfc6 @ lookbook-xilem-example-57aa3af25ad35cc9.js:332
  ```
</details>

This looks to be related to `Pod::downcast_mut`, where the underlying element type is incorrect.
I'm also not using any conditionals or a `ViewSequence` so this is definitely confusing to me, maybe `AnyView` is storing the wrong type?

Philipp-M:
This ports xilem_web to the new xilem_core.

There's also a lot of cleanup internally:
* Get rid of all of the complex macros to support DOM interfaces, and instead use associated type bounds on the `View::Element`.
* Introduce an extendable modifier based system, which should also work on top of memoization (`Arc`, `Memoize`) and `OneOf`  views with an intersection of the modifiable properties.
    * This modifier based system gets rid of the hacky way to propagate attributes to elements, and takes inspiration by masonrys `WidgetMut` type to apply changes.
    * Currently that means `Attributes`, `Classes`  and `Styles` to reflect what xilem_web previously offered.

Downsides (currently, needs some investigation):

~~Due to more internal type complexity via associated types this suffers from https://github.com/rust-lang/rust/issues/105900. The new trait solver should hopefully mitigate some of that, but it seems currently it completely stalls in the todomvc example (not in other examples).~~
~~The deep, possibly completely static composition via associated type-bounds of the view and element tree unfortunately can take some time to compile, this gets (already) obvious in the todomvc example. The other examples don't seem to suffer that bad yet from that issue, probably because they're quite simple.~~

~~I really hope we can mitigate this mostly, because I think this is the idiomatic (and more correct) way to implement what the previous API has offered.~~

One idea is to add a `Box<dyn AnyViewSequence>`, as every element takes a "type-erased" `ViewSequence` as parameter, so this may solve most of the issues (at the slight cost of dynamic dispatch/allocations).

Edit: idea was mostly successful, see comment right below.

I think it also closes #274

It's a draft, as there's a lot of changes in xilem_core that should be upstreamed (and cleaned up) via separate PRs and I would like to (mostly) fix the slow-compile time issue.


Philipp-M:
> but it seems currently it completely stalls in the todomvc example

Just to clarify, as I think there was already some misunderstanding, this is only true for the new (unfinished) trait-solver, not for the current one. The todomvc example takes without the new changes explained below, about 5s to compile in release mode vs 0.75s with current xilem_web. It'll be interesting to see how the new trait-solver will handle this, when it's finished, maybe we can convert it back to being completely static/stack-based again then (wishful thinking...).

So indeed the idea of type-erasing the `ViewSequence` mostly fixes the compile time issue, which makes sense, since it chops down the otherwise completely static composition of the view tree at each element boundary, since every DOM element takes a `ViewSequence` as parameter. But as expected this results in regression of either the performance (very slightly, less concerning than binary size at least) as well as binary size unfortunately. Compared to old xilem_web this binary size change is about 5-10% bigger (the lower end when it's compressed). Though I noticed, the bigger the app gets, the less of an issue this is (but we still need evidence of really big apps...) 

I do think it's worth it to have this kind of regression, with that we are roughly at the same compilation times as before and IMO the other advantages (disregarding the actual port to new xilem_core) outweigh the slight regression. I also think that it (the compile times) should scale this way, as the static type composition will not grow "infinitely". With incremental compilation, probably also big apps will compile somewhat fast. But that needs to be proven with bigger apps than the todomvc example... 

I've also started documenting xilem_web. So every item directly visible at in docs.rs should now at least hint what's going on.

xilem_web/src/interfaces.rs
```diff
@@ -204,449 +232,2100 @@ where
...
-                };
-            }
-            #[allow(unused_imports)]
-            pub(crate) use [<for_all_ $interface:snake _descendents>];
-        }
+// #[cfg(feature = "HtmlAnchorElement")]
+pub trait HtmlAnchorElement<State, Action = ()>:
+    HtmlElement<State, Action, DomNode: AsRef<web_sys::HtmlAnchorElement>>
+{
+}
```
Philipp-M, line 247: This whole file is mostly generated (with a local hacky webidl based generator).
I have avoided macros because I expect this to be continuously extended overtime, and as most of us, I'm not a fan of macros, as I think DX suffers.
I've added all these commented out feature gates because I think, when some of the element props are extended (e.g. a video element has then `VideoElementProps` which composes `ElementProps` + additional stuff relevant to the video element), that inevitably the wasm binary size will grow, if we don't constrain this with features.
Because I have not yet found a way that is reasonable and robust at least to do this just with static typing, such that these additional props are optimized away when they're not needed.
I don't think it's a good idea to let all the users suffer when they don't need such features.
Currently this is not necessary because as seen below, all is basically depending on the `ElementProps` and I think the `Element` interface/trait should be included by default.

xilem_web/src/interfaces.rs
```diff
@@ -204,449 +232,2100 @@ where
...
+pub trait HtmlSelectElement<State, Action = ()>:
+    HtmlElement<State, Action, DomNode: AsRef<web_sys::HtmlSelectElement>>
+{
+    /// A string representing the value of the HTMLOptionElement, i.e. the value attribute of the equivalent `<option>`.
+    /// If this is not specified, the value of text is used as the value, e.g. for the associated `<select>` element's value when the form is submitted to the server.
+    ///
+    /// See <https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/value> for more details
+    fn value(self, value: impl IntoAttributeValue) -> Attr<Self, State, Action> {
+        Attr::new(self, Cow::from("value"), value.into_attr_value())
+    }
```
Philipp-M, line 1038: My vision of the API is obviously builder/modifier-pattern like as seen with all these DOM interface extension traits, and the examples.
So I added all these attributes (seen in the `blog` example, which doesn't use any `Element::attr`) as methods, currently just sugaring the `Attr` view.
These are of course not complete, I only added the attributes needed in the `blog` example, and added MDN docs on top of some.
I expect this to be extended over time (together with the feature gates explained above).

xilem_web/src/attribute.rs
```diff
@@ -1,55 +1,302 @@
...
+            }
+            None => {
+                self.attribute_modifiers
+                    .push(AttributeModifier::EndMarker(self.start_idx));
+            }
+        }
+        self.idx += 1;
+        self.start_idx = self.idx;
+    }
+}
```
Philipp-M, line 175: This is one of the more interesting parts of this PR, a modifier based system (`Styles`, `Classes` and this `Attributes`), which is designed in such a way, that `Memoize` or `Arc` doesn't offer weird surprises (as well as just overwriting attributes, when they're previously set, which would have problems, when just directly setting attributes on the DOM elements).

As well as `OneOf` views being able to intersect on the elements and their props in a strong/statically-typed way. For example when there is `OneOf2::A(html::div(()))` and `OneOf2::B(mathml::mrow(())`, this `OneOf2` view fulfills the contract of the `Element` trait, but not of the `HtmlElement`, because the trait bound `AsRef<HtmlElement>` is not satisfied for `mathml::mrow`, so e.g. `HtmlElement::style` is not possible to use (currently at least, because MathML *should* be able to support style, this needs extension in wasm_bindgen I think.), but all the methods of `Element` are possible on `mathml::mrow` as well.

The modifier system works in such a way, that it has stateful `start` and `end` methods, which add markers into a `Vec` for the modifier type on its props to know where the modifier view has started/ended.
Additional modifiers for this property (e.g. `AttributeModifier::Set` or `AttributeModifier::Remove`) are then pushed into this `Vec` until the next modifier starts. 
All of this happens in either `View::build` as well as `View::rebuild`, and these attributes are added on every reconciliation (i.e. `View::rebuild`), so that it can detect changes (so all these attributes should be added always in the same order, to avoid unnecessary updates).
On each reconciliation pass, it is checked, whether the attribute at the same index has changed, and if so, the relevant props or attributes are marked dirty, such that they're correctly updated when applying the props to the actual DOM element.
This should be robust to underlying element changes in e.g. a `OneOf` which would build the new view, while being in a `View::rebuild` of a composing modifier view, as the actual modifiers have to be applied on the up-traversal and only the `end` method actually tells via an `EndMarker` where the modifier has started (which is read by the `start` method on the down traversal, to jump to the next modifier).

All the possible accumulated changes are then applied when `SuperElement::upcast`ing the element, which usually happens on a `ViewSequence` boundary.
These modifiers are described via traits such as (currently), `WithAttributes`, `WithClasses` and `WithStyles` and are also working with the intersection of `OneOf` types.

Philipp-M (commented):
So I think this is now in a good state for reviewing, I added more documentation (including simple doc tests in the interfaces, to test whether doc is up-to-date), it should now cover the more important parts of the architecture/API.
I have also added a new example (`blog`), which shows at least one thing that wasn't possible before (extending memoized elements).
The last things this is depending on is #402 and the new `blog` example depends also on #175, so after these are merged, this would be ready to merge as well (I'll do a merge from main into this branch before though).

As this is quite a big PR, and I doubt someone has the time or will to dive through all of this in detail while reviewing,
I'll give some hints via code-comments to the more interesting, and possibly discussion worthy things of it, as I still think this may be a reasonable direction for Xilem itself as well.

I guess, when there's use for it, I could also add some modified version of the comments as document such as `ARCHITECTURE.md` or something.

xilem_web/web_examples/blog/src/main.rs
DJMcNab, line 1: I've not done much review, but the name of this example is concerning. If you're doing a blog, you really should be doing server side rendering, imo.

I'm not sure though, I think it's good to get another sizeable example.

Philipp-M: I'm open for alternative suggestions, could also be something like "news" or what I'm probably planning is a simple hackernews implementation. I actually want to revisit the hydration support I've experimented with quite a long time ago. So I think in the future this likely will be extended with SSR (I think it's reasonable to support SSR in xilem_web isn't it?).

DJMcNab: Yeah, SSR and hydration in Xilem Web would be *incredible* and is something we should support sooner rather than later.

DJMcNab: I think reframing it in terms of e.g. a "ticker", so having the concept that the feed will be live-updating relatively frequently might be good enough. I don't want anyone getting the idea that we want people to be using Xilem Web as-is for a blog.

Philipp-M: Yeah you're probably right, a blog is not the most interactive example and should mostly be SSR, at least short term this would be misleading. I'll update that example as you said, to something that is more interactive.

Philipp-M: I was too lazy to add another example, removed it for now, I guess this somewhat "fixes" the issue. I may revisit it someday later though.

Philipp-M:
All of the dependent PRs are merged, it got another minor cleanup round, I think it's ready to merge now.

xilem_web/README.md
```diff
@@ -1,9 +1,32 @@
-# `xilem_web` prototype
+# `xilem_web`
```
DJMcNab, line 1: We could plausibly follow the usual readme patterns here, but that probably isn't a job for this pr

xilem_web/README.md
```diff
@@ -1,9 +1,32 @@
...
 
-This is an early prototype of a potential implementation of the Xilem architecture using DOM elements
-as Xilem elements (unfortunately the two concepts have the same name). This uses xilem_web_core under the hood,
-which is a legacy version of xilem_core.
+This is a prototype of a potential implementation of the Xilem architecture (implementing `xilem_core`) using DOM elements as Xilem elements (unfortunately the two concepts have the same name).
```
DJMcNab, line 3:
```suggestion
This is a prototype implementation of the Xilem architecture (through `xilem_core`) using DOM elements as Xilem elements (unfortunately the two concepts have the same name).
```

xilem_web/src/app.rs
```diff
@@ -1,44 +1,48 @@
...
-use crate::{
-    context::Cx,
-    view::{DomNode, View},
-    Message,
-};
-use xilem_core::{Id, MessageResult};
+use crate::{DomView, DynMessage, PodMut};
+use xilem_core::{MessageResult, ViewId};
+
+type IdPath = Vec<ViewId>;
```
DJMcNab, line 10: Have you considered `Arc<[ViewId]>` for this?

xilem_web/src/app.rs
```diff
@@ -1,44 +1,48 @@
...
+
+pub(crate) struct AppMessage {
+    pub id_path: IdPath,
+    pub body: DynMessage,
+}
 
 /// The type responsible for running your app.
-pub struct App<T, V: View<T>, F: FnMut(&mut T) -> V>(Rc<RefCell<AppInner<T, V, F>>>);
+pub struct App<T, V: DomView<T>, F: FnMut(&mut T) -> V>(Rc<RefCell<AppInner<T, V, F>>>);
```
DJMcNab, line 18: I'd obviously much prefer this avoided interior mutability, but I'll assume it's necessary - I don't have enough context to know whether it's avoidable

xilem_web/src/message.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{any::Any, fmt::Debug, ops::Deref};
+
+/// A dynamically typed message for the [`View`] trait.
+///
+/// Mostly equivalent to `Box<dyn Any>`, but with support for debug printing.
+// We can't use intra-doc links here because of
```
DJMcNab, line 9: I guess I should fix this sentence fragment in both places...

xilem_web/src/message.rs
```diff
@@ -0,0 +1,79 @@
...
+/// The primary interface for this type is [`dyn Message::downcast`](trait.Message.html#method.downcast).
+///
+/// These messages must also be [`Send`].
+/// This makes using this message type in a multithreaded context easier.
+/// If this requirement is causing you issues, feel free to open an issue
```
DJMcNab, line 14: And we should remove this section in both places too

DJMcNab (approved):
I have not reviewed any of the actual logic here, but found a few things on a very quick skim

xilem_web/README.md
Philipp-M: You're right, I've copied a lot of the Xilem Core Readme and actually used the same pattern to include it in the `lib.rs` now, you probably want to look quickly at it again before I merge.

xilem_web/src/app.rs
Philipp-M: I haven't but it definitely makes sense and was actually a simple change (with `Rc` though, because `Arc` isn't necessary (yet)). Thanks for the hint.

xilem_web/src/app.rs
Philipp-M: That's actually one of the few things from the old code-base that I haven't paid too much attention to, nor written it myself. But I do think it's necessary, since the `App` itself is "forgotten" to avoid blocking the event-loop, and just lives on within the view context (and message thunks).

xilem_web/src/message.rs
Philipp-M: Yeah, I'm not really sure what you mean with it^^ (I suspect it has something to do with `dyn`...)

xilem_web/src/message.rs
Philipp-M: Done.

xilem_core/src/message.rs
```diff
@@ -46,13 +46,6 @@ impl<A, Message> MessageResult<A, Message> {
 // We can't use intra-doc links here because of
```
DJMcNab, line 46:
```suggestion
// We can't use intra-doc links here because rustdoc doesn't understand impls on `dyn Message`
```

xilem_core/src/message.rs
```diff
@@ -46,13 +46,6 @@ impl<A, Message> MessageResult<A, Message> {
...
 /// The primary interface for this type is [`dyn Message::downcast`](trait.Message.html#method.downcast).
 ///
-/// These messages must also be [`Send`].
```
DJMcNab, line 49: This is still true in `xilem_core`.

xilem_web/README.md
```diff
@@ -1,17 +1,43 @@
...
+</div>
+
+<div align="center">
+
+**Experimental implementation of the Xilem architecture for the Web**
+
+[![Latest published version.](https://img.shields.io/crates/v/xilem_web.svg)](https://crates.io/crates/xilem_web)
+[![Documentation build status.](https://img.shields.io/docsrs/xilem_web.svg)](https://docs.rs/xilem_web)
+[![Apache 2.0 license.](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](#license)
+
```
DJMcNab, line 18:
```suggestion
\
```
This makes the badge lines become closer together (a line break rather than a paragraph break)

xilem_web/src/message.rs
DJMcNab: Yeah; it's the `dyn Message`. Rustdoc doesn't seem to have a syntax to create an intra-doc-link to this item

xilem_web/src/lib.rs
```diff
@@ -1,6 +1,23 @@
...
+[LICENSE]: https://github.com/linebender/xilem/blob/main/xilem_web/LICENSE
+
+<!-- intra-doc-links go here -->
+
+<style>
+.rustdoc-hidden { display: none; }
+</style>
+
+<!-- Hide the header section of the README when using rustdoc -->
+<div style=\"display:none\">
```
DJMcNab, line 16: This format of the README is out-of-date as compared to https://github.com/linebender/linebender.github.io/pull/55

But I'll update both of them later, so this is still fine

DJMcNab (approved):
A few tweaks on the comment feedback, but otherwise we should be good to land!

xilem_web/README.md
DJMcNab: I'll update both once this is merged. Thanks!

PR #405, opened 2024-06-18
Move to a single global `parley::LayoutContext`

DJMcNab:
Based on a suggestion from @dfrg.

This appears to have improved performance with the 30_000 hello's example from ~$\frac{1}{20}$ms to $\frac{1}{160}$ms
So an approx 8x increase in throughput for that scene.


jaredoconnell (approved):
I don't see any red flags, and the examples work fine.

dfrg:
> So an approx 8x increase for that scene.

I’ll take it :)

LGTM

Issue #44, opened 2023-02-16
Code Quality CI Checks


nicoburns:
Inspired by [this comment](https://github.com/linebender/xilem/pull/40#pullrequestreview-1300367479), I thought it might be a good idea to add more code quality checks to Xilem's CI. I am thinking we should have:

- Tests `cargo test`
- Clippy `cargo clippy --workspace -- -D warnings`
- Rustfmt `cargo fmt --all -- --check`
- Documentation `RUSTDOCFLAGS="-D warnings" cargo doc`

Additionally we may want to enforce documentation requirements more strongly. The following

```rust
#![warn(missing_docs)]
#![warn(clippy::missing_docs_in_private_items)]
```

will enforce doc comments on all items. Which seems to me a sensible way to go about things if we want to have an approachable codebase that is welcoming to newcomers.

We might also want to consider:

```rust
#![deny(unsafe_code)]
#![forbid(unsafe_code)]
```
To completely ban `unsafe`. I'm not sure if that's appropriate for Xilem, but I checked it doesn't *currently* have any `unsafe`. So perhaps it would be sensible to include that at least until someone makes a case that they need `unsafe`?

derekdreery:
You can always `#![warn(unsafe_code)]` if you want to be notified when it's added.

xStrom:
Yes, I think this should all happen. So for CI I think we can go far by adopting [the Druid CI config](https://github.com/linebender/druid/blob/master/.github/workflows/ci.yml) which is quite battle tested and has the tasks you listed. Of course it will still need some modifications, mostly deletions.

Banning `unsafe` seems like a good idea to me. We did that for `druid` and I don't think we ever ran into this as a problem.

Of course I'm also all for enforcing documentation. The private items one is a slight question though, as in does this apply to struct fields? Need to test it I guess, but that might be a bit heavyweight. Public API stuff should definitely all 100% be documented though.

nicoburns:
> The private items one is a slight question though, as in does this apply to struct fields?

Yes it does. And enum variants.

> that might be a bit heavyweight. Public API stuff should definitely all 100% be documented though.

I definitely think private items should also be documented for the most part. The way I see it, the purpose of documentation on private items is different the purpose of documentation of public items (it's aimed at developers reading the code not end users), but it's equally valuable.

[...]

PR #406, opened 2024-06-18
Deny unsafe code globally

DJMcNab:
Related to #44

We don't have any necessary unsafe code, and I don't think any of us want to lose that.

(Vello is also no longer laundering the unsafety wrt surface creation thanks to a recent wgpu release)


PoignardAzur (approved):
LGTM

PR #407, opened 2024-06-21
Move a log line to avoid false positives and reduce noise

PoignardAzur:


DJMcNab (approved):
This should definitely help

PR #408, opened 2024-06-21
xilem_core: Support generic `Message`s

Philipp-M:
This adds yet another generic parameter `Message` to the `View`, `ViewSequence` and `AnyView` trait, such that implementors have more freedom about what the message is. This is necessary for xilem_web, as a `Send` bound is not possible to use with wasm_bindgen IIRC. (At least I had to remove it in the `DynMessage` to get it to compile...).

It was fortunately straight-forward to just add the `Message` param with the default `DynMessage`. Basically search replace... (I hope I haven't missed anything, but I went through it twice...)

### DIFF
--- a/xilem_core/src/any_view.rs
+++ b/xilem_core/src/any_view.rs
@@ -7,7 +7,9 @@ use core::any::Any;
 
 use alloc::boxed::Box;
 
-use crate::{AnyElement, DynMessage, MessageResult, Mut, View, ViewId, ViewPathTracker};
+use crate::{
+    AnyElement, DynMessage, MessageResult, Mut, View, ViewElement, ViewId, ViewPathTracker,
+};
 
 /// A view which can have any view type where the [`View::Element`] is compatible with
 /// `Element`.
@@ -24,7 +26,7 @@ use crate::{AnyElement, DynMessage, MessageResult, Mut, View, ViewId, ViewPathTr
 ///
 /// Libraries using `xilem_core` are expected to have a type alias for their own `AnyView`, which specifies
 /// the `Context` and `Element` types.
-pub trait AnyView<State, Action, Context, Element: crate::ViewElement> {
+pub trait AnyView<State, Action, Context, Element: ViewElement, Message = DynMessage> {
     /// Get an [`Any`] reference to `self`.
     fn as_any(&self) -> &dyn Any;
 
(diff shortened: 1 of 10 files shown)


xilem_core/src/views/adapt.rs
```diff
@@ -89,21 +99,30 @@ where
...
+pub fn adapt<
+    ParentState,
+    ParentAction,
+    ChildState,
+    ChildAction,
+    Context,
+    ChildView,
+    Message,
+    ProxyFn,
+>(
```
Philipp-M, line 111: Generic params are getting a little bit out of hand, but I guess this is what we signed up for, with a generic xilem_core...

jplatte: Could some of these possibly be "bundled" together under a marker type that implements a trait that only consists of associated types?

DJMcNab (approved):
So obviously this does have some consequences for the ergonomics of Xilem Web, but makes Xilem "proper" more ergonomic.

I think that's the right choice, but we should document that in the README of xilem core.
i.e. that it supports being used outside of Xilem, but that the crate prioritises the ergonomics for users of Xilem.

xilem_core/src/views/adapt.rs
Philipp-M: Sounds interesting, but I'm not exactly sure what you mean, could you show an example (or when you have something more concrete in mind, PR)?

xilem_core/src/any_view.rs
```diff
@@ -206,20 +212,23 @@ where
...
-        message: crate::DynMessage,
+        message: Message,
         app_state: &mut State,
-    ) -> crate::MessageResult<Action> {
+    ) -> crate::MessageResult<Action, Message> {
         self.dyn_message(view_state, id_path, message, app_state)
     }
 }
 
 // TODO: IWBN if we could avoid this
```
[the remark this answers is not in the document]
Philipp-M: I think you mean the duplication of all the variations with `Send` and `Sync` for `dyn AnyView`. A different way to do this would be with our beloved macros I guess, but I think it's the right call to not do this with 4 repetitions...

Philipp-M:
> So obviously this does have some consequences for the ergonomics of Xilem Web, but makes Xilem "proper" more ergonomic.

 I don't think it's too bad, when it doesn't directly concern the end-user, maybe all of this is less of an "issue" when we get something like trait aliases. In any way I agree.

xilem_core/src/any_view.rs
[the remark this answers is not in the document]
DJMcNab: Ah yeah, right. That makes sense. It's not clear without the context of the nearby code. I think it's *probably* fine as-is

xilem_core/src/views/adapt.rs
jplatte: I'll try it after this PR is merged, and post a PR if it works out.

PR #409, opened 2024-06-22
Move another log line to reduce noise

PoignardAzur:


DJMcNab (approved):
There's still a ton of noise on first paint, but this is much better.

jaredoconnell (approved):
It definitely makes sense to only paint when it's actually painting.

Is it worth having all of this logging, though, when you can set breakpoints when you are looking for an issue?

PoignardAzur:
The main rationale is that you don't know ahead of time when you'll encounter a bug. If we get to a point where we get record-and-replay by default, we'll probably ease off on the logs a lot.

Also, my personal experience is that breakpoints don't work too well with GUI-type apps, because often the same breakpoint can be triggered hundreds of time before you reach the context you're actually interested in. You can use conditional breakpoints sometimes, but you need good visibility into what your app is doing to set useful conditions, hence logs.

PR #410, opened 2024-06-24
Don't create a log file and print its name unless it will be used

DJMcNab:
Previously, this had the potential to be quite confusing.


jaredoconnell (approved):
Seems logical.

PR #413, opened 2024-06-27
Support `text_size` in `Label`/`Prose` views

DJMcNab:
Format:
```rust
label("everything").text_size(42.)
```
and the same with `prose`.

I've also added `font_size` aliases for these, because people will expect it to have these names.


Philipp-M (approved):
LGTM, Thanks.

PR #415, opened 2024-06-28
Use the cleaner format for READMEs

DJMcNab:


Philipp-M (approved):
Yeah that looks definitely cleaner.

PR #416, opened 2024-06-28
masonry: Fix `unused_qualifications` lints

waywardmonkeys:


DJMcNab (approved):
I mean, I think these are improvements.

But these suggestions concern me; this is not what the lint is documented to do...

waywardmonkeys:
I should have been more clear: I fixed some by removing the qualifications, and some by ... not that the lint suggested such a thing (in case you thought that was what I meant).

This lint isn't necessarily safe to enable though as I've seen another crate where it yields false positives (that I haven't reported).

PR #417, opened 2024-06-30
Allow external event loop to drive masonry and xilem

cfagot:
These changes allow you to create a masonry or xilem app that is driven by an external event loop.

## Masonry

Existing method for creating masonry app:
```
    masonry::event_loop_runner::run(
        masonry::event_loop_runner::EventLoop::with_user_event(),
        window_attributes,
        root_widget,
        app_driver,
    )
    .unwrap();
```

Instead you can now do this:

```
    let masonry_state = MasonryState::new(window_attributes, &event_loop, root_widget);

    let mut app = AppInterface {
        masonry_state,
        app_driver: Box::new(driver),
    };
    event_loop.run_app(&mut app)
```

Where AppInterface implements the winit ApplicationHandler<accesskit_winit::Event> trait.

## Xilem

Existing method:
```
    let app = Xilem::new(state, app_logic);
    app.run_windowed(EventLoop::with_user_event(), title)?;
```

Now:
```
    let xilem = Xilem::new(0, app_logic);
    let (root_widget, app_driver) = xilem.split();
    let parts = xilem.split();
    let (root_widget, app_driver) = (parts.root_widget, parts.driver)

   // and then create masonry app just like above using root_widget and app_driver
```

Also adds example/external_event_loop.rs which duplicates example/flex.rs but with an external event loop.

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -54,9 +54,9 @@ pub enum WindowState<'a> {
     },
 }
 
-/// The state of the Masonry application. If you run masonry from an external event loop, create a
-/// MasonryState via MasonryState::new and forward winit events to it via the appropriate method (e.g.,
-/// masonry_state.handle_window_event via window_event on the winit ApplicationHandler).
+/// The state of the Masonry application. If you run Masonry from an external Winit event loop, create a
+/// `MasonryState` via [`MasonryState::new`] and forward events to it via the appropriate method (e.g.,
+/// calling [`handle_window_event`](MasonryState::handle_window_event) in [`window_event`](ApplicationHandler::window_event)).
 pub struct MasonryState<'a> {
     render_cx: RenderContext,
     render_root: RenderRoot,
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -54,6 +54,9 @@ pub enum WindowState<'a> {
     },
 }
 
+/// The state of the Masonry application. If you run masonry from an external event loop, create a
+/// MasonryState via MasonryState::new and forward winit events to it via the appropriate method (e.g.,
+/// masonry_state.handle_window_event via window_event on the winit ApplicationHandler).
 pub struct MasonryState<'a> {
     render_cx: RenderContext,
     render_root: RenderRoot,
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -12,7 +12,10 @@ use vello::{peniko::Color, AaSupport, RenderParams, Renderer, RendererOptions, S
 use wgpu::PresentMode;
 use winit::application::ApplicationHandler;
 use winit::error::EventLoopError;
-use winit::event::{MouseButton as WinitMouseButton, WindowEvent as WinitWindowEvent};
+use winit::event::{
+    DeviceEvent as WinitDeviceEvent, DeviceId, MouseButton as WinitMouseButton,
+    WindowEvent as WinitWindowEvent,
+};
 use winit::event_loop::{ActiveEventLoop, EventLoopProxy};
 use winit::window::{Window, WindowAttributes, WindowId};
 
@@ -51,11 +54,10 @@ pub enum WindowState<'a> {
     },
 }
 
-struct MainState<'a> {
+pub struct MasonryState<'a> {
     render_cx: RenderContext,
     render_root: RenderRoot,
     pointer_state: PointerState,
-    app_driver: Box<dyn AppDriver>,
     renderer: Option<Renderer>,
     // TODO: Winit doesn't seem to let us create these proxies from within the loop
     // The reasons for this are unclear
@@ -64,6 +66,11 @@ struct MainState<'a> {
     // Per-Window state
     // In future, this will support multiple windows
     window: WindowState<'a>,
+}
+
+struct MainState<'a> {
+    masonry_state: MasonryState<'a>,
+    app_driver: Box<dyn AppDriver>,
 }
 
 /// The type of the event loop used by Masonry.
@@ -96,18 +103,9 @@ pub fn run_with(
     root_widget: impl Widget,
     app_driver: impl AppDriver + 'static,
 ) -> Result<(), EventLoopError> {
-    let render_cx = RenderContext::new();
-    // TODO: We can't know this scale factor until later?
-    let scale_factor = 1.0;
     let mut main_state = MainState {
-        render_cx,
-        render_root: RenderRoot::new(root_widget, WindowSizePolicy::User, scale_factor),
-        renderer: None,
-        pointer_state: PointerState::empty(),
+        masonry_state: MasonryState::new(window, &event_loop, root_widget),
         app_driver: Box::new(app_driver),
-        proxy: event_loop.create_proxy(),
-
-        window: WindowState::Uninitialized(window),
     };
 
     // If there is no default tracing subscriber, we set our own. If one has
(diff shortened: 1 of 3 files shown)


cfagot:
Added fix for clippy. xilem.split() now returns struct rather than tuple.

cfagot:
Converting to draft in order to resolve merge conflicts.

cfagot:
Resolved merge conflicts.

masonry/src/event_loop_runner.rs
```diff
@@ -51,11 +54,10 @@ pub enum WindowState<'a> {
     },
 }
 
-struct MainState<'a> {
+pub struct MasonryState<'a> {
```
DJMcNab, line 57: We should have a doc comment here explaining how to integrate with an external `winit` event loop (i.e. call `::new`, and then pass through all events as needed.

masonry/src/event_loop_runner.rs
```diff
@@ -206,8 +290,76 @@ impl ApplicationHandler<accesskit_winit::Event> for MainState<'_> {
...
+        };
+        let render_params = RenderParams {
+            base_color: Color::BLACK,
+            width,
+            height,
+            antialiasing_method: vello::AaConfig::Area,
+        };
+        self.renderer
+            .get_or_insert_with(|| Renderer::new(device, renderer_options).unwrap())
+            .render_to_surface(device, queue, scene_ref, &surface_texture, &render_params)
```
DJMcNab, line 349: This unconditionally rendering to a surface here does limit the usefulness of this API somewhat.

To be clear, this is still the right API at the moment, but it does limit the use cases

xilem/examples/external_event_loop.rs
DJMcNab, line 1: I wonder if there's a canonical folder for "less important" examples.

It is important that we allow this, but this API design is liable to change quite a bit (e.g. to make Masonry not just own the Window's rendering entirely). And so I don't really want to encourage building anything starting from this example.

To be clear, I think this is fine where it is, but I think it's worth some thinking

xilem/examples/external_event_loop.rs
```diff
@@ -0,0 +1,147 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
DJMcNab, line 3:
```suggestion

//! Shows driving a Xilem application from a pre-existing Winit event loop.
//! Currently, this supports running as its own window alongside an existing application, or
//! accessing raw events from winit.
//! Support for more custom embeddings would be welcome, but needs more design work
```

xilem/examples/external_event_loop.rs
```diff
@@ -0,0 +1,147 @@
...
+        big_button("+", |data| {
+            *data += 1;
+        }),
+    ))
+    .direction(xilem::Axis::Horizontal)
+    .cross_axis_alignment(CrossAxisAlignment::Center)
+    .main_axis_alignment(MainAxisAlignment::Center)
+}
+
+struct AppInterface {
```
DJMcNab, line 43: `AppInterface` is not a very evocative name here. I think `App` might be better, but still not great. Maybe we could use `ExistingApp` :P

xilem/src/lib.rs
```diff
@@ -38,6 +38,16 @@ where
     driver: MasonryDriver<State, Logic, View, View::ViewState>,
 }
 
+// This struct is same as the Xilem struct, except it exposes members as public but implements
+// no methods. It is used to split Xilem into parts for use with external event loops.
+pub struct XilemParts<State, Logic, View>
```
DJMcNab, line 43: What's the reasoning for having this struct? The other options I can see are:
- Returning a tuple
- Making the fields on `Xilem` public

DJMcNab (commented):
A few observations mostly around names, but overall this is looking really good

masonry/src/event_loop_runner.rs
cfagot: Absolutely, but just to be clear that's the existing code, I just moved it. I have a couple more changes that I was going to add in a followup pull request that expose just enough in order to run rendering from an external app (and integrate with other render passes). That method ends up just using the render root to get the scene and then handling everything else in the external render code.

This is what the external render code ends up looking like (note the vello_ext module is just some vello code that was private copy pasted into the project):
https://github.com/cfagot/space_survival/blob/main/src/xilem_render.rs

I could add those few other hooks to masonry sate now if you wanted (I mostly want this pull request to be integrated quickly though, because any change to the event loop runner file ends up being a pain to merge).

DJMcNab: > To be clear, this is still the right API at the moment

Was attempting to indicate that this is a non-blocking concern

xilem/examples/external_event_loop.rs
cfagot: I think this might be an important example if you want people building product now and growing with xilem. Would it be enough to add extensive comments in this example to make it clear this is not a permanent api. I don't think anyone building on this is going to end up at a dead end. They'll just have to move to a more general api.

xilem/src/lib.rs
cfagot: Yeah, initially I returned a tuple and clippy complained it was too complicated. The reason I didn't make the fields public is to make the fact that these are different types clear (i.e., you do different things with each).

But whatever is sensible to you is fine with me. The nature of the type already makes it impossible to access the fields after you call run. I'll just make the fields public.

xilem/examples/external_event_loop.rs
cfagot: Makes sense. Will do.

masonry/src/event_loop_runner.rs
cfagot: Ok, added.

masonry/src/event_loop_runner.rs
```diff
@@ -51,11 +54,13 @@ pub enum WindowState<'a> {
     },
 }
 
-struct MainState<'a> {
+/// The state of the Masonry application. If you run masonry from an external event loop, create a
+/// MasonryState via MasonryState::new and forward winit events to it via the appropriate method (e.g.,
+/// masonry_state.handle_window_event via window_event on the winit ApplicationHandler).
```
DJMcNab, line 59:
```suggestion
/// The state of the Masonry application. If you run Masonry from an external Winit event loop, create a
/// `MasonryState` via [`MasonryState::new`] and forward events to it via the appropriate method (e.g.,
/// calling [`handle_window_event`](MasonryState::handle_window_event) in [`window_event`](ApplicationHandler::window_event)).
```
Makes the doc comments follow the expected patterns

xilem/examples/external_event_loop.rs
```diff
@@ -0,0 +1,150 @@
...
+        big_button("+", |data| {
+            *data += 1;
+        }),
+    ))
+    .direction(xilem::Axis::Horizontal)
+    .cross_axis_alignment(CrossAxisAlignment::Center)
+    .main_axis_alignment(MainAxisAlignment::Center)
+}
+
+struct ExternalApp {
```
DJMcNab, line 47:
```suggestion
/// An application not managed by Xilem, but which wishes to embed Xilem.
struct ExternalApp {
```

xilem/src/lib.rs
```diff
@@ -34,8 +34,8 @@ pub struct Xilem<State, Logic, View>
 where
     View: WidgetView<State>,
 {
-    root_widget: RootWidget<View::Widget>,
-    driver: MasonryDriver<State, Logic, View, View::ViewState>,
+    pub root_widget: RootWidget<View::Widget>,
+    pub driver: MasonryDriver<State, Logic, View, View::ViewState>,
```
DJMcNab, line 38: I am going to change this API in #423, effectively back to your original suggestion.
But this is perfectly fine for now

DJMcNab (approved):
Some small tweaks to add, but I'll merge once those are done

xilem/src/lib.rs
cfagot: Sounds good.

PR #418, opened 2024-06-30
Remove key_without_modifiers call on iOS

johanholmerin:
I had trouble getting masonry to compile for iOS due to a winit platform method call that's not supported. There's already a cfg for android so I just added iOS to it.

Reproduction:

```
cargo check --target aarch64-apple-ios
```

### DIFF
--- a/masonry/src/text2/selection.rs
+++ b/masonry/src/text2/selection.rs
@@ -278,12 +278,12 @@ impl<T: Selectable> TextWithSelection<T> {
 ///
 /// `key_without_modifiers` is only available on some platforms
 fn shortcut_key(key: &winit::event::KeyEvent) -> winit::keyboard::Key {
-    #[cfg(not(target_os = "android"))]
+    #[cfg(all(not(target_os = "android"), not(target_os = "ios")))]
     {
         use winit::platform::modifier_supplement::KeyEventExtModifierSupplement;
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
Our position on iOS support is that we make no guarantees, but we're happy to allow community contributions. This looks like a reasonable change towards that.

Thanks!

masonry/src/text2/selection.rs
```diff
@@ -278,12 +278,12 @@ impl<T: Selectable> TextWithSelection<T> {
 ///
 /// `key_without_modifiers` is only available on some platforms
 fn shortcut_key(key: &winit::event::KeyEvent) -> winit::keyboard::Key {
-    #[cfg(not(target_os = "android"))]
+    #[cfg(all(not(target_os = "android"), not(target_os = "ios")))]
```
DJMcNab, line 281:
```suggestion
    #[cfg(not(any(target_os = "android", target_os = "ios")))]
```

To march the other condition

DJMcNab:
Thanks! Consider posting any further progress in [#xilem > iOS support](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/iOS.20support) - I'd be interested to see what you end up with

johanholmerin:
Will do! And thanks for the quick response

PR #420, opened 2024-07-01
Emoji picker/selector example

DJMcNab:
Xilem Emoji Picker.

Some notes:
1) The accessibility of this example probably isn't great. Not sure what to do about this.
2) Our layout primitives aren't great; we use a grid, but ideally the number of rows would be reactive to the available space.
3) The pagination is slightly hacked together - it should really try and give you page numbers. I'm not planning to address this, unless someone provides the algorithm

This was originally created to act as a screenshot for https://github.com/linebender/linebender.github.io/pull/56


jaredoconnell:
I tested this, and the fonts are not visible on Mac OS. Maybe the font isn't COLR format.

DJMcNab:
Indeed.  This is called out in linebender/linebender.github.io#56, for example.

DJMcNab:
Moving this into draft due to the lack of bitmap emoji support.

DJMcNab:
This should start working properly once we have Vello 0.3 and Parley 0.2

DJMcNab:
Hmm... It supports bitmap and COLR emoji - it works on my desktop Linux machine.

@dfrg do we expect this to work with the Emoji support in Parley/Vello as-is?

dfrg:
I think we make an attempt to do proper emoji font selection. It should work fairly well on Mac/Windows. Linux will depend on installed fonts and fontconfig setup.

DJMcNab:
@jaredoconnell then I'm going to have to ask you to debug this. I don't have a Mac to test with

jaredoconnell:
Are there any resources or documentation I should use to debug this?

dfrg:
Actually, @LaurenzV discovered a bug in the sbix bitmap code here: https://github.com/linebender/vello/blob/98192612d9f4b7aed9b3223680527473ec449ee9/vello/src/scene/bitmap.rs#L165

The tag should be “png “. This will need to be fixed to properly handle Apple Color Emoji.

DJMcNab:
@LaurenzV are you willing to make that PR to Vello?

LaurenzV:
If you could do it, that would be great! If not I'll try to do it in the next few days.

xilem/examples/emoji_picker.rs
```diff
@@ -0,0 +1,293 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0 AND MIT
```
DJMcNab, line 2: Oops. This is wrong. I'll fix this before merge.

xilem/examples/emoji_picker/main.rs
```diff
@@ -0,0 +1,173 @@
...
+                    Box::new(view)
+                }
+                None => Box::new(flex(())),
+            };
+            row_contents.push(sized_box(view).width(dimensions).height(dimensions));
+        }
+        result.push(flex(row_contents).direction(Axis::Horizontal));
+    }
+
+    flex(result)
```
jaredoconnell, line 96: This would be a good place to use a grid. The items can easily get out of alignment with the current implementation.

jaredoconnell (approved):
Confirmed as working on Mac and Windows.
Ideally the example would use a grid instead of nested flex.

DJMcNab:
Re-requested review because I've not too familiar with the grid, and maybe there's some trick I'm missing.

jaredoconnell (approved):
The grid makes it much more polished. Looks good to me.
I'll follow up with a PR that makes a few polish improvements.

PR #421, opened 2024-07-01
Use window.outer_size on iOS

### LINKED ISSUE #419: Scale issue on iOS
I'm experiencing some rendering issues when running xilem on iOS. The content is stretched vertically compared to how it looks on android and macOS. The touch targets also differ from how the content is rendered so you have to tap above the buttons for them to work.

Repo: https://github.com/johanholmerin/xilem-cross-platform

<details>
  <summary>iOS</summary>

xilem ios

</details>

<details>
  <summary>android</summary>

xilem android

</details>

johanholmerin:
As I understand it, on iOS the outer_size corresponds to the size of the window, meaning when the surface is rendered using inner_size(the safe/non-obscured size) the elements get stretched.

There's an open issue in winit about clarifying/standardizing the different sizes, but until that's done switching to outer_size fixes the issue.

The touch positions now also match the rendering

Winit issue: https://github.com/rust-windowing/winit/issues/2308

Fixes #419

<details>
  <summary>Before</summary>

xilem ios

</details>

<details>
  <summary>After</summary>

ios after

</details>

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -135,6 +135,10 @@ impl ApplicationHandler<accesskit_winit::Event> for MainState<'_> {
                 let adapter = Adapter::with_event_loop_proxy(&window, self.proxy.clone());
                 window.set_visible(visible);
                 let window = Arc::new(window);
+                // https://github.com/rust-windowing/winit/issues/2308
+                #[cfg(target_os = "ios")]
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
We'll want to properly handle the "safe area" at some point, but that seems to be blocked on the linked winit issue. This seems like a good short term fix.

PR #422, opened 2024-07-02
Fix the `head` elements of Xilem Web

DJMcNab:
These got mixed up in https://github.com/linebender/xilem/pull/142


Philipp-M (approved):
Looks all good to me, and I just tested all the examples.

PR #423, opened 2024-07-04
Integrate `tokio` for async communication with Xilem

DJMcNab:
Supercedes https://github.com/linebender/xilem/pull/411

This is designed with #417 in mind, to not lock-in to our event loop.

### DIFF
--- a/masonry/examples/calc.rs
+++ b/masonry/examples/calc.rs
@@ -7,8 +7,6 @@
 #![windows_subsystem = "windows"]
 #![allow(clippy::single_match)]
 
-use std::sync::Arc;
-
 use accesskit::{DefaultActionVerb, Role};
 use masonry::app_driver::{AppDriver, DriverCtx};
 use masonry::dpi::LogicalSize;
@@ -156,7 +154,7 @@ impl Widget for CalcButton {
             }
             PointerEvent::PointerUp(_, _) => {
                 if ctx.is_active() && !ctx.is_disabled() {
-                    ctx.submit_action(Action::Other(Arc::new(self.action)));
+                    ctx.submit_action(Action::Other(Box::new(self.action)));
                     ctx.request_paint();
                     trace!("CalcButton {:?} released", ctx.widget_id());
                 }
@@ -176,7 +174,7 @@ impl Widget for CalcButton {
         if event.target == ctx.widget_id() {
             match event.action {
                 accesskit::Action::Default => {
-                    ctx.submit_action(Action::Other(Arc::new(self.action)));
+                    ctx.submit_action(Action::Other(Box::new(self.action)));
                     ctx.request_paint();
                 }
                 _ => {}
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -1,8 +1,15 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use masonry::{app_driver::AppDriver, widget::RootWidget};
-use xilem_core::MessageResult;
+use std::sync::Arc;
+
+use masonry::{
+    app_driver::AppDriver,
+    event_loop_runner::{self, EventLoopProxy, MasonryUserEvent},
+    widget::RootWidget,
+    WidgetId,
+};
+use xilem_core::{DynMessage, Message, MessageResult, Proxy, ProxyError, ViewId};
 
 use crate::{ViewCtx, WidgetView};
 
@@ -10,9 +17,48 @@ pub struct MasonryDriver<State, Logic, View, ViewState> {
     pub(crate) state: State,
     pub(crate) logic: Logic,
     pub(crate) current_view: View,
-    pub(crate) view_ctx: ViewCtx,
+    pub(crate) ctx: ViewCtx,
     pub(crate) view_state: ViewState,
+    #[allow(unused)]
+    // Keep the runtime alive
+    // Is it more idiomatic to stick this in the `ViewCtx`?
+    pub(crate) runtime: tokio::runtime::Runtime,
 }
+
+/// The `WidgetId` which async events should be sent to.
+pub const ASYNC_MARKER_WIDGET: WidgetId = WidgetId::reserved(0x1000);
+
+/// The action which should be used for async events
+pub fn async_action(path: Arc<[ViewId]>, message: Box<dyn Message>) -> masonry::Action {
+    masonry::Action::Other(Box::<MessagePackage>::new((path, message)))
+}
+
+type MessagePackage = (Arc<[ViewId]>, DynMessage);
+
+impl Proxy for MasonryProxy {
+    fn send_message(&self, path: Arc<[ViewId]>, message: DynMessage) -> Result<(), ProxyError> {
+        match self
+            .0
+            .send_event(event_loop_runner::MasonryUserEvent::Action(
+                async_action(path, message),
+                ASYNC_MARKER_WIDGET,
+            )) {
+            Ok(()) => Ok(()),
+            Err(err) => {
+                let MasonryUserEvent::Action(masonry::Action::Other(res), _) = err.0 else {
+                    unreachable!(
+                        "We know this is the value we just created, which matches this pattern"
+                    )
+                };
+                Err(ProxyError::DriverFinished(
+                    res.downcast::<MessagePackage>().unwrap().1,
+                ))
+            }
+        }
+    }
+}
+
+pub(crate) struct MasonryProxy(pub(crate) EventLoopProxy);
 
 impl<State, Logic, View> AppDriver for MasonryDriver<State, Logic, View, View::ViewState>
 where
@@ -21,47 +67,56 @@ where
 {
     fn on_action(
         &mut self,
-        ctx: &mut masonry::app_driver::DriverCtx<'_>,
+        masonry_ctx: &mut masonry::app_driver::DriverCtx<'_>,
         widget_id: masonry::WidgetId,
         action: masonry::Action,
     ) {
-        if let Some(id_path) = self.view_ctx.widget_map.get(&widget_id) {
-            let message_result = self.current_view.message(
+        let message_result = if widget_id == ASYNC_MARKER_WIDGET {
+            let masonry::Action::Other(action) = action else {
+                panic!();
+            };
+            let (path, message) = *action.downcast::<MessagePackage>().unwrap();
+            // Handle an async path
+            self.current_view
+                .message(&mut self.view_state, &path, message, &mut self.state)
+        } else if let Some(id_path) = self.ctx.widget_map.get(&widget_id) {
+            self.current_view.message(
                 &mut self.view_state,
                 id_path.as_slice(),
                 Box::new(action),
                 &mut self.state,
-            );
-            let rebuild = match message_result {
-                MessageResult::Action(()) => {
-                    // It's not entirely clear what to do here
-                    true
-                }
-                MessageResult::RequestRebuild => true,
-                MessageResult::Nop => false,
-                MessageResult::Stale(_) => {
-                    tracing::info!("Discarding message");
-                    false
-                }
-            };
-            if rebuild {
-                let next_view = (self.logic)(&mut self.state);
-                let mut root = ctx.get_root::<RootWidget<View::Widget>>();
-
-                self.view_ctx.view_tree_changed = false;
-                next_view.rebuild(
-                    &self.current_view,
-                    &mut self.view_state,
-                    &mut self.view_ctx,
-                    root.get_element(),
-                );
-                if cfg!(debug_assertions) && !self.view_ctx.view_tree_changed {
-                    tracing::debug!("Nothing changed as result of action");
-                }
-                self.current_view = next_view;
-            }
+            )
         } else {
             eprintln!("Got action {action:?} for unknown widget. Did you forget to use `with_action_widget`?");
+            return;
+        };
+        let rebuild = match message_result {
+            MessageResult::Action(()) => {
+                // It's not entirely clear what to do here
+                true
+            }
+            MessageResult::RequestRebuild => true,
+            MessageResult::Nop => false,
+            MessageResult::Stale(_) => {
+                tracing::info!("Discarding message");
+                false
+            }
+        };
+        if rebuild {
+            let next_view = (self.logic)(&mut self.state);
+
+            let mut root = masonry_ctx.get_root::<RootWidget<View::Widget>>();
+
+            next_view.rebuild(
+                &self.current_view,
+                &mut self.view_state,
+                &mut self.ctx,
+                root.get_element(),
+            );
+            if cfg!(debug_assertions) && !self.ctx.view_tree_changed {
+                tracing::debug!("Nothing changed as result of action");
+            }
+            self.current_view = next_view;
         }
     }
 }
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -19,10 +19,6 @@ pub struct MasonryDriver<State, Logic, View, ViewState> {
     pub(crate) current_view: View,
     pub(crate) ctx: ViewCtx,
     pub(crate) view_state: ViewState,
-    #[allow(unused)]
-    // Keep the runtime alive
-    // Is it more idiomatic to stick this in the `ViewCtx`?
-    pub(crate) runtime: tokio::runtime::Runtime,
 }
 
 /// The `WidgetId` which async events should be sent to.
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -9,7 +9,7 @@ use masonry::{
     widget::RootWidget,
     WidgetId,
 };
-use xilem_core::{DynMessage, Message, MessageResult, Proxy, ProxyError, ViewId};
+use xilem_core::{DynMessage, Message, MessageResult, ProxyError, RawProxy, ViewId};
 
 use crate::{ViewCtx, WidgetView};
 
@@ -35,7 +35,7 @@ pub fn async_action(path: Arc<[ViewId]>, message: Box<dyn Message>) -> masonry::
 
 type MessagePackage = (Arc<[ViewId]>, DynMessage);
 
-impl Proxy for MasonryProxy {
+impl RawProxy for MasonryProxy {
     fn send_message(&self, path: Arc<[ViewId]>, message: DynMessage) -> Result<(), ProxyError> {
         match self
             .0
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -28,11 +28,12 @@ pub struct MasonryDriver<State, Logic, View, ViewState> {
 /// The `WidgetId` which async events should be sent to.
 pub const ASYNC_MARKER_WIDGET: WidgetId = WidgetId::reserved(0x1000);
 
-/// The action which should be used for async events
+/// The action which should be used for async events.
 pub fn async_action(path: Arc<[ViewId]>, message: Box<dyn Message>) -> masonry::Action {
     masonry::Action::Other(Box::<MessagePackage>::new((path, message)))
 }
 
+/// The type used to send a message for async events.
 type MessagePackage = (Arc<[ViewId]>, DynMessage);
 
 impl RawProxy for MasonryProxy {
@@ -58,7 +59,13 @@ impl RawProxy for MasonryProxy {
     }
 }
 
-pub(crate) struct MasonryProxy(pub(crate) EventLoopProxy);
+pub struct MasonryProxy(pub(crate) EventLoopProxy);
+
+impl MasonryProxy {
+    pub fn new(proxy: EventLoopProxy) -> Self {
+        Self(proxy)
+    }
+}
 
 impl<State, Logic, View> AppDriver for MasonryDriver<State, Logic, View, View::ViewState>
 where
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -35,6 +35,9 @@ pub use driver::{async_action, MasonryDriver, MasonryProxy, ASYNC_MARKER_WIDGET}
 
 pub mod view;
 
+/// Re-export of tokio as the async driver for Masonry
+pub use tokio;
+
 pub struct Xilem<State, Logic> {
     state: State,
     logic: Logic,
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -17,7 +17,7 @@ use winit::{
     window::{Window, WindowAttributes},
 };
 use xilem_core::{
-    AsyncCtx, MessageResult, Proxy, SuperElement, View, ViewElement, ViewId, ViewPathTracker,
+    AsyncCtx, MessageResult, RawProxy, SuperElement, View, ViewElement, ViewId, ViewPathTracker,
 };
 
 pub use masonry::{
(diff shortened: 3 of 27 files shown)


xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+/// This can be used for asynchronous event handling.
+/// For example, to get the result of a `Future` or a channel into
+/// the view, which then will ultimately.
+///
+/// In the Xilem crate, this will wrap an `EventLoopProxy` from Winit.
+///
+/// ## Lifetimes
+///
+/// It is valid for a [`Proxy`] to outlive the [`View`](crate::View) it is associated with.
+pub trait RawProxy: Send + Sync + 'static {
```
Philipp-M, line 32: I wonder, if we can avoid the `Send` bound, such that this could also be used in xilem_web. I do think we also want some real async in xilem_web (e.g. via web-workers), but I could also imagine, using something like `setInterval` similar as `async_repeat` which AFAIK doesn't need real async, and could e.g. use `web_sys::Event`. (This is probably more relevant to the `DynMessage` type used in the method).

xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Send `message` to the `View` which created this `MessageProxy`
+    pub fn message(&self, message: M) -> Result<(), ProxyError> {
+        self.proxy
+            .send_message(self.path.clone(), Box::new(message))
+    }
+}
+
+/// A [`View`] which has no element type.
+pub trait PhantomView<State, Action, Context, Message = DynMessage>:
+    View<State, Action, Context, Element = NoElement>
```
Philipp-M, line 77:
```suggestion
pub trait PhantomView<State, Action, Context, Message = DynMessage>:
    View<State, Action, Context, Message, Element = NoElement>
```

xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+where
+    Context: ViewPathTracker,
+{
+}
+
+impl<State, Action, Context, Message, V> PhantomView<State, Action, Context, Message> for V
+where
+    V: View<State, Action, Context, Element = NoElement>,
```
Philipp-M, line 85:
```suggestion
    V: View<State, Action, Context, Message, Element = NoElement>,
```

Philipp-M (commented):
Left some comments, I hope I find more time soon to dive a little deeper.

xilem_core/src/element.rs
```diff
@@ -82,3 +82,15 @@ where
...
     fn replace_inner(this: Self::Mut<'_>, child: Child) -> Self::Mut<'_>;
 }
+
+/// Element type for views which don't impact the element tree.
+///
+/// Views with this element type can be included in any [`ViewSequence`](crate::ViewSequence) (with the
+/// correct `State` and `Action` types), as they do not need to actually add an element to the sequence.
+///
+/// These views can also as the `alongside_view` in [`fork`](crate::fork).
+pub struct NoElement;
```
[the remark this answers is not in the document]
DJMcNab: The reason that `NoElement` exists is so that it can be produced from nothing, and so can be inserted anywhere into the view tree.

I don't see how it can be meaningful for `NoElement` to contain data?

xilem_core/src/views/fork.rs
```diff
@@ -0,0 +1,95 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{Mut, NoElement, View, ViewId, ViewPathTracker};
+
+/// Create a view which acts as `active_view`, whilst also running `alongside_view`, without inserting it into the tree.
+///
+/// `alongside_view` must be a View with an element type of [`NoElement`].
+pub fn fork<Active, Alongside>(
```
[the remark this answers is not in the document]
DJMcNab: Yeah, `fork` is not a compelling name - it gets the job done, but it's not good.

Maybe `sidecar` :laughing:

xilem_core/src/element.rs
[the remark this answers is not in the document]
Philipp-M: I mean certainly not for `NoElement`, but a view could create e.g. an `AsyncElement` which may contain state relevant to that element, or something along those lines, which is managed by a `View` such as `fork`, i.e. having a custom view runner for different elements inside such a view. But right now, it probably doesn't make sense.

xilem_core/src/views/fork.rs
```diff
@@ -0,0 +1,139 @@
...
+    AppendVec, ElementSplice, Mut, NoElement, View, ViewId, ViewPathTracker, ViewSequence,
+};
+
+/// Create a view which acts as `active_view`, whilst also running `alongside_view`, without inserting it into the tree.
+///
+/// `alongside_view` must be a `ViewSequence` with an element type of [`NoElement`].
+pub fn fork<Active, Alongside, Marker>(
+    active_view: Active,
+    alongside_view: Alongside,
+) -> Fork<Active, Alongside, Marker> {
```
Philipp-M, line 16: I wonder whether this should have all the bounds necessary for `View` to avoid misusage?

xilem_core/src/views/fork.rs
```diff
@@ -0,0 +1,139 @@
...
+    active_view: Active,
+    alongside_view: Alongside,
+    marker: PhantomData<Marker>,
+}
+
+impl<State, Action, Context, Active, Alongside, Marker, Message>
+    View<State, Action, Context, Message> for Fork<Active, Alongside, Marker>
+where
+    Active: View<State, Action, Context, Message>,
+    Alongside: ViewSequence<State, Action, Context, NoElement, Marker, Message>,
```
Philipp-M, line 35: Neat trick, to allow multiple views here. But are those `unreachable!`s really unreachable below? I guess they are because of the `ViewSequence` impl above?

xilem_core/src/views/run_once.rs
```diff
@@ -0,0 +1,119 @@
...
+/// fn log_data(app: &mut AppData) -> impl PhantomView<AppData, (), ViewCtx> {
+///     let val = app.data;
+///     run_once(move || println!("{}", val))
+/// }
+/// # // We need to call the function to make the inline constant be evaluated
+/// # let _ = log_data(&mut AppData { data: 10 });
+/// ```
+pub fn run_once<F>(once: F) -> RunOnce<F>
+where
+    F: Fn() + 'static,
```
Philipp-M, line 46: I could also imagine this taking the mutable app state, triggered via a message that is sent to this view, but I guess this could also be a separate view, and maybe even take an `async` function then. Something like [this](https://github.com/Philipp-M/trui/blob/dd24a2921eb0587dfffbbafea4a3e62b92f39c43/examples/async_event_handler_stream.rs#L67) (in that context the `EventHandler` trait introduced there also gets more interesting again...)

xilem_core/src/views/run_once.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/// The view type for [`run_once`].
+///
+/// This is a [`NoElement`] view.
+pub struct RunOnce<F> {
+    once: F,
+}
+
+impl<F, State, Action, Context> View<State, Action, Context> for RunOnce<F>
```
Philipp-M, line 76:
```suggestion
impl<F, State, Action, Context, Message> View<State, Action, Context, Message> for RunOnce<F>
```

I think it would be useful to add a simple example to the `mason` example (maybe just the `log_lifecycle` there).

xilem_core/src/views/run_once.rs
```diff
@@ -0,0 +1,119 @@
...
+        _: crate::Mut<'_, Self::Element>,
+    ) {
+        // Nothing to do
+    }
+
+    fn message(
+        &self,
+        (): &mut Self::ViewState,
+        _: &[crate::ViewId],
+        message: DynMessage,
```
Philipp-M, line 113:
```suggestion
        message: Message,
```

xilem_core/src/views/run_once.rs
```diff
@@ -0,0 +1,119 @@
...
+        _: &mut State,
+    ) -> MessageResult<Action> {
```
Philipp-M, line 115:
```suggestion
    ) -> MessageResult<Action, Message> {
```

xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// them, of the type they expect.
+    /// It is expected for [`View`](crate::View)s to panic otherwise, and the routing
+    /// will prefer to send stable.
+    ///
+    /// # Errors
+    ///
+    /// This method may error if the driver is no longer running, and in any other
+    /// cases directly documented on the context which was used to create this proxy.
+    /// It may also fail silently.
+    // TODO: Do we want/need a way to asynchronously report errors back to the caller?
```
Philipp-M, line 45: I guess that would be quite nice, so that apps, could show some indicator in the ui, when it failed.

xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+use alloc::{boxed::Box, sync::Arc};
+
+use crate::{DynMessage, Message, NoElement, View, ViewId, ViewPathTracker};
+
+/// A `Context` for a [`View`](crate::View) implementation which supports
+/// asynchronous message reporting.
+pub trait AsyncCtx: ViewPathTracker {
+    /// Get a [`Proxy`] for this context.
+    // TODO: Maybe store the current path within this Proxy?
+    fn proxy(&mut self) -> Arc<dyn RawProxy>;
```
Philipp-M, line 15: Maybe the `RawProxy` can be the `MessageProxy<M>` (taking that `TODO` into account)? (Haven't looked into that though whether that's feasable.)

So something like:


```suggestion
    fn proxy<M>(&mut self) -> Arc<dyn MessageProxy<M>>;
```

masonry/examples/calc.rs
```diff
@@ -156,7 +154,7 @@ impl Widget for CalcButton {
             }
             PointerEvent::PointerUp(_, _) => {
                 if ctx.is_active() && !ctx.is_disabled() {
-                    ctx.submit_action(Action::Other(Arc::new(self.action)));
+                    ctx.submit_action(Action::Other(Box::new(self.action)));
```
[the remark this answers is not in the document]
DJMcNab: No, this is needed so that ownership of the inner message can be passed into `Xilem`, rather than using interior mutability for that

xilem/src/driver.rs
```diff
@@ -1,17 +1,70 @@
...
 pub struct MasonryDriver<State, Logic, View, ViewState> {
     pub(crate) state: State,
     pub(crate) logic: Logic,
     pub(crate) current_view: View,
-    pub(crate) view_ctx: ViewCtx,
+    pub(crate) ctx: ViewCtx,
     pub(crate) view_state: ViewState,
+    #[allow(unused)]
+    // Keep the runtime alive
+    // Is it more idiomatic to stick this in the `ViewCtx`?
```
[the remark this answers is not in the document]
DJMcNab: The question I'm raising here is whether the context needs direct access to the runtime - according to tokio's docs, there is no difference between `Handle` and `&Runtime`

xilem_core/src/views/fork.rs
DJMcNab: This has problems with ambiguity. Essentially, there is no link between the method call and the implementation of `ViewSequence` which gets selected.

I think it is possible to work around this, but it requires adding `State` and `Action` generics to each view which uses this trick. The same thing applies to e.g. `label`, for example.

I think we could do this, but it should happen in a follow-up and be applied to all views

xilem_core/src/views/fork.rs
DJMcNab: I've added an explanatory comment.

xilem_core/src/views/run_once.rs
DJMcNab: As in make this a view which sends a message to itself?
It's an interesting idea, although it makes things more complicated; it means that the `ctx` needs to be able to send itself a message, which feels a bit weird.

This taking an async function is kind of missing the point of this view, which is that it can exist in `xilem_core`.

xilem_core/src/deferred.rs
DJMcNab: Yeah, there is a possible need to gracefully handle the stale messages.

At the moment, the future does get cancelled when the message would go stale, so there's only a very small window where this could error - practically, I think it's not really needed.

xilem_core/src/deferred.rs
DJMcNab: I'm not sure I understand what you're suggesting here.

xilem/src/driver.rs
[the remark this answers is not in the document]
Philipp-M: Yeah good point, I mean I don't think it's too important then. But I guess we can still avoid an additional attribute. I mean we can still put it back to the driver in case we need the `Runtime` there for whatever reason...

xilem_core/src/views/fork.rs
Philipp-M: Sounds good to me, and reflects what I have commented in #428. (Though that issue there might not occur here, I'm not sure though).

xilem_core/src/views/run_once.rs
Philipp-M: Yeah as said, this is a different kind of view, this view is fine to me, I guess it's more for debugging purposes?

>  it means that the ctx needs to be able to send itself a message, which feels a bit weird.

Is it though? We don't have access to the app state in `build` or `rebuild` so we kind of need to do it in `View::message`, when we want access to app state.

> This taking an async function is kind of missing the point of this view, which is that it can exist in xilem_core.

Good point

xilem_core/src/deferred.rs
Philipp-M: I'm also not entirely sure ^^ (or at least whether that is working for either xilem and xilem_web).
I guess what I mean, is that the `RawProxy` takes a strongly typed `message` in `send_message`, without the path, as it is within the implementor (i.e. the current `MessageProxy` struct as a trait)

xilem_core/src/deferred.rs
Philipp-M: Yeah we can see how long we can get away without this. Though it's probably not relevant for this PR

xilem_core/src/views/run_once.rs
DJMcNab: I don't think that new capability belongs in this PR. You can use `run_once_raw` to *read* the app state, and it feels like it's asking for infinite loops to modify the app state in response to creating a view.

I'm hopeful that we can avoid needing this entirely.

xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+    ///
+    /// This likely requires async error handling to happen.
+    ViewExpired(DynMessage, Arc<[ViewId]>),
+    #[allow(missing_docs)]
+    Other(&'static str),
+    // TODO: When core::error::Error is stabilised
+    // Other(Box<dyn core::error::Error + Send>),
+}
+
+// Is it fine to use thiserror in this crate?
```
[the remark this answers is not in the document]
DJMcNab: `thiserror` doesn't support `no_std`, so it's not currently usable in this case. The error handling story still needs tidying up, so I think we can defer this.

xilem/src/driver.rs
```diff
@@ -1,67 +1,125 @@
...
+                };
+                Err(ProxyError::DriverFinished(
+                    res.downcast::<MessagePackage>().unwrap().1,
+                ))
+            }
+        }
+    }
+}
+
+pub struct MasonryProxy(pub(crate) EventLoopProxy);
```
Philipp-M, line 58: I'm not sure if we should "abuse" `EventLoopProxy`, and whether we should have our "own" async event handling mechanism in xilem to have more control over the whole (async) data flow and view rendering, which combines either `MasonryDriver::on_action` as well as async handling of xilem, which would also make (most) of the changes in masonry unnecessary.

I'm mostly thinking about the (unfortunately a little bit complex) logic [here](https://github.com/linebender/xilem/blob/xilem_classic_archive/src/app.rs#L324-L387), which is somewhat explained [here](https://github.com/raphlinus/raphlinus.github.io/issues/74) (paragraph starting with "Ideal is hybrid").

I don't want to block on this, as this could as well be done in a follow-up PR, just something to think about...

xilem/src/lib.rs
```diff
@@ -27,36 +29,32 @@ pub use xilem_core as core;
...
 pub mod view;
 
-pub struct Xilem<State, Logic, View>
-where
-    View: WidgetView<State>,
-{
-    pub root_widget: RootWidget<View::Widget>,
-    pub driver: MasonryDriver<State, Logic, View, View::ViewState>,
+/// Re-export of tokio as the async driver for Masonry
+pub use tokio;
```
Philipp-M, line 39: I wonder how feasible it is to be generic over the async runtime (remembering a TODO comment in the classic archive). But it probably results in "overabstraction"...

Also is it "Masonry" or "Xilem" (which IMO would make more sense?)?

xilem_core/src/deferred.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// cases directly documented on the context which was used to create this proxy.
+    /// It may also fail silently.
+    // TODO: Do we want/need a way to asynchronously report errors back to the caller?
+    //
+    // e.g. an `Option<Arc<dyn FnMut(ProxyError, ProxyMessageId?)>>`?
+    fn send_message(&self, path: Arc<[ViewId]>, message: Message) -> Result<(), ProxyError>;
+}
+
+/// A way to send a message of an expected type to a specific view.
+pub struct MessageProxy<M: Message> {
```
Philipp-M, line 52: That `Message` bound is a little bit unfortunate for xilem_web, as it makes it incompatible with `web_sys` types (mostly relevant for `web_sys::Event` and derivatives).
If we want to make this compatible, something like `RawProxy<ErasedMessage = DynMessage>` would also be necessary for this. But at that point, as mentioned in the final review comment, I think it should rather be a trait (as the impl of `message` doesn't work that easily anymore).

Philipp-M (approved):
I wonder, how much benefit the async stuff (mostly `MessageProxy`, `RawProxy` and `AsyncCtx`) in xilem_core brings us, and if all of this should just reside in xilem itself? It's really unfortunate that we can't easily be generic over `Send` and `Sync` (IIRC this is also an issue for async traits).

I found a way for the `trait MessageProxy<M>` (emulating the current `MessageProxy` struct) but that obviously requires importing the trait in user code, to access `proxy.message()`. (which also begs for the question of a prelude, in which case I wouldn't think it's too bad). I could add that in a follow-up PR if the above isn't a deal-breaker (when there's a reason for the first question of this comment).

But I don't want to block on this further...

xilem_core/src/views/run_once.rs
Philipp-M: Yeah, I think we can postpone all of this until it really will be necessary or brings a "measurable" benefit.

Though I don't think a `RunOnce` view with access to the app state should function like this view (i.e. it should rather send a message to itself, in build, so that it can "cleanly" access app state in `View::message` and change it, but as you say, an infinite loop could be created anyway. I think `run_once_raw` can be removed then?

xilem/src/lib.rs
DJMcNab: There isn't a compelling ecosystem story about abstracting over this - you need to make a choice, and the default choice is certainly Tokio.

Thanks for catching the comment

xilem/src/driver.rs
DJMcNab: Needing to hard-code such a delay is one of those problems which goes away if we properly implement frame pacing. Actually doing that is of course a hard problem, but it's not an impossible one.

It is possible to move `app_logic`, etc. off of the main thread. However, I don't expect the cost of `app_logic` to be very high, so I strongly suspect it's not worth doing.

I'm afraid I don't really understand the rest of the suggestion you're making - we need some way to communicate with the main thread, and that mechanism is `EventLoopProxy`.

Philipp-M: Ah I think I get what you mean, i.e. we need access to the root widget, which is only possible in `on_action`. Ok, well then forget about it, I also think, that the rest could of course be done with custom messages only targeted for the driver impl (like it is already with `ASYNC_MARKER_WIDGET`), but as you say that may be done with frame pacing properly.

DJMcNab:
Thanks for the thorough reviews. To clarify my thinking about Web support; I treat it as effectively tier 2 target.

That is, I'm happy for us to make sure we don't actively regress it, but I don't want to require that every change has to be designed with the constrained environment the web platform provides. It is possible that in the future the web becomes a stronger focus for us, but I don't want to end up in a situation where we have changes needed for Xilem.

I think that every platform other than the web will be able to use these async abstractions, which are only exposed to users of Xilem Core, rather than directly reachable to end-users anyway. A context type for a backend doesn't need to implement  `AsyncContext`.

xilem/src/view/async_repeat.rs
```diff
@@ -0,0 +1,90 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{future::Future, marker::PhantomData, sync::Arc};
+
+use tokio::task::JoinHandle;
+use xilem_core::{DynMessage, Message, MessageProxy, NoElement, View, ViewId, ViewPathTracker};
+
+use crate::ViewCtx;
+
+pub fn async_repeat<M, F, H, State, Action, Fut>(
```
Philipp-M, line 11: Isn't the name a little bit too specific and misleading?

I think this is more general isn't it, something like `messenger`, `async_messenge`, `proxy` or actually `fork` (as it "forks" the current "thread" with a new closure returning the content of the thread as a future). I mean I'm not sure about the name either^^ but I think `async_repeat` lets the reader/user think, that this is only for repeating stuff (at least that's what I just did, revisiting the code), while this actually starts a different thread with a handle to send messages to change the state while this view is within the view tree.
Where the message could be an enum, being similar to async elm again.

xilem/src/view/async_repeat.rs
```diff
@@ -0,0 +1,90 @@
...
+    type ViewState = JoinHandle<()>;
+
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        let path: Arc<[ViewId]> = ctx.view_path().into();
+
+        let proxy = ctx.proxy.clone();
+        let handle = ctx
+            .runtime()
+            .spawn((self.future_future)(MessageProxy::new(proxy, path)));
+        // TODO: Clearly this shouldn't be a label here
```
Philipp-M, line 52: I don't understand this, do you mean `NoElement`?

xilem/src/view/async_repeat.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    fn teardown(
+        &self,
+        _: &mut Self::ViewState,
+        _: &mut ViewCtx,
+        _: xilem_core::Mut<'_, Self::Element>,
+    ) {
+        // Nothing to do
+        // TODO: Our state will be dropped, finishing the future
```
Philipp-M, line 73: Could this just be:

```suggestion
handle.abort();
// let _ = ctx.runtime().block_on(handle); not sure about that, I don't like blocking here...
```

Or does this need more sophisticated handling?

xilem/src/view/async_repeat.rs
DJMcNab: I didn't put much thought. Happy to see a PR which renames it.

The main thing to note about this future is that there's no way for it to get a message "back" from the UI thread, so it's quite limited in what it can do.

xilem/src/view/async_repeat.rs
DJMcNab: This comment is vestigal from before I added NoElement. Thanks for catching

xilem/src/view/async_repeat.rs
DJMcNab: Ooh, I didn't see that `abort` method. That's better

PR #425, opened 2024-07-12
xilem_web: Add Elm Architecture example

flosse:
I have intentionally left out the `xilem_core::adapt` variant here in order to have a clear classic Elm example. I would rather show the `adapt` variant in a separate example.

### DIFF
--- a/xilem_web/web_examples/elm/Cargo.toml
+++ b/xilem_web/web_examples/elm/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "elm"
+version = "0.0.0" # not versioned
+publish = false
+license.workspace = true
+edition.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+console_error_panic_hook = "0.1.7"
+console_log = { version = "1.0.0", features = ["color"] }
+log = "0.4.22"
+xilem_web = { path = "../.." }
--- a/xilem_web/web_examples/elm/src/main.rs
+++ b/xilem_web/web_examples/elm/src/main.rs
@@ -1,0 +1,58 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Xilem supports several patterns for creating modular components.
+//! You can also emulate the elm architecture for a subset of your app.
+//! Though usually it's more idiomatic to modularize state with
+//! [`map_state`](xilem_web::core::map_state) or
+//! [`adapt`](xilem_web::core::adapt).
+
+use xilem_web::{
+    core::map_action,
+    document_body,
+    elements::html as el,
+    interfaces::{Element, HtmlDivElement},
+    Action, App,
+};
+
+#[derive(Debug, Default)]
+struct Model {
+    count: i32,
+}
+
+#[derive(Debug)]
+enum Message {
+    Increment,
+    Decrement,
+}
+
+impl Action for Message {}
+
+fn update(model: &mut Model, message: Message) {
+    log::debug!("Update model {model:?} by {message:?}");
+    match message {
+        Message::Increment => model.count += 1,
+        Message::Decrement => model.count -= 1,
+    }
+    log::debug!("Model updated: {model:?}");
+}
+
+fn app_logic(model: &mut Model) -> impl HtmlDivElement<Model> {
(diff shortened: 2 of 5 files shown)


xilem_web/web_examples/elm/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    document_body,
+    elements::html as el,
+    interfaces::{Element, HtmlDivElement},
+    Action, App,
+};
+
+// ----- ------
+//    Model
+// ----- ------
+
```
Philipp-M, line 15: I think these kinds of comments are somewhat redundant and take quite a bit of real-estate for this otherwise short example

xilem_web/web_examples/elm/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+        Message::Decrement => model.count -= 1,
+    }
+    log::debug!("Model updated: {model:?}");
+}
+
+// ----- ------
+//    View
+// ----- ------
+
+fn view(model: &mut Model) -> impl HtmlDivElement<Model> {
```
Philipp-M, line 50: I think I get why this function is called `view`, though just to be consistent with the other examples (and those in xilem itself), also because technically the updating logic is contained within this function.
```suggestion
fn app_logic(model: &mut Model) -> impl HtmlDivElement<Model> {
```

Philipp-M (approved):
Apart from the nits, looks good to me, thanks.

Just want to link to [this](https://github.com/linebender/xilem/pull/401#pullrequestreview-2124661139), as I agree with it, so it probably makes sense to add a comment to that example (similar to the one in xilem) as well.

flosse:
@Philipp-M Thanks for you review. I force-pushed the changes.

xilem_web/web_examples/elm/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+console_error_panic_hook = "0.1.7"
+console_log = { version = "1.0.0", features = ["color"] }
+log = "0.4.22"
+xilem_web = { path = "../.." }
+xilem_core = { workspace = true }
```
Philipp-M, line 16: I don't think this dependency is necessary here, you can just use `xilem_web::core` which is a reexport of `xilem_core`

xilem_web/web_examples/elm/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Xilem supports several patterns for creating modular components.
+//! You can also emulate the elm architecture for a subset of your app.
+//! Though usually it's more idiomatic to modularize state with
+//! [`map_state`](xilem_core::map_state),
+//! [`map_action`](xilem_core::map_action) or
+//! [`adapt`](xilem_core::adapt).
+
+use xilem_core::map_action;
```
Philipp-M, line 11: See comment above

flosse:
> Just want to link to [this](https://github.com/linebender/xilem/pull/401#pullrequestreview-2124661139)

BTW: I have created a Xilem ideomatic example for performance measurement: https://github.com/krausest/js-framework-benchmark/pull/1716
If you have any thoughts on this, perhaps we can discuss it (in another place) :relaxed:

Philipp-M:
> I have created a Xilem ideomatic example for performance measurement

Oh well, I've done that as well some time ago, and updated it to the new xilem_web recently locally, I just comitted it and pushed it [here](https://github.com/Philipp-M/js-framework-benchmark/tree/ff7ae0340ac3965e554e26ebc8a6b2415bb1e40c/frameworks/non-keyed/xilem_web-rewrite). Should've probably be more public about that (I think I did want to support the keyed bench before being more public with it, but I guess that's an issue for duplicated work...) 
Relevant [zulip topic](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/xilem_html.3A.20js-framework-benchmark)
You may want to compare your version with mine (like all variables, wasm binary size, raw performance and code-quality) and either merge those, or take one or the other version. I'm still not sure to already publicly add it to the bench since xilem_web is still quite experimental, but I guess we can do that...


Anyway, merging this, thanks!

PR #426, opened 2024-07-12
[masonry] Update to accesskit 0.16

waywardmonkeys:
There was a breaking change in this release that renamed `role::StaticText` to `role::Label`.

### DIFF
--- a/masonry/src/widget/prose.rs
+++ b/masonry/src/widget/prose.rs
@@ -284,7 +284,7 @@ impl Widget for Prose {
     }
 
     fn accessibility_role(&self) -> Role {
-        Role::StaticText
+        Role::Paragraph
     }
 
     fn accessibility(&mut self, ctx: &mut AccessCtx) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,15 +20,15 @@ checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
 name = "accesskit"
-version = "0.15.0"
+version = "0.16.0"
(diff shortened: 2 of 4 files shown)


masonry/src/widget/prose.rs
```diff
@@ -285,7 +285,7 @@ impl Widget for Prose {
     }
 
     fn accessibility_role(&self) -> Role {
-        Role::StaticText
+        Role::Label
```
DJMcNab, line 288: Is this the right role for `Prose`? I think `Paragraph` might be better (still not completely correct, of course, but closer to what the Prose actually represents).

DJMcNab (approved):
Just a lingering question about the accessibility role of `Prose`.

masonry/src/widget/prose.rs
waywardmonkeys: I wondered the same thing and found it deeply unsatisfying to use `Role::Label` there. But I don't think there's a single right answer and it will depend on what the prose is for ... there are several roles that could apply and I think it'll need to be configurable.

DJMcNab: I don't think `Label` can ever meaningfully apply to `Prose`, but I also don't know what the better role would be. The thing to test is what role does the browser use for something like the main article text of a body (e.g. the body of this comment

which has multiple paragraphs but is logically one node).

DJMcNab: Of course, the web splits each paragraph into its own node/widget. Maybe we need to be doing that?
I think for now, we need to choose one of paragraph or label. I'd lean *slightly* towards paragraph, but it's not a well-reasoned position.

DJMcNab (approved):

PR #427, opened 2024-07-13
xilem_web: Add fetch example

flosse:
I was wondering what the integration of asynchronous calls looks like.
Here is an example, which works, but everything is more like a dirty hack.

Therefore, this PR should first serve as a basis for discussion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -178,6 +178,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.86"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
+
+[[package]]
 name = "arrayref"
 version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1057,6 +1063,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "fetch"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "console_error_panic_hook",
+ "console_log",
+ "futures",
+ "gloo-net",
+ "log",
+ "serde",
+ "thiserror",
+ "wasm-bindgen",
+ "wasm-bindgen-futures",
+ "web-sys",
+ "xilem_web",
+]
+
+[[package]]
 name = "flate2"
 version = "1.0.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


DJMcNab:
If at all practical, this should try and use APIs similar to those in #423

flosse:
> If at all practical, this should try and use APIs similar to those in #423

Thank you for the hint!
I'll take a closer look at that.

flosse:
My first question on the subject is what the **API** should ideally look like.
If I understand the example in #423 correctly, a pseudo view element is created which executes an asynchronous context:

https://github.com/linebender/xilem/blob/79bc28c6f6e5945cde180d599c47e9c2c568356b/xilem/examples/mason.rs#L71-L82<

But how could a async task be executed that was triggered by an event?

```rust
div(()).on_click(|state, event|{
   // TODO: how to spawn an async task and process the outcome?
})
```
Return a task/command with the future and an update function?
```rust
div(()).on_click(|state, event|{
    let task = async move {
        do_something().await
    };
    Command::AsyncTask(task, |state, output|{
        state.value = output;
    })
})
```

Philipp-M:
> But how could a async task be executed that was triggered by an event?

I see two ways currently to deal with that:

1. Similar as I have done that with an `EventHandler` trait (which is basically a `View` so it may or may not be just an "event" `View`, with an additional generic parameter for the `Event` type), which allows more flexibility for dealing with events, as already linked in that PR, like [this](https://github.com/Philipp-M/trui/blob/dd24a2921eb0587dfffbbafea4a3e62b92f39c43/examples/async_event_handler_stream.rs#L67-L82)
2. Similar as you have already shown, via an `Action` type (doesn't need to be necessarily an enum though), which is composed within a handler-view that handles the returned action.

I slightly lean to 1. as it's less boilerplate and I think looks a little bit cleaner and less convoluted (and in my eyes more idiomatic), but it's not 100% optimal with inference in rust-analyzer currently, and *may* create more complex errors.

A third, but in my eyes not ideal option (because not that straight-forward and convoluted), would be to have a flag that is set in the event-callback, which in turn inserts such a pseudo-view in the view-tree (like `fork(..,Some(run_once()))`).

flosse:
> 1. Similar as I have done that with an `EventHandler` trait (which is basically a `View` so it may or may not be just an "event" `View`, with an additional generic parameter for the `Event` type), which allows more flexibility for dealing with events, as already linked in that PR, like [this](https://github.com/Philipp-M/trui/blob/dd24a2921eb0587dfffbbafea4a3e62b92f39c43/examples/async_event_handler_stream.rs#L67-L82)

As a novice and from a user's point of view, this looks like a very intuitive and elegant solution to me :+1: 
I will definitely take a closer look at your approach.

DJMcNab:
A composed async action handler is a really neat idea!

DJMcNab:
As a note on my philosophy for this async work - I really want to try and keep the async work declarative if at all possible - that has some really nice properties around cancellation and similar.

Philipp-M:
> A composed async action handler is a really neat idea!

Just for clarification, do you mean my first or my second point?

> As a note on my philosophy for this async work - I really want to try and keep the async work declarative if at all possible

I think we're aligned in this.

flosse:
I started playing with the EventHandler trait idea, but got stuck in a few places.
Unfortunately I'm away for a week now but will come back to it afterwards.

DJMcNab:
> Just for clarification, do you mean my first or my second point?

The second point.

Philipp-M:
@flosse I think you may be interested in #440 which does now implement the first of my points

flosse:
I am closing this PR in favor of #440 and #448.
Thanks @Philipp-M for taking up this topic!

PR #428, opened 2024-07-15
Add a `FlexElement` to support explicit spacers and flex parameters for `Flex` in xilem

Philipp-M:
Takes ideas from #235 (in particular the masonry side).

This implements support for explicit spacers in the `Flex` view and flex parameters for views (see examples `flex` and `mason` for more details).
It also adds a way (`AnyFlexChild`) to dynamically switch between spacers and widgets (with optional flex parameters).
`masonry::widget::Flex::layout` is also updated to be in sync with current druid.


xilem/src/lib.rs
```diff
@@ -139,6 +141,13 @@ pub trait WidgetView<State, Action = ()>:
     View<State, Action, ViewCtx, Element = Pod<Self::Widget>> + Send + Sync
 {
     type Widget: Widget;
+
+    fn flex(self, params: impl Into<FlexParams>) -> FlexItem<Self, State, Action>
```
Philipp-M, line 145: Having a builder method for this on the `WidgetViewTrait, is obviously open for discussion, I'm not sure of it yet:

What may or may not make sense from an API standpoint is to support these kinds of methods, when the returned element is also a `WidgetView`, `FlexItem` is not a `WidgetView`, so it can't be used outside of the context of a `Flex` container view.

masonry/src/widget/flex.rs
```diff
@@ -384,25 +362,20 @@ impl<'a> WidgetMut<'a, Flex> {
...
+        tracing::info!("{:?}", p);
+        let child = new_flex_child(p, child);
         self.widget.children.insert(idx, child);
-        // TODO
-        self.ctx.widget_state.children_changed = true;
-        self.ctx.widget_state.needs_layout = true;
+        self.ctx.children_changed();
     }
 
     // TODO - remove
```
Philipp-M, line 381: I guess that could be removed now?

masonry/src/widget/flex.rs
```diff
@@ -518,27 +572,49 @@ impl Widget for Flex {
...
                 Child::Fixed { widget, alignment } => {
-                    any_use_baseline &= *alignment == Some(CrossAxisAlignment::Baseline);
+                    // The BoxConstrains of fixed-children only depends on the BoxConstrains of the
+                    // Flex widget.
+                    let child_size = if bc_changed || ctx.child_needs_layout(widget) {
+                        let alignment = alignment.unwrap_or(self.cross_alignment);
+                        any_use_baseline |= alignment == CrossAxisAlignment::Baseline;
+
+                        let old_size = ctx.widget_state.layout_rect().size();
+                        let child_bc = self.direction.constraints(&loosened_bc, 0.0, 9000.0); // TODO Infinity does lead to NaN and then crashes the app, if this is too big, it leads to floating point issues
```
Philipp-M, line 596: I think I remember discussions about this, to avoid crashes (`NaN` and the likes), I set it to a high but not too high number (as vello goes crazy than, with weird floating point artifacts), the issue also exists for the current `layout` method

xilem/src/view/flex.rs
```diff
@@ -139,61 +141,255 @@ where
...
         ret
     }
 
     fn skip(&mut self, n: usize) {
-        self.ix += n * 2;
+        self.idx += n;
+    }
+}
+
+pub struct FlexItem<V, State, Action> {
```
Philipp-M, line 268: Because of parameterization issues, I needed to add these generic parameters.

I guess we need to do this a lot more for other views as well, to seamlessly support stuff like `Adapt` or in general inference for these variables, when it's not explicit somewhere...

masonry/src/widget/flex.rs
PoignardAzur: > to avoid crashes (NaN and the likes), I set it to a high but not too high number

Pretty sure there's [a Codeless Code fable](http://thecodelesscode.com/case/115) about this. :smiley: 

The fundamental problem is that Masonry's layout model is confused, not that Flex had the wrong maximum constraint. Though I guess this works in the meantime.

Philipp-M (commented):
Yeah that's better.

I think in the future it would be great to somehow implement a similar mechanism as `WidgetMut` to access the `Child` to avoid needing access to the parent/flex-container, but I don't think it should be done in this PR, and I'm still not sure if it's worth it...

xilem/src/view/flex.rs
```diff
@@ -139,61 +138,263 @@ where
...
+    fn message(
+        &self,
+        _: &mut Self::ViewState,
+        _: &[xilem_core::ViewId],
+        _: DynMessage,
+        _: &mut State,
+    ) -> xilem_core::MessageResult<Action> {
+        unreachable!()
     }
 }
```
DJMcNab, line 400: I think we're missing `AnyFlexView` here.

We could also support a `FlexChild` view, which would be:
```rust
enum FlexChild<V>
{
    Child(V, ...),
    Spacer(...)
}
```
although I don't think that has much value if we have `AnyFlexView`.

DJMcNab (approved):
I haven't carefully examined the Masonry level changes, but I also think that's fine, because we're planning on changing `Flex` to use Taffy at some point anyway.

I think that Masonry flex *needs* to support something akin to CSS `gap`, because currently this significantly regresses all of our examples. That doesn't need to happen in this PR, but it does need to happen extremely soon after this lands.

The Xilem level changes look as I'd expect. If you want to add `AnyFlexView` in this PR, that would be very useful @Philipp-M, but I'm not going to block on it.

I would like to get @PoignardAzur's sign-off on `WidgetMut::reborrow` before landing.

xilem/src/view/flex.rs
Philipp-M: You mean, that one could do something like this?:

```rust
fn my_flex_elem() -> FlexChild {...}
```

which would be dynamic, so that it could be for example used in a `Vec`?

DJMcNab: Well, either:
```rust
fn my_flex_elem() -> FlexChild<Box<AnyWidgetView<MyState>>> {...}
```
or
```rust
fn my_flex_elem() -> AnyFlexView<MyState> {...}
```
But essentially that, yes.

PoignardAzur:
Re: `reborrow_mut`, I wonder if you could skip the `is_reborrow` value and just apply merge_up multiple time. It's supposed to be idempotent anyway.

In theory it could have surprising effects if some other code reads the parent state and doesn't expect it to change; in practice the fact that WidgetMut holds a mut reference to said parent state means that will never happen.

PoignardAzur:
Otherwise, LGTM.

PoignardAzur (approved):

DJMcNab:
> Re: reborrow_mut, I wonder if you could skip the is_reborrow value and just apply merge_up multiple time. It's supposed to be idempotent anyway.

The `is_reborrow` method is entirely an optimisation, because merge_up is a relatively heavy operation

PoignardAzur:
> because merge_up is a relatively heavy operation

It's not that heavy. About a dozen boolean flags getting OR-ed. (That's 16 bytes with bools, 2 with bitblags).

The two non-trivial bits of logic are text registrations and cursor changes, both of which should probably be moved out of merge_up.

The more concerning part is this:

```rust
        self.text_registrations
            .append(&mut child_state.text_registrations);
```

Which we probably don't want to duplicate. Once we remove that line, we can remove the `is_reborrow` flag. In the meantime I'd suggest adding a todo comment, but we can merge either way.

Philipp-M:
> we can remove the `is_reborrow` flag.

Does it hurt otherwise, I mean it's just little bit of extra mental overhead, and I assume `WidgetMut::Drop` will get more expensive in the future.

Btw. I have started with the `AnyFlexView`, it's more complicated than I thought (requiring all the generational id stuff), I want to get that in here as well, as I see value for it too.

And I will likely also add the `FlexExt` stuff, which I agree with

PoignardAzur:
> Does it hurt otherwise, I mean it's just little bit of extra mental overhead,

Fair enough, it's a minor point.

> and I assume `WidgetMut::Drop` will get more expensive in the future.

I don't expect it to get much more expensive, but it's a fair point too.

xilem/src/view/flex.rs
```diff
@@ -131,69 +131,579 @@ where
...
+    /// ```
+    /// use xilem::{view::{flex, label, FlexSpacer, FlexExt, AnyFlexChild}};
+    /// # use xilem::{WidgetView};
+    ///
+    /// # fn view<State: 'static>() -> impl WidgetView<State> {
+    /// flex([label("a label").into_any_flex(), AnyFlexChild::Spacer(FlexSpacer::Fixed(1.0))])
+    /// # }
+    ///
+    /// ```
+    fn into_any_flex(self) -> AnyFlexChild<State, Action>
```
Philipp-M, line 314: I've tried to do this via a blanket impl, but it doesn't work, since I can't specialize over `View<Element = FlexElement>` and `WidgetView`, and otherwise that impl intersects, so I hardcoded these directly on either `V: WidgetView` as well as for either `FlexItem` and `FlexSpacer`...

xilem/examples/mason.rs
```diff
@@ -29,9 +32,30 @@ fn app_logic(data: &mut AppData) -> impl WidgetView<AppData> {
...
+                if data.active {
+                    FlexSpacer::Flex(x as f64)
+                } else {
+                    FlexSpacer::Fixed((count - x) as f64)
+                },
+            )
+        })
         .collect::<Vec<_>>();
+
+    let flexy_flex_sequence = [2, 3, 4].map(|c| {
```
Philipp-M, line 48: I guess this is our catch-all playground example? This also uses `[impl ViewSequence; N]` btw.

Philipp-M (commented):
I've started documenting (including doc tests, which at least check if all the typing checks out), added the `WidgetView::boxed` as I think it's an uncontroversal change, and the `FlexExt` trait, and of course added an `AnyFlexChild` view, which in its `impl View` logic is very similar to `impl ViewSequence for Option<impl ViewSequence>`. I've also pulled the `Axis` into `View` and reexported relevant enums for `flex_item` from masonry.

Since this is a bigger update, @DJMcNab I think it would make sense if you check at least the `impl View for AnyFlexChild`.

xilem/src/view/flex.rs
DJMcNab: Did you try `View<Element = Pod<W>>` and `View<Element = FlexElement>`? The first should be equivalent to `WidgetView`, but might be more friendly to the compiler

xilem/src/view/flex.rs
```diff
@@ -269,23 +270,110 @@ impl ElementSplice<FlexElement> for FlexSplice<'_> {
...
+impl<State, Action, V: WidgetView<State, Action>> FlexExt<State, Action> for V {}
+
+/// A `WidgetView` that can be used within a [`Flex`] [`View`]
 pub struct FlexItem<V, State, Action> {
     view: V,
     params: FlexParams,
     phantom: PhantomData<fn() -> (State, Action)>,
 }
 
+/// Applies [`impl Into<FlexParams>`](`FlexParams`) to the [`View`] `V`, can be used as child of a [`Flex`] [`View`]
```
DJMcNab, line 333:
```suggestion
/// Applies [`impl Into<FlexParams>`](`FlexParams`) to the [`View`] `V`, can be used as child of a [`Flex`] View
```

xilem/src/view/flex.rs
Philipp-M: Yep I tried, I did not make it to work, unfortunately...

xilem/src/view/flex.rs
```diff
@@ -269,23 +270,110 @@ impl ElementSplice<FlexElement> for FlexSplice<'_> {
...
+/// # fn view<State: 'static>() -> impl WidgetView<State> {
+/// flex((
+///     flex_item(button("click me", |_| ()), 2.0),
+///     FlexSpacer::Fixed(2.0),
+///     flex_item(label("a label"), CrossAxisAlignment::Fill),
+///     FlexSpacer::Fixed(2.0),
+/// ))
+/// # }
+///
+/// ```
```
DJMcNab, line 349: Ideally we'd have an image here. We really should set up `linebender.org/assets` at some point

(To be clear, this is fine as-is)

xilem/src/view/flex.rs
```diff
@@ -394,7 +489,221 @@ impl<State, Action> View<State, Action, ViewCtx> for FlexSpacer {
...
+        ctx: &mut ViewCtx,
+        mut element: Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element> {
+        match (prev, self) {
+            (AnyFlexChild::Item(prev), AnyFlexChild::Item(this)) => ctx
+                .with_id(ViewId::new(view_state.generation), |ctx| {
+                    this.rebuild(prev, view_state.inner.as_mut().unwrap(), ctx, element)
+                }),
+            (AnyFlexChild::Spacer(prev), AnyFlexChild::Spacer(this)) => {
+                View::<(), (), ViewCtx>::rebuild(this, prev, &mut (), ctx, element)
```
DJMcNab, line 607: We should have a comment at the spacer view that it is an invariant that this view doesn't deal with messages.

xilem/src/view/flex.rs
```diff
@@ -131,69 +131,580 @@ where
...
     }
 }
 
-struct FlexSplice<'f, 'w> {
-    ix: usize,
-    element: &'f mut WidgetMut<'w, widget::Flex>,
-    scratch: AppendVec<Pod<Box<dyn Widget>>>,
+pub enum FlexElement {
+    // Avoid making the enum massive for the spacer cases by boxing
+    Child(Box<Pod<Box<dyn Widget>>>, FlexParams),
```
DJMcNab, line 144: I think `FlexElement` is only transitory, so it doesn't matter if the spacer case is huge.
That is, we're never storing a `Vec<FlexElement>` for very long.

xilem/src/view/flex.rs
DJMcNab: Yeah, this is a choice we have to make, and I think adding them to the View type is the right one, to keep the type inference failures local.

xilem/examples/mason.rs
```diff
@@ -66,7 +90,7 @@ fn app_logic(data: &mut AppData) -> impl WidgetView<AppData> {
             toggleable(data),
             button("Decrement", |data: &mut AppData| data.count -= 1),
             button("Reset", |data: &mut AppData| data.count = 0),
-            flex(sequence).direction(axis),
+            flex((flex_sequence, flexy_flex_sequence)).direction(axis),
```
DJMcNab, line 93:
```suggestion
            flex((flexy_flex_sequence, flex_sequence)).direction(axis),
```

xilem/examples/mason.rs
DJMcNab: Yeah - it's fine to stick anything in here :laughing:

xilem/examples/mason.rs
```diff
@@ -32,9 +35,30 @@ fn app_logic(data: &mut AppData) -> impl WidgetView<AppData> {
...
+    let flexy_flex_sequence = [2, 3, 4].map(|c| {
+        if data.count.abs() % c == 0 {
+            FlexSpacer::Fixed(30.0 * c as f64).into_any_flex()
+        } else {
+            button(format!("flexy +{c}"), move |data: &mut AppData| {
+                data.count += c;
+            })
+            .into_any_flex()
+        }
+    });
```
DJMcNab, line 60: This feels `FizzBuzz` adjacent - not sure that we should action that though

DJMcNab (approved):
Thanks! A few minor feedback things, then we can land this.

xilem/src/view/flex.rs
Philipp-M: Well actually I had it before without double boxing, this is your code from #235, but I'm also fine without it.

xilem/src/view/flex.rs
Philipp-M: Yeah our documentation story can certainly be improved, but I hope this PR does improve it a little bit at least.

xilem/examples/mason.rs
Philipp-M: Ah yeah, lets make it `FizzBuzz` ^^

xilem/src/view/flex.rs
Philipp-M: I also wonder, if the doc-tests could be combined with snapshots, such that this really also tests the logic (while reducing the amount of manual tests) (But I doubt that this is simple, if at all possible within in reason)

masonry/src/widget/flex.rs
```diff
@@ -384,25 +362,20 @@ impl<'a> WidgetMut<'a, Flex> {
...
+    }
+
+    pub fn insert_flex_child_pod(
+        &mut self,
+        idx: usize,
+        child: WidgetPod<Box<dyn Widget>>,
+        params: impl Into<FlexParams>,
+    ) {
+        let p = params.into();
+        tracing::info!("{:?}", p);
```
DJMcNab, line 375: This needs to be removed.

PR #430, opened 2024-07-16
Update to Vello 0.2.1 and wgpu 0.20.1

DJMcNab:
Mostly to double check that the new release is suitable, but formalising the update can't hurt


Philipp-M (approved):
Works for me

PR #431, opened 2024-07-16
Allow Masonry to compile in release mode

DJMcNab:
Cc @xStrom for the CI change

Cc @PoignardAzur for the Masonry changes

### DIFF
--- a/masonry/src/widget/widget_pod.rs
+++ b/masonry/src/widget/widget_pod.rs
@@ -895,12 +895,18 @@ impl<W: Widget> WidgetPod<W> {
                 let (child_state, _) = state_token
                     .get_child(child_id)
                     .unwrap_or_else(|| panic!("widget #{child_id} not found"));
+                // The `widget_name` field is only available under debug_assertions - we don't want to
+                // #[cfg] out this entire block so that the same borrow-checking applies in release and debug.
+                #[cfg(not(debug_assertions))]
+                let widget_name = "UNREACHABLE";
+                #[cfg(debug_assertions)]
+                let widget_name = &child_state.widget_name;
                 if child_state.is_expecting_place_child_call {
                     debug_panic!(
                         "Error in '{}' #{}: missing call to place_child method for child widget '{}' #{}. During layout pass, if a widget calls WidgetPod::layout() on its child, it then needs to call LayoutCtx::place_child() on the same child.",
                         widget.short_type_name(),
                         id,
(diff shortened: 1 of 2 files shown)


masonry/src/widget/widget_pod.rs
```diff
@@ -895,12 +895,18 @@ impl<W: Widget> WidgetPod<W> {
                 let (child_state, _) = state_token
                     .get_child(child_id)
                     .unwrap_or_else(|| panic!("widget #{child_id} not found"));
+                // The `widget_name` field is only available under debug_assertions - we don't want to
+                // #[cfg] out this entire block so that the same borrow-checking applies in release and debug.
+                #[cfg(not(debug_assertions))]
+                let widget_name = "UNREACHABLE";
```
ratmice, line 901: looking at the `debug_panic!`  macro it panics in debug mode, but logs to tracing otherwise, I guess I'm concerned that this is going to actually log *UNREACHABLE*, where *UNKNOWN* might be more apt.
Otherwise, not knowing much about this I wonder whether the widget_name field really should be `cfg(debug_assertions)` if it is going to be used with `cfg(not(debug_assertions)))`.

ratmice: You are right I was speaking of the else branch of `if cfg!(debug_assertions)` in the `debug_panic!` macro,
it didn't occur to me that there would be another *runtime* check on the caller side here.

I agree that it is unreachable/resolving.

ratmice: Unresolving (though this is sort of a nit/suggestion that you should feel free to ignore)

I wonder if it wouldn't be better to use a `#[cfg!(debug_assertions)] { block` in place of `if cfg!(debug_assertions) {` removing the entire need for this `#[cfg(not(debug_assertions)] let widget_name = "UNREACHABLE";` and the other `#[cfg(debug_assertions)]`.

AFAICT since there isn't an `else` branch it should be possible to make this statically conditional.

DJMcNab: This is what I was trying to express in the comment I have added, and indeed why I have requested Olivier's review.

That would be my preferred initial code style, but I can definitely see the appeal in this code style as well. A conditionally compiled block is quite a weird thing, and e.g. this form plays more nicely with rust-analyzer and/or clippy.

PoignardAzur (approved):

PR #432, opened 2024-07-17
Import our intended set of lints

DJMcNab:
See discussion in [#linebender > Standard Lint set](https://xi.zulipchat.com/#narrow/stream/419691-linebender/topic/Standard.20Lint.20set)

Of note: I have taken steps to ensure that this can be practically reviewed by *not* applying most of the lints.
The commented out lints make good follow-ups


masonry/src/util.rs
```diff
@@ -15,15 +15,14 @@ use std::hash::Hash;
 ///
 /// It's useful when a backtrace would aid debugging but a crash can be avoided in release.
 macro_rules! debug_panic {
-    () => { ... };
-    ($msg:expr) => {
+    // () => { ... }; // TODO: What was this rule for?
```
PoignardAzur, line 18:
```suggestion
```

No idea where that's from.

Since it's not used anywhere and I'm pretty sure it wouldn't compile anyway, we can remove it.

PoignardAzur (approved):
LGTM.

Cargo.toml
```diff
@@ -37,6 +36,70 @@ rust.unexpected_cfgs = { level = "warn", check-cfg = [
...
+rust.non_local_definitions = "forbid"
+
+rust.unused_lifetimes = "warn"
+rust.unit_bindings = "warn"
+rust.unused_import_braces = "warn"
+rust.trivial_numeric_casts = "warn"
+rust.unused_macro_rules = "warn"
+rust.variant_size_differences = "warn"
+
+clippy.allow_attributes_without_reason = "warn"
```
Philipp-M, line 51: Hmm this, pollutes my diagnostics in my editor, without being able to action about it, since this does seem to be experimental, and thus CI fails when adding a reason (as seen [here](https://github.com/linebender/xilem/actions/runs/9999775115/job/27641321539#step:7:398)) , should we disable that for now, and enable it, when it's stable, maybe with directly fixing all of this, so that it doesn't contribute to OCD^^?

DJMcNab: We can comment this out.

However, the behaviour you're seeing is a bug in clippy. This should be changed in clippy to properly handle the MSRV for where this feature can be used.

Philipp-M: Hmm that's interesting (just updated nightly, and still the issue).
I don't use rust via rustup, but via [rust-overlay](https://github.com/oxalica/rust-overlay) maybe that's an issue, as there's no rustup-binary on my system. I mean it's not critical to me, I can just comment it out locally.

PR #435, opened 2024-07-18
Use `cargo-rdme` to sync README and documentation root

PoignardAzur:
This was remarkably painless.

I especially appreciate @taiki-e providing an install-action which does 90% of the CI work for us.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ on:
 
 jobs:
   fmt:
-    name: cargo fmt
+    name: check formatting
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -78,6 +78,15 @@ jobs:
 
       - name: check copyright headers
         run: bash .github/copyright.sh
+
+      - name: install cargo-rdme
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-rdme
+
+      - name: cargo rdme
+        run: |
+          cargo rdme --check --workspace-project=masonry
 
   clippy-stable:
     name: cargo clippy
--- a/masonry/README.md
+++ b/masonry/README.md
@@ -1,9 +1,5 @@
 <div align="center">
 
-# Masonry
-
-**Foundational framework for Rust GUI libraries**
-
 [![Latest published version.](https://img.shields.io/crates/v/masonry.svg)](https://crates.io/crates/masonry)
 [![Documentation build status.](https://img.shields.io/docsrs/masonry.svg)](https://docs.rs/masonry)
 [![Apache 2.0 license.](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](#license)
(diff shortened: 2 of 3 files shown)


PoignardAzur:
Not sure what's up with the "expected check". I guess this is because I renamed one of the CI tasks, but I don't know if that will block a merge or what the proper way to do a rename is.

masonry/README.md
```diff
@@ -12,12 +8,18 @@
 [![GitHub Actions CI status.](https://img.shields.io/github/actions/workflow/status/linebender/xilem/ci.yml?logo=github&label=CI)](https://github.com/linebender/xilem/actions)
 [![Dependency staleness status.](https://deps.rs/crate/masonry/latest/status.svg)](https://deps.rs/crate/masonry)
 
+# Masonry
```
DJMcNab, line 11: It's not nice for this to not be above the badges

masonry/README.md
```diff
@@ -12,12 +8,18 @@
...
+
 </div>
 
-Masonry gives you a platform to create a window (using [winit] as a backend) with a tree of widgets.
-It also gives you tools to inspect that widget tree at runtime, write unit tests on it, and generally have an easier time debugging and maintaining your app.
+<!-- cargo-rdme start -->
+
+**A foundational framework for Rust GUI libraries.**
```
DJMcNab, line 17: My personal preference is for this line to also be above the badges. That does require moving the badges into the lib.rs documentation, but I think that's fine.

Either way, I think this should still be centered

.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ on:
 
 jobs:
   fmt:
-    name: cargo fmt
+    name: check formatting
```
DJMcNab, line 60: I think I'd slightly prefer a more "pithy" name - this is already a GitHub "check".

```suggestion
    name: formatting
```
I do agree that this has outgrown the `cargo fmt` name.

DJMcNab (approved):
Looks good. Once we settle on the new name, I can fixup the branch protection

I do want to reorder some of the items slightly, but we can have this discussion on Zulip afterwards if needed.

PoignardAzur:
Some screenshots to feed the discussion:

## Masonry README before the change

image

## Masonry README after the change

image

## Doc page with the blurb centered

image

## Doc page with badges and the blurb, both centered

image

---

Overall my takeaways are:

- Anything centered looks bad in the generated doc.
- In the README, having badges below the title and blurb does look better.
- I don't think the doc really needs the blurb.

Will create follow-up PR soon based on those takeaways.

PoignardAzur:
New version pushed.

[RENDERED.](https://github.com/linebender/xilem/blob/1bd6d376575529277196a3add91a313638b65e58/masonry/README.md)

masonry/src/lib.rs
```diff
@@ -1,15 +1,18 @@
 // Copyright 2018 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//! A framework that aims to provide the foundation for Rust GUI libraries.
-//!
 //! Masonry gives you a platform to create windows (using [winit] as a backend) each with a tree of widgets. It also gives you tools to inspect that widget tree at runtime, write unit tests on it, and generally have an easier time debugging and maintaining your app.
```
DJMcNab, line 4: I think this first sentence should be rewritten, but that's not required in this PR.

DJMcNab (approved):
We're still blocking on the name for the CI check. If you want to leave it as-is, please indicate as such, and we can land this.

PoignardAzur:
> We're still blocking on the name for the CI check

Changed the name as per your commit. I'd say this is good to merge.

PR #436, opened 2024-07-19
Change the `OneOf` views to not use macros

DJMcNab:
This still needs to be implemented in Xilem.

But I think this is an improvement overall

### DIFF
--- a/xilem_core/src/views/one_of.rs
+++ b/xilem_core/src/views/one_of.rs
@@ -8,7 +8,9 @@ use hidden::OneOfState;
 
 /// This trait allows, specifying a type as `ViewElement`, which should never be constructed or used,
 /// but allows downstream implementations to adjust the behaviour of [`PhantomElementCtx::PhantomElement`],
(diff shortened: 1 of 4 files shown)


xilem_core/src/views/one_of.rs
```diff
@@ -110,36 +135,438 @@ pub trait OneOfCtx<
...
+
+    impl<State, Action, Context: PhantomElementCtx, Message> View<State, Action, Context, Message>
+        for Never
+    {
+        type Element = Context::PhantomElement;
+
+        type ViewState = Never;
+
+        fn build(&self, _: &mut Context) -> (Self::Element, Self::ViewState) {
+            match *self {}
```
Philipp-M, line 533: That's a nice trick!

Philipp-M (approved):
Nice, that definitetly improves all of that, my local tests also work as expected, great improvement!

xilem_core/src/views/one_of.rs
DJMcNab: Yeah - uninhabited types have some neat properties. We're proving that an impossible type has been constructed, and therefore that this code path was never entered.

PR #438, opened 2024-07-19
Correctly set `rust-version` in all `Cargo.toml`s

DJMcNab:
Discovered in #432; this would otherwise cause errors for users of nightly (or 1.81 beta, due to be released next week)


Philipp-M (approved):
LGTM, I think it also aligns with our version matrix RFC IIRC.

PR #439, opened 2024-07-19
Expose control over window background color

azymohliad:
Adds `Xilem::background_color` builder method and the relevant changes to Masonry. I've also modified `mason` example to demonstrate the usage of the new API. 

[Zulip discussion](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Window.20background.20color)

### DIFF
--- a/xilem/examples/mason.rs
+++ b/xilem/examples/mason.rs
@@ -140,8 +140,9 @@ fn run(event_loop: EventLoopBuilder) {
         active: false,
     };
 
-    let app = Xilem::new(data, app_logic);
-    app.run_windowed(event_loop, "First Example".into())
+    Xilem::new(data, app_logic)
+        .background_color(Color::rgb8(0x20, 0x20, 0x20))
+        .run_windowed(event_loop, "First Example".into())
         .unwrap();
 }
 
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -61,6 +61,7 @@ where
         }
     }
 
+    /// Sets main window background color.
     pub fn background_color(mut self, color: Color) -> Self {
         self.background_color = color;
         self
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -43,6 +43,7 @@ pub struct Xilem<State, Logic> {
     state: State,
     logic: Logic,
     runtime: tokio::runtime::Runtime,
+    background_color: Color,
 }
 
 impl<State, Logic, View> Xilem<State, Logic>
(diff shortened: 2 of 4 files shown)


xilem/src/lib.rs
```diff
@@ -56,9 +57,15 @@ where
             state,
             logic,
             runtime,
+            background_color: Color::BLACK,
         }
     }
 
+    pub fn background_color(mut self, color: Color) -> Self {
```
DJMcNab, line 64: This should have a very brief documentation comment

xilem/examples/mason.rs
```diff
@@ -140,8 +140,9 @@ fn run(event_loop: EventLoopBuilder) {
         active: false,
     };
 
-    let app = Xilem::new(data, app_logic);
```
DJMcNab, line 143: I do agree with the change to not use the `app` binding here, it's unnecessary.

xilem/examples/mason.rs
```diff
@@ -140,8 +140,9 @@ fn run(event_loop: EventLoopBuilder) {
...
-    app.run_windowed(event_loop, "First Example".into())
+    Xilem::new(data, app_logic)
+        .background_color(Color::rgb8(0x20, 0x20, 0x20))
```
DJMcNab, line 144: This change isn't that obvious when viewing the example. I'd suggest maybe instead updating the `flex` example to have a white background with black text.

DJMcNab (approved):
This looks good, thanks. As I mentioned on Zulip, this API is very likely to change at some point, but meanwhile this is a useful capability to expose.

There are a few peripheral tweaks I've suggested, but the actual code change is exactly how I would implement this.

xilem/examples/mason.rs
azymohliad: I'm ready to commit the change to the `flex` and revert this, but are you sure you want white background there? Black buttons would look a bit out of place there until they can be styled too. I think it's fine if the change isn't very obvious visually, it's more important to demonstrate the code. And IMO dark grey looks better than black for window background anyway (at least it's more common), so such change would still have some value. Wdyt?

azymohliad: Here's how the three options look:
- Current:
Screenshot from 2024-07-22 19-35-39
- `Color::rgb8(0x20, 0x20, 0x20)`
Screenshot from 2024-07-22 19-36-01
- `Color::WHITE`
Screenshot from 2024-07-22 19-36-38

azymohliad: Seems like not only Xilem Button view, but also Masonry Button widget do not expose control over its inner label text size. I guess I'll only commit the docstring for `Xilem::background_color`, and leave the examples as is then?

PR #442, opened 2024-07-22
xilem: Fix intradoc link.

waywardmonkeys:


waywardmonkeys:
We're not denying warnings from `cargo doc` because masonry still has a lot of warnings when building docs.

Philipp-M (approved):
> We're not denying warnings from cargo doc because masonry still has a lot of warnings when building docs.

Yeah I guess we should fix that sooner than later... (so that something like this is not happening)

PR #443, opened 2024-07-22
Tweak `async_repeat` and add some docs

DJMcNab:
Addresses post-merge review comments in #423.

Also added some stub-like docs.

### DIFF
--- a/xilem/src/view/async_repeat.rs
+++ b/xilem/src/view/async_repeat.rs
@@ -18,7 +18,7 @@ use crate::ViewCtx;
 /// Note that this task will not be updated if the view is rebuilt, so `future_future`
 /// cannot capture.
 // TODO: More thorough documentation.
-/// See [run_once](crate::core::run_once) for details.
+/// See [`run_once`](crate::core::run_once) for details.
 pub fn async_repeat<M, F, H, State, Action, Fut>(
     future_future: F,
     on_event: H,
@@ -36,6 +36,27 @@ where
             To ignore this warning, use `async_repeat_raw`."
         );
     };
+    AsyncRepeat {
+        future_future,
+        on_event,
+        message: PhantomData,
+    }
+}
+
+/// Launch a task which will run until the view is no longer in the tree.
+///
+/// This is [`async_repeat`] without the capturing rules.
+/// See `async_repeat` for full documentation.
+pub fn async_repeat_raw<M, F, H, State, Action, Fut>(
+    future_future: F,
+    on_event: H,
+) -> AsyncRepeat<F, H, M>
+where
+    F: Fn(MessageProxy<M>) -> Fut,
+    Fut: Future<Output = ()> + Send + 'static,
+    H: Fn(&mut State, M) -> Action + 'static,
+    M: Message + 'static,
+{
     AsyncRepeat {
         future_future,
         on_event,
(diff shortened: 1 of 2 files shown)


xilem/src/view/async_repeat.rs
```diff
@@ -8,6 +8,17 @@ use xilem_core::{DynMessage, Message, MessageProxy, NoElement, View, ViewId, Vie
...
+/// This `MessageProxy` can be used to send a message to `on_event`, which can then update
+/// the app's state.
+///
+/// For exampe, this can be used with the time functions in [`crate::tokio::time`].
+///
+/// Note that this task will not be updated if the view is rebuilt, so `future_future`
+/// cannot capture.
+// TODO: More thorough documentation.
+/// See [run_once](crate::core::run_once) for details.
 pub fn async_repeat<M, F, H, State, Action, Fut>(
```
Philipp-M, line 22: Regarding the name, maybe this:
```suggestion
pub fn relay<M, F, H, State, Action, Fut>(
```
or `spam`^^ or maybe something like `emit_proxy`.

Hmm difficult to find a good name for this...

xilem/src/view/async_repeat.rs
```diff
@@ -18,6 +29,13 @@ where
     H: Fn(&mut State, M) -> Action + 'static,
     M: Message + 'static,
 {
+    const {
+        assert!(
+            core::mem::size_of::<F>() == 0,
+            "`async_repeat` will not be ran again when its captured variables are updated.\n\
+            To ignore this warning, use `async_repeat_raw`."
```
Philipp-M, line 36: Yeah that's good to have, but there's no `async_repeat_raw`, and I don't think we should have it.
Same still applies to the `run_once_raw` IMO, though it doesn't hurt I guess, as it's not something the user reaches for first, and the user is warned with this assertion at least.

xilem/src/view/async_repeat.rs
```diff
@@ -65,12 +82,11 @@ where
 
     fn teardown(
         &self,
-        _: &mut Self::ViewState,
+        view_state: &mut Self::ViewState,
```
Philipp-M, line 85: I mean that's probably one of the nittiest nits:
```suggestion
        join_handle: &mut Self::ViewState,
```
as I think, then it's more obvious, what the view state is (as it's otherwise obvious via the type that it's the view state).

xilem/src/view/async_repeat.rs
```diff
@@ -65,12 +82,11 @@ where
...
         _: &mut ViewCtx,
         _: xilem_core::Mut<'_, Self::Element>,
     ) {
-        // Nothing to do
-        // TODO: Our state will be dropped, finishing the future
+        view_state.abort();
```
Philipp-M, line 89:
```suggestion
        join_handle.abort();
```

Philipp-M (approved):
Yeah that's better.

xilem/src/view/async_repeat.rs
DJMcNab: Yeah. I don't like either of the options for different reasons, but I think this is better

xilem/src/view/async_repeat.rs
DJMcNab: I do agree that this name is not ideal.
I want to avoid it being something too abstract.

DJMcNab: I'm going to merge and we follow-up with a rename if needed

PR #445, opened 2024-07-23
Make `OneOf` work in Xilem Masonry

DJMcNab:
This uses the interface in Xilem Core from #394 and #436

Some thoughts:
1) You only need to specify the number of items in a single match arm. I.e.
    ```rust
    fn my_view(){
        match x {
            0 => OneOf3(...),
            1=> OneOf9(...),
            _=> OneOf9(...),
        }
    }
    ```
    works. We probably should rename `OneOf9` back again in that case.
2) The example currently isn't that interesting. Any suggestions for a more suitable state machine would be welcome.

### DIFF
--- a/xilem/src/one_of.rs
+++ b/xilem/src/one_of.rs
@@ -6,7 +6,7 @@
 use accesskit::Role;
 use masonry::{
     AccessCtx, AccessEvent, BoxConstraints, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx,
-    PointerEvent, Size, StatusChange, TextEvent, Widget, WidgetId, WidgetPod,
+    Point, PointerEvent, Size, StatusChange, TextEvent, Widget, WidgetId, WidgetPod,
 };
 use smallvec::{smallvec, SmallVec};
 use vello::Scene;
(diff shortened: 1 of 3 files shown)


Philipp-M:
> We probably should rename `OneOf9` back again in that case.

I need to check that in detail later, but a quick thought, I think it currently works (I might be wrong, did only a super quick skim), because we type erase it for the other elements, so I think the intersection trait stuff I've implemented in xilem_web would not work then?

DJMcNab:
I think it would work equivalently?
I'm not sure I understand what you're saying

xilem/src/one_of.rs
```diff
@@ -0,0 +1,373 @@
...
+            OneOfWidget::G(w) => elem_mut.ctx.remove_child(w),
+            OneOfWidget::H(w) => elem_mut.ctx.remove_child(w),
+            OneOfWidget::I(w) => elem_mut.ctx.remove_child(w),
+        }
+        elem_mut.ctx.children_changed();
+    }
+}
+
+impl crate::core::one_of::PhantomElementCtx for ViewCtx {
+    type PhantomElement = Pod<Box<dyn Widget>>;
```
Philipp-M, line 190: I guess when we would do something similar as in xilem_web (like `WithStyle`, `WithClasses` etc.) we would need some `PhantomWidget` which implements all kinds of these traits, similar as `Noop` in xilem_web.

xilem/src/one_of.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+#[allow(unnameable_types)] // Public because of trait visibility rules, but has no public API.
+pub enum OneOfWidget<A, B, C, D, E, F, G, H, I> {
```
Philipp-M, line 194: I guess the orphan rule has hit us again?

Philipp-M (approved):
> I think it would work equivalently?

Ah yeah, I misread/misunderstood what you meant I think? Like you still need `OneOfN` where each variant in `N` needs to have a concrete type, but you can then just use `OneOf9` everywhere, as long as you only use those specified variants of `N`.
Sure lets just rename it to `OneOf` again. I'm not sure though if it makes sense to mix `OneOf` and `OneOfN`, even if it's possible, might be confusing for the user...

> Any suggestions for a more suitable state machine would be welcome.

Hmm, after a quick google, maybe something like [this](https://www.cs.cmu.edu/~dst/Tekkotsu/Tutorial/images/statemachine1.png). I think we could add some async here as well, like if it's in the bark state, it creates an `async_repeat` which "barks", in the "wait" state, something like adding some points: ... to a text. Though I think something like `run_once_async` would be more appropriate for autoswitching the states (but we could do that just with the buttons).
Anyway just some ideas, I don't think examples need to be particularly interesting as long as they effectively communicate how to use stuff (but ideally they're also interesting...).


Looks good otherwise, feel free to merge, when you're satisfied.

xilem/src/one_of.rs
DJMcNab: Yes, indeed. That would be unneeded complexity as-is, though.

xilem/src/one_of.rs
DJMcNab: You're suggesting that we would implement `Widget` for `OneOf` directly? I suppose we *could* do that, but it seems hard to make the docs sensible in that case.

xilem/src/one_of.rs
Philipp-M: > That would be unneeded complexity as-is

You mean something like `Noop: Widget`, since `Box<dyn Widget>` already implements `Widget`?

xilem/src/one_of.rs
Philipp-M: I'm not suggesting anything, just noticing.

But this:

> but it seems hard to make the docs sensible in that case.

is a good argument against reusing `OneOf` as widget.
Though now I'm curious, how `impl Widget for OneOf` should work, since either of those types are defined in different crates?

Philipp-M: I mean apart from a wrapper type...

DJMcNab: Yeah, it couldn't work anyway - I'm just saying another reason it doesn't make much sense.

xilem/src/one_of.rs
Philipp-M: Yeah I agree. That should only be done, when it really makes sense.

PR #448, opened 2024-07-24
xilem_web: Add a `MemoizedAwait` view

Philipp-M:
This is the `rerun_on_change` view described in #440, I named it `MemoizedAwait` as I think that fits its functionality.

It got also additional features `debounce_ms` (default `0`) and `reset_debounce_on_update` (default `true`) as I think that's quite useful, in case a lot of updates are happening.

When `reset_debounce_on_update == false`, `debounce` is more a throttle than an actual debounce.

This also adds a more heavily modified version of the example added in #427, also showing `OneOf` in xilem_web (which didn't have an example yet).

This *could* be considered as alternative to #440 (at the very least the example) but those approaches could also well live next to each other.


xilem_web/src/attribute_value.rs
```diff
@@ -11,6 +11,7 @@ pub enum AttributeValue {
     True, // for the boolean true, this serializes to an empty string (e.g. for <input checked>)
     I32(i32),
     U32(u32),
+    Usize(usize),
```
Philipp-M, line 14: I don't think that change is worth its own PR, I might add more types in a different PR though, as I think all values that make sense here should be supported IMO.

xilem_web/src/lib.rs
```diff
@@ -53,6 +54,7 @@ pub use attribute_value::{AttributeValue, IntoAttributeValue};
 pub use class::{AsClassIter, Class, Classes, ElementWithClasses, WithClasses};
 pub use context::ViewCtx;
 pub use element_props::ElementProps;
+pub use memoized_await::{memoized_await, MemoizedAwait};
```
Philipp-M, line 57: I'm still not sure about the actual namespacing in xilem_web (and I think this is certainly also relevant for xilem), this exports `MemoizedAwait` which is probably nice for explicitely naming types, though I doubt that will likely happen (mostly because of the countless generic params necessary), So I'm not sure whether that type should even be exported, as it pollutes the namespace, but I slightly lean to exporting it.

I'm not sure, whether I should just create a new `mod` `view` and put all things that implement a `View` + things relevant for those into it, that means also `elements`, but I could also imagine just namespacing views in the root, e.g. with this view in `xilem_web::concurrent`...

xilem_web/src/memoized_await.rs
```diff
@@ -0,0 +1,248 @@
...
+    #[allow(clippy::type_complexity)]
+    phantom: PhantomData<fn() -> (State, Action, OA, F, FOut)>,
+}
+
+impl<State, Action, OA, InitFuture, Data, CB, F, FOut>
+    MemoizedAwait<State, Action, OA, InitFuture, Data, CB, F, FOut>
+{
+    /// Debounce the `init_future` function, when `data` updates,
+    /// when `reset_debounce_on_update == false` then this throttles updates each `millisecconds`
+    pub fn debounce(mut self, milliseconds: usize) -> Self {
```
Philipp-M, line 28: So this doesn't debounce when this view is spawned (e.g. as in example `(state.cats_to_fetch < TOO_MANY_CATS).then_some(memoized_await(..))`), I'm not sure whether it actually *should* debounce in `View::build`, so that it has the same behavior when it's in the view tree, as well as when it's being spawned.

Philipp-M (commented):
I can add the same Api (or different depending on discussion on this PR?) for xilem itself, as I think it can be ported mostly 1:1 to xilem (with additional `Send + Sync` bounds I think, `debounce` could probably also be a `Duration`).

xilem_web/src/memoized_await.rs
```diff
@@ -0,0 +1,248 @@
...
+        self.debounce = milliseconds;
+        self
+    }
+
+    /// When `reset` is `true`, everytime `data` updates, the debounce timeout is cleared until `init_future` is invoked.
+    /// This is only effective when `debounce > 0`
+    pub fn reset_debounce_on_update(mut self, reset: bool) -> Self {
```
Philipp-M, line 35: Thinking about it, mabye this should be: 
```suggestion
    pub fn reset_debounce_timout_on_update(mut self, reset: bool) -> Self {
```

xilem_web/web_examples/fetch/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+};
+
+use gloo_net::http::Request;
+use serde::{Deserialize, Serialize};
+
+const TOO_MANY_CATS: usize = 8;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Cat {
+    pub url: String,
```
DJMcNab, line 20: We should have the width and the height as well, to avoid the jumping around

DJMcNab (approved):
Only given a cursory review as this is web code, but it's looking ok.

I feel like the fetching state management is more imperative than I'd like it to be, but I'm not sure how that can be resolved.

xilem_web/src/lib.rs
Philipp-M: Continuing loud thoughts: I've moved the view now in `xilem_web::concurrent`, as I think that this cleans it up a little bit and is more future proof.

xilem_web/src/memoized_await.rs
Philipp-M: I've did the debounce now in `View::build` and slightly refactored the `View` impl as well, as I think this should be able to be controlled.

When this behavior is not desired, it could be disabled by the user with `.debounce(0)` at the beginning and reenabled in the future callback (not that this is super awesome, but it at a least allows the user to decide). In the example it "feels" better with having the debounce in the beginning (when going over the `TOO_MANY_CATS` boundary).

xilem_web/src/memoized_await.rs
```diff
@@ -0,0 +1,248 @@
...
+use xilem_core::{MessageResult, Mut, NoElement, View, ViewId, ViewPathTracker};
+
+use crate::{DynMessage, OptionalAction, ViewCtx};
+
+/// Await a future returned by `init_future` invoked with the argument `data`, `callback` is called with the output of the future. `init_future` will be invoked again, when `data` changes. Use [`memoized_await`] for construction of this [`View`]
+pub struct MemoizedAwait<State, Action, OA, InitFuture, Data, Callback, F, FOut> {
+    init_future: InitFuture,
+    data: Data,
+    callback: Callback,
+    debounce: usize,
```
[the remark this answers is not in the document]
Philipp-M: I guess it should indeed be (I guess I was optimizing at the wrong location), I think that answers the question above for a potential xilem implementation.

The only thing, I don't really like about having `Duration` here, apart from minimal more boilerplate, is that it's not accurate below 1millisecond and is an integer value representing the time in milliseconds in `setTimeout`, so `usize` is at least from a technical perspective more correct I think.

xilem_web/web_examples/fetch/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+                    td(label("Debounce timeout in ms:").for_("debounce-timeout-duration")),
+                    td(input(())
+                        .id("debounce-timeout-duration")
+                        .attr("type", "number")
+                        .attr("min", 0)
+                        .attr("value", state.debounce_in_ms)
+                        .on_input(|state: &mut AppState, ev: web_sys::Event| {
+                            state.debounce_in_ms = input_target(&ev).value().parse().unwrap_or(0);
+                        })),
+                )),
```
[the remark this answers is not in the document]
Philipp-M: You mean the amount of code for the controls, or the ability to change these variables? I think it shows the whole API of the new `MemoizedAwait` view in an interactive way to quickly see what those variables are about, and directly serves as (yet another) slightly bigger example in action?

xilem_web/web_examples/fetch/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+                tr((
+                    td(label("How many cats would you like?").for_("cat-count")),
+                    td(input(())
+                        .id("cat-count")
+                        .attr("type", "number")
+                        .attr("min", 0)
+                        .attr("value", state.cats_to_fetch)
+                        .on_input(|state: &mut AppState, ev: web_sys::Event| {
+                            if !state.cats_are_being_fetched {
+                                state.cats.clear();
```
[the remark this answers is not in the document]
Philipp-M: Yeah you're right, though I rather wanted to have a clear visual state when images are fetched (i.e. I consider that a feature not a bug^^), to see when a new request is started, so that they don't slowly pop in over the previous images, when they're being loaded, I think the "weirdness" will be further amplified, when setting width/height on each.

Ideally, I think the `fetch_cats` prefetches the images, and when it's resolved all images are changed at once in one frame. But that requires quite a bit [more code](https://users.rust-lang.org/t/a-future-for-loading-images-via-web-sys/42370/2) (I actually tried just something like the equivalent of js: `Promise.all([fetch(), fetch()])` but that didn't seem to be enough, as the images were redownloaded, probably different headers or something like that).

xilem_web/web_examples/fetch/src/main.rs
Philipp-M: Oh, I somehow missed, that this is provided by the API, yeah I agree, thanks for the catch.

Philipp-M:
> I feel like the fetching state management is more imperative than I'd like it to be, but I'm not sure how that can be resolved.

Hmm you mean in the example, like `cats_are_being_fetched`? Maybe with more sugarized and specialized views, but I'm also not sure, I probably look into more sugarized views too, which may help for common use-cases. But I think there needs to be some state for the user that they know when the future has started, and when it has ended...

Any thoughts on a similar API regarding the `MemoizedAwait` for xilem itself?

xilem_web/web_examples/fetch/src/main.rs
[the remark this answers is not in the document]
DJMcNab: It was not very obvious what changing the parameters actually did, and it's quite a lot of code for things which are peripheral to the actual logic.

Maybe the right solution is to move this further down the file?

DJMcNab:
It would make sense to have a similar API in Xilem.

flosse:
@Philipp-M Great work!

One small idea: If the fetch state logic is separated from the view, then the code is easier to read for me.
Here is how I'd do this:

https://github.com/linebender/xilem/compare/main...slowtec:xilem:use-fetch-state

If you like it, I can open it as a PR :wink:

Philipp-M:
Sure that's probably slightly more readable, I agree, so you could indeed upstream that if you want.

flosse:
Here it is: #454

PR #450, opened 2024-07-26
Update dependencies and bump to Rust 1.80

DJMcNab:
There is one "semi-false-positive" lint triggered, which I have fixed.
Otherwise, the required
```
cargo upgrade --ignore-rust-version
cargo update
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,41 +26,60 @@ homepage = "https://xilem.dev/"
 # Remove assigning_clones once it's allowed by default in stable Rust
 # https://github.com/rust-lang/rust-clippy/pull/12779
 clippy.assigning_clones = "allow"
+# Remove assigning_clones once it's allowed by default in stable Rust
+# https://github.com/rust-lang/rust-clippy/pull/12779
+clippy.allow_attributes_without_reason = "warn"
+# Remove assigning_clones once it's allowed by default in stable Rust
+# https://github.com/rust-lang/rust-clippy/pull/12779
+clippy.collection_is_never_read = "warn"
+# Remove assigning_clones once it's allowed by default in stable Rust
+# https://github.com/rust-lang/rust-clippy/pull/12779
+clippy.debug_assert_with_mut_call = "warn"
+# Remove assigning_clones once it's allowed by default in stable Rust
+# https://github.com/rust-lang/rust-clippy/pull/12779
+clippy.fn_to_numeric_cast_any = "forbid"
+# Remove assigning_clones once it's allowed by default in stable Rust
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -26,42 +26,61 @@ homepage = "https://xilem.dev/"
 # Remove assigning_clones once it's allowed by default in stable Rust
 # https://github.com/rust-lang/rust-clippy/pull/12779
 clippy.assigning_clones = "allow"
+# Remove assigning_clones once it's allowed by default in stable Rust
+# https://github.com/rust-lang/rust-clippy/pull/12779
```
Philipp-M, line 30: Why all the comments (including below)?

xilem_core/examples/filesystem.rs
```diff
@@ -166,11 +166,11 @@ impl<State, Action> View<State, Action, ViewCtx> for File {
     ) -> Mut<'el, Self::Element> {
         if prev.name != self.name {
             let new_path = ctx.current_folder_path.join(&*self.name);
-            let _ = std::fs::rename(&element, &new_path);
+            let _ = std::fs::rename(&*element, &new_path);
```
Philipp-M, line 169: :+1:  (I was already about to do this as PR, as I was annoyed in my editor, but was too lazy...)

Philipp-M (approved):
LGTM, but I'd probably change the title, such that it also reflects the other package updates

Cargo.toml
DJMcNab: Oh seriously? I trusted `cargo upgrade` to actually do the right thing. That's incredibly annoying.

Philipp-M: Ok, that's a weird, probably open an issue in cargo upstream then?

waywardmonkeys:
Why bundle the update to Rust with the other? (Just asking.)

DJMcNab:
This is the policy we decided upon in https://github.com/linebender/rfcs/pull/5

Cargo.toml
DJMcNab: It's part of [cargo-edit](https://github.com/killercup/cargo-edit). I believe that the in-place toml editing doing stupid things with comments/ordering is known, although I can't find where it's tracked at the moment.

Philipp-M:
> Why bundle the update to Rust with the other? (Just asking.)

> This is the policy we decided upon in https://github.com/linebender/rfcs/pull/5

I believe he meant, why not separate PRs for package-updates and rust-version, but I think having all of this in one PR is good for me too, as those update changes seem to be largely uncontroversial/maintenance

PR #451, opened 2024-07-27
xilem_core: Refactor Memoization (add `Frozen` view, fix force-rebuild in `Arc`, and cleanup `Memoize`)

Philipp-M:
This adds a `force` which is basically a specialized `Memoize` without any bound data.
This also cleans up `Memoize` a little bit (mostly code-aesthetic stuff, but also adds `State` and `Action` params to the view, so that it doesn't offer weird surprises when composing it with views like `Adapt`).

The `Arc` view is also fixed, as it didn't support force rebuilding yet, which is relevant for e.g. async, e.g. the `MemoizedAwait` view would not work inside an `Arc<impl View>` currently.

### DIFF
--- a/xilem/examples/memoization.rs
+++ b/xilem/examples/memoization.rs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::sync::Arc;
-use xilem::view::{button, flex, memoize};
+use xilem::{
+    core::{memoize, static_view},
+    view::{button, flex},
+};
 use xilem::{AnyWidgetView, EventLoop, WidgetView, Xilem};
 
 // There are currently two ways to do memoization
@@ -47,8 +50,9 @@ fn decrease_button(state: &AppState) -> impl WidgetView<AppState> {
     })
 }
 
+// When there's no access to the app state, i.e. the view is "static", you can use `static_view` which is basically a Memoize view without data
 fn reset_button() -> impl WidgetView<AppState> {
-    button("reset", |data: &mut AppState| data.count = 0)
+    static_view(|| button("reset", |data: &mut AppState| data.count = 0))
 }
 
 fn app_logic(state: &mut AppState) -> impl WidgetView<AppState> {
--- a/xilem/examples/memoization.rs
+++ b/xilem/examples/memoization.rs
@@ -50,8 +50,9 @@ fn decrease_button(state: &AppState) -> impl WidgetView<AppState> {
     })
 }
 
-// When there's no access to the app state, i.e. the view is "static", you can use `frozen` which is basically a Memoize view without data
 fn reset_button() -> impl WidgetView<AppState> {
+    // The contents of this view never changes, so we use `frozen` to avoid unnecessary rebuilds.
+    // This is a special case of memoization for when the view doesn't depend on any data.
     frozen(|| button("reset", |data: &mut AppState| data.count = 0))
 }
 
--- a/xilem/examples/memoization.rs
+++ b/xilem/examples/memoization.rs
@@ -3,7 +3,7 @@
 
 use std::sync::Arc;
 use xilem::{
-    core::{memoize, static_view},
+    core::{frozen, memoize},
     view::{button, flex},
 };
 use xilem::{AnyWidgetView, EventLoop, WidgetView, Xilem};
@@ -50,9 +50,9 @@ fn decrease_button(state: &AppState) -> impl WidgetView<AppState> {
     })
 }
 
-// When there's no access to the app state, i.e. the view is "static", you can use `static_view` which is basically a Memoize view without data
+// When there's no access to the app state, i.e. the view is "static", you can use `frozen` which is basically a Memoize view without data
 fn reset_button() -> impl WidgetView<AppState> {
-    static_view(|| button("reset", |data: &mut AppState| data.count = 0))
+    frozen(|| button("reset", |data: &mut AppState| data.count = 0))
 }
 
 fn app_logic(state: &mut AppState) -> impl WidgetView<AppState> {
--- a/xilem_core/src/view.rs
+++ b/xilem_core/src/view.rs
@@ -180,6 +180,11 @@ where
     }
 }
 
+pub struct ArcState<ViewState> {
+    view_state: ViewState,
+    dirty: bool,
+}
+
 /// An implementation of [`View`] which only runs rebuild if the states are different
 impl<State, Action, Context, Message, V> View<State, Action, Context, Message> for Arc<V>
 where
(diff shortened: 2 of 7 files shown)


xilem_core/src/views/memoize.rs
```diff
@@ -1,65 +1,68 @@
...
 where
-    F: Fn(&D) -> V,
+    InitView: Fn(&Data) -> V,
 {
-    const ASSERT_CONTEXTLESS_FN: () = {
-        assert!(
-            core::mem::size_of::<F>() == 0,
-            "
+    /// Create a new `Memoize` view.
+    pub fn new(data: Data, init_view: InitView) -> Self {
```
DJMcNab, line 29: I don't think this should be `pub`?

xilem_core/src/view.rs
```diff
@@ -224,7 +238,12 @@ where
         message: Message,
         app_state: &mut State,
     ) -> MessageResult<Action, Message> {
-        self.deref()
-            .message(view_state, id_path, message, app_state)
+        let message_result =
+            self.deref()
+                .message(&mut view_state.view_state, id_path, message, app_state);
+        if matches!(message_result, MessageResult::RequestRebuild) {
```
DJMcNab, line 244: Can you point towards the equivalent logic in Xilem Classic?

Just trying to ascertain if this is new semantics being added or just something I missed when porting this through.
Currently, I'm not aware of `MessageResult` having any semantics.

xilem_core/src/views/memoize.rs
```diff
@@ -116,15 +119,123 @@ where
...
+}
+
+/// Specialized version of [`Memoize`], which doesn't take any data at all, the closure is evaluated only once and when a child view forces a rebuild
+pub struct Static<InitView, State, Action> {
+    init_view: InitView,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+/// Specialized version of [`memoize`], which doesn't take any data at all, the closure is evaluated only once and when a child view forces a rebuild
+pub fn static_view<State, Action, Context, Message, V, InitView>(
```
DJMcNab, line 142: I thought against this in Bevy, and I'm going to fight against this here; we do not want components to have names ending in `_view` .

I know that we can't use `static` because that's a keyword. Maybe `eternal` works?

xilem_core/src/views/memoize.rs
Philipp-M: I guess we could also remove this entirely und just use a struct initializer in `memoize` as I think we settled for the `function()` syntax for views.

xilem_core/src/view.rs
Philipp-M: > Just trying to ascertain if this is new semantics being added or just something I missed when porting this through.

I think missed? This is/was already in the `Memoize` view (now and then), and was used similarly in the `async_list` in the xilem_tokio branch in druid. It makes sense, since views don't have the same access in `View::message` to stuff as in `View::rebuild`, so they need to tell the parent views, via the `MessageResult`, that even though they don't seem to have changed, that they need a rebuild, so that the logic in `View::rebuild` can handle the rest. I guess it's related to [this comment](https://github.com/linebender/xilem/pull/456#issuecomment-2254721485)

xilem_core/src/views/memoize.rs
Philipp-M: `eternal` sounds epic at least^^ Yeah I'm also not happy with `static_view` ideal would be `static` or something. Though `eternal` probably suggests that this will live forever...

xilem_core/src/view.rs
DJMcNab: Yeah, fair enough. This does seem right.

It just seems a shame to lose the simplicity of the `Arc` view; I had been under the impression that the inner view would never be rebuilt if there was no change.

xilem_core/src/views/memoize.rs
DJMcNab: `static` also suggests it will live forever. We definitely can't end up in a state where this is `_view` :sweat_smile:

xilem_core/src/view.rs
Philipp-M: Yeah I know what you mean, though I expect when we go to something like an `Arc` based signal/dirty-tracking system as described [here](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/.60impl.20View.20for.20Arc.3CV.3A.20View.3E.60) (which I still haven't found a satisfying solution for though), that something like that would need to be done anyways.

xilem_core/src/views/memoize.rs
Philipp-M: Also true... Yeah finding good names is hard, probably the hardest thing in programming^^ I'll try to think about it, but I agree, we should remove `_view`.

xilem_core/src/views/memoize.rs
```diff
@@ -1,65 +1,76 @@
...
 
-impl<D, V, F> Memoize<D, F>
-where
-    F: Fn(&D) -> V,
-{
-    const ASSERT_CONTEXTLESS_FN: () = {
-        assert!(
-            core::mem::size_of::<F>() == 0,
-            "
+macro_rules! non_capturing_closure_string {
```
DJMcNab, line 19: We need a comment to explain why we're writing that this way. I'd expect `assert!(..., "{}", ASSERTION_STRING)` to be the idiomatic Rust here, where the value is stored in a constant.

Did that run into the rustfmt issue?

xilem_core/src/views/memoize.rs
```diff
@@ -115,16 +126,101 @@ where
...
-    data: Data,
-    view: ViewFn,
-) -> Memoize<Data, ViewFn>
+/// Specialized version of [`Memoize`], which doesn't take any data at all, the closure is evaluated only once and when a child view forces a rebuild
+pub struct Frozen<InitView, State, Action> {
+    init_view: InitView,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+/// Specialized version of [`memoize`], which doesn't take any data at all, the closure is evaluated only once and when a child view forces a rebuild
```
DJMcNab, line 135: We should explain this in terms of why you would use it, not in reference to any other value.

xilem/examples/memoization.rs
```diff
@@ -47,8 +50,9 @@ fn decrease_button(state: &AppState) -> impl WidgetView<AppState> {
     })
 }
 
+// When there's no access to the app state, i.e. the view is "static", you can use `frozen` which is basically a Memoize view without data
 fn reset_button() -> impl WidgetView<AppState> {
```
DJMcNab, line 54: Again, we should explain `frozen` standalone
```suggestion
fn reset_button() -> impl WidgetView<AppState> {
    // The contents of this view never changes, so we use `frozen` to avoid unnecessary rebuilds.
    // This is a special case of memoization for when the view doesn't depend on any data.
```

DJMcNab (approved):
Some code style suggestions, actual PR content looks good!

xilem_core/src/views/memoize.rs
Philipp-M: Yeah thinking about it, that's probably better, no I didn't run into it, just haven't thought about it.

xilem_core/src/views/memoize.rs
Philipp-M: Probably skim again, I also added an example doc-comment for `Memoize`

DJMcNab (approved):

xilem_core/src/views/memoize.rs
flosse: I like `frozen` and `immutable`.
`fixed` reminds me more of positioning like in CSS `position: fixed`

PR #452, opened 2024-07-27
xilem_web: Add a `await_once` view, and a simple example `suspense` showing it in action.

Philipp-M:
This adds a view similar as `xilem_web::concurrent::memoized_await` with the difference, that it's not having any memoized data, and instead takes the app state as first parameter, the future is run just once.

This is demonstrated within a new `suspense` example, which shows, that this view could be used similar as something like Reacts `Suspense`. This doesn't necessarily mean, that there won't be a more specialized view for this use-case though.
It also shows a slightly more interesting use-case of `Either` (accessing the `Class` view for both `p` and `h1`).

### DIFF
--- a/xilem_web/src/concurrent/await_once.rs
+++ b/xilem_web/src/concurrent/await_once.rs
@@ -1,0 +1,118 @@
+// Copyright 2024 the Xilem Authors and the Druid Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{future::Future, marker::PhantomData};
+
+use wasm_bindgen::UnwrapThrowExt;
+use wasm_bindgen_futures::spawn_local;
+use xilem_core::{MessageResult, Mut, NoElement, View, ViewId};
+
+use crate::{DynMessage, OptionalAction, ViewCtx};
+
+/// Await a future returned by `init_future`, `callback` is called with the output of the future. `init_future` will only be invoked once. Use [`await_once`] for construction of this [`View`]
+pub struct AwaitOnce<InitFuture, Callback, State, Action> {
+    init_future: InitFuture,
+    callback: Callback,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+/// Await a future returned by `init_future`, `callback` is called with the output of the future. `init_future` will only be invoked once.
+///
+/// # Examples
+///
+/// ```
+/// use xilem_web::{core::fork, concurrent::await_once, elements::html::div, interfaces::Element};
+///
+/// fn app_logic(state: &mut i32) -> impl Element<i32> {
+///     fork(
+///         div(*state),
+///         await_once(
+///             |_state: &mut i32| std::future::ready(42),
+///             |state: &mut i32, meaning_of_life| *state = meaning_of_life,
+///         )
+///     )
+/// }
+/// ```
+pub fn await_once<State, Action, FOut, F, InitFuture, OA, Callback>(
+    init_future: InitFuture,
+    callback: Callback,
+) -> AwaitOnce<InitFuture, Callback, State, Action>
+where
+    State: 'static,
+    Action: 'static,
+    FOut: std::fmt::Debug + 'static,
+    F: Future<Output = FOut> + 'static,
+    InitFuture: Fn(&mut State) -> F + 'static,
+    OA: OptionalAction<Action> + 'static,
+    Callback: Fn(&mut State, FOut) -> OA + 'static,
+{
+    AwaitOnce {
+        init_future,
+        callback,
+        phantom: PhantomData,
+    }
+}
+
+pub struct AwaitOnceState<F> {
+    future: Option<F>,
+}
+
+impl<State, Action, InitFuture, F, FOut, Callback, OA> View<State, Action, ViewCtx, DynMessage>
+    for AwaitOnce<InitFuture, Callback, State, Action>
+where
+    State: 'static,
+    Action: 'static,
+    FOut: std::fmt::Debug + 'static,
+    F: Future<Output = FOut> + 'static,
+    InitFuture: Fn(&mut State) -> F + 'static,
+    OA: OptionalAction<Action> + 'static,
+    Callback: Fn(&mut State, FOut) -> OA + 'static,
+{
+    type Element = NoElement;
+
+    type ViewState = AwaitOnceState<F>;
+
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        let thunk = ctx.message_thunk();
+        // we can't directly push the initial message, as this would be executed directly (not in the next microtick), which in turn means that the already mutably borrowed `App` would be borrowed again.
+        // So we have to delay this with a spawn_local
+        spawn_local(async move { thunk.push_message(None::<FOut>) });
+        (NoElement, AwaitOnceState { future: None })
+    }
+
+    fn rebuild<'el>(
+        &self,
+        _prev: &Self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut ViewCtx,
+        (): Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element> {
+        if let Some(future) = view_state.future.take() {
+            let thunk = ctx.message_thunk();
+            spawn_local(async move { thunk.push_message(Some(future.await)) });
+        }
+    }
+
+    fn teardown(&self, _: &mut Self::ViewState, _: &mut ViewCtx, _: Mut<'_, Self::Element>) {}
+
+    fn message(
+        &self,
+        view_state: &mut Self::ViewState,
+        id_path: &[ViewId],
+        message: DynMessage,
+        app_state: &mut State,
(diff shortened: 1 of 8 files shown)


casey:
Just tried this locally, and it seems like a nice addition!

I was able to rewrite this:

```rust
memoized_await(
  (),
  |()| async {
    Api::default()
      .packages()
      .await
      .unwrap()
      .into_iter()
      .collect::<Vec<(Hash, Manifest)>>()
  },
  |state: &mut State, output| state.packages = output,
),
memoized_await(
  (),
  |()| async { Api::default().handlers().await.unwrap() },
  |state: &mut State, output| state.handlers = output,
),
```

Into:

```rust
await_once(
  |_| async {
    Api::default()
      .packages()
      .await
      .unwrap()
      .into_iter()
      .collect::<Vec<(Hash, Manifest)>>()
  },
  |state: &mut State, output| state.packages = output,
),
await_once(
  |_| async { Api::default().handlers().await.unwrap() },
  |state: &mut State, output| state.handlers = output,
),
```

I like the name `await_once`, and it might be nice if `memoized_await` were renamed `await_memoized` to make their similarity slightly more obvious.

casey:
One thought, is it necessary to give `await_once` access to the initial state?

Instead of:

```rust
await_once(
  |state| async { async_function(state).await },
  |state: &mut State, output| state.handlers = output,
),
```

Could it be:

```rust
await_once(
  async_function(state), // since we're probably calling await_once in a context where `state` is in scope
  |state: &mut State, output| state.handlers = output,
),
```

Philipp-M:
> I like the name `await_once`, and it might be nice if `memoized_await` were renamed `await_memoized` to make their similarity slightly more obvious.

Yeah I think that makes sense.

> Could it be:

I'm not sure, as the `app_logic` is evaluated quite often, so that `async_function(state)` is run multiple times, and may lead to weird/unexpected behavior, when it changes the state.

Philipp-M:
Regarding the `state` as initial argument, see this [zulip topic](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/.60Suspense.60.20view) for more thoughts, or in other words I'm not yet sure, but it may be interesting for the user to track some state, when this is evaluated first. Though the whole async story is still quite in flux I think.

DJMcNab:
I think the version directly accepting a future makes sense. It would need to `move` the fields it needs from `state`, but that's fine.

Philipp-M:
I was more thinking about having an "event" when the future is actually initialized, but as said I'm not sure about it, so having a future there directly maybe makes sense as well. Hmm I'm not sure about especially this init future event, I think it could certainly be useful, to really know then a future has started, but as noticed in the fetch example, also leads to more imperative code, not sure how we should handle these kind of "events" in general...

Another issue with having a future here directly is also, that it may lead to surprising behavior when the future itself changes, as it would not have any effect, as it would then be invoked in `View::build`.

casey:
Actually, thinking more about it, I think requiring a function/closure which takes the current state and returns a future is perhaps best for clarity, i.e., reinforcing that the closure will only be called once, as opposed to taking a future each time, but only evaluating that future once.

xilem_web/src/concurrent/await_once.rs
```diff
@@ -0,0 +1,118 @@
...
+{
+    type Element = NoElement;
+
+    type ViewState = AwaitOnceState<F>;
+
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        let thunk = ctx.message_thunk();
+        // we can't directly push the initial message, as this would be executed directly (not in the next microtick), which in turn means that the already mutably borrowed `App` would be borrowed again.
+        // So we have to delay this with a spawn_local
+        spawn_local(async move { thunk.push_message(None::<FOut>) });
```
flosse, line 79: As far as I can remember, we already have this situation in several use cases. I wonder if it is time to either extend the `MessageThunk` with e.g. `fn send_message_asynchronously`, or if there is another common solution?

xilem_web/src/concurrent/await_once.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    fn rebuild<'el>(
+        &self,
+        _prev: &Self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut ViewCtx,
+        (): Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element> {
+        if let Some(future) = view_state.future.take() {
```
flosse, line 90:
```suggestion
        let Some(future) = view_state.future.take() else {
            return;
        };
```

xilem_web/src/concurrent/await_once.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn teardown(&self, _: &mut Self::ViewState, _: &mut ViewCtx, _: Mut<'_, Self::Element>) {}
+
+    fn message(
+        &self,
+        view_state: &mut Self::ViewState,
+        id_path: &[ViewId],
+        message: DynMessage,
+        app_state: &mut State,
+    ) -> MessageResult<Action, DynMessage> {
+        assert!(id_path.is_empty()); // `debug_assert!` instead? to save some bytes in the release binary?
```
flosse, line 105: ..yes, I almost always use `debug_assert`, but mostly in projects where the potential bugs should actually be noticed during development. I don't know enough about the internals of Xilem to judge the likelihood of this.
In any case, it is an advantage to create a panic as early as possible if something is obviously fundamentally wrong.
Since Xilem is still under development, an early bug report from users is probably also very desirable.
In other words: at the moment, it would be even more important to get early feedback on errors than to save bytes :wink:

xilem_web/src/concurrent/await_once.rs
```diff
@@ -0,0 +1,118 @@
...
+/// }
+/// ```
+pub fn await_once<State, Action, FOut, F, InitFuture, OA, Callback>(
+    init_future: InitFuture,
+    callback: Callback,
+) -> AwaitOnce<InitFuture, Callback, State, Action>
+where
+    State: 'static,
+    Action: 'static,
+    FOut: std::fmt::Debug + 'static,
```
flosse, line 43:
```suggestion
    FOut: fmt::Debug + 'static,
```
....and of course `use std::fmt;` at the top. Not important, just a personal preference :wink:

xilem_web/src/concurrent/await_once.rs
```diff
@@ -0,0 +1,118 @@
...
+use crate::{DynMessage, OptionalAction, ViewCtx};
+
+/// Await a future returned by `init_future`, `callback` is called with the output of the future. `init_future` will only be invoked once. Use [`await_once`] for construction of this [`View`]
+pub struct AwaitOnce<InitFuture, Callback, State, Action> {
+    init_future: InitFuture,
+    callback: Callback,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+/// Await a future returned by `init_future`, `callback` is called with the output of the future. `init_future` will only be invoked once.
```
flosse, line 19: Maybe a sentence about the fact that a change of `&mut AppState` has no effect at all and that if it should be so, then `async_repeat` must be used for it?

flosse (approved):
OK, looks good so far, i just added a few low-priority thoughts.

flosse:
> I was more thinking about having an "event" when the future is actually initialized, but as said I'm not sure about it, so having a future there directly maybe makes sense as well. Hmm I'm not sure about especially this init future event, I think it could certainly be useful, to really know when a future has started, but as noticed in the fetch example, also leads to more imperative code, not sure how we should handle these kind of "events" in general...
> 
> Another issue with having a future here directly is also, that it may lead to surprising behavior when the future itself changes, as it would not have any effect, as it would then be invoked in `View::build`.

I have to admit, I'm still a bit confused about the Xilem approach, the feeling that everything in the world is a view.
But maybe I still don't understand the approach.
I mean, the probability that I have a large number of asynchronous tasks that don't have anything directly to do with the GUI at first, but actually only the effects of it, is very high, isn't it?
Here you would be tempted to pack all asynchronous tasks into views, which is not really architecturally clean, is it?

xilem_web/src/concurrent/await_once.rs
Philipp-M: I mean a change in `AppState` has effect (in a rerender)? It just won't have an effect when using interior mutability within e.g. the future. (Which we should probably generally document).

That was actually a reason why I wanted to include it in the `init_future` so that some kind of `is_loading` state could be set.

xilem_web/src/concurrent/await_once.rs
Philipp-M: Yes I've thought about this as well, makes sense. Maybe we want to rename `push_message` to `handle_message` (so that it's more clear, that it will be handled immediately, which is probably relevant for events) and add a new method `enqueue_message` or `defer_message` with this inside.

xilem_web/src/concurrent/await_once.rs
Philipp-M: Yes, my thinkings similar, it's probably better to get bug reports when this is somehow triggered.

PR #453, opened 2024-07-27
masonry: add pressure sensitivity to the pen's touch on the screen

werickdasilva:
device test
* huion kamvas 13 tablet windows 10 platform


DJMcNab (approved):
Our API for touch has not been fully designed, however this should be able to unblock use cases in the near term. It's likely that this API will change in the medium term, however.

Thanks.

PR #458, opened 2024-07-28
xilem_web: Remove hardcoded `prevent_default`/`stop_propagation` as the user should decide that behavior

### LINKED ISSUE #457: xilem_web: Error running `counter` example: Unable to preventDefault inside passive event listener inovcation
I was trying out the counter example, and got an error in the console whenever I pressed a button. The error is "Unable to preventDefault inside passive event listener inovcation", screenshot below:

Screenshot 2024-07-28 at 1 07 53 PM

Philipp-M:
I think we could also create some additional sugar around that, but since the event is given to the user, they can handle that behavior anyway.

Fixes #457


DJMcNab (approved):

PR #463, opened 2024-07-28
xilem: Add async_worker view

azymohliad:
In this PR I'm trying to explore an alternative `async_repeat` implementation, which accepts `FnOnce` closure and hence allows more use-cases. I called it `async_worker` as a more general-purpose component.

It does require interior mutability currently, but only to workaround the immutable `&self` reference in `View::build`. (Btw, I'm sure there are many implications I haven't considered, but would it be a terrible idea to have `&mut self` in `View::build`?) Anyway, the interior mutability seems not too inappropriate here, because the framework calls `View::build` or `View::rebuild` once per lifetime of the view, but this isn't expressed in the type system.

The TL;DR on the implementation is that it stores the task closure as an interior-mutable `Option<FnOnce>` so it just `take`-s it out of the option in `View::build` before calling. 

I think it should also be a suitable base to build wrappers on top, such as [this `async_receiver`](https://github.com/linebender/xilem/pull/463/files#diff-d98aa1b6f994304a53f2404b78455da6463792748856bcdb9800dc009b370525R15-R40) view, but I couldn't figure out the type parameters yet. UPD: Not sure if it's actually possible, help wanted :pray:.

See also [this](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/How.20to.20do.20frequent.20data.20updates.20from.20another.20thread.3F) Zulip thread, the PR is introduced in [this](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/How.20to.20do.20frequent.20data.20updates.20from.20another.20thread.3F/near/454681976) message.

### DIFF
--- a/xilem/Cargo.toml
+++ b/xilem/Cargo.toml
@@ -40,6 +40,7 @@ tokio = { version = "1.39.1", features = ["rt", "rt-multi-thread", "time"] }
 
 [dev-dependencies]
 accesskit_winit.workspace = true
+tokio = { version = "1.39.1", features = ["sync"] }
 
 [target.'cfg(target_os = "android")'.dev-dependencies]
 winit = { features = ["android-native-activity"], workspace = true }
--- a/xilem/examples/data_thread.rs
+++ b/xilem/examples/data_thread.rs
@@ -1,0 +1,92 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{thread, time};
+
+use xilem::{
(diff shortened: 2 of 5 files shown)


DJMcNab (commented):
The other reason for `View::build` and `rebuild` being `&self` based is for cross-process use cases (such as hot reloading). I don't want to build too many things up here before we make that available.

I think this is a discussion which needs to involve Raph.

azymohliad:
I'll close it for now to keep the PR list clean, but will keep the branch around for a while just in case.

Issue #462, opened 2024-07-28
xilem_web: Should helper functions for known attributes be added to elements?


casey:
I noticed that `HtmlButtonElement` has a `disabled` helper method for setting the disabled state, but that, for example, `HtmlIFrameElement` doesn't have a `src` helper method for setting the `<iframe>` `src` attribute. Is it desired to have such helper methods? If so, I'd be happy to submit a PR.

Philipp-M:
Yeah that's actually the plan, and I'm happy for PRs implementing this, e.g. the `HtmlImageElement` does have `src` already, currently just wrapping `Attr` is fine. Ideally this should access the DOM attributes directly of the underlying element, but I have concerns regarding the resulting binary size with the code that's necessary for this (which is why there's a commented out feature gate above those traits).

PR #466, opened 2024-07-29
xilem_web: Add function to set HtmlIFrameElement src attribute

casey:
Apropos of #462. Let me know if this should have a test or an example!

### DIFF
--- a/xilem_web/src/interfaces.rs
+++ b/xilem_web/src/interfaces.rs
@@ -637,6 +637,10 @@ where
 pub trait HtmlIFrameElement<State, Action = ()>:
     HtmlElement<State, Action, DomNode: AsRef<web_sys::HtmlIFrameElement>>
 {
+    /// See <https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/src> for more details
+    fn src(self, value: impl IntoAttributeValue) -> Attr<Self, State, Action> {
+        Attr::new(self, "src".into(), value.into_attr_value())
+    }
 }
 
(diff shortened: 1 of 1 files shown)


Philipp-M:
I don't think either is necessary (as this is a quite trivial change), having more examples, e.g. showing the "vision" of a more typed API is probably worth it anyway (I actually started something in #403, which is why there are a few methods in the interfaces traits, but that example was not optimal because of yet missing SSR).

I do think that should maybe be done in a separate PR though, maybe showing also other stuff like the `Box<AnyDomView>`.

What I think generally makes sense is to have some doc-comment examples, also serving as tests of the API, but I think in this case it's trivial, so I don't think necessary, that will be more interesting, when it's more than just a simple `impl IntoAttributeValue` as parameter, e.g. like with `Element::class`.

Philipp-M (approved):
LGTM

casey:
I had a thought: Should the value of this attribute be a [Uri](https://docs.rs/http/latest/http/uri/struct.Uri.html)? It can't be a URL, because relative paths are allowed.

Philipp-M:
Hmm, out of a typing perspective I think it should be.
I'm not sure, though, if we should have something like `impl IntoAttributeValue for Uri` instead, as HTML/DOM is so messy in its specification, and this will likely further increase boilerplate both in wasm binary size as well as visual, in the end it will be converted back to a string again anyway.

Philipp-M:
Anyway, we can just go with this first, and do further adjustments of the (typed) API later I think. Merging, thanks.

PR #469, opened 2024-07-29
chore: Update dependencies for serde_json and tokio

JRedrupp:


DJMcNab:
Thanks for your interest.

I see that these updates are only patch releases. Is this addressing a real bug? Our normal policy is to perform dependency updates when updating our Rust version or before making a release - see e.g. https://github.com/linebender/xilem/pull/450

If you need these higher versions in your project, you should be able to use `cargo update`.

PR #472, opened 2024-07-31
Use `ViewMarker` trait instead of the generic `Marker` to allow `impl ViewSequence` as return type

Philipp-M:
This allows returning `impl ViewSequence` as this was previously not possibly as the `Marker` generic parameter couldn't be named (easily) in `ViewSequence<..., Marker>`.

This also provides specialized `ViewSequence`s for xilem (`WidgetViewSequence` and `FlexSequence`) as well as for xilem_web (`DomFragment`). Additional doc-tests/documentation and a small example (in `counter`) for xilem_web is provided as well.

This has the drawback to not being able to reeimplement `ViewSequence` for types that already implement `View`, which was previously possible (as seen by the now removed `NoElementView` `ViewSequence` implementation).
And additionally by introducing more boilerplate by having to implement `ViewMarker` for every type that implements `View`.

### DIFF
--- a/xilem_core/src/view.rs
+++ b/xilem_core/src/view.rs
@@ -9,6 +9,10 @@ use alloc::{boxed::Box, sync::Arc};
 
 use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 
+// TODO better doc-comment
+/// To avoid ambiguity for `impl ViewSequence for impl View`
+pub trait ViewMarker {}
+
 /// A lightweight, short-lived representation of the state of a retained
 /// structure, usually a user interface node.
 ///
@@ -34,7 +38,9 @@ use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 ///
 /// In order to support the default open-ended [`DynMessage`] type as `Message`, this trait requires an
 /// allocator to be available.
-pub trait View<State, Action, Context: ViewPathTracker, Message = DynMessage>: 'static {
+pub trait View<State, Action, Context: ViewPathTracker, Message = DynMessage>:
+    ViewMarker + 'static
+{
     /// The element type which this view operates on.
     type Element: ViewElement;
     /// State that is used over the lifetime of the retained representation of the view.
@@ -137,6 +143,7 @@ pub trait ViewPathTracker {
     }
 }
 
+impl<V: ?Sized> ViewMarker for Box<V> {}
 impl<State, Action, Context, Message, V> View<State, Action, Context, Message> for Box<V>
 where
     Context: ViewPathTracker,
@@ -180,6 +187,7 @@ where
     }
 }
 
+impl<V: ?Sized> ViewMarker for Arc<V> {}
 /// An implementation of [`View`] which only runs rebuild if the states are different
 impl<State, Action, Context, Message, V> View<State, Action, Context, Message> for Arc<V>
 where
--- a/xilem_core/src/view.rs
+++ b/xilem_core/src/view.rs
@@ -11,7 +11,7 @@ use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 
 /// A type which can be a [`View`]. Imposes no requirements on the underlying type.
 /// Should be implemented alongside every `View` implementation:
-/// ```rust
+/// ```ignore
 /// impl<...> ViewMarker for Button<...> {}
 /// impl<...> View<...> for Button<...> {...}
 /// ```
@@ -49,7 +49,7 @@ pub trait ViewMarker {}
 /// Due to restrictions of the [orphan rules](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules),
 /// `ViewMarker` needs to be implemented for every type that implements `View`, see [`ViewMarker`] for more details.
 /// For example:
-/// ```rust
+/// ```ignore
 /// impl<...> ViewMarker for Button<...> {}
 /// impl<...> View<...> for Button<...> {...}
 /// ```
--- a/xilem_core/src/view.rs
+++ b/xilem_core/src/view.rs
@@ -9,8 +9,11 @@ use alloc::{boxed::Box, sync::Arc};
 
 use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 
-/// This trait is a necessary workaround of the orphan rules to be able to have `impl ViewSequence for impl View` to avoid ambiguity.
-/// It needs to be implemented for every type that implements [`View`].
+/// Because [`View`] is generic, Rust [allows you](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules) to implement this trait for certain non-local types.
+/// These non-local types can include `Vec<_>` and `Option<_>`.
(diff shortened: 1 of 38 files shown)


xilem_core/src/view.rs
```diff
@@ -9,6 +9,10 @@ use alloc::{boxed::Box, sync::Arc};
 
 use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 
+/// This trait is a necessary workaround of the orphan rules to be able to have `impl ViewSequence for impl View` to avoid ambiguity.
+/// It needs to be implemented for every type that implements [`View`].
+pub trait ViewMarker {}
```
DJMcNab, line 14: This probably needs much better docs, and needs to be mentioned in the docs for `View`. 

This will easily become the most confusing part of Xilem, especially if we don't explain this properly.

DJMcNab (commented):
As I mentioned in Office hours, I think this is the right thing to do. I don't want to land this until the docs are improved, however.

It's very frustrating that this is necessary, because we're making a selection which isn't actually real ambiguity. But I don't have the strength to go finding a solution at the Rust level for this.

xilem_core/src/view.rs
Philipp-M: > This will easily become the most confusing part of Xilem, especially if we don't explain this properly.

I don't think so, unless someone really wants to know why this is necessary, otherwise it's just (for me at least) "ok I also need to implement that type for the view", but yeah the drive/curiosity to understand *why* this is necessary may be there... 

Do you have suggestions for the documentation (or some direction at least)? I'm not sure how much we need to go into detail here.

The error messages, when removing that requirement in the `ViewSequence` impl also doesn't make sense:

```
conflicting implementation for `Option<_>`
    |
    = note: downstream crates may implement trait `view::View<_, _, _, _>` for type `core::option::Option<_>`
```

How should downstream crates implement that trait anyway... argh, confusing edge cases of rusts type-system...

-------

Lol, I just tried removing `ViewMarker` in the `ViewSequence` impl with `RUSTFLAGS="-Znext-solver"` and it works :) (at least in xilem_web, xilem currently doesn't compile, because a dependent crate (`async-io`) fails type-checking)

So maybe we should just put some warning, that this is a workaround of the current rust compiler, but that this limitation will be lifted in the future trait solver, to avoid people loosing their sanity trying to understand this? (I hope at least, that the new trait-solver doesn't implement that guard) Maybe really something to ask the types-team of Rust about.

DJMcNab: How does this sound:

Because [`View`] is generic, Rust [allows you](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules) to implement this trait for certain non-local types.
These non-local types can include `Vec<_>` and `Option<_>`.
If this trait were not present, those implementations of `View` would conflict with those types' implementations of `ViewSequence`.
This is because every `View` type also implementations `ViewSequence`.
Since `ViewMarker` is not generic, these non-local implementations are not permitted for this trait, which means that the conflicting implementation cannot happen.

Philipp-M: Yeah that makes sense, I'll add that. I just checked, removing `ViewMarker` from the super-bound of `View` produces that error again :/ but makes sense thinking about it...

Philipp-M: I've added that for the doc-comment of `ViewMarker`, should the doc-comment for `View` be more detailed?

xilem_core/src/view.rs
```diff
@@ -9,6 +9,13 @@ use alloc::{boxed::Box, sync::Arc};
 
 use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 
+/// Because [`View`] is generic, Rust [allows you](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules) to implement this trait for certain non-local types.
```
DJMcNab, line 12:
```suggestion
/// A type which can be a [`View`]. Imposes no requirements on the underlying type.
/// Should be implemented alongside every `View` implementation:
/// ```rust
/// impl<...> ViewMarker for Button<...> {}
/// impl<...> View<...> for Button<...> {}
/// ```
///
/// ## Details
///
/// Because `View` is generic, Rust [allows you](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules) to implement this trait for certain non-local types.
```

xilem_core/src/view.rs
```diff
@@ -30,11 +37,16 @@ use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 /// During message handling, mutable access to the app state is given to view nodes,
 /// which will in turn generally expose it to callbacks.
 ///
+/// Due to restrictions of the [orphan rules](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules),
+/// `ViewMarker` needs to be implemented for every type that implements `View`, see [`ViewMarker`] for more details.
+///
```
DJMcNab, line 42: It might be worth repeating the same code block here as well

DJMcNab (approved):
Let's land this and see what comes of it.

xilem_core/src/view.rs
Philipp-M: Something like in the last commit?

xilem_core/src/view.rs
```diff
@@ -9,6 +9,22 @@ use alloc::{boxed::Box, sync::Arc};
 
 use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 
+/// A type which can be a [`View`]. Imposes no requirements on the underlying type.
+/// Should be implemented alongside every `View` implementation:
+/// ```rust
+/// impl<...> ViewMarker for Button<...> {}
+/// impl<...> View<...> for Button<...> {}
```
DJMcNab, line 16:
```suggestion
/// impl<...> View<...> for Button<...> {...}
```

xilem_core/src/view.rs
```diff
@@ -30,11 +46,21 @@ use crate::{message::MessageResult, DynMessage, Mut, ViewElement};
 /// During message handling, mutable access to the app state is given to view nodes,
 /// which will in turn generally expose it to callbacks.
 ///
+/// Due to restrictions of the [orphan rules](https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules),
+/// `ViewMarker` needs to be implemented for every type that implements `View`, see [`ViewMarker`] for more details.
+/// For example:
+/// ```rust
+/// impl<...> ViewMarker for Button<...> {}
+/// impl<...> View<...> for Button<...> {}
```
DJMcNab, line 54:
```suggestion
/// impl<...> View<...> for Button<...> {...}
```

xilem_core/src/view.rs
DJMcNab: Yeah, exactly.

DJMcNab (approved):
One last little tweak

Philipp-M:
Thanks for the review, yeah, it's a little nuisance, but I think it's worth it having to avoid discussions, why we can't return `ViewSequence` the same way as a `View`.

PR #475, opened 2024-08-01
Use `git lfs` to store Masonry screenshots

DJMcNab:
Also restores caching for Rust code in merge queues and in new PRs, which was lost when we migrated to merge queues.


PoignardAzur (approved):
The image diff previews all say "Unable to render rich display. Invalid image source." Is that normal?

Otherwise, LGTM.

DJMcNab:
Yeah. GitHub doesn't know how to handle the transition into lfs properly.

Right, I'll merge this into #233 and then land that, I think. Thanks for the approval.

Issue #449, opened 2024-07-25
Fix / remove warnings in Masonry documentation


waywardmonkeys:
Separate from #389, the warnings when running `cargo doc -p masonry` need to be fixed.

Most of them are about things that no longer exist within masonry or have perhaps changed names, so it should be a pretty quick pass.

This is blocking re-enabling denying warnings within CI for xilem documentation and has already led to an issue in that regard with a broken link making its way into there.

waywardmonkeys:
This is even worse with `cargo clippy -- -W clippy::doc_markdown` and that has even more regressions in `xilem_core`, `xilem_web`, and `xilem`.

PoignardAzur:
This should be fixed by #815. Are there still errors?

DJMcNab:
This issue also tracks re-enabling these issues in CI. Is that resolved?

PoignardAzur:
Well I didn't chnage CI after fixing the doc errors, so probably not.

DJMcNab:
It looks like https://github.com/linebender/xilem/pull/815/files#diff-27b19fb95b25f32e39ea044b6ab68ee4a84d37944078b20ea52985fcca779e7aL112-L115 should have addressed this, I think.

The only thing that's still wrong is https://github.com/linebender/xilem/blob/d77b31b892cfc5390c3fd49bc2f41da4101e3f24/masonry/src/app/tracing_backend.rs#L4

which is redundant I think. It looks like these were being suppressed by `allow`s rather than anything in the CI script.

waywardmonkeys:
@DJMcNab I'd removed that locally yesterday but didn't get to submitting the PR. Done now as #855.

PR #478, opened 2024-08-01
Fix doc warnings in xilem_core, xilem_web

waywardmonkeys:


jaredoconnell (approved):

DJMcNab (approved):
Can we automate this being checked?

That is, how did you determine these errors?

It's ok to add lints to lib.rs if they are crate specific, so long as we document that in the root Cargo.toml

waywardmonkeys:
Sure, we do in every other repo. This is blocked by #449 in theory, but I've decided to take a different approach to this today once this PR another lands.

PR #479, opened 2024-08-02
Fix some `clippy::doc_markdown` lints in masonry.

waywardmonkeys:
This is a start on #449.

### DIFF
--- a/masonry/src/testing/harness.rs
+++ b/masonry/src/testing/harness.rs
@@ -336,7 +336,7 @@ impl TestHarness {
 
     // --- MARK: EVENT HELPERS ---
 
-    /// Move an internal mouse state, and send a MouseMove event to the window.
+    /// Move an internal mouse state, and send a [`PointerMove`](PointerEvent::PointerMove) event to the window.
     pub fn mouse_move(&mut self, pos: impl Into<Point>) {
         // FIXME - Account for scaling
         let pos = pos.into();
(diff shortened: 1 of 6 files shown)


masonry/src/testing/harness.rs
```diff
@@ -336,7 +336,7 @@ impl TestHarness {
 
     // --- MARK: EVENT HELPERS ---
 
-    /// Move an internal mouse state, and send a MouseMove event to the window.
+    /// Move an internal mouse state, and send a [`PointerEvent::PointerMove`] event to the window.
```
DJMcNab, line 339:
```suggestion
    /// Move an internal mouse state, and send a [`PointerMove`](PointerEvent::PointerMove) event to the window.
```
might be slightly nicer formatting.

masonry/src/text/edit.rs
```diff
@@ -75,7 +75,7 @@ impl<T: EditableText> TextEditor<T> {
 
     /// Rebuild the text.
     ///
-    /// See also [TextLayout::rebuild](crate::text2::TextLayout::rebuild) for more comprehensive docs.
+    /// See also [`TextLayout::rebuild()`](crate::text::TextLayout::rebuild) for more comprehensive docs.
```
DJMcNab, line 78: I'm not really sure I like the function call brackets here. E.g. https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#indexing doesn't use that form, unless it cares about the value of the argument.

DJMcNab (approved):

masonry/src/text/edit.rs
waywardmonkeys: I tend to do them partly out of a habit from places where you might refer to either a member or a function, and from Doxygen which auto-links things that use `()` ... but I have got them into lots of Rust code now!

Philipp-M: I think it makes sense to stay consistent, and I don't know of crates that use `()` for doc-links.

waywardmonkeys:
Updated per comments.

PR #480, opened 2024-08-02
Enable `clippy::doc_markdown` for non-masonry crates.

waywardmonkeys:


waywardmonkeys:
This lets our clippy builds check `clippy::doc_markdown` in xilem and friends but suppresses it in `masonry` since otherwise it would fail.

This is a result of #449.

waywardmonkeys:
This is also a precursor to having a discussion about enabling `RUSTDOCFLAGS="-D warnings" in all of the Linebender repos.

Cargo.toml
```diff
@@ -49,6 +49,8 @@ rust.variant_size_differences = "warn"
 clippy.allow_attributes_without_reason = "warn"
 clippy.collection_is_never_read = "warn"
 clippy.debug_assert_with_mut_call = "warn"
+# This is marked as `allow` within masonry due to #449.
```
DJMcNab, line 52: We should use full links here - see e.g. #412.
i.e. 
```suggestion
# This is marked as `allow` within masonry due to https://github.com/linebender/xilem/issues/449.
```

masonry/src/lib.rs
```diff
@@ -80,6 +80,8 @@
 //! [Druid]: https://crates.io/crates/druid
 //! [Xilem]: https://crates.io/crates/xilem
 
+// TODO: #449: Remove this once the issues within masonry are fixed.
```
DJMcNab, line 83:
```suggestion
// TODO: Remove this once the issues within masonry are fixed. Tracked in https://github.com/linebender/xilem/issues/449
```

DJMcNab (approved):
Nice!

PR #481, opened 2024-08-02
xilem_web: Add an `AfterBuild`, `AfterRebuild` and `BeforeTeardown` view, which reflects `Ref` in React

flosse:


flosse:
@Philipp-M I cherry-picked [your commit](https://github.com/Philipp-M/xilem/commit/f7d351ff970556c84560327cf5cab2d9cf0eb7c7#diff-4e6ca7cc46c5f661102bd9d26f1a9c10f0d729c3b2157affe3eb92beb23830c5) and tried to adjust it to the current code base.
To be honest, I haven't figured it all out yet :see_no_evil: 
Could you imagine helping out a little here?

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,120 @@
...
+    E: Element<State, Action>,
+    E::Element: Clone + PartialEq,
+    F: Fn(&mut State, &E::Element) + 'static,
+{
+    type Element = E::Element;
+
+    type ViewState = AfterUpdateState<Self::Element, E::ViewState>;
+
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        let id = ViewId::new(0); // FIXME: what's the right number here?
```
Philipp-M, line 38: This should be fine, our whole view id system has changed to generational indexing (and the generation doesn't change in this view).
Thus views don't have a real id anymore, and are instead "unique" by the view path that is created by `ViewPathTracker::with_id()` while traversing in one of these methods (implemented by `ViewCtx`), but as an id could be changing in the ascendents (e.g. because an `AnyView` has changed its type), which should be kept in mind.

The child view below thus also doesn't need an explicit id anymore as well.

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,120 @@
...
+        where
+            E: $crate::interfaces::$dom_interface<T, A>,
+            E::Element: Clone + PartialEq,
+            F: Fn(&mut T, &E::Element),
+        {
+        }
+    };
+}
+
+//FIXME: for_all_dom_interfaces!(impl_dom_interface_for_attr);
```
Philipp-M, line 120: All of this isn't necessary anymore, as the previous rather hacky way to implement those interfaces is now encoded in the type-system directly via associated type bounds within `interfaces.rs`, it's now basically only dependent on the `View::Element` (and its trait implementations).
```suggestion
```

Philipp-M (commented):
As I said on zulip, I do think this whole view (or something similar) needs a rethink/overhaul just from the API side, because I think it would be quite nice to have "events" for the lifetime of such an element.

An obvious way to do this is with closures that don't have any mutable access to `State`. (i.e. something like `after_build`, `after_rebuild`, `after_teardown`) running those methods in the respective `View` methods. but I'm not sure how clean it is when those closures capture stuff (see e.g. the assertion in `Memoize`, although that case there isn't as relevant here, since we're more talking about events).

And I would really like to avoid this extra map in the `ViewCtx` as this creates unnecessary bloat, which is not used in the common case. Ideally this would only reside in the view itself, so that it could be optimized away when it's not used.

Philipp-M:
> running those methods in the respective `View` methods.

btw. before that, I think the props should be applied (via `DomNode::apply_props`), so that it's really up-to-date before such methods would be executed.

flosse:
@Philipp-M It's now a simple hack that works for me.

xilem_web/src/after_update.rs
```diff
@@ -1,55 +1,107 @@
...
+    pub fn new(element: E, callback: F) -> AfterRebuild<E, F> {
+        Self { element, callback }
+    }
+}
+
+impl<E, F> BeforeTeardown<E, F> {
+    pub fn new(element: E, callback: F) -> BeforeTeardown<E, F> {
+        Self { element, callback }
+    }
+}
```
Philipp-M, line 39: We [leaned towards](https://github.com/linebender/xilem/pull/451#discussion_r1694948958) using functions directly for constructing views, so maybe rather use `after_build`, `after_rebuild` and `before_teardown`.

I also feel like we should document these, to explain, what this is about, and why these views shouldn't be used ^^.

xilem_web/src/after_update.rs
```diff
@@ -1,55 +1,107 @@
...
+    F: Fn(&V::DomNode) + 'static,
+    V: DomView<State> + 'static,
+{
+    type Element = V::Element;
+
+    type ViewState = V::ViewState;
+
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        let (el, view_state) = self.element.build(ctx);
+        (self.callback)(&el.node);
```
Philipp-M, line 57: I think this is fine for now, as it doesn't break the modifier system (and the issue explained here is less worrisome), but as explained earlier, this should have the props applied to the element before calling this IMO.
Probably worth adding a `TODO` comment.

xilem_web/src/after_update.rs
```diff
@@ -59,29 +111,67 @@ where
...
-                .entry(id)
-                .and_modify(|e| e.0 = true)
-                .or_insert((true, view_path));
-
-            rebuild_outcome
-        })
+        let element = self
+            .element
+            .rebuild(&prev.element, view_state, ctx, element);
+        (self.callback)(element.node);
```
Philipp-M, line 117: But we can do that at least here without breaking stuff (needs the trait `DomNode` in scope):
```suggestion
        element.node.apply_props(element.props);
        (self.callback)(element.node);
```

xilem_web/src/interfaces.rs
```diff
@@ -80,8 +80,25 @@ pub trait Element<State, Action = ()>:
...
+        F: Fn(&Self::DomNode),
+    {
+        AfterRebuild::new(self, callback)
+    }
+
+    fn before_teardown<F>(self, callback: F) -> BeforeTeardown<Self, F>
+    where
+        F: Fn(&Self::DomNode),
+    {
+        BeforeTeardown::new(self, callback)
```
Philipp-M, line 101: As said on zulip, These should be in the `DomView` trait (ideally with doc-comments)

xilem_web/web_examples/after_update/src/main.rs
```diff
@@ -1,26 +1,49 @@
 // Copyright 2023 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use xilem_web::{document_body, elements::html, interfaces::Element, App};
+use xilem_web::{core::one_of::Either, document_body, elements::html, interfaces::Element, App};
 
-type AppState = bool;
+#[derive(Default)]
+struct AppState {
+    focus: bool,
```
Philipp-M, line 8: I think this should be either a `Rc<Cell<bool>>` or `Rc<AtomicBool>` (not sure which is more idiomatic in this context TBH), so that the focus value can be reset in the closure. to avoid calling `focus` on the element every rebuild.

You can see what I mean in action when adding somewhere something like this while typing in the input when `focus == true`:
```rust
            html::input(()).on_input(|_, _| {}),
```

I think somewhere in this example should probably also be comments, why this is an aweful hack (e.g. it doesn't trigger a rebuild after the callback was run), and just a necessary workaround to access the underlying DOM node.

xilem_web/web_examples/after_update/Cargo.toml
```diff
@@ -1,12 +1,13 @@
 [package]
-name = "after_update"
+name = "after_updtae"
```
Philipp-M, line 2: Typo, but maybe we should call this example something like "Raw DOM access", so that it's obvious what it's about.

Philipp-M (commented):
Looks mostly good, but especially these views need some documentation.

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,207 @@
...
+    callback: F,
+}
+
+pub struct BeforeTeardown<E, F> {
+    element: E,
+    callback: F,
+}
+
+/// This adds a callback to a DOM node
+/// that is invoked after it has been created.
```
Philipp-M, line 24:
```suggestion
/// Invokes the `callback` after the inner `element` [`DomView`] was created.
```

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,207 @@
...
+/// no properties have been applied to the node.
+///
+/// The use of this function should be avoided and
+/// should only be utilized in exceptional cases!
+pub fn after_build<E, F>(element: E, callback: F) -> AfterBuild<E, F> {
+    AfterBuild { element, callback }
+}
+
+/// This adds a callback to a DOM node
+/// that is invoked after it has been rebuild.
```
Philipp-M, line 37:
```suggestion
/// Invokes the `callback` after the inner `element` [`DomView<State>`] was rebuild, which usually happens after anything has changed in the `State` .
///
/// Memoization can prevent `callback` being called.
```

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,207 @@
...
+/// The use of this function should be avoided and
+/// should only be utilized in exceptional cases!
+pub fn after_rebuild<E, F>(element: E, callback: F) -> AfterRebuild<E, F> {
+    AfterRebuild { element, callback }
+}
+
+/// This adds a callback to a DOM node
+/// that is invoked before it is destroyed.
+/// This adds a callback to the view
+/// that is called before a DOM node is destroyed.
```
Philipp-M, line 49:
```suggestion
/// Invokes the `callback` before the inner `element` [`DomView`] (and its underlying DOM node) is destroyed.
```

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,207 @@
...
+/// The callback has a reference to the raw DOM node as its only parameter.
+///
+/// The use of this function should be avoided and
+/// should only be utilized in exceptional cases!
```
Philipp-M, line 53: Maybe (and if you agree, probably in `after_build` and `after_rebuild` as well)?:

```suggestion
/// As accessing the underlying raw DOM node can mess with the inner logic of xilem_web,
/// this should only be used as an escape-hatch for properties not supported by xilem_web. E.g. to be interoperable with external javascript libraries.
```

btw. I think adding bounds here actually helps with documenting (and prevents misuse) what should be used as `element` and can be given to the `callback`

Philipp-M (approved):
Apart from doc-nits looks good to me, thanks :)

xilem_web/src/after_update.rs
flosse: You mean something like this?:

```rust
pub fn before_teardown<E, F, State>(element: E, callback: F) -> BeforeTeardown<E, F>
  where
    E: DomView<State>,
    F: Fn(&web_sys::Element) + 'static,
{
    BeforeTeardown { element, callback }
}
```

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,204 @@
+// Copyright 2023 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use xilem_core::{MessageResult, Mut, View, ViewId, ViewMarker};
+
+use crate::{DomNode, DomView, DynMessage, ViewCtx};
+
+pub struct AfterBuild<E, F> {
```
Philipp-M, line 8: These likely also need the generic vars `State` and `Action` as attribute `phantom: PhantomData<fn() -> (State, Action)>`

We had some inference issues with these two variables in the past, when e.g. using `Adapt` around such a type, as that needed explicit typing then...

(same for all the other views)

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,204 @@
...
+/// The use of this function should be avoided and
+/// should only be utilized in exceptional cases!
+pub fn after_rebuild<E, F>(element: E, callback: F) -> AfterRebuild<E, F> {
+    AfterRebuild { element, callback }
+}
+
+/// Invokes the `callback` before the inner `element` [`DomView`] (and its underlying DOM node) is destroyed.
+/// As accessing the underlying raw DOM node can mess with the inner logic of `xilem_web`,
+/// this should only be used as an escape-hatch for properties not supported by `xilem_web`.
+/// E.g. to be interoperable with external javascript libraries.
```
Philipp-M, line 50: As you applied this here, maybe as suggested in `after_rebuild` and `after_build` as well?

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,204 @@
...
+pub fn before_teardown<E, F>(element: E, callback: F) -> BeforeTeardown<E, F> {
+    BeforeTeardown { element, callback }
+}
+
+impl<E, F> ViewMarker for AfterBuild<E, F> {}
+impl<E, F> ViewMarker for AfterRebuild<E, F> {}
+impl<E, F> ViewMarker for BeforeTeardown<E, F> {}
+
+impl<State, V, F> View<State, (), ViewCtx, DynMessage> for AfterBuild<V, F>
```
Philipp-M, line 59:
```suggestion
impl<State, Action, V, F> View<State, Action, ViewCtx, DynMessage> for AfterBuild<V, F>
```

I think I missed this, this should be generic over `Action` so that it's usable in other contexts. See also the other views here.

xilem_web/src/after_update.rs
```diff
@@ -0,0 +1,204 @@
...
+where
+    State: 'static,
```
Philipp-M, line 61: Likely needs a bound then as well
```suggestion
    State: 'static,
    Action: 'static,
```

Philipp-M (commented):
More like this:
```rust 
pub fn before_teardown<State, Action, E, F>(element: E, callback: F) -> BeforeTeardown<E, F, State, Action>
where
    E: DomView<State, Action>,
    F: Fn(&E::DomNode) + 'static,
    State: 'static,
    Action: 'static,
{
    BeforeTeardown { element, callback }
}
```

Similar bounds for the trait methods in `DomView`

xilem_web/src/after_update.rs
flosse: Of course, I just forgot it :see_no_evil:

Philipp-M (approved):
I think this is in a good state to merge, thanks!

PR #482, opened 2024-08-02
xilem_web: Allow `DomFragment` instead of `DomView` as `app_logic`

### LINKED ISSUE #461: xilem_web: Is it possible for a update function to return two dom nodes without a parent contianer?
I have an app I'm rewriting using `xilem_web` whose HTML was formerly:

```html
<body>
  <nav>…</nav>
  <main>…</main>
</body>
```

With `xilem_web`, I'm doing:

```html
fn update(state: &mut State) -> impl DomView<State> {
  div((html::nav(()), html::main(())))
}
```

Which results in an extra wrapper div:

```html
<body>
  <div>
    <nav>…</nav>
    <main>…</main>
  </div>
</body>
```

I tried:

```html
fn update(state: &mut State) -> impl DomView<State> {
  (html::nav(()), html::main(()))
}
```

As well as fooling around with `xilem_core::ViewSequence`, to no avail. Is it possible to avoid the extra wrapper div? It's not a huge deal, but an extra wrapper div does sometimes make CSS more complex (`body > div > nav` instead of `body > nav`) and is an extra div to manually expand when inspecting the DOM.

Philipp-M:
Should fix #461. This allows a `ViewSequence` (called `DomFragment`) of `DomView`s as root component.

The `counter` example is updated to show this new behavior.

### DIFF
--- a/xilem_web/src/app.rs
+++ b/xilem_web/src/app.rs
@@ -1,11 +1,12 @@
 // Copyright 2023 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::{DomNode, ViewCtx};
+use crate::{elements::DomChildrenSplice, AnyPod, DomFragment, ViewCtx};
 use std::{cell::RefCell, rc::Rc};
 
-use crate::{DomView, DynMessage, PodMut};
-use xilem_core::{MessageResult, ViewId};
+use crate::DynMessage;
+use wasm_bindgen::UnwrapThrowExt;
+use xilem_core::{AppendVec, MessageResult, ViewId};
 
 pub(crate) struct AppMessage {
     pub id_path: Rc<[ViewId]>,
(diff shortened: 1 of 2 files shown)


xilem_web/src/app.rs
```diff
@@ -1,27 +1,32 @@
...
+use xilem_core::{AppendVec, MessageResult, ViewId};
 
 pub(crate) struct AppMessage {
     pub id_path: Rc<[ViewId]>,
     pub body: DynMessage,
 }
 
 /// The type responsible for running your app.
-pub struct App<T, V: DomView<T>, F: FnMut(&mut T) -> V>(Rc<RefCell<AppInner<T, V, F>>>);
+pub struct App<State, Fragment: DomFragment<State>, InitFragment>(
```
Philipp-M, line 17: I'm not sure about the name `InitFragment` for the `app_logic`, I'm open for other suggestions, it's also related to [this](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/.22app_logic.22)

Philipp-M:
This coincidentally also reduces a little bit of wasm bloat btw. :)

DJMcNab (approved):
Optimistic approval for Xilem web

PR #483, opened 2024-08-02
masonry: `TextEditor` add `Action::TextChanged` for backspace & delete

edgy-sphere:
`masonry::text::edit::TextEditor` should emit an `Action::TextChanged(...)` also on `Backspace` and `Delete` if anything was deleted, so that e.g. `xilem::view::textbox::Textbox` receives this action. This is already implemented for `Ctrl+Backspace` and `Ctrl+Delete`.

The code changes are not elegant, but intended to be minimally invasive.

Local `cargo test` failed for the modified and an unmodified code, but sometimes with different errors&mdash;which I do not quite understand.

I apologize in advance for any mistakes on my part, I've never before worked on open source repos and did not find any guideline for your project.

### DIFF
--- a/masonry/src/text/edit.rs
+++ b/masonry/src/text/edit.rs
@@ -124,6 +124,9 @@ impl<T: EditableText> TextEditor<T> {
                                     self.text_mut().edit(selection.range(), "");
                                     self.inner.selection =
                                         Some(Selection::caret(selection.min(), Affinity::Upstream));
+
+                                    let contents = self.text().as_str().to_string();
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
Thank you! This code has always had some layering issues, and so is likely to be changed at some point in the future; however this addresses a real issue now.

The Xilem text code has not been carefully designed yet. It's amongst the things we need to deal with.

DJMcNab:
The test failures is likely due to #233; we will be fixing that soon, but it's waiting on review for #475.

PR #484, opened 2024-08-02
xilem_web: Fix `DomChildrenSplice::with_scratch`

Philipp-M:
I'm surprised that this was not noticed yet, `DomChildrenSplice::with_scratch` previously only appended new elements to the end of the children of an element in the DOM tree, whereas it should've inserted it at the current index of the `ElementSplice`. This should fix this, using a `DocumentFragment` which is shared on the `ViewCtx` to avoid unnecessary allocations.

### DIFF
--- a/xilem_web/Cargo.toml
+++ b/xilem_web/Cargo.toml
@@ -31,6 +31,7 @@ features = [
   "console",
   "CssStyleDeclaration",
   "Document",
+  "DocumentFragment",
   "DomTokenList",
   "Element",
   "Event",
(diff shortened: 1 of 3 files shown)


DJMcNab (approved):
I don't understand this code, but doing optimistic approval for Xilem Web

Philipp-M:
Thanks, yeah a `DocumentFragment` is just a pseudo container, which will be emptied as soon as it will be inserted in a real DOM node, I wanted to avoid some Rust tricks, and this may be faster anyway (but as MDN says, not necessarily).

Thanks for the approval.

PR #485, opened 2024-08-05
xilem_web: Add async_repeat

flosse:


flosse:
This is a first attempt to port `xilem::async_repeat` to `xilem_web`.
But for some reason the example does not compile:

```shell
error[E0277]: the trait bound `AsyncRepeat<{closure@xilem_web/web_examples/async_repeat/src/main.rs:21:9: 21:16}, {closure@xilem_web/web_examples/async_repeat/src/main.rs:27:9: 27:49}, Message>: View<AppState, (), ViewCtx, Box<(dyn xilem_web::Message + 'static)>>` is not satisfied
  --> xilem_web/web_examples/async_repeat/src/main.rs:19:39
   |
19 | fn app_logic(state: &mut AppState) -> impl Element<AppState> {
   |                                       ^^^^^^^^^^^^^^^^^^^^^^ the trait `View<AppState, (), ViewCtx, Box<(dyn xilem_web::Message + 'static)>>` is not implemented for `AsyncRepeat<{closure@main.rs:21:9}, {closure@main.rs:27:9}, Message>`, which is required by `Fork<xilem_web::elements::html::Div<_, _>, AsyncRepeat<{closure@xilem_web/web_examples/async_repeat/src/main.rs:21:9: 21:16}, {closure@xilem_web/web_examples/async_repeat/src/main.rs:27:9: 27:49}, Message>>: Element<AppState>`
```

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,111 @@
...
+use xilem_core::{DynMessage, Message, MessageResult, NoElement, View, ViewMarker};
+
+use crate::{context::MessageThunk, ViewCtx};
+
+pub fn async_repeat<M, F, H, State, Action, Fut>(
+    future_future: F,
+    on_event: H,
+) -> AsyncRepeat<F, H, M>
+where
+    F: Fn(MessageThunk) -> Fut + 'static,
```
Philipp-M, line 15: I'm not sure if we should expose this implementation detail of xilem_web, maybe wrap it? But I'm not sure if it's worth it to wrap.
I'm thinking about using `requestAnimationFrame` in the main loop to batch stuff, and throttle events and other messages before doing an actual `View::rebuild` but that may produce issues with input-latency (I'd need to test e.g. with https://www.vsynctester.com/ ) So not sure if worth it at least in the near-term, low-priority for me currently.

It's also somewhat dangerous, as currently the whole runtime of xilem_web relies on interior mutability, and when this is used in the same event loop frame, this'll lead to a crash because it's already borrowed (but I'm not sure if it's possible to provoke this though). I'm thinking about whether it's a good idea to just `queueMicrotask` (which I think is just `wasm_bindgen_futures::spawn_local`) in `MessageThunk::push_message` to defuse this risk.

Philipp-M (commented):
So the reason why this doesn't compile has to do with the two different `Message` types (as well as `DynMessage`). One being of xilem_core, which requires `Send` and the other of xilem_web which doesn't require `Send` (which is necessary to be interoperable with e.g. `web_sys::Event`). See other views of xilem_web for more details.

We should probably document this, make it more clear to distinct between those two, because I've also had rust-analyzer suggesting me `xilem_core::Message` while I did want to use `crate::Message`.

Hmm the error message is really aweful, I wonder if it's somehow possible to improve this...

flosse:
> So the reason why this doesn't compile has to do with the two different `Message` types (as well as `DynMessage`). 

Thank you! Now it works :smile:

Philipp-M:
> Thank you! Now it works 😄

Good to know, I hope it's the same issue with `AfterBuild` etc.

The clippy suggestion is pretty dumb why CI fails... As if an endless loop doesn't have any side-effects...

Anyway, that example is somewhat redundant, as that should be a use-case for the new `interval` view now.
(Also I don't think we should use `async_repeat_raw` in an example, it's IMO mostly a hack to be able to capture stuff.)

flosse:
> Anyway, that example is somewhat redundant, as that should be a use-case for the new `interval` view now.

But now I added the "abort handle", that might be specific to async tasks.

xilem_web/src/concurrent/async_repeat.rs
flosse: > I'm not sure if we should expose this implementation detail of xilem_web, maybe wrap it? 

There is now a `AsyncRepeatProxy` struct as wrapper.
BTW: What about the `xilem_core::MessageProxy`? It's not used in `xilem_web`. Isn't that something we're looking for here?

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl AsyncRepeatProxy {
+    pub fn send_message<M>(&self, message: M)
+    where
+        M: Message,
+    {
+        let thunk = Arc::clone(&self.thunk);
+        spawn_local(async move {
+            thunk.push_message(message);
+        });
```
flosse, line 74: I'm not sure if this makes any difference to running `thunk.push_message` synchronously here?

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,143 @@
...
+use xilem_core::{MessageResult, Mut, NoElement, View, ViewId, ViewMarker};
+
+use crate::{context::MessageThunk, DynMessage, Message, ViewCtx};
+
+pub fn async_repeat<M, F, H, State, Action, Fut>(
+    future_future: F,
+    on_event: H,
+) -> AsyncRepeat<F, H, M>
+where
+    F: Fn(AsyncRepeatProxy, oneshot::Receiver<()>) -> Fut + 'static,
```
flosse, line 17: There might be a better API than passing `oneshot::Receiver<()>` as an abort handle, but for now at least it works :stuck_out_tongue_winking_eye:

Philipp-M: Oh right, it's not possible to cancel promises without something like this, unfortunate (though on the other hand also allows a more graceful shutdown)...
Yeah in that case that's probably ok. I wonder whether that should be a new type wrapper to make it more obvious, but I fear that needs quite a bit more code.

So this should probably documented then.

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,143 @@
...
+{
+    AsyncRepeat {
+        future_future,
+        on_event,
+        message: PhantomData,
+    }
+}
+
+pub struct AsyncRepeat<F, H, M> {
+    future_future: F,
```
Philipp-M, line 53: As Daniel said, I think
```suggestion
    init_future: F,
```

is probably more clear.

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    fn teardown(
+        &self,
+        view_state: &mut Self::ViewState,
+        _: &mut ViewCtx,
+        _: Mut<'_, Self::Element>,
+    ) {
+        let Some(tx) = view_state.abort_tx.take() else {
+            // TODO: Is that even allowed to happen?
```
Philipp-M, line 123: I think we can just `unwrap_throw` so that it's clear that this is a bug, when this happens...

xilem_web/web_examples/async_repeat/index.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Async repeat example | Xilem Web</title>
```
Philipp-M, line 4: I wonder if we should call the example different (though it's most descriptive for the view at least), like interval or ping or (news-)ticker as was suggested [here](https://github.com/linebender/xilem/pull/403#discussion_r1650740496), and maybe add the alternative (`interval`) there as well, so that people don't come to the idea to use `async_repeat`, when their use case is `interval`. At the very least, it should mention `interval` for this use-case.

Philipp-M (commented):
Same here, looks mostly good, but I think we should strive towards documenting all this stuff at this point

xilem_web/src/concurrent/async_repeat.rs
Philipp-M: I *guess* we could use it here, but it would make things more difficult with `web_sys` types, as it requires the `Send` bound, so I think having a separate proxy makes more sense and creates less headaches, unless we really want to support cross-web-worker stuff...

xilem_web/src/concurrent/async_repeat.rs
flosse: > I wonder whether that should be a new type wrapper to make it more obvious, but I fear that needs quite a bit more code.

I made a simple wrapper that should help a bit :smile:

xilem_web/web_examples/async_repeat/index.html
flosse: I renamed it to `spawn_tasks`, but that's probably not really appropriate either :thinking:

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,191 @@
...
+use std::{future::Future, marker::PhantomData, rc::Rc};
+
+use futures::{channel::oneshot, FutureExt};
+use wasm_bindgen::UnwrapThrowExt;
+use wasm_bindgen_futures::spawn_local;
+use xilem_core::{MessageResult, Mut, NoElement, View, ViewId, ViewMarker};
+
+use crate::{context::MessageThunk, DynMessage, Message, ViewCtx};
+
+/// Spawn a future.
```
Philipp-M, line 13:
```suggestion
/// Spawn an async task to update state asynchronously
```

xilem_web/src/concurrent/async_repeat.rs
```diff
@@ -0,0 +1,191 @@
...
+///
+/// The `init_future` function is given a [`AsyncRepeatProxy`] and a [`ShutdownSignal`].
```
Philipp-M, line 15:
```suggestion
/// The `init_future` function is given an [`AsyncRepeatProxy`] and a [`ShutdownSignal`].
```

xilem_web/web_examples/spawn_tasks/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+async fn create_ping_task(proxy: AsyncRepeatProxy, shutdown_signal: ShutdownSignal) {
+    log::debug!("Start ping task");
+    let mut abort = shutdown_signal.into_future().fuse();
+
+    #[allow(clippy::infinite_loop)]
+    loop {
+        // NOTE:
+        // This is just to simulate some async behavior.
+        // If you just need an interval, you should use
+        // [`interval`](xilem_web::concurrent::interval) instead.
```
Philipp-M, line 39: When using doc links, maybe put this as doc-comment above `create_ping_task`?

Philipp-M (approved):
Apart of nits, looks good, nice job with the `ShutdownSignal`.

flosse:
> Apart of nits, looks good, nice job with the `ShutdownSignal`.

thank you 😊

Philipp-M (approved):
I think you should've received an invitation to contribute.

Our merge-policy is usually that the author merges (after approval from someone else).

Otherwise looks good, so feel free to merge.

PR #486, opened 2024-08-05
xilem_web: Add `interval` `View`, which somewhat reflects `setInterval`

Philipp-M:
Simple but convenient `View`, e.g. for some kind of ticker. (Maybe needs an additional example)

### DIFF
--- a/xilem_web/src/concurrent/interval.rs
+++ b/xilem_web/src/concurrent/interval.rs
@@ -1,0 +1,137 @@
+// Copyright 2024 the Xilem Authors and the Druid Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::marker::PhantomData;
+
+use wasm_bindgen::{closure::Closure, JsCast, UnwrapThrowExt};
+use xilem_core::{MessageResult, Mut, NoElement, View, ViewId, ViewMarker};
+
+use crate::{DynMessage, OptionalAction, ViewCtx};
+
+/// Start an interval which invokes `callback` every `ms` milliseconds
+pub struct Interval<Callback, State, Action> {
+    ms: i32,
+    callback: Callback,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+/// Start an interval which invokes `callback` every `ms` milliseconds
+///
+/// `ms` is of type `i32` because of the weird choice for it in [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/setInterval#sect2)
+///
+/// # Examples
+///
+/// ```
+/// use xilem_web::{core::fork, concurrent::interval, elements::html::div, interfaces::Element};
+///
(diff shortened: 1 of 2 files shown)


xilem_web/src/concurrent/interval.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn rebuild<'el>(
+        &self,
+        prev: &Self,
+        view_state: &mut Self::ViewState,
+        _: &mut ViewCtx,
+        (): Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element> {
+        if prev.ms != self.ms {
+            clear_interval(view_state.interval_handle);
+            view_state.interval_handle = start_interval(&view_state.interval_fn, self.ms);
```
DJMcNab, line 110: There's a discussion here about what the expected behaviour is. Whether you should start a new task from fresh, or act as-if the previous task had been started with the new time.

DJMcNab (approved):
Seems like a reasonable design for a view. There are some complications, I'm sure you see.

xilem_web/src/concurrent/interval.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Start an interval which invokes `callback` every `ms` milliseconds
+pub struct Interval<Callback, State, Action> {
+    ms: i32,
+    callback: Callback,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+/// Start an interval which invokes `callback` every `ms` milliseconds
+///
+/// `ms` is of type `i32` because of the weird choice for it in [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/setInterval#sect2)
```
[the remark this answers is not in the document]
Philipp-M: I had `u32` before, and wasn't sure about it and changed it to `i32`, but as you say it, I'll probably change it back, as I see the reasoning (and agree with it)...

xilem_web/src/concurrent/interval.rs
Philipp-M: You mean you initiated the discussion or is there really somewhere a discussion somewhere?^^ Or do you mean the stuff in `memoized_await`?

Philipp-M: I don't think it's trivial though, might require a `setTimeout` with the remaining time + the change of the interval I guess...

DJMcNab: So, as in this is a common discussion with this kind of API. I don't think we've had it within Linebender.

We need to document it carefully

Philipp-M: We may extend that behavior in the future, because I see reasons for both behaviors, I chose this because it was more simple, but I agree, I'll add this to the doc-comment, so that it's clear

PR #487, opened 2024-08-05
Improve `calc` and xilem_web's `mathml_svg` example

Philipp-M:
Reduces a little bit of the boilerplate in `calc`

### DIFF
--- a/xilem/examples/calc.rs
+++ b/xilem/examples/calc.rs
@@ -187,6 +187,12 @@ impl Calculator {
 const DISPLAY_FONT_SIZE: f32 = 30.;
 const GRID_GAP: f64 = 2.;
 fn app_logic(data: &mut Calculator) -> impl WidgetView<Calculator> {
+    let num_row = |nums: [&'static str; 3], operator| {
+        flex_row((
+            nums.map(|num| digit_button(num).flex(1.)),
+            operator_button(operator).flex(1.),
+        ))
+    };
     flex((
         // Display
         centered_flex_row((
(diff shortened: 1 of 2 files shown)


xilem/examples/calc.rs
```diff
@@ -187,6 +187,12 @@ impl Calculator {
 const DISPLAY_FONT_SIZE: f32 = 30.;
 const GRID_GAP: f64 = 2.;
 fn app_logic(data: &mut Calculator) -> impl WidgetView<Calculator> {
+    let num_row = |nums: [&'static str; 3], operator| {
+        flex_row((
+            nums.map(|num| digit_button(num).flex(1.)),
```
Philipp-M, line 192: This is slightly more complex to avoid 2 additional lines, I would in practice rather use a little bit of repetition here, but I think it shows another interesting aspect, when it's not just an array with 3 elements...

xilem_web/web_examples/mathml_svg/src/main.rs
DJMcNab, line 1: This *probably* should be in a different PR...

xilem/examples/calc.rs
DJMcNab: I think this change should use a different free component, since this doesn't actually use any state from within the function justifying being a closure.

DJMcNab (approved):
Thanks.

xilem/examples/calc.rs
Philipp-M: I thought about it, but than it adds more boilerplate to add all the typing, do you think that's worth it anyway?

xilem_web/web_examples/mathml_svg/src/main.rs
Philipp-M: Should I extract this as a different PR? Or isn't this worth it?

PoignardAzur:
This PR appears to have added a regression on my machine. Here's what my calc example looks like:

image

PR #488, opened 2024-08-05
Implement event and update_pointer passes from pass spec RFC

PoignardAzur:
This is a first step in implementing the "Pass Specification" RFC:
https://github.com/linebender/rfcs/pull/7

Create a `passes` module.
Create event passes.
Create the update_pointer pass.

Remove `WidgetPod::update_hot_state` method.
Move mouse-cursor-handling code to update_pointer pass.
Implement pointer capture.
Refactor the TreeArena code.


PoignardAzur:
Unit tests pass, examples seem to work fine, and the last commit fixed a performance-cliff-in-waiting. If we want to implement the RFC step by step, now would be a good point to merge.

masonry/src/contexts.rs
```diff
@@ -333,6 +326,7 @@ impl_context_method!(
 // --- MARK: CURSOR ---
 // Cursor-related impls.
 impl_context_method!(EventCtx<'_>, {
+    // TODO - Rewrite doc
```
Philipp-M, line 329: Not sure if there's something missing yet in the code, but it would help reviewing this IMO.

masonry/src/contexts.rs
```diff
@@ -627,18 +608,35 @@ impl_context_method!(
 pub struct TimerToken;
 
 impl EventCtx<'_> {
+    // TODO - Document
+    // TODO - Figure out cases where widget should be notified of pointer capture
+    // loss
+    // TODO - Panic when event isn't PointerDown
```
Philipp-M, line 614: I wonder whether this should probably be done in this PR? It would make reviewing easier at least. (Same for the other methods)

Philipp-M (commented):
Just started looking into this. And quickly tested the examples in xilem. The Textbox widget seems to lead to a panic when hovering over it. Should be easy to reproduce by running the `mason` example.

PoignardAzur:
Will look into the crash.

masonry/src/widget/checkbox.rs
```diff
@@ -74,24 +74,19 @@ impl Widget for Checkbox {
...
             }
             PointerEvent::PointerUp(_, _) => {
-                if ctx.is_active() && !ctx.is_disabled() {
-                    if ctx.is_hot() {
-                        self.checked = !self.checked;
-                        ctx.submit_action(Action::CheckboxChecked(self.checked));
-                        trace!("Checkbox {:?} released", ctx.widget_id());
-                    }
-                    ctx.request_paint();
+                if ctx.is_active() && ctx.is_hot() && !ctx.is_disabled() {
```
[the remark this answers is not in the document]
PoignardAzur: No, it's a drive-by fix.

masonry/src/widget/align.rs
```diff
@@ -85,17 +85,11 @@ impl Align {
 
 // --- MARK: IMPL WIDGET ---
 impl Widget for Align {
-    fn on_pointer_event(&mut self, ctx: &mut EventCtx, event: &PointerEvent) {
-        self.child.on_pointer_event(ctx, event);
-    }
+    fn on_pointer_event(&mut self, _ctx: &mut EventCtx, _event: &PointerEvent) {}
```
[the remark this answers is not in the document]
PoignardAzur: I'm split. On the one hand, some widgets won't need to implement them. On the other hand, filling in default implementations take like two minutes, and it's good to be mindful about what methods you implement. I lean towards "no unless >95% of widgets are expected not to implement the method".

masonry/src/passes/event.rs
```diff
@@ -0,0 +1,344 @@
...
+
+impl<'a> PassCtx<'a> {
+    fn parent(&self) -> Option<WidgetId> {
+        let parent_id = self.widget_state.parent_id?;
+        let parent_id = parent_id.try_into().unwrap();
+        Some(WidgetId(parent_id))
+    }
+}
+
+// TODO - Merge copy-pasted code
```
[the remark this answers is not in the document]
PoignardAzur: After I implement the other passes.

masonry/src/event.rs
```diff
@@ -247,6 +247,13 @@ impl PointerEvent {
         }
     }
 
+    pub fn position(&self) -> Option<LogicalPosition<f64>> {
+        match self {
+            PointerEvent::PointerLeave(_) | PointerEvent::HoverFileCancel(_) => None,
```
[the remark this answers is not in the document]
PoignardAzur: Mostly because they're associated with a mouse position.

masonry/src/passes/event.rs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        merge_state_up(&mut root.widget_arena, widget_id, root_state);
+        target_widget_id = parent_id;
+    }
+
+    Handled::from(is_handled)
+}
+
+#[cfg(FALSE)]
```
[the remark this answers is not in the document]
PoignardAzur: It's scratchpad stuff. I think I can remove the first two, but the latter two do need to be re-implemented, so I'll keep them around until the whole RFC is implemented, with a comment for context.

masonry/src/tree_arena.rs
```diff
@@ -76,12 +99,15 @@ impl<Item> TreeArena<Item> {
     ///
     /// Using [`insert_child`](TreeArenaTokenMut::insert_child) on this token
```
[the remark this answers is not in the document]
PoignardAzur: Yup.

masonry/src/passes/event.rs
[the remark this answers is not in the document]
PoignardAzur: Done.

masonry/src/tree_arena.rs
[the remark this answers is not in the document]
PoignardAzur: Fixed.

masonry/src/render_root.rs
```diff
@@ -300,16 +362,17 @@ impl RenderRoot {
...
             .expect("root widget not in widget tree");
-        let (widget, widget_token) = root_widget_token
+        let widget_ref = root_widget_token
             .into_child_mut(self.root.id().to_raw())
             .expect("root widget not in widget tree");
 
         // Our WidgetArena stores all widgets as Box<dyn Widget>, but the "true"
         // type of our root widget is *also* Box<dyn Widget>. We downcast so we
         // don't add one more level of indirection to this.
+        let widget = widget_ref.item;
```
waywardmonkeys, line 375: why not just combine with the line below?

```
let widget = widget_ref.item.as_mut_dyn_any()...
```

masonry/src/widget/checkbox.rs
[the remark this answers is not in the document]
waywardmonkeys: I really don't like drive-by fixes much, but recognize I may well be in the minority ...

Why submit 1 PR when you can submit 5? :)

But this won't block things ...

masonry/src/widget/align.rs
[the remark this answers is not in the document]
waywardmonkeys: On the other hand, it is a bunch of boilerplate and it means you can't search for meaningful implementations of the method rather than "all implementations, even the empty ones".

waywardmonkeys:
I'm essentially okay with this in the current state as an incremental step towards a better place. I haven't run it myself, just reviewed code and RFCs. I think it would benefit strongly from additional documentation and comments within the code, even if it were largely adapted from some of the RFC text.

I'll defer to @DJMcNab or @Philipp-M for an actual approval as they're more up to speed on some of this than I am.

masonry/src/passes/event.rs
```diff
@@ -0,0 +1,277 @@
...
+fn run_event_pass<E>(
+    root: &mut RenderRoot,
+    root_state: &mut WidgetState,
+    target: Option<WidgetId>,
+    event: &E,
+    pass_fn: impl FnMut(&mut dyn Widget, &mut EventCtx, &E),
+) -> Handled {
+    let mut pass_fn = pass_fn;
+
+    let mut target_widget_id = target;
```
Philipp-M, line 208: Nit: you could just use `target`/`pass_fn` directly, by declaring it `mut` in the function signature (or is this intentional?)

masonry/src/passes/event.rs
```diff
@@ -0,0 +1,277 @@
...
+                widget.short_type_name(),
+                widget_id.to_raw(),
+            );
+
+            pass_fn(widget, &mut ctx, event);
+            is_handled = ctx.is_handled;
+        }
+
+        merge_state_up(&mut root.widget_arena, widget_id, root_state);
+        target_widget_id = parent_id;
```
Philipp-M, line 235: So unlike browser behavior, this doesn't do a capturing phase (yet?), only bubbling.

(Just noticing)

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,220 @@
...
+// TODO - Replace LifecycleCtx with UpdateCtx
+pub(crate) fn run_targeted_update_pass(
+    root: &mut RenderRoot,
+    root_state: &mut WidgetState,
+    target: Option<WidgetId>,
+    pass_fn: impl FnMut(&mut dyn Widget, &mut LifeCycleCtx),
+) {
+    let mut pass_fn = pass_fn;
+
+    let mut target_widget_id = target;
```
Philipp-M, line 84: Same Nit as before

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,220 @@
...
+            .find_widget_at_pos(pos)
+            .map(|widget| widget.id())
+    } else {
+        None
+    };
+    if let Some(capture_target) = root.state.pointer_capture_target {
+        if next_hovered_widget != Some(capture_target) {
+            next_hovered_widget = None;
+        }
+    }
```
Philipp-M, line 120:
```suggestion
    let pointer_capture_target = root.state.pointer_capture_target;
    if pointer_capture_target.is_some() && pointer_capture_target != next_hovered_widget {
        next_hovered_widget = None;
    }
```

(could reuse that `pointer_capture_target` below)

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,220 @@
...
+        hovered_set.insert(*widget_id);
+    }
+
+    trace!("prev_hovered_path: {:?}", prev_hovered_path);
+    trace!("next_hovered_path: {:?}", next_hovered_path);
+
+    // This algorithm is written to be resilient to future changes like reparenting and multiple
+    // cursors. In theory it's O(Depth² * CursorCount) in the worst case, which isn't too bad
+    // (cursor count is usually 1 or 2, depth is usually small), but in practice it's almost
+    // always O(Depth * CursorCount) because we only need to update the hovered status of the
```
Philipp-M, line 136: Isn't that then more like the average case?

masonry/src/tree_arena.rs
```diff
@@ -181,15 +220,18 @@ impl<'a, Item> TreeArenaToken<'a, Item> {
...
+                return Some(ArenaRef {
+                    parent_id: self.id,
+                    id: Some(child.id),
+                    item: &child.item,
+                    children: ArenaRefChildren {
+                        id: Some(child.id),
                         children: &child.children,
                     },
-                ));
+                });
```
Philipp-M, line 234: I think a lot of that duplicated boilerplate could be avoided (same for `ArenaMut`)?

masonry/src/widget/align.rs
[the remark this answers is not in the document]
Philipp-M: Hmm I also think default implementations may reduce unnecessary clutter and I think are one of the main advantages of avoiding the hierarchical child-calling we had previously.
Instead I think good documentation should help the implementer of that trait what they actually need to implement for their widget.

Reasons for default impls get even more vaild, when we add more of these methods to the trait (which I'm pretty sure will happen). So we could extend the trait more deliberately without having to think about all the boilerplate or breaking changes that it causes.

masonry/src/widget/checkbox.rs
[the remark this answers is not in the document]
Philipp-M: I think it depends on the complexity (i.e. is it worth it to notify a lot of people because of some mini-code-aesthetic issue), but I also usually think it's better to isolate PRs

Philipp-M (commented):
Looks mostly good (Daniel should probably look over this as well, I likely have missed stuff (so far)).

I'm all for more incremental PRs of that RFC, though it would be nice, when they're more atomic/single isolated changes like the arena changes (and ideally including already doc-comments, and a little bit more polished).

It would make reviewing easier (and likely faster) and more effective, like splitting single changes that seem to be mostly consolidated out of a working branch. I know that may increase a little bit the workload, but I think also results in higher code-quality merged in main (and less numerous half-finished building-blocks).
(I'll review more tomorrow, way too late here...)

DJMcNab:
I think the PR description needs some details, and a link to the RFC.

masonry/src/passes/event.rs
DJMcNab: We decided against a capturing phase, in the hope that it won't be necessary.

masonry/src/passes/update.rs
PoignardAzur: It's not average or amortized, because by default after you visit a widget you immediately visit the next one in `next_hovered_path` (its parent).

To be specific, worst case for the algorithm below that comment is `O(Depth * CursorCount)` *unless* a widget has been reparented (which currently isn't possible but might be in the future) in which case it's `O(Depth² * CursorCount)`.

(The above assumes that accessing a widget is `O(1)`, which isn't the case but will be in the future.)

masonry/src/tree_arena.rs
PoignardAzur: I'm not worried about duplicated boilerplate in library code. Especially for TreeArena, the error surface isn't that large; since the rest of the codebase uses it everywhere, it has extensive test coverage.

Philipp-M: It's less about that it's well tested or errornous, but just IMO unnecessary clutter in code that may be simplified by containing it in a well named function, to focus on the more important bits?

masonry/src/passes/event.rs
```diff
@@ -0,0 +1,344 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
PoignardAzur: I'll move a lot of the RFC into documentation once the port is complete. Right now I'm focusing on implementation.

masonry/src/widget/align.rs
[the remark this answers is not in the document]
PoignardAzur: I'm not so sure, but I guess it doesn't hurt to add a default impl.

Empty implementations can be removed in a follow-up PR.

masonry/src/widget/checkbox.rs
```diff
@@ -30,15 +30,15 @@ impl Checkbox {
     pub fn new(checked: bool, text: impl Into<ArcStr>) -> Checkbox {
         Checkbox {
             checked,
-            label: WidgetPod::new(Label::new(text)),
+            label: WidgetPod::new(Label::new(text).with_skip_pointer(true)),
```
DJMcNab, line 33: When would a label not skip pointer handling?

I guess maybe if we get link support within labels?

masonry/src/testing/harness.rs
```diff
@@ -570,7 +582,7 @@ impl TestHarness {
             // Remove '<test_name>.new.png' file if it exists
             let _ = std::fs::remove_file(&new_path);
             new_image.save(&new_path).unwrap();
-            panic!("No reference file");
+            panic!("Snapshot test '{test_name}' failed: No reference file");
```
DJMcNab, line 588: Great. I am planning on making these snapshot tests *much* more user-friendly, but this is a good start.

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,163 @@
...
+    root.widget_arena
+        .widget_states
+        .get_id_path(widget_id.to_raw())
+        .iter()
+        .map(|&id| WidgetId(id.try_into().unwrap()))
+        .collect()
+}
+
+// TODO - Replace LifecycleCtx with UpdateCtx
+pub(crate) fn run_targeted_update_pass(
```
DJMcNab, line 27:
```suggestion
fn run_targeted_update_pass(
```

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    let hovered_widget = root.state.hovered_path.last().copied();
+    let mut next_hovered_widget = if let Some(pos) = pointer_pos {
+        // TODO - Apply scale?
+        root.get_root_widget()
+            .find_widget_at_pos(pos)
+            .map(|widget| widget.id())
+    } else {
+        None
+    };
```
DJMcNab, line 63: I wonder if the optimisation of "start from the previous hovered widget" is worthwhile.
i.e. special case the common case where the mouse stays within the same widget.

(Notice that we need to be careful to handle any children of that hovered widget, and any reparenting; i.e. this isn't really for this PR )

masonry/src/passes/update.rs
DJMcNab: Ok, so the factor of $$\text{Depth}^{2}$$ occurs if at every item in the tree, the parent has changed. I.e. the tree used to be:
```mermaid
graph TB;
    A;
    B;
    A-->C;
    A-->D;
    C-->E;
```
where E was hovered over, and is now something like:
```mermaid
graph TB;
    A;
    A-->D;
    B-->C;
    D-->E;
```
even if E is still hovered over?

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,163 @@
...
+            let is_hot = hovered_set.contains(&ctx.widget_id());
+
+            if ctx.widget_state.is_hot != is_hot {
+                widget.on_status_change(ctx, &StatusChange::HotChanged(is_hot));
+            }
+            ctx.widget_state.is_hot = is_hot;
+        });
+    }
+
+    // TODO - Make sure widgets are iterated from the bottom up
```
DJMcNab, line 103:
```suggestion
    // TODO - Make sure widgets are iterated from the bottom up for efficiency.
```

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,163 @@
...
+                != hovered_set.contains(widget_id)
+        {
+            update_hovered_status_of(root, *widget_id, &hovered_set);
+        }
+    }
+
+    root.state.hovered_path = next_hovered_path;
+
+    // -- UPDATE CURSOR --
+    // TODO - Rewrite more cleanly
```
DJMcNab, line 132: How?

PoignardAzur:
I've pushed a refactor that was probably overdue. There's a lot less duplicated code now, addressing one of the main points of feedback, and it's overall more readable; it should also be easier to iterate on.

I've prepared the PR's commit message as per Daniel's request.

That said, yeah, this is still a work in progress, and documenting everything in real time would slow me down more than a little. At that point it's either small-ish PRs (this one is already quite big!) with some missing doc and boilerplate code, or one huge extremely polished PR one month from now.

I can't maintain a high level of polish while I'm refactoring major parts of the engine, and honestly I find it a bit harsh that I'm being held to that standard.

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,163 @@
...
+    let cursor_changed = next_hovered_widget
+        .is_some_and(|id| root.widget_arena.get_state_mut(id).item.cursor_changed);
```
DJMcNab, line 134: Can this be a `mem::take` of the cursor_changed?

masonry/src/passes/update.rs
```diff
@@ -0,0 +1,163 @@
...
+    if hovered_widget != next_hovered_widget || cursor_changed {
+        let cursor;
+        if let Some(capture_target) = root.state.pointer_capture_target {
```
DJMcNab, line 137: I'm presuming the case where the capture target's `cursor_changed` is set is handled elsewhere.

masonry/src/passes/event.rs
DJMcNab, line 1: I really like this code! It's reasonably clear what's happening, and why.

I think there could be a few more comments to make understanding a bit easier.

DJMcNab (commented):
I've not reasoned about this from the top down, but I really like the core of the new code!

Not approving to allow time for review feedback to be handled.

DJMcNab:
Seems like that was poor timing - just did a review pass.

DJMcNab (approved):
Looking at it again, nothing in these review comments is especially controversial, and so we should be fine to land once these are addressed.

(I would however like to echo the concern about review comments not being appropriately actioned)

masonry/src/passes/update.rs
PoignardAzur: It's not just for efficiency, we want iteration order to be predictable. We'll probably want to document it too.

masonry/src/passes/update.rs
PoignardAzur: I don't think so?

masonry/src/passes/update.rs
PoignardAzur: ... Crap, I don't think it is. Good catch.

masonry/src/passes/update.rs
PoignardAzur: > I wonder if the optimisation of "start from the previous hovered widget" is worthwhile.

I doubt it. For anything `O(Depth)` the overhead is already small enough that any optimization that makes the code more complex / adds hidden assumptions is a bad trade.

masonry/src/passes/update.rs
PoignardAzur: Yup.

masonry/src/passes/update.rs
PoignardAzur: In that case it's because I don't want to merge that change. I don't think it improves readability.

It was kind of a dick move to mark it as resolved without explaining the reason, sorry.

I think every other comment you'd made I've either addressed in a follow-up commit or explained why I wouldn't?

masonry/src/widget/checkbox.rs
PoignardAzur: Links and selecting text (though I think maybe we have the Prose widget for that now?).

I guess we could go ahead and say they're *always* skipped and only Prose gets to have links. I haven't really thought this through and I don't think it's a priority.

masonry/src/passes/update.rs
PoignardAzur: (Note that sometimes I mark a comment as resolved after I wrote the change that addresses it, but before I've pushed it. It helps me keep track of what to do, but maybe it's not worth the confusion it creates for reviewers.)

masonry/src/passes/update.rs
PoignardAzur: I don't really know, but I do want to leave that comment for a future cleanup.

masonry/src/passes/update.rs
Philipp-M: Yeah, I'm fine with a short comment, explaining that (*although* I disagree), generally I'd prefer, when the author of a review-comment resolves it, unless it's obviously resolved, as (and don't take this personal) it looks like silencing concerns.

> I think every other comment you'd made I've either addressed

No, the nits for example not (although being nits, it's not being really important), but a short answer to e.g. "or is this intentional?" would've been nice.

masonry/src/passes/event.rs
PoignardAzur: Done for pass_fn. Renamed target_widget_id for something clearer instead.

masonry/src/contexts.rs
PoignardAzur: I'm leaving this for a later PR.

masonry/src/contexts.rs
PoignardAzur: I've added a quick-and-dirty assertion. Documentation will be for a later PR.

masonry/src/passes/update.rs
PoignardAzur: Done.

masonry/src/passes/update.rs
PoignardAzur: Oh, you meant to skip the later part that set this to false? Yeah, in retrospect that was a good idea. It's moot now.

masonry/src/passes/update.rs
PoignardAzur: Re-reading this, I think I was assuming `capture_target` would always be the same as `hovered_widget`, but that's not actually the case (`hovered_widget` could be None it we'd still want to check the capture target).

In the end I just removed the `cursor_changed` part for expediency.

masonry/src/passes/update.rs
DJMcNab: (I also don't think that this suggestion is better)

DJMcNab (approved):

waywardmonkeys:
Yay!

PR #489, opened 2024-08-06
Fix non-standard textbox behavior in Windows

### LINKED ISSUE #465: xilem_web: Possible bug when using `.boxed` and changing type of HTML node
I think I found an issue when using `.boxed` combinator from #460 and using it to dynamically change the type of HTML element.

I created a minimal reproduction [here](https://github.com/casey/xilem-update-bug/blob/master/src/main.rs), which looks like this:

```rust
use xilem_web::{document_body, elements::html, interfaces::Element, App, DomView};

fn update(state: &mut bool) -> impl DomView<bool> {
    html::div((
        html::button("click me!").on_click(|state: &mut bool, _| *state = !*state),
        html::br(()),
        if *state {
            html::div("div!").boxed()
        } else {
            html::span("span!").boxed()
        },
    ))
}

fn main() {
    App::new(document_body(), false, update).run();
}
```

The boxed element keeps it's initial type and text, even when the button is clicked.

RagibHasin:


masonry/src/text/selection.rs
```diff
@@ -184,8 +184,16 @@ impl<T: Selectable> TextWithSelection<T> {
...
                         _ => Handled::No,
                     },
-                    winit::keyboard::Key::Unidentified(_) => todo!(),
-                    winit::keyboard::Key::Dead(_) => todo!(),
+                    winit::keyboard::Key::Unidentified(key) => {
+                        // TODO: figure out correct behavior
+                        tracing::warn!(?key, "got unidentified key, did not handle");
+                        Handled::No
+                    }
+                    winit::keyboard::Key::Dead(key) => {
```
DJMcNab, line 192: In selection code, there's no need to handle a dead key, so we can just use `Handled::No`

masonry/src/text/edit.rs
```diff
@@ -188,23 +188,18 @@ impl<T: EditableText> TextEditor<T> {
...
-                                // We have just added this character, so we are "affined" with it
-                                Affinity::Downstream,
-                            ));
-                            let contents = self.text().as_str().to_string();
-                            ctx.submit_action(Action::TextChanged(contents));
-                            Handled::Yes
+                            self.insert_text(event.text.as_ref().unwrap_or(c), ctx)
+                        }
+                        Key::Unidentified(_) => {
+                            if cfg!(windows) {
```
DJMcNab, line 194: Why is this `cfg!(windows)`? This at least needs an explanatory comment.

masonry/src/text/selection.rs
```diff
@@ -184,8 +184,16 @@ impl<T: Selectable> TextWithSelection<T> {
                         }
                         _ => Handled::No,
                     },
-                    winit::keyboard::Key::Unidentified(_) => todo!(),
-                    winit::keyboard::Key::Dead(_) => todo!(),
+                    winit::keyboard::Key::Unidentified(key) => {
+                        // TODO: figure out correct behavior
+                        tracing::warn!(?key, "got unidentified key, did not handle");
```
DJMcNab, line 189: I suspect that just doing `Handled::No` here is best.

It's bad form to log the keys that a user has input, in case they have e.g. tried to enter a password.

DJMcNab (commented):
This seems reasonable, thanks! Using the `text` field has been on my mental list of things to get around to, but I haven't been able to prioritise.

I've not tested the change myself yet. Will do so once review comments are handled.

masonry/src/text/edit.rs
RagibHasin: I actually don't know that. I almost exclusively use Windows and do not own a Mac. So I don't know what the behavior should be for Linux or macOS.

DJMcNab:
Can you please apply:
```patch
From d8f24005fb02dabb8b82837940fc2c192f480908 Mon Sep 17 00:00:00 2001
From: Daniel McNab <36049421+DJMcNab@users.noreply.github.com>
Date: Tue, 6 Aug 2024 17:31:11 +0100
Subject: [PATCH] Remove the conditional for windows

---
 masonry/src/text/edit.rs | 16 ++++------------
 1 file changed, 4 insertions(+), 12 deletions(-)

diff --git a/masonry/src/text/edit.rs b/masonry/src/text/edit.rs
index bf4b28dd..50e41a11 100644
--- a/masonry/src/text/edit.rs
+++ b/masonry/src/text/edit.rs
@@ -190,18 +190,10 @@ impl<T: EditableText> TextEditor<T> {
                         Key::Character(c) => {
                             self.insert_text(event.text.as_ref().unwrap_or(c), ctx)
                         }
-                        Key::Unidentified(_) => {
-                            // Ensures conformance with native Windows behavior in case some
-                            // key event hook sets unidentifed key code and some text for pickup
-                            if cfg!(windows) {
-                                event
-                                    .text
-                                    .as_ref()
-                                    .map_or(Handled::No, |c| self.insert_text(c, ctx))
-                                Handled::No
-                        }
+                        Key::Unidentified(_) => match event.text.as_ref() {
+                            Some(text) => self.insert_text(text, ctx),
+                            None => Handled::No,
+                        },
                         Key::Dead(d) => {
                             eprintln!("Got dead key {d:?}. Will handle");
                             Handled::No
-- 
2.45.2

```

Because you have chosen to create this PR from your default branch, I can't push to it

masonry/src/text/edit.rs
RagibHasin: Thanks!

RagibHasin: Thanks!

DJMcNab (approved):
Thank you! We are aware that this code will still have some janky and unexpected behaviour, but it's good to close some holes, anyway.

PR #491, opened 2024-08-06
xilem: Fix calc example (missing `flex` property)

### LINKED ISSUE #490: Regression in calc example
Here's what the calc example looks like in main branch right now (commit 99d61603):

image

A quick bisect tells me #487 is responsible for the regression.

My system is 64 bit PopOS 22.04.

Philipp-M:
Fixes #490

### DIFF
--- a/xilem/examples/calc.rs
+++ b/xilem/examples/calc.rs
@@ -192,6 +192,7 @@ fn app_logic(data: &mut Calculator) -> impl WidgetView<Calculator> {
             nums.map(|num| digit_button(num).flex(1.)),
             operator_button(operator).flex(1.),
         ))
+        .flex(1.0)
     };
     flex((
         // Display


jaredoconnell (approved):
Right when I saw the screenshot of the regression, I knew it was a missing flex call. Tested, and this fixes it.

xilem/examples/calc.rs
```diff
@@ -192,6 +192,7 @@ fn app_logic(data: &mut Calculator) -> impl WidgetView<Calculator> {
             nums.map(|num| digit_button(num).flex(1.)),
             operator_button(operator).flex(1.),
         ))
+        .flex(1.0)
```
DJMcNab, line 195: I would add this to the individual items, rather than inside the creation function.
That matches the existing items; the vertical `flex` is a property of the container the row is in, not inherent to the item itself.

Philipp-M: You're probably right...

PR #492, opened 2024-08-08
Xilem stopwatch example

jaredoconnell:
This pull request adds a stopwatch example with start, stop, lap, and reset buttons.
When the stopwatch is active, the stop and lap buttons are visible. When the stopwatch is not active, the reset and start buttons are example.
To reduce resource usage, I set the timer period to 50ms, and only show 1 decimal place.

Screenshot 2024-08-08 at 10 55 49 AM

This is an older screenshot of what it looks like when it exceeds a day:

Screenshot 2024-08-07 at 10 21 43 PM

Also tested on Windows:
image

Let me know if there are any improvements you see that I can make.

I included the android boilerplate, but I haven't tested it.

### DIFF
--- a/xilem/examples/stopwatch.rs
+++ b/xilem/examples/stopwatch.rs
@@ -60,6 +60,8 @@ impl Stopwatch {
         self.active = false;
         self.last_start_time = None;
         self.added_duration = Duration::ZERO;
+        self.completed_lap_splits.clear();
+        self.split_start_time = Duration::ZERO;
         self.update_display();
     }
 
--- a/xilem/examples/stopwatch.rs
+++ b/xilem/examples/stopwatch.rs
@@ -14,6 +14,7 @@ use masonry::widget::{Axis, CrossAxisAlignment, MainAxisAlignment};
 use xilem::{WidgetView, Xilem};
 use xilem::view::{AnyFlexChild, async_repeat, button, flex, FlexExt, FlexSpacer, label};
 use tracing::warn;
+use xilem_core::fork;
 
 /// The state of the entire application.
 ///
@@ -98,34 +99,36 @@ fn get_formatted_duration(dur: &Duration) -> String {
 }
 
 fn app_logic(data: &mut Stopwatch) -> impl WidgetView<Stopwatch> {
-    flex((
-        FlexSpacer::Fixed(5.0),
-        label(get_formatted_duration(&data.displayed_duration)).text_size(70.0),
+    fork(
         flex((
-            lap_reset_button(data),
-            start_stop_button(data),
-        ),
-        ).direction(Axis::Horizontal),
-        laps_section(data),
-        label(data.displayed_error.as_ref()),
+                 FlexSpacer::Fixed(5.0),
+                 label(get_formatted_duration(&data.displayed_duration)).text_size(70.0),
+                 flex((
+                          lap_reset_button(data),
+                          start_stop_button(data),
+                      ),
+                 ).direction(Axis::Horizontal),
+                 laps_section(data),
+                 label(data.displayed_error.as_ref()),
+        ))
+            .main_axis_alignment(MainAxisAlignment::Start)
+            .cross_axis_alignment(CrossAxisAlignment::Center),
         async_repeat(
-            |proxy| async move {
-                let mut interval = time::interval(Duration::from_millis(50));
-                loop {
-                    interval.tick().await;
-                    let Ok(()) = proxy.message(()) else {
-                        break;
-                    };
-                }
-            },
-            |data: &mut Stopwatch, ()|
-                if data.active {
-                    data.update_display();
-                },
-        ),
-    ))
-        .main_axis_alignment(MainAxisAlignment::Start)
-        .cross_axis_alignment(CrossAxisAlignment::Center)
+             |proxy| async move {
+                 let mut interval = time::interval(Duration::from_millis(50));
+                 loop {
+                     interval.tick().await;
+                     let Ok(()) = proxy.message(()) else {
+                         break;
+                     };
+                 }
+             },
+             |data: &mut Stopwatch, ()|
+             if data.active {
+                 data.update_display();
+             },
+        )
+    )
 }
 
 fn laps_section(data: &mut Stopwatch) -> Vec<AnyFlexChild<Stopwatch>> {
--- a/xilem/examples/stopwatch.rs
+++ b/xilem/examples/stopwatch.rs
@@ -145,17 +145,14 @@ fn laps_section(data: &mut Stopwatch) -> impl FlexSequence<Stopwatch> {
         items.push(single_lap(i, split_dur, &total_dur));
     }
     let current_split_duration = data.get_current_duration().sub(total_dur);
-    let mut reversed = Vec::new();
-    // Add the current lap first so that it shows up on the top.
-    reversed.push(single_lap(
+    // Add the current lap, which is not stored in completed_lap_splits
+    items.push(single_lap(
         current_lap,
         &current_split_duration,
         &data.get_current_duration(),
     ));
-    for item in items.into_iter().rev() {
-        reversed.push(item);
-    }
-    reversed
+    items.reverse();
+    items
 }
 
 fn single_lap(
--- a/xilem/examples/stopwatch.rs
+++ b/xilem/examples/stopwatch.rs
@@ -15,6 +15,7 @@ use winit::window::Window;
 use xilem::view::{async_repeat, button, flex, label, AnyFlexChild, FlexExt, FlexSpacer};
 use xilem::{WidgetView, Xilem};
 use xilem_core::fork;
+use xilem_core::one_of::Either;
 
 /// The state of the entire application.
 ///
@@ -113,9 +114,7 @@ fn app_logic(data: &mut Stopwatch) -> impl WidgetView<Stopwatch> {
             FlexSpacer::Fixed(1.0),
             laps_section(data),
             label(data.displayed_error.as_ref()),
-        ))
-        .main_axis_alignment(MainAxisAlignment::Start)
-        .cross_axis_alignment(CrossAxisAlignment::Center),
+        )),
         async_repeat(
             |proxy| async move {
                 let mut interval = time::interval(Duration::from_millis(50));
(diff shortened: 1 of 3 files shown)


xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,263 @@
...
+            data.stop();
+        })
+        .into_any_flex()
+    } else {
+        button("Start", |data: &mut Stopwatch| {
+            data.start();
+        })
+        .into_any_flex()
+    }
+}
```
DJMcNab, line 192:
```suggestion
fn start_stop_button(data: &mut Stopwatch) -> impl WidgetView<Stopwatch> {
    if data.active {
        Either::A(button("Stop", |data: &mut Stopwatch| {
            data.stop();
        }))
    } else {
        Either::B(button("Start", |data: &mut Stopwatch| {
            data.start();
        }))
    }
}
```

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,263 @@
...
+    split_dur: &Duration,
+    total_dur: &Duration,
+) -> impl WidgetView<Stopwatch> {
+    flex((
+        FlexSpacer::Flex(1.0),
+        label(format!("Lap {}", lap_id + 1)).flex(1.0),
+        label(get_formatted_duration(split_dur)).flex(1.0),
+        label(get_formatted_duration(total_dur)).flex(1.0),
+    ))
+    .direction(Axis::Horizontal)
```
DJMcNab, line 175: I think you might be able to use `must_fill_major_axis(true)` here?

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,263 @@
...
+                if data.active {
+                    data.update_display();
+                }
+            },
+        ),
+    )
+}
+
+/// Creates a list of items that shows the lap number, split time, and total cumulative time.
+fn laps_section(data: &mut Stopwatch) -> Vec<AnyFlexChild<Stopwatch>> {
```
DJMcNab, line 139: This is a key use case for the changes in #472.

You should be able to return `impl WidgetViewSequence<StopWatch>` here, which means that you wouldn't need to use the box-based `AnyFlexChild` (since all the items in the sequence are of the same type)

DJMcNab (commented):
Some first thoughts, from not having run this

xilem/examples/stopwatch.rs
jaredoconnell: That fixed it. Thanks!

xilem/examples/stopwatch.rs
jaredoconnell: That's what I needed. Thank you. This example will be a good example for future users of Xilem.
I did this for both buttons.

xilem/examples/stopwatch.rs
jaredoconnell: I'm not sure the right way to use that. It may make sense for you to do that post-merge.

Philipp-M: I do think, in that simple case it could also be `Vec<impl WidgetView<StopWatch>>` but for being an example more educating about how this can be used in a more general sense, an `impl WidgetViewSequence<StopWatch>` is probably indeed better.

You should be able to use this as return type, when removing the explicit type for `let mut items = ...`

And by removing all the `into_any_flex` in there...

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,260 @@
...
+    // The absolute time of the last start for calculating elapsed time.
+    last_start_time: Option<SystemTime>,
+    // The duration displayed; updated by by `update_display()`
+    displayed_duration: Duration,
+    // An error string to display if there is an error.
+    displayed_error: String,
+    // A list of the length of all completed splits. Does not include the current split.
+    completed_lap_splits: Vec<Duration>,
+    // The duration of the main timer when the split was started.
+    split_start_time: Duration,
```
Philipp-M, line 38: I guess these could also be doc-comments, for better documentation support in editors?

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+fn app_logic(data: &mut Stopwatch) -> impl WidgetView<Stopwatch> {
+    fork(
+        flex((
+            FlexSpacer::Fixed(5.0),
+            label(get_formatted_duration(&data.displayed_duration)).text_size(70.0),
+            flex((lap_reset_button(data), start_stop_button(data))).direction(Axis::Horizontal),
+            FlexSpacer::Fixed(1.0),
+            laps_section(data),
```
Philipp-M, line 115: I wonder whether there should be a "legend-row" above this, explaining the two different times?

Philipp-M (approved):
Looks good (apart of nits), I haven't tested the android part, so maybe someone else should do this?

xilem/examples/stopwatch.rs
jaredoconnell: That would really benefit from a future grid layout type. Getting the alignment right with the current options would be a little fragile.

xilem/examples/stopwatch.rs
jaredoconnell: Done.

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,261 @@
...
+            &current_split_duration,
+            &data.get_current_duration(),
+        )
+        .into_any_flex(),
+    );
+    for item in items.into_iter().rev() {
+        reversed.push(item);
+    }
+    reversed
+}
```
Philipp-M, line 162:
```suggestion
fn laps_section(data: &mut Stopwatch) -> FlexSequence<Stopwatch> {
    let mut items = Vec::new();
    let mut total_dur = Duration::ZERO;
    let current_lap = data.completed_lap_splits.len();

    for (i, split_dur) in data.completed_lap_splits.iter().enumerate() {
        total_dur = total_dur.add(*split_dur);
        items.push(single_lap(i, split_dur, &total_dur));
    }
    let current_split_duration = data.get_current_duration().sub(total_dur);
    let mut reversed = Vec::new();
    reversed.push(single_lap(
        current_lap,
        &current_split_duration,
        &data.get_current_duration(),
    ));
    for item in items.into_iter().rev() {
        reversed.push(item);
    }
    reversed
}
```

So this is what I mean, I have tried `impl WidgetViewSequence<Stopwatch> `, but I think it's incompatible in contexts requiring a `FlexSequence` (the `SuperElement` magic only seems to work I think when it's not yet explicitely typed yet, so inference can kick in). So this above would work instead, or alternatively also `Vec<impl WidgetView<Stopwatch>>` as return type.

(Requires imports obviously)

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,261 @@
...
+    let mut items: Vec<AnyFlexChild<Stopwatch>> = Vec::new();
+    let mut total_dur = Duration::ZERO;
+    let current_lap = data.completed_lap_splits.len();
+
+    for (i, split_dur) in data.completed_lap_splits.iter().enumerate() {
+        total_dur = total_dur.add(*split_dur);
+        items.push(single_lap(i, split_dur, &total_dur).into_any_flex());
+    }
+    let current_split_duration = data.get_current_duration().sub(total_dur);
+    let mut reversed = Vec::new();
```
DJMcNab, line 149: Rather than using a different vector, could we use [`Vec::reverse`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reverse).

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,261 @@
...
+    } else {
+        Either::B(button("Start", |data: &mut Stopwatch| {
+            data.start();
+        }))
+    }
+}
+
+fn lap_reset_button(data: &mut Stopwatch) -> impl WidgetView<Stopwatch> {
+    if data.active {
+        Either::A(button("  Lap  ", |data: &mut Stopwatch| {
```
DJMcNab, line 196: :laughing:

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,261 @@
...
+                    warn!("error calculating elapsed time: {}", err.to_string());
+                    self.added_duration
+                }
+            },
+            _ => self.added_duration,
+        }
+    }
+
+    fn update_display(&mut self) {
+        self.displayed_duration = self.get_current_duration();
```
DJMcNab, line 97: Can we just call `self.get_current_duration()` whenever `self.displayed_duration` would be used?

jaredoconnell: I don't think that would work right. The timestamp doesn't change because it's an absolute time. The async repeat call needs to trigger a change for the label to detect.

xilem/examples/stopwatch.rs
jaredoconnell: That works. Thanks! Rust's typing is a little unusual compared to what I'm used to.

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,261 @@
...
+    ))
+    .direction(Axis::Horizontal)
+    .cross_axis_alignment(CrossAxisAlignment::Center)
+    .main_axis_alignment(MainAxisAlignment::Start)
+    .must_fill_major_axis(true)
+}
+
+fn start_stop_button(data: &mut Stopwatch) -> impl WidgetView<Stopwatch> {
+    if data.active {
+        Either::A(button("Stop", |data: &mut Stopwatch| {
```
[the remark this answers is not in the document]
jaredoconnell: Perhaps, if we were designing this from scratch. But every other app has "stop". Anyway, it's a STOPwatch. 😁

xilem/examples/stopwatch.rs
jaredoconnell: The downside of that is that I need to get the first item to be the current lap, which I add to the beginning after building the first vector. So this wouldn't allow that without redesigning something else.

xilem/examples/stopwatch.rs
jaredoconnell: Addressed by implementing Philipp's changes.

xilem/examples/stopwatch.rs
jaredoconnell: Didn't need to use a sized_box because of that 😁

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,258 @@
...
+    // Add the current lap first so that it shows up on the top.
+    reversed.push(single_lap(
+        current_lap,
+        &current_split_duration,
+        &data.get_current_duration(),
+    ));
+    for item in items.into_iter().rev() {
+        reversed.push(item);
+    }
+    reversed
```
Philipp-M, line 158: I mean that's nit-picking at this point, but I think a reverse may avoid an allocation.

```suggestion
    items.reverse();
    // Add the current lap first so that it shows up on the top.
    let current_lap = single_lap(
        current_lap,
        &current_split_duration,
        &data.get_current_duration(),
    );
    items.insert(0, current_lap);
    items
```

jaredoconnell: Good idea. I implemented a similar change, but I put it into the data structure before reversing, since it's probably more efficient to swap locations rather than shift everything forward one, since I believe Vec is array backed.

Philipp-M: Yeah that's even better.

xilem/examples/stopwatch.rs
```diff
@@ -0,0 +1,255 @@
...
+                        interval.tick().await;
+                        let Ok(()) = proxy.message(()) else {
+                            break;
+                        };
+                    }
+                },
+                |data: &mut Stopwatch, ()| {
+                    data.update_display();
+                },
+            )
```
Philipp-M, line 133: I mean totally unrelated to this PR, just some food for thought, as I'm currently thinking about something like a `request_animation_frame` view in xilem_web.
(or even let xilem_webs event loop run inside something like `request_animation_frame` with batched events/mutations, but I need to check how that impacts input latency).

I think it might be interesting to have a view that is synced with the framerate/vsync/compositor, to e.g. have a smooth increasing counter (or more useful-cases such as a canvas animating/moving stuff).

jaredoconnell: That would definitely be ideal for this example.

PR #493, opened 2024-08-08
xilem_web: Add support for opacity in `svg::Fill` and dashes, opacity in `svg::Stroke`

Philipp-M:
Uses `ViewState` for a little bit of optimization (of allocations mostly).


DJMcNab (approved):

This probably should be used in an example

Philipp-M:
You're right, I've added it to the svgtoy example, and also added `stroke-dashoffset` support (also in in example).

PR #494, opened 2024-08-08
xilem_web: Change `events::OnResize` to be a `ResizeObserver`

Philipp-M:
Since [`on_resize`](https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event) on an `Element` is not really useful as it won't be called, we can reuse this event listener with a [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/ResizeObserver) on itself, to be actually useful.


xilem_web/src/events.rs
```diff
@@ -508,3 +507,109 @@ event_definitions!(
...
+    Callback: Fn(&mut State, web_sys::ResizeObserverEntry) -> OA + 'static,
+    V: View<State, Action, ViewCtx, DynMessage>,
+    V::Element: ElementAsRef<web_sys::Element>,
+{
+    type Element = V::Element;
+
+    type ViewState = OnResizeState<V::ViewState>;
+
+    fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+        ctx.with_id(ViewId::new(0), |ctx| {
```
DJMcNab, line 541: I wonder if it would be more idiomatic to use a different number here.

e.g. something like.
```rust
const ON_RESIZE_VIEW_ID: ViewId = ViewId::new(0x1234_ABCD);
```

Which might be slightly more likely to catch a mistake in a different view.

DJMcNab (approved):

xilem_web/src/events.rs
Philipp-M: Yeah makes sense, needed a `const` change in xilem_core. I applied that for all the event views now.

Philipp-M: And good, that I just tested this quickly before merging... I introduced a small bug (which is now fixed)...

xilem_web/src/events.rs
```diff
@@ -3,11 +3,13 @@
 
 use std::{borrow::Cow, marker::PhantomData};
 use wasm_bindgen::{prelude::Closure, throw_str, JsCast, UnwrapThrowExt};
-use web_sys::AddEventListenerOptions;
+use web_sys::{js_sys, AddEventListenerOptions};
 use xilem_core::{MessageResult, Mut, View, ViewId, ViewMarker, ViewPathTracker};
 
 use crate::{DynMessage, ElementAsRef, OptionalAction, ViewCtx};
 
+const ON_EVENT_VIEW_ID: ViewId = ViewId::new(0x2357_1113);
```
DJMcNab, line 11: Probably should have a doc comment here...

Philipp-M: I've added that in #493 (as I think a PR is overkill for this little change).

PR #495, opened 2024-08-08
xilem_web: Add hydration support, and a new view using it `Templated`

Philipp-M:
This is mostly a first step towards SSR.
But we can use hydration to speed up creation of long (non-virtualized) lists with the `Templated` view.
`Templated` stores a deep copy of the DOM node of the first occurence of its `impl DomView` based on the `TypeId` in the `ViewCtx`,
and reuses it on every further invocation, where it will be hydrated and updated based on the new attributes.
As it uses an `Rc` to achieve this, it also supports memoization, similar to an `Arc<impl View>`.

Hydration is currently feature-gated, as it produces a little bit more binary bloat. Though it's little enough to this being justified as default.


DJMcNab:
It's not at all clear how this is meant to be used.

Philipp-M:
This is mostly a first step towards SSR. But we can use hydration to speed up creation of long (non-virtualized) lists. It's actually pretty simple to use it:
Just wrap any `view` with `as_template(view)` that is to be expected to be used a lot (e.g. in a `Vec`).

I'm not sure if an additional example makes sense.
I'm thinking about adding the js-framework-bench here as example anyway, where it will be used (since that contains one of the primary use-cases, a long list of always the same view with slightly different values), this was also one of the motivations for the view (apart testing hydration), as I've seen that this was one of the optimizations to land in the upper percentiles (mostly the `create x rows` benches) in that bench...

> I would say that `TypeId`s are not stable across platforms

I don't think that's an issue, as long as the `TypeId` doesn't change while running the app? It'll be only used at runtime. I.e. when this view is created first to cache the DOM node with a unique identifier for that type.
It *won't* be used with actual hydration of an SSR rendered app.

xilem_web/src/as_template.rs
```diff
@@ -0,0 +1,132 @@
...
+///
+/// # Examples
+///
+/// ```
+/// use xilem_web::{as_template, elements::html};
+///
+/// # use xilem_web::interfaces::Element;
+/// # fn component() -> impl Element<()> {
+/// as_template(html::p("This will likely created a lot"))
+/// # }
```
Philipp-M, line 125: Would this help understanding one of the use-cases?
It's some form of instancing, the view could also be called as such (`instanced`)?
```suggestion
/// use xilem_web::{as_template, elements::html, DomFragment};
///
/// fn long_list_fragment() -> impl DomFragment<()> {
///     (0..1000)
///         // Performance increase will be larger with a deeper child views
///         .map(|num| as_template(html::li(format!("number: {num}"))))
///         .collect::<Vec<_>>()
/// }
```

xilem_web/src/text.rs
```diff
@@ -15,13 +17,17 @@ macro_rules! impl_string_view {
 
             fn orphan_build(
                 view: &$ty,
-                _ctx: &mut ViewCtx,
+                #[allow(unused_variables)] ctx: &mut ViewCtx,
```
flosse, line 20:
```suggestion
                #[cfg_attr(not(feature = "hydration"), allow(unused_variables))]
                ctx: &mut ViewCtx,
```

xilem_web/src/text.rs
```diff
@@ -71,13 +77,17 @@ macro_rules! impl_to_string_view {
...
```
flosse, line 80:
```suggestion
                #[cfg_attr(not(feature = "hydration"), allow(unused_variables))]
                ctx: &mut ViewCtx,
```

xilem_web/src/context.rs
```diff
@@ -29,6 +29,13 @@ pub struct ViewCtx {
     id_path: Vec<ViewId>,
     app_ref: Option<Box<dyn AppRunner>>,
     pub(crate) fragment: Rc<web_sys::DocumentFragment>,
+    #[cfg(feature = "hydration")]
+    hydration_node_stack: Vec<web_sys::Node>,
+    #[cfg(feature = "hydration")]
+    is_hydrating: bool,
+    #[cfg(feature = "hydration")]
+    pub(crate) templates:
+        crate::vecmap::VecMap<std::any::TypeId, (web_sys::Node, Rc<dyn std::any::Any>)>,
```
flosse, line 38: Maybe
```suggestion
        VecMap<TypeId, (web_sys::Node, Rc<dyn Any>)>,
```
and at the top

```rust
#[cfg(feature = "hydration")]
use crate::vecmap::VecMap;
#[cfg(feature = "hydration")]
std::any::{Any, TypeId};
```

xilem_web/src/as_template.rs
```diff
@@ -0,0 +1,132 @@
...
+/// ```
+pub fn as_template<State, Action, E>(view: impl Into<Rc<E>>) -> AsTemplate<E>
```
flosse, line 127: Isn't this more a `into_template`?

xilem_web/src/as_template.rs
```diff
@@ -0,0 +1,132 @@
...
+        if matches!(message_result, MessageResult::RequestRebuild) {
+            view_state.dirty = true;
+        }
+        message_result
+    }
+}
+
+/// This view creates an internally cached deep-clone of the underlying DOM node.
+/// When the inner view is created again, this will be done more efficiently.
+/// It's recommended to use this as wrapper, when it's expected that the inner `view` will be created a lot, for example in a long list
```
flosse, line 113: Maye add a sentence when it's not recommended to use?

flosse (approved):
In general, LGTM :+1:

xilem_web/src/as_template.rs
Philipp-M: @DJMcNab @flosse what are your thoughts on this (see also my other new comment)? (also the name `instanced`?)

xilem_web/src/as_template.rs
Philipp-M: Hmm not sure, it's not *only* being used as template, the initial instance of this is cached, any follow-up is deep-cloned and rebuild by the new values.

I mean the name could be either `instanced`, `as_template`, `into_template` or something else (I had `templated` as well, hmm finding good names for views is hard).

(Thanks for the review btw. will address the suggestions soon)

xilem_web/src/as_template.rs
flosse: The example in the doc comments definitely helps :+1: 

I'm not sure about the name :thinking:  
Both `as_tamplate` and `instanced` are fine, but maybe it could be a bit more descriptive?
I wouldn't suggest `cached_deep_clone`, but something more along those words?

xilem_web/src/as_template.rs
flosse: I think `templated` is actually even better than `as_template` and `into_template`.

> hmm finding good names for views is hard.

Yes, one of the most difficult things when it comes to public APIs, which should ideally not change and be intuitively understandable for everyone.
Ideally, all views should also be consistent in their naming.

Philipp-M: I've renamed it now `templated` seems to be a good compromise.

Philipp-M:
Thanks for the review. Merging.

xilem_web/src/as_template.rs
flosse: Nice! :+1:

PR #496, opened 2024-08-09
Mason Only Add While Active

jaredoconnell:
It's jarring for the example to automatically add to the button count, so this makes it so it only increments while active is true.

### DIFF
--- a/xilem/examples/mason.rs
+++ b/xilem/examples/mason.rs
@@ -92,6 +92,8 @@ fn app_logic(data: &mut AppData) -> impl WidgetView<AppData> {
             button("Reset", |data: &mut AppData| data.count = 0),
             flex((fizz_buzz_flex_sequence, flex_sequence)).direction(axis),
         )),
+         // The following `async_repeat` view only exists whilst the example is in the "active" state, so
+         // the updates it performs will only be running whilst we are in that state.
         data.active.then(|| {
             async_repeat(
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
I'd probably prefer if we made the whole `async_repeat` block be toggled by this, to demonstrate the stopping and starting futures works.

jaredoconnell:
> I'd probably prefer if we made the whole `async_repeat` block be toggled by this, to demonstrate the stopping and starting futures works.

Thanks for the help. I've implemented this.

xilem/examples/mason.rs
```diff
@@ -92,18 +92,22 @@ fn app_logic(data: &mut AppData) -> impl WidgetView<AppData> {
...
-                loop {
-                    interval.tick().await;
-                    let Ok(()) = proxy.message(()) else {
-                        break;
-                    };
-                }
-            },
-            |data: &mut AppData, ()| data.count += 1,
-        ),
+        data.active.then(|| {
```
DJMcNab, line 95: We should have a quick comment here.
Something like:
```suggestion
         // The following `async_repeat` view only exists whilst the example is in the "active" state, so
         // the updates it performs will only be running whilst we are in that state.
        data.active.then(|| {
```

DJMcNab (approved):
Looks good! This should make the example much less jarring.

PR #497, opened 2024-08-09
Restore animation support - add `spinner` view

DJMcNab:
This is the smallest change possible to make animations work in Masonry.

Essentially, we treat every redraw as a potential animation frame, so that animations do work.

I've also added the `spinner` view in Xilem to test this out. The `state_machine` examples uses this.

This is work done for the variable fonts demo.

### DIFF
--- a/xilem/src/view/spinner.rs
+++ b/xilem/src/view/spinner.rs
@@ -1,0 +1,59 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use masonry::{widget, Color};
+use xilem_core::{Mut, ViewMarker};
+
+use crate::{MessageResult, Pod, View, ViewCtx, ViewId};
+
+pub fn spinner() -> Spinner {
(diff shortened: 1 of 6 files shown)


xilem/src/view/spinner.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    fn rebuild<'el>(
+        &self,
+        prev: &Self,
+        (): &mut Self::ViewState,
+        _: &mut ViewCtx,
+        mut element: Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element> {
+        if prev.color != self.color {
+            // TODO: Don't duplicate the default colour from Masonry
```
PoignardAzur, line 41: Can you expand on that TODO?

PoignardAzur (approved):
LGTM.

Honestly, I'm surprised it's that simple.

jaredoconnell (approved):
Looks like it will need some polish in the future, but I tested this and it works.

PR #499, opened 2024-08-09
Fix doc links by either correcting the item name, or making these public

Philipp-M:
I don't think anything speaks against making the items public in masonry?


DJMcNab (approved):
The Xilem changes are definitely uncontroversial.

I don't think that `WorkerCtx` is really a useful type any more, but I also don't think it has any public API, so there's no harm in exposing it to remove it later.

PR #502, opened 2024-08-09
Cleanup of unnecessary tracing and "external" `.flex()` of `num_row` in the `calc` example

Philipp-M:


jaredoconnell (approved):
Tested. Works.
I don't have any background on the benefit of the info trace, so I don't have any comments on that removal.

Philipp-M:
That removal came because of [this](https://github.com/linebender/xilem/pull/428#discussion_r1711769519) comment. Thanks for the review!

PR #503, opened 2024-08-10
masonry: Remove ExtEvent, Promise, WorkerCtx, WorkerFn, SpawnWorker.

waywardmonkeys:
These either aren't used, aren't implemented, or are just stubs. A new implementation may or may not take the same approach given changes within the ecosystem, and if we want these back, they're in the history.


DJMcNab (approved):
I have long found these to be cluttering things up, and agree that they should go.

We could wait another day or so to see if we've missed a compelling to reason for these to stay, but I'd also be happy to just land this now..

PoignardAzur (approved):

PR #504, opened 2024-08-10
xilem_web: Make MessageThunk public

flosse:


flosse:
In order to send messages based on a JS callback, I need access to the `MessageThunk` to do something like this:

```rust
let thunk = ctx.message_thunk();
js_obj.on_event(move |ev|{
  let value = process_event(ev);
  thunk.push_message(JsMessage::Foo());
});
```

Philipp-M:
Yeah I guess we can export that, but expect things to break in the future.

flosse:
> Yeah I guess we can export that, but expect things to break in the future.

sure. But for now it would help in some situations :relaxed:

Philipp-M (approved):
Feel free to merge

PR #505, opened 2024-08-11
xilem_web: Fix issue with downcasting to a wrong type transitively from `AnyView`

Philipp-M:
This should fix the issue described in #440 by adding the type of the `Children` `ViewSequence` to the type of element, which results in unique types, so `AnyView` should not cause issues anymore. Fortunately there's no regression with the compilation times here, so an `AnyViewSequence` doesn't seem to be necessary in that case.


Philipp-M:
Thanks, yeah it was already confirmed that this fixes the issue

PR #506, opened 2024-08-11
xilem_web: Add event_target_value helper function

flosse:


flosse:
I took the opportunity to create a new `util` module in which all such helpers can end up.

xilem_web/src/util.rs
```diff
@@ -0,0 +1,40 @@
...
+///
+/// Returns `None` if the event isn't a valid input event or conversions fail.
+pub fn event_target_value<T>(event: &T) -> Option<String>
+where
+    T: JsCast,
+{
+    event
+        .unchecked_ref::<web_sys::Event>()
+        .target()?
+        .dyn_into::<web_sys::HtmlInputElement>()
```
Philipp-M, line 36: I wonder, if that function should be called `event_target_value` when it's only valid for `HtmlInputElement` as target? AFAIK, there's quite a few other elements (e.g. `HtmlOptionElement`) having a `.value()`. I also don't think it needs to be generic over `T` as it's only used for `Deref<Target=web_sys::Event>`, so `&web_sys::Event` seems to be sufficient.

flosse: > I wonder, if that function should be called `event_target_value` when it's only valid for `HtmlInputElement` as target?

You're right, i just took it over from leptos out of habit :see_no_evil: 

> AFAIK, there's quite a few other elements (e.g. `HtmlOptionElement`) having a `.value()`. 

:thinking: We can either create separate functions (as it is right now) or we can try out all possible elements within a generic function. Or is there another possibility?

> I also don't think it needs to be generic over `T` as it's only used for `Deref<Target=web_sys::Event>`, so `&web_sys::Event` seems to be sufficient.

:+1:

Philipp-M: I think we can keep it separate functions for now, less magic (which I think is good in this situation). In the future, e.g. with #440 we could filter events, in this case based on value.

Philipp-M (approved):
Thanks for the cleanup. I wonder though, whether that module should be named generically `util`. Maybe be more explicit, like `dom_util` or `dom_helpers` or something like that?

PR #507, opened 2024-08-12
Label with animated variable font parameters

DJMcNab:
This adds a new `VariableLabel` widget, which animates its weight in a linear fashion. Also adds support for this in Xilem, and a new `variable_clock` example. This example also runs on Android.

[Variable Fonts Lorem ipsum](https://github.com/user-attachments/assets/7a426029-0c54-40c5-88f8-1cd1160edf84)

Outstanding issues:

- [ ] Hacks in support for "Roboto Flex", by always loading it from the local file
- [ ] It's not clear where we should get Roboto Flex from, as it is too big to store in this repository
- [ ] The variable font animation is quite "niche"
- [ ] The only supported variable axis is `wgth`


Philipp-M:
I think you forgot to add a font?
(`resources/fonts/roboto_flex/RobotoFlex-Subset.ttf` to be precise)

DJMcNab:
Apologies that I didn't make this clear - I was intending to get review/feedback on all the parts but the example, as I am planning on changing the example today. I now recognise that it's hard to review without a running example...

I hadn't yet added the font, as it's not clear what subset we want to use, and changing the subset eats up "valuable" repository history space.
But adding a font which is later removed in the same PR should be fine, since we use squashing.

waywardmonkeys:
Probably could have an assets repo that is a shallow clone as a sub module with it … apart from everyone hating submodules.

DJMcNab:
The subsetted font is only on the order of 100-200KB, depending on the subset chosen. Which is quite a lot, but is still manageable.

Philipp-M:
I have not yet looked into the code, but I'm having some issues with jumpiness, when the font is not numbers (or colons it seems?)


https://github.com/user-attachments/assets/edaa5234-3d5b-4b59-826f-56e8d8b5c6a3

DJMcNab:
Apologies, I have not made the current status clear enough.

> It just looked like it jumped to the new value at the end of what was supposed to be the animation period.

To be clear, is this applying to the numbers as well? That would be quite a serious bug if so, and I don't have any idea how that could be happening.

It is expected that the "letter" text doesn't have smooth weight (except on Android), because the font subset which was created doesn't support letters, to avoid committing too large of a file. This is documented both in the font's README file and in the example itself (above the definition of `LOREM`).

As I have also said, I am planning on changing the example's content, and the parts which are ready for review are everything but the example code itself.

masonry/src/render_root.rs
```diff
@@ -188,8 +188,9 @@ impl RenderRoot {
                 // See https://github.com/linebender/druid/issues/85 for discussion.
                 let last = self.last_anim.take();
                 let elapsed_ns = last.map(|t| now.duration_since(t).as_nanos()).unwrap_or(0) as u64;
-
-                if self.root_state().request_anim {
+                let root_state = self.root_state();
+                if root_state.request_anim {
+                    root_state.request_anim = false;
```
Philipp-M, line 193: Is this a drive-by fix? Was this a bug previously?

masonry/src/widget/variable_label.rs
```diff
@@ -0,0 +1,393 @@
...
+pub(super) const LABEL_X_PADDING: f64 = 2.0;
+
+/// An `f32` value which can move towards a target value at a linear rate over time.
+#[derive(Clone, Debug)]
+pub struct AnimatedF32 {
+    /// The value which self will eventually reach.
+    target: f32,
+    /// The current value
+    value: f32,
+    // TODO: Provide different easing functions, instead of just linear
```
Philipp-M, line 35: Maybe a future case for interpoli?

I think this could also be achieved with the `Animatable` trait I did in trui (to be more generally customizable in xilem and I think also in masonry, but I haven't looked into that yet... (also because I have not yet gotten real feedback there)).

masonry/src/widget/variable_label.rs
```diff
@@ -0,0 +1,393 @@
...
+            value,
+            rate_per_millisecond: 0.,
+        }
+    }
+
+    /// Move this value to the `target` over `target` milliseconds.
+    pub fn move_to(&mut self, target: f32, over_millis: f32) {
+        assert!(target.is_finite());
+        self.target = target;
+        self.rate_per_millisecond = (self.target - self.value) / over_millis;
```
Philipp-M, line 55: I think we should either add a division by zero guard here (epsilon?), or warn the user somewhere (with a doc-comment). (I lean towards defusing this)

masonry/src/widget/widget_pod.rs
```diff
@@ -395,7 +395,10 @@ impl<W: Widget> WidgetPod<W> {
 
                 true
             }
-            LifeCycle::AnimFrame(_) => true,
+            LifeCycle::AnimFrame(_) => {
+                state.request_anim = false;
```
Philipp-M, line 399: Same here, was animation never cleared before (i.e. always requested)?

xilem/Cargo.toml
```diff
@@ -58,7 +68,9 @@ accesskit.workspace = true
 tokio = { version = "1.39.1", features = ["rt", "rt-multi-thread", "time"] }
 
 [dev-dependencies]
-accesskit_winit.workspace = true
+# accesskit_winit.workspace = true
```
Philipp-M, line 71: Not necessary because this is already in masonry? (Could be removed in that case I guess)

xilem/examples/variable_clock.rs
```diff
@@ -0,0 +1,221 @@
...
+struct TimeZone {
+    /// An approximate region which this offset applies to.
+    region: &'static str,
+    /// The offset from UTC
+    offset: time::UtcOffset,
+}
+
+impl TimeZone {
+    fn view(&self, data: &mut Clocks) -> impl WidgetView<Clocks> {
+        let date_time_in_self = data.now_utc.to_offset(self.offset);
```
Philipp-M, line 46: Should this be updated every second?

xilem/examples/variable_clock.rs
```diff
@@ -0,0 +1,221 @@
...
+        .height(72.)
+    }
+}
+
+fn app_logic(data: &mut Clocks) -> impl WidgetView<Clocks> {
+    flex((
+        // HACK: We add a spacer at the top for Android. See https://github.com/rust-windowing/winit/issues/2308
+        FlexSpacer::Fixed(40.),
+        controls(),
+        TIMEZONES.iter().map(|it| it.view(data)).collect::<Vec<_>>(),
```
Philipp-M, line 94: I mean, not super important, but would it make sense to do this on 2 columns until we have a scroll view (to reduce the chance of this being cut-off)

xilem/examples/variable_clock.rs
```diff
@@ -0,0 +1,221 @@
...
+            // This does nothing because of `time`'s soundness checks...
+            data.local_offset = UtcOffset::current_local_offset().or(data.local_offset);
+        }),
+    ))
+    .direction(Axis::Horizontal)
+}
+
+/// A subset of [Roboto Flex](https://fonts.google.com/specimen/Roboto+Flex), used under the OFL.
+/// This is a variable font, and so can be.
+/// The version in the repository supports the numbers 0-9 and `:`, as it is included for this example,
```
Philipp-M, line 122: Nice, I hope this avoids further misunderstandings (and good that there's no unsupported letter directly in the example, as I think that just produces avoidable discussions)

xilem/src/view/variable_label.rs
```diff
@@ -0,0 +1,176 @@
...
+        style::{FontFamily, FontStack, GenericFamily},
+    },
+    text::TextBrush,
+    widget, ArcStr,
+};
+use xilem_core::{Mut, ViewMarker};
+
+use crate::{Color, MessageResult, Pod, TextAlignment, View, ViewCtx, ViewId};
+
+pub fn variable_label(label: impl Into<ArcStr>) -> VariableLabel {
```
Philipp-M, line 16: I wonder (not necessarily relevant for this PR), if this should be a "transformer" modifier on the `Label` (i.e. moves all relevant values of `label()` to this type). (And not a separate view constructor). 

so that we could do something like this instead:

```rust
label("label").target_weight(..)
```

(Not sure if more complex typing to achieve this with traits etc. makes sense here)

Philipp-M (approved):
(slightly optimistic approval). Looks good so far. A few comments, but nothing fancy.

masonry/src/render_root.rs
DJMcNab: It was a bug previously, although it's not really a change in observable behaviour, just a missed optimisation. I don't think it's very important to land this more expediently than this change.

masonry/src/widget/variable_label.rs
DJMcNab: Yeah, this would be the type of thing I expect Interpoli to provide.
I haven't had time to dig into that Animatable work, unfortunately.

masonry/src/widget/variable_label.rs
DJMcNab: Is that not achieved by the very next statement? I do agree that this should be documented, good point

masonry/src/render_root.rs
Philipp-M: (Was just a question, agree)

masonry/src/widget/variable_label.rs
Philipp-M: I could produce a crash by using `0.0` in the xilem example

masonry/src/widget/widget_pod.rs
DJMcNab: Yeah, exactly. It would just get stuck in an infinite loop.

xilem/Cargo.toml
DJMcNab: It's not clear to me why this dependency was required. I presume it was part of the #325/ #417 stuff (seriously, GitHub has deleted that?)

xilem/examples/variable_clock.rs
DJMcNab: There are two sides to this:
1) This was my original plan
2) I think this version looks better on a "normal resolution" Android device, which is where this example is primarily targeted.

masonry/src/widget/variable_label.rs
DJMcNab: Oh I see. Yeah, if `over_millis` is zero, it makes sense to just set the value directly. I need to document that possibility properly.

xilem/src/view/variable_label.rs
Philipp-M: Oh I might have missed that, nice.

xilem/Cargo.toml
Philipp-M: What happened there? Bug in Github, outage? I don't think they usually delete PRs?

xilem/examples/variable_clock.rs
DJMcNab: Indeed. This is planned, you just reviewed before I got around to it :)

xilem/Cargo.toml
DJMcNab: The PR author's account got deleted.

Philipp-M: Do you know what the PR was about?

Philipp-M: (answering my own question, it's the "cfagot" stuff, argh they should just recover his account...)

It was "Allow external event loop to drive masonry and xilem (#417)"

xilem/examples/variable_clock.rs
```diff
@@ -0,0 +1,267 @@
...
+        local_time(data),
+        controls(),
+        // TODO: When we get responsive layouts, move this into a two-column view.
+        TIMEZONES.iter().map(|it| it.view(data)).collect::<Vec<_>>(),
+    ));
+    fork(
+        view,
+        async_repeat(
+            |proxy| async move {
+                // TODO: Synchronise with the actual "second" interval. This is expected to show the wrong second
```
Philipp-M, line 54: Yeah I think this is certainly good enough for an example for now, we're not building an atomic clock here^^

Philipp-M (approved):
Seems good to land for me. Thanks. looks like a nice addition showing off the font-abilities of our text-stack.

PoignardAzur:
A git bisect tells me this added a regression in the masonry calc example (changing the hover state no longer updates the widget's appearance).

One more reason to test examples, I guess.

PoignardAzur:
After looking further, it seems that the example only worked because Masonry kept pumping animation frames. This PR makes the animation frames only happen when necessary, which surfaced the bug. See #526.

PR #508, opened 2024-08-12
Fix view overflow in the calc example on mobile

rustui:
On mobile platforms, setting the logical width of the window to 400 can cause the view to overflow on phones with smaller screen widths (such as the iPhone SE 2nd generation).
<details>
  <summary>screenshot of the view overflow</summary>
  screenshot of the view overflow
</details>

It is more intuitive not to manually set the window size. Additionally, winit does not support set_title and set_resizable on iOS and Android (calling them does not have any effect).

### DIFF
--- a/xilem/examples/calc.rs
+++ b/xilem/examples/calc.rs
@@ -300,6 +300,14 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
         .with_resizable(true)
         .with_min_inner_size(min_window_size)
         .with_inner_size(window_size);
+    // On iOS, winit has unsensible handling of `inner_size`
+    // See https://github.com/rust-windowing/winit/issues/2308 for more details
+    #[cfg(target_os = "ios")]
+    let window_attributes = {
+        let mut window_attributes = window_attributes; // to avoid `unused_mut`
+        window_attributes.inner_size = None;
+        window_attributes
+    };
     app.run_windowed_in(event_loop, window_attributes)?;
     Ok(())
 }


jaredoconnell (commented):
This makes sense. It looks good, but I'll let the maintainers of the repository have an opportunity to feedback instead of approving.

xilem/examples/calc.rs
```diff
@@ -293,13 +292,19 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
...
-        .with_resizable(true)
-        .with_min_inner_size(min_window_size)
-        .with_inner_size(window_size);
+    #[cfg(not(any(target_os = "ios", target_os = "android")))]
+    let window_attributes = {
+        use winit::dpi::LogicalSize;
+        let min_window_size = LogicalSize::new(200., 200.);
+        let window_size = LogicalSize::new(400., 500.);
+        Window::default_attributes()
+            .with_title("Calculator")
```
DJMcNab, line 301: We definitely should still call `with_title` on mobile.

xilem/examples/calc.rs
```diff
@@ -293,13 +292,19 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
...
+            .with_resizable(true)
+            .with_min_inner_size(min_window_size)
```
DJMcNab, line 303: I don't see why we should remove this on mobile either.

xilem/examples/calc.rs
```diff
@@ -293,13 +292,19 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
...
 
     let app = Xilem::new(data, app_logic);
-    let min_window_size = LogicalSize::new(200., 200.);
-    let window_size = LogicalSize::new(400., 500.);
-    let window_attributes = Window::default_attributes()
-        .with_title("Calculator")
-        .with_resizable(true)
-        .with_min_inner_size(min_window_size)
-        .with_inner_size(window_size);
+    #[cfg(not(any(target_os = "ios", target_os = "android")))]
```
DJMcNab, line 295: This example works fine on my Android device - I believe that this is instead a winit bug on iOS

DJMcNab (commented):
Thanks for your interest. I think that a much more targeted fix for this issue should be used, as it only actually exists on ios.

Overall, I'd prefer this to just be of the form:
```rust
// On iOS, winit has unsensible handling of `inner_size`
// [Link to the relevant winit issue, which you will create if necessary]
#[cfg(target_os = "ios")]
let mut window_attributes = window_attributes;
#[cfg(target_os = "ios")]
window_attributes.inner_size = None;
```
The calc example works completely normally for me on Android

This code could also then possibly live inside `run_windowed`?

DJMcNab (approved):
Thanks! I'm not certain that the linked issue necessarily covers this exact case, but if this helps people understand an issue and fix it for their apps, that would be good.

We don't make any support guarantees for ios, but we're happy to keep it working reasonably well.

rustui:
> ```rust
> // On iOS, winit has unsensible handling of `inner_size`
> // [Link to the relevant winit issue, which you will create if necessary]
> #[cfg(target_os = "ios")]
> let mut window_attributes = window_attributes; // to avoid `unused_mut`
> #[cfg(target_os = "ios")]
> window_attributes.inner_size = None;
> ```

The error message “attributes on expressions are experimental” appeared, so I had to place it inside a block.

> This code could also then possibly live inside `run_windowed`?

`run_windowed` only sets min_inner_size without specifying inner_size. Based on testing, on iOS, even if min_inner_size is set (regardless of whether it’s set larger or smaller), as long as inner_size is not set, there won’t be any overflow. So, might it not be necessary to modify `run_windowed`?

DJMcNab:
But what I mean is that if it's meaningless to ever set inner_size on iOS, then can we can unset it centrally - it might even go into Masonry itself

But I wasn't going to block on that, since adding to just one example is ok as well

rustui:
If it is unset within Masonry or `run_windowed_in`, the WindowAttributes created by the user might also be modified, could this lead to some unexpected surprises for the user? In `run_windowed`, the WindowAttributes are created locally, and inner_size is not involved.

DJMcNab:
It seems to me that the behaviour of ignoring the size of the screen and instead strictly following the provided initial size would be more surprising than this behaviour.

But as I say, I don't know what iOS users expect; maybe the screenshot you have shown is what users would want in most cases, and it's only the calc example which should change?
I don't need any further action to be taken here, but I'm suggesting an alternative design which I suspect would be better.

PR #509, opened 2024-08-13
masonry: reuse redundant `ArenaRef/Mut` creation code

Philipp-M:
@PoignardAzur this is what I meant with the redundancy.


waywardmonkeys (commented):
Just chiming in to say that this all reads so much better to me.

PoignardAzur (approved):
Looks good!

PR #510, opened 2024-08-13
Implement mutate pass from pass spec RFC

PoignardAzur:
This is part of the "Pass Specification" RFC:
https://github.com/linebender/rfcs/pull/7

Rename WidgetCtx to MutateCtx.
Add a mutate pass.
Add a `mutate_later` context method to trigger that pass.
Refactor `edit_root_widget` to use a version of that pass.
Add a separate constructor for the synthetic WidgetState created in RenderRoot.

### DIFF
--- a/masonry/examples/calc_masonry.rs
+++ b/masonry/examples/calc_masonry.rs
@@ -146,22 +146,22 @@ impl Widget for CalcButton {
             PointerEvent::PointerDown(_, _) => {
                 if !ctx.is_disabled() {
                     let color = self.active_color;
+                    // See on_status_change for why we use `mutate_later` here.
                     ctx.mutate_later(&mut self.inner, move |mut inner| {
                         inner.set_background(color);
                     });
                     ctx.set_active(true);
-                    ctx.request_paint();
                     trace!("CalcButton {:?} pressed", ctx.widget_id());
                 }
             }
             PointerEvent::PointerUp(_, _) => {
                 if ctx.is_active() && !ctx.is_disabled() {
                     let color = self.base_color;
+                    // See on_status_change for why we use `mutate_later` here.
                     ctx.mutate_later(&mut self.inner, move |mut inner| {
                         inner.set_background(color);
                     });
                     ctx.submit_action(Action::Other(Box::new(self.action)));
-                    ctx.request_paint();
                     trace!("CalcButton {:?} released", ctx.widget_id());
                 }
                 ctx.set_active(false);
@@ -186,18 +186,21 @@ impl Widget for CalcButton {
     }
 
     fn on_status_change(&mut self, ctx: &mut LifeCycleCtx, event: &StatusChange) {
+        // Masonry doesn't let us change a widget's attributes directly.
+        // We use `mutate_later` to get a mutable reference to the inner widget
+        // and change its border color. This is a simple way to implement a
+        // "hovered" visual effect, but it's somewhat non-idiomatic compared to
+        // implementing the effect inside the "paint" method.
         match event {
             StatusChange::HotChanged(true) => {
                 ctx.mutate_later(&mut self.inner, move |mut inner| {
                     inner.set_border(Color::WHITE, 3.0);
                 });
-                ctx.request_paint();
             }
             StatusChange::HotChanged(false) => {
                 ctx.mutate_later(&mut self.inner, move |mut inner| {
                     inner.set_border(Color::TRANSPARENT, 3.0);
                 });
-                ctx.request_paint();
             }
             _ => (),
         }
--- a/masonry/examples/calc_masonry.rs
+++ b/masonry/examples/calc_masonry.rs
@@ -145,8 +145,10 @@ impl Widget for CalcButton {
         match event {
             PointerEvent::PointerDown(_, _) => {
                 if !ctx.is_disabled() {
-                    ctx.get_mut(&mut self.inner)
-                        .set_background(self.active_color);
+                    let color = self.active_color;
+                    ctx.mutate_later(&mut self.inner, move |mut inner| {
+                        inner.set_background(color);
+                    });
                     ctx.set_active(true);
                     ctx.request_paint();
                     trace!("CalcButton {:?} pressed", ctx.widget_id());
@@ -154,11 +156,14 @@ impl Widget for CalcButton {
             }
             PointerEvent::PointerUp(_, _) => {
                 if ctx.is_active() && !ctx.is_disabled() {
+                    let color = self.base_color;
+                    ctx.mutate_later(&mut self.inner, move |mut inner| {
+                        inner.set_background(color);
+                    });
                     ctx.submit_action(Action::Other(Box::new(self.action)));
                     ctx.request_paint();
                     trace!("CalcButton {:?} released", ctx.widget_id());
                 }
-                ctx.get_mut(&mut self.inner).set_background(self.base_color);
                 ctx.set_active(false);
             }
             _ => (),
@@ -183,12 +188,15 @@ impl Widget for CalcButton {
     fn on_status_change(&mut self, ctx: &mut LifeCycleCtx, event: &StatusChange) {
         match event {
             StatusChange::HotChanged(true) => {
-                ctx.get_mut(&mut self.inner).set_border(Color::WHITE, 3.0);
+                ctx.mutate_later(&mut self.inner, move |mut inner| {
+                    inner.set_border(Color::WHITE, 3.0);
+                });
                 ctx.request_paint();
             }
             StatusChange::HotChanged(false) => {
-                ctx.get_mut(&mut self.inner)
-                    .set_border(Color::TRANSPARENT, 3.0);
+                ctx.mutate_later(&mut self.inner, move |mut inner| {
+                    inner.set_border(Color::TRANSPARENT, 3.0);
+                });
                 ctx.request_paint();
             }
             _ => (),
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -495,6 +495,7 @@ impl_context_method!(
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
     {
+        // TODO - Remove from MutateCtx?
         /// Queue a callback that will be called with a [`WidgetMut`] for this widget.
         ///
         /// The callbacks will be run in the order they were submitted during the mutate pass.
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -374,6 +374,16 @@ impl<'a> MutateCtx<'a> {
             is_reborrow: false,
         }
     }
+
+    pub(crate) fn reborrow_mut(&mut self) -> MutateCtx<'_> {
+        MutateCtx {
+            global_state: self.global_state,
+            parent_widget_state: self.parent_widget_state,
+            widget_state: self.widget_state,
+            widget_state_children: self.widget_state_children.reborrow_mut(),
+            widget_children: self.widget_children.reborrow_mut(),
+        }
+    }
 }
 
 // --- MARK: UPDATE FLAGS ---
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -34,12 +34,12 @@ macro_rules! impl_context_method {
 /// A context provided inside of [`WidgetMut`].
 ///
 /// When you declare a mutable reference type for your widget, methods of this type
-/// will have access to a `WidgetCtx`. If that method mutates the widget in a way that
+/// will have access to a `MutateCtx`. If that method mutates the widget in a way that
 /// requires a later pass (for instance, if your widget has a `set_color` method),
 /// you will need to signal that change in the pass (eg `request_paint`).
 ///
 // TODO add tutorial - See https://github.com/linebender/xilem/issues/376
-pub struct WidgetCtx<'a> {
+pub struct MutateCtx<'a> {
     pub(crate) global_state: &'a mut RenderRootState,
     pub(crate) parent_widget_state: &'a mut WidgetState,
     pub(crate) widget_state: &'a mut WidgetState,
@@ -110,7 +110,7 @@ pub struct AccessCtx<'a> {
 // --- MARK: GETTERS ---
 // Methods for all context types
 impl_context_method!(
-    WidgetCtx<'_>,
+    MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
@@ -158,7 +158,7 @@ impl_context_method!(
 
 // Methods for all mutable context types
 impl_context_method!(
-    WidgetCtx<'_>,
+    MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
@@ -184,7 +184,7 @@ impl_context_method!(
 // Methods on all context types except LayoutCtx
 // These methods access layout info calculated during the layout pass.
 impl_context_method!(
-    WidgetCtx<'_>,
+    MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     PaintCtx<'_>,
@@ -226,7 +226,7 @@ impl_context_method!(
 // Methods on all context types except LayoutCtx
 // Access status information (hot/active/disabled/etc).
 impl_context_method!(
-    WidgetCtx<'_>,
+    MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     PaintCtx<'_>,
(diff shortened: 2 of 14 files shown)


PoignardAzur:
Note: this was split off from #498 and rebased after #488 was merged.

PoignardAzur:
I think this is ready to merge.

masonry/src/contexts.rs
```diff
@@ -540,11 +488,34 @@ impl_context_method!(WidgetCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
 // --- MARK: OTHER METHODS ---
 // Methods on all context types except PaintCtx and AccessCtx
 impl_context_method!(
-    WidgetCtx<'_>,
+    MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
     {
+        pub fn mutate_self_later(
```
waywardmonkeys, line 496: Would be good to have a doc comment for this method and the one below. Probably a pretty short one.

masonry/src/render_root.rs
```diff
@@ -295,56 +306,73 @@ impl RenderRoot {
...
+
+        root_state.merge_up(self.widget_arena.get_state_mut(self.root.id()).item);
+        self.post_event_processing(&mut root_state);
 
+        res.unwrap()
+    }
+
+    /// Get a [`WidgetMut`] to a specific widget.
+    ///
+    /// Because of how `WidgetMut` works, it can only be passed to a user-provided callback.
```
waywardmonkeys, line 348: If mutations are required to several widgets, is there an inefficiency here? (Question, not looking for a change, I think.)

PoignardAzur: If you mutate several widgets with the same call to `edit_widget`, no. If you call `edit_widget` hundreds of times per frame, yes, a bit.

It's an escape hatch first and foremost, it's not supposed to substitute for the pass system.

masonry/src/passes/mutate.rs
```diff
@@ -0,0 +1,56 @@
...
+    let (widget_mut, state_mut) = root.widget_arena.get_pair_mut(id);
+
+    let _span = info_span!("mutate_widget").entered();
+    let root_widget = WidgetMut {
+        ctx: MutateCtx {
+            global_state: &mut root.state,
+            // FIXME - This works in practice, because the loop below will merge the
+            // state up to the root. But it's semantically awkward.
+            // parent_widget_state should probably be an Option.
+            parent_widget_state: &mut dummy_state,
```
Philipp-M, line 26: Yeah good question, I think it depends on whether there should be "dangling" widgets, in which case an `Option` certainly makes sense. But when not, I think this is fine, would reduce boilerplate everywhere else.

masonry/src/render_root.rs
Philipp-M: So when I'm getting it right, later only `edit_root_widget` and `mutate_later` should be used?

Is this method even necessary (outside of tests)? I only see usage of it in the tests, so should this be `pub(crate)` instead?

masonry/src/widget/widget_state.rs
```diff
@@ -148,20 +148,20 @@ impl WidgetState {
...
             baseline_offset: 0.0,
             is_hot: false,
-            needs_layout: false,
-            needs_paint: false,
-            needs_accessibility_update: false,
-            needs_window_origin: false,
+            needs_layout: true,
+            needs_paint: true,
+            needs_accessibility_update: true,
+            needs_window_origin: true,
```
Philipp-M, line 154: I wonder why this was not necessary before, or was this a bug, that was just never surfaced? (xilem "classic" had this too btw. with the Changeflags)

masonry/src/widget/widget_state.rs
```diff
@@ -148,20 +148,20 @@ impl WidgetState {
...
             has_focus: false,
             request_anim: false,
```
Philipp-M, line 156: should this also be:
```suggestion
            request_anim: true,
```
?
I see that this is set in the `Spinner` widget at `WidgetAdded`. Or should this explicitly be set somewhere else for widgets needing animation directly on creation (like the `Spinner`)?

Philipp-M (approved):
I wonder if we slowly want to remove the `get_raw_mut` stuff? (I think that is planned isn't it?)

(I agree with Bruce, I think the methods could have some doc-comments ideally when it makes sense with doc-tests, but I don't want to block on that, when you add these later, it would just also help with reviewing.
Thanks for keeping this PR more focused, that makes reviewing easier and more effective)

Thanks looks good apart from a few more open questions.

masonry/src/passes/mutate.rs
DJMcNab, line 26: I think the move to `Option` should happen in this PR. AIUI, the only thing this impacts is the `Drop` implementation, right? So it should be a relatively undisruptive change.

masonry/src/passes/mutate.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub(crate) fn mutate_widget(
+    root: &mut RenderRoot,
+    id: WidgetId,
+    mutate_fn: impl FnOnce(WidgetMut<'_, Box<dyn Widget>>),
+) {
+    let mut dummy_state = WidgetState::root(root.root.id(), root.get_kurbo_size());
+    let (widget_mut, state_mut) = root.widget_arena.get_pair_mut(id);
+
+    let _span = info_span!("mutate_widget").entered();
```
DJMcNab, line 19: Would it be worth this being
```suggestion
    let _span = info_span!("mutate_widget", name = state_mut.short_name).entered();
```
(or whatever the actually correct name is)

masonry/src/passes/mutate.rs
```diff
@@ -0,0 +1,56 @@
...
+            widget_state: state_mut.item,
+            widget_state_children: state_mut.children,
+            widget_children: widget_mut.children,
+        },
+        widget: widget_mut.item,
+        is_reborrow: false,
+    };
+
+    mutate_fn(root_widget);
+
```
DJMcNab, line 36:
```suggestion

// Merge any state changes up to the root.
```

masonry/src/passes/mutate.rs
```diff
@@ -0,0 +1,56 @@
...
+    let mut current_id = Some(id);
+    while let Some(id) = current_id {
+        let parent_id = root.widget_arena.parent_of(id);
+        merge_state_up(&mut root.widget_arena, id);
+        current_id = parent_id;
+    }
+}
+
+pub(crate) fn run_mutate_pass(root: &mut RenderRoot, root_state: &mut WidgetState) {
```
DJMcNab, line 45:
```suggestion
/// Apply any deferred mutations (created using [...Ctx::mutate_later](crate::LayoutCtx::mutate_later)).
pub(crate) fn run_mutate_pass(root: &mut RenderRoot, root_state: &mut WidgetState) {
```

masonry/src/contexts.rs
```diff
@@ -540,11 +488,34 @@ impl_context_method!(WidgetCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
 // --- MARK: OTHER METHODS ---
 // Methods on all context types except PaintCtx and AccessCtx
 impl_context_method!(
-    WidgetCtx<'_>,
+    MutateCtx<'_>,
```
DJMcNab, line 491: Are these meaningful on `MutateCtx`?

masonry/src/passes/mutate.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::passes::merge_state_up;
+use crate::render_root::RenderRoot;
+use crate::widget::WidgetMut;
+use crate::{MutateCtx, Widget, WidgetId, WidgetState};
+
+pub(crate) fn mutate_widget(
+    root: &mut RenderRoot,
+    id: WidgetId,
+    mutate_fn: impl FnOnce(WidgetMut<'_, Box<dyn Widget>>),
+) {
```
DJMcNab, line 15:
```suggestion
pub(crate) fn mutate_widget<R>(
    root: &mut RenderRoot,
    id: WidgetId,
    mutate_fn: impl FnOnce(WidgetMut<'_, Box<dyn Widget>>) -> R,
) -> R {
```

masonry/src/widget/widget_state.rs
```diff
@@ -170,6 +170,20 @@ impl WidgetState {
         }
     }
 
+    pub(crate) fn root(id: WidgetId, size: Size) -> WidgetState {
```
DJMcNab, line 173: I find this `root` thing quite confusing, because my general understanding is that the root widget has its own state.

DJMcNab (approved):
A clear step towards the required results.

masonry/src/render_root.rs
PoignardAzur: @waywardmonkeys requested it for his use-cases.

It's not necessary for Xilem, but it's easy to implement and I can easily imagine use-cases for it for consumers of Masonry.

masonry/src/passes/mutate.rs
PoignardAzur: Yeah, it's a whole thing.

Other reason to have parent state as an Option would be for reborrowed WidgetMut, to replaced the `is_reborrow` flag.

masonry/src/widget/widget_state.rs
PoignardAzur: The default constructor in xilem classic's `ChangeFlags` would set these to true.

The way it worked in Druid and Masonry is that the default constructor set flags used by the composition root (eg `Window` for Druid, `RenderRoot` for Masonry) and then the `WidgetAdded` event set a bunch of flags to true.

masonry/src/passes/mutate.rs
PoignardAzur: I guess that makes sense, since where your "root" is would be pretty important information in that case.

masonry/src/widget/widget_state.rs
PoignardAzur: It's useful for lifetime purposes; but yeah, I think at this point it can probably be considered vestigial. I might remove it once other passes are implemented.

masonry/src/contexts.rs
PoignardAzur: Maybe.

Either way, I'm trying to consider these methods less in terms of "does this context absolutely need it" and more "does it break something if this context has it". I think the context APIs are easier to reason about that way.

masonry/src/widget/widget_state.rs
PoignardAzur: No, I think you're right. Having all the machinery run once by default is easier. Fewer footguns for users that way.

PoignardAzur:
I think I've addressed all reviews. Will merge this tonight if nobody raises objections.

masonry/examples/calc_masonry.rs
```diff
@@ -145,20 +145,25 @@ impl Widget for CalcButton {
         match event {
             PointerEvent::PointerDown(_, _) => {
                 if !ctx.is_disabled() {
-                    ctx.get_mut(&mut self.inner)
-                        .set_background(self.active_color);
+                    let color = self.active_color;
+                    ctx.mutate_later(&mut self.inner, move |mut inner| {
```
DJMcNab, line 149: I think that this should have a comment, explaining why we can't access the background/box directly.

masonry/src/contexts.rs
```diff
@@ -361,86 +361,36 @@ impl<'a> WidgetCtx<'a> {
...
-            widget_state_children: child_state_mut.children,
-            widget_children: child_mut.children,
-        };
-        WidgetMut {
-            ctx: child_ctx,
-            widget: child_mut.item.as_mut_dyn_any().downcast_mut().unwrap(),
-            is_reborrow: false,
+            // We don't don't reborrow parent_widget_state. This means a WidgetMut
+            // made from this won't merge state on Drop, which is usually what you want
+            // from a reborrowed WidgetMut.
```
DJMcNab, line 382: Maybe something like this would be clearer.
```suggestion
            // We don't don't reborrow `parent_widget_state`. This avoids running
            // `merge_up` in `Drop` multiple times for this widget (it will still be called when
            // `self` is dropped)
```

masonry/src/passes/mutate.rs
```diff
@@ -0,0 +1,58 @@
...
+    // TODO - Factor out into a "pre-event" function?
+    // root.state.next_focused_widget = root.state.focused_widget;
+
+    let callbacks = std::mem::take(&mut root.state.mutate_callbacks);
+    for callback in callbacks {
+        mutate_widget(root, callback.id, callback.callback);
+    }
+
+    root_state.merge_up(root.widget_arena.get_state_mut(root.root.id()).item);
+    // root.post_event_processing(&mut root_state);
```
DJMcNab, line 57: We should reuse the allocation here (comment on the implicit `drop(callbacks)`)

Of course, we need to be aware of the fact that new callbacks could have been added, but that won't be the general case.

masonry/src/contexts.rs
DJMcNab: It makes reuse of the `mutate_callbacks` vector potentially fraught, and the semantics aren't especially clear.

Effectively, if you're calling this, you should be using a different method on the `MutateCtx` which you already have...

masonry/src/render_root.rs
```diff
@@ -295,83 +306,94 @@ impl RenderRoot {
...
-        let handled = root_on_pointer_event(self, &mut root_state, &event);
-        run_update_pointer_pass(self, &mut root_state);
+        // TODO - Factor out into a "pre-event" function?
+        self.state.next_focused_widget = self.state.focused_widget;
 
-        self.post_event_processing(&mut root_state);
+        let handled = root_on_pointer_event(self, &mut dummy_state, &event);
+        run_update_pointer_pass(self, &mut dummy_state);
+
+        self.post_event_processing(&mut dummy_state);
```
DJMcNab, line 373: It's not clear to me what role we expect `post_event_processing` to have "long-term".

masonry/src/render_root.rs
```diff
@@ -295,83 +306,94 @@ impl RenderRoot {
...
 
-        let res = {
-            let _span = info_span!("edit_root_widget").entered();
-            f(root_widget)
-        };
-        self.post_event_processing(&mut fake_widget_state);
+        let mut res = None;
+        mutate_widget(self, id, |widget_mut| {
+            res = Some(f(widget_mut));
+        });
```
DJMcNab, line 355:
```suggestion
        let res = mutate_widget(self, id, f);
```

DJMcNab (commented):
A few more details, nothing architecturally problematic

masonry/src/render_root.rs
PoignardAzur: I'll remove it once all the passes are split out from RenderRoot and WidgetPod.

masonry/src/passes/mutate.rs
PoignardAzur: Hard disagree. Skipping one allocation per frame is not worth even one line of code. Especially if we need additional logic in case new callbacks have been added (though I understand you think that shouldn't be possible).

DJMcNab: It's not necessarily just one allocation, if this feature is used widely. But sure, I'm not going to block on this.

masonry/src/contexts.rs
Philipp-M: Agree, I don't see benefits of this yet (other than being confusing, or allure users to "hacky" workarounds). I think we should see how far we get without it on `MutateCtx`. (I'm specifically also talking about `mutate_callbacks` not sure about the rest, that may be useful)

masonry/src/passes/mutate.rs
Philipp-M: Yeah likely more allocations, in case there are added a few callbacks (which seems likely). I don't yet see how your case applies, it's sync isn't it? It could also be double-buffered with a second `Vec` and `std::mem::swap` alternatively. But yeah also no strong feelings here (just seems to be a cheap optimization, which can still be reversed later)...

Philipp-M (approved):
Apart from the two comments seems good to land.

masonry/src/passes/mutate.rs
PoignardAzur: Right.

Even then, `mutate_callbacks` is a Vec of boxes, so we're talking about saving `O(log(N))` reallocs per frame in a context where we have `O(N)` allocations minimum. There's no situation in which that meaningfully impacts performance.

And I think `render_root` is *the* file where you can to keep complexity down as much as you can. It's a complexity bottleneck, so to speak, both because it naturally tends to amass more code as features get added, and because anyone who wants to global view of the code needs to read it.

masonry/src/contexts.rs
PoignardAzur: Will keep as-is for now, with a TODO comment. We can reconsider when passes are all implemented.

PR #511, opened 2024-08-13
Reduce logs in tests

PoignardAzur:
Stop debug log lines from showing up between test indicators when running `cargo test`.

### DIFF
--- a/masonry/src/tracing_backend.rs
+++ b/masonry/src/tracing_backend.rs
@@ -6,7 +6,6 @@ use std::time::UNIX_EPOCH;
 
 use time::macros::format_description;
 use tracing::subscriber::SetGlobalDefaultError;
-use tracing::Level;
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::fmt::time::UtcTime;
 use tracing_subscriber::prelude::*;
--- a/masonry/src/tracing_backend.rs
+++ b/masonry/src/tracing_backend.rs
@@ -6,6 +6,7 @@ use std::time::UNIX_EPOCH;
 
 use time::macros::format_description;
 use tracing::subscriber::SetGlobalDefaultError;
+use tracing::Level;
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::fmt::time::UtcTime;
 use tracing_subscriber::prelude::*;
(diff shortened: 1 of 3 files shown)


masonry/src/tracing_backend.rs
```diff
@@ -99,10 +101,10 @@ pub(crate) fn try_init_layered_tracing() -> Result<(), SetGlobalDefaultError> {
     Ok(())
 }
 
-pub(crate) fn try_init_tracing() -> Result<(), SetGlobalDefaultError> {
+pub(crate) fn try_init_tracing(testing: bool) -> Result<(), SetGlobalDefaultError> {
```
DJMcNab, line 104: A bool parameter here isn't very clear. Can we instead split this into:
`init_test_tracing` and `try_init_tracing`?

I'd consider also making `try_init_layered_tracing` take a level instead, since we probably want to duplicate this in the `try_init_wasm_tracing` version

DJMcNab (commented):
I don't see this when using `cargo nextest`, which I'd probably recommend doing anyway.

However, this seems like a reasonable improvement, but I'd like to see the implementation be slightly different.

masonry/src/tracing_backend.rs
```diff
@@ -34,11 +35,15 @@ pub(crate) fn try_init_wasm_tracing() -> Result<(), SetGlobalDefaultError> {
 }
 
 #[cfg(not(target_arch = "wasm32"))]
-pub(crate) fn try_init_layered_tracing() -> Result<(), SetGlobalDefaultError> {
-    // Default level is DEBUG in --dev, INFO in --release
+pub(crate) fn try_init_layered_tracing(
+    default_level: Option<LevelFilter>,
+) -> Result<(), SetGlobalDefaultError> {
```
DJMcNab, line 39:
```suggestion
fn try_init_layered_tracing(
    default_level: LevelFilter,
) -> Result<(), SetGlobalDefaultError> {
```
I'd like us to move the fallback cases into `try_init_tracing` instead. This would also allow the WASM version to have the same signature and avoid duplicating the fallback logic.

DJMcNab (approved):
Looks good. I'd slightly tweak the separation of concerns, but it seems like you want this to be a different way

masonry/src/tracing_backend.rs
PoignardAzur: > I'd slightly tweak the separation of concerns, but it seems like you want this to be a different way

The advantage of doing things this way is that it keeps all the comments in a single function.

(Also, meta comment. For me seeing a suggestion block communicates "This is a self-contained change you can commit right now." I'd advise not using them for more open-ended suggestions that will fail to compile without other changes. Regular code blocks are fine for that.)

PoignardAzur:
Alright, thanks for the review! And yeah, like you say, I prefer the separation of concerns this way. It's glue code anyway.

PR #512, opened 2024-08-14
Implement compose pass

PoignardAzur:
Add `Widget::compose` method.
Add `ComposeCtx` type.
Add `needs_compose`, `request_compose`, `translation_changed` flags to WidgetState.
Add `window_origin` attribute to WidgetState.
Add convenience methods to TreeArena.
Remove `needs_window_origin` flag and `parent_window_origin` attribute.

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -887,6 +887,9 @@ impl ComposeCtx<'_> {
         self.widget_state.needs_compose
     }
 
+    /// Set a translation for the child widget.
+    ///
+    /// The translation is applied on top of the position from [`LayoutCtx::place_child`].
     pub fn set_child_translation(&mut self, translation: Vec2) {
         if self.widget_state.translation != translation {
             self.widget_state.translation = translation;
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -6,6 +6,7 @@
 use std::time::Duration;
 
 use accesskit::{NodeBuilder, TreeUpdate};
+use kurbo::Vec2;
 use parley::{FontContext, LayoutContext};
 use tracing::{trace, warn};
 
@@ -84,6 +85,13 @@ pub struct LayoutCtx<'a> {
     pub(crate) mouse_pos: Option<Point>,
 }
 
+pub struct ComposeCtx<'a> {
+    pub(crate) global_state: &'a mut RenderRootState,
+    pub(crate) widget_state: &'a mut WidgetState,
+    pub(crate) widget_state_children: ArenaMutChildren<'a, WidgetState>,
+    pub(crate) widget_children: ArenaMutChildren<'a, Box<dyn Widget>>,
+}
+
 /// A context passed to paint methods of widgets.
 pub struct PaintCtx<'a> {
     pub(crate) global_state: &'a mut RenderRootState,
@@ -114,6 +122,7 @@ impl_context_method!(
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
+    ComposeCtx<'_>,
     PaintCtx<'_>,
     AccessCtx<'_>,
     {
@@ -187,6 +196,7 @@ impl_context_method!(
     MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
+    ComposeCtx<'_>,
     PaintCtx<'_>,
     AccessCtx<'_>,
     {
@@ -229,6 +239,7 @@ impl_context_method!(
     MutateCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
+    ComposeCtx<'_>,
     PaintCtx<'_>,
     AccessCtx<'_>,
     {
@@ -412,6 +423,17 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
         self.widget_state.needs_layout = true;
     }
 
+    // TODO - Document better
+    /// Request a compose pass.
+    ///
+    /// The compose pass is often cheaper than the layout pass, because it can only transform individual widgets' position.
+    /// [`compose`]: crate::Widget::compose
+    pub fn request_compose(&mut self) {
+        trace!("request_compose");
+        self.widget_state.needs_compose = true;
+        self.widget_state.request_compose = true;
+    }
+
     pub fn request_accessibility_update(&mut self) {
         trace!("request_accessibility_update");
         self.widget_state.needs_accessibility_update = true;
@@ -494,6 +516,7 @@ impl_context_method!(
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
+    ComposeCtx<'_>,
     {
         // TODO - Remove from MutateCtx?
         /// Queue a callback that will be called with a [`WidgetMut`] for this widget.
@@ -847,7 +870,7 @@ impl LayoutCtx<'_> {
         self.assert_layout_done(child, "place_child");
         if origin != self.get_child_state_mut(child).origin {
             self.get_child_state_mut(child).origin = origin;
-            self.get_child_state_mut(child).needs_window_origin = true;
+            self.get_child_state_mut(child).translation_changed = true;
         }
         self.get_child_state_mut(child)
             .is_expecting_place_child_call = false;
@@ -856,6 +879,19 @@ impl LayoutCtx<'_> {
             .widget_state
             .local_paint_rect
             .union(self.get_child_state(child).paint_rect());
+    }
+}
+
+impl ComposeCtx<'_> {
+    pub fn needs_compose(&self) -> bool {
+        self.widget_state.needs_compose
+    }
+
+    pub fn set_child_translation(&mut self, translation: Vec2) {
+        if self.widget_state.translation != translation {
+            self.widget_state.translation = translation;
+            self.widget_state.translation_changed = true;
+        }
     }
 }
 
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -6,9 +6,9 @@
 use std::time::Duration;
 
 use accesskit::{NodeBuilder, TreeUpdate};
-use kurbo::Vec2;
 use parley::{FontContext, LayoutContext};
 use tracing::{trace, warn};
+use vello::kurbo::Vec2;
 
 use crate::action::Action;
 use crate::render_root::{MutateCallback, RenderRootSignal, RenderRootState};
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -424,7 +424,7 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
     }
 
     // TODO - Document better
-    /// Request a compose pass.
+    /// Request a [`compose`] pass.
     ///
     /// The compose pass is often cheaper than the layout pass, because it can only transform individual widgets' position.
     /// [`compose`]: crate::Widget::compose
--- a/masonry/src/passes/compose.rs
+++ b/masonry/src/passes/compose.rs
@@ -51,9 +51,6 @@ fn compose_widget(
 ) {
     let moved = parent_moved || state.item.translation_changed;
     let translation = parent_translation + state.item.translation + state.item.origin.to_vec2();
-    state.item.needs_compose = false;
-    state.item.request_compose = false;
-    state.item.translation_changed = false;
     state.item.window_origin = Point::new(translation.x, translation.y);
 
     let mut ctx = ComposeCtx {
@@ -65,6 +62,10 @@ fn compose_widget(
     if ctx.widget_state.request_compose {
         widget.item.compose(&mut ctx);
     }
+
+    state.item.needs_compose = false;
+    state.item.request_compose = false;
+    state.item.translation_changed = false;
 
     let id = state.item.id;
     let parent_state = state.item;
--- a/masonry/src/passes/compose.rs
+++ b/masonry/src/passes/compose.rs
@@ -1,0 +1,102 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use kurbo::{Point, Vec2};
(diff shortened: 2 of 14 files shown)


masonry/src/widget/widget_state.rs
```diff
@@ -5,11 +5,17 @@
 
 use std::sync::atomic::{AtomicBool, Ordering};
 
+use kurbo::Vec2;
```
waywardmonkeys, line 8: `use crate::kurbo::Vec2;`

waywardmonkeys: Actually, can combine with the import below.

waywardmonkeys (commented):
Some initial comments...

masonry/src/contexts.rs
```diff
@@ -879,6 +902,19 @@ impl LayoutCtx<'_> {
     }
 }
 
+impl ComposeCtx<'_> {
+    pub fn needs_compose(&self) -> bool {
+        self.widget_state.needs_compose
+    }
+
+    pub fn set_child_translation(&mut self, translation: Vec2) {
+        if self.widget_state.translation != translation {
```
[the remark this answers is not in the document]
PoignardAzur: Nah. Worst case, you do superfluous work.

masonry/src/passes/compose.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    for child_id in widget.item.children_ids() {
+        let widget = widget
+            .children
+            .get_child_mut(child_id.to_raw())
+            .unwrap_or_else(|| {
+                panic!(
+                    "Error in '{}' #{}: cannot find child #{} returned by children_ids()",
+                    parent_name,
+                    parent_id.to_raw(),
```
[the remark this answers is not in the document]
PoignardAzur: Sure, but that's a future PR.

masonry/src/passes/compose.rs
```diff
@@ -0,0 +1,103 @@
...
+fn compose_widget(
+    global_state: &mut RenderRootState,
+    mut widget: ArenaMut<'_, Box<dyn Widget>>,
+    mut state: ArenaMut<'_, WidgetState>,
+    parent_moved: bool,
+    parent_translation: Vec2,
+) {
+    let moved = parent_moved || state.item.translation_changed;
+    let translation = parent_translation + state.item.translation + state.item.origin.to_vec2();
+    state.item.window_origin = Point::new(translation.x, translation.y);
```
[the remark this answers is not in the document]
PoignardAzur: (For what it's worth, you can create an inline suggestion with the "Add a suggestion button"; it's the one with the file icon with a +-, next to the "Preview" tab.)

masonry/src/contexts.rs
```diff
@@ -6,6 +6,7 @@
 use std::time::Duration;
 
 use accesskit::{NodeBuilder, TreeUpdate};
+use kurbo::Vec2;
```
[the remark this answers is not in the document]
PoignardAzur: I think it doesn't make a difference, they resolve to the same symbol. (Otherwise I think rustc would complain.)

I don't know if we have a coding style for this, but either way, we're already using both version throughout Masonry, so I'll keep the ones you pointed out as-is.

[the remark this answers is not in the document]
PoignardAzur: I think it doesn't make a difference, they resolve to the same symbol. (Otherwise I think rustc would complain.)

I don't know if we have a coding style for this, but either way, we're already using both version throughout Masonry, so I'll keep the ones you pointed out as-is.

masonry/src/widget/widget_state.rs
```diff
@@ -61,6 +67,11 @@ pub struct WidgetState {
     // TODO - Document
     pub(crate) is_portal: bool,
 
+    pub(crate) request_compose: bool,
+    // TODO - Handle matrix transforms
```
DJMcNab, line 71: Matrices are tricky with text, due to optical size of text (which we don't yet handle). I think changing the scale of a child can only correctly happen in `layout`

DJMcNab (commented):
It's hard to judge exactly what this is doing with no widgets actually using it...

Code seems about as expected, I believe.

PoignardAzur:
I think this is ready to merge?

masonry/src/contexts.rs
```diff
@@ -847,7 +870,7 @@ impl LayoutCtx<'_> {
         self.assert_layout_done(child, "place_child");
         if origin != self.get_child_state_mut(child).origin {
             self.get_child_state_mut(child).origin = origin;
-            self.get_child_state_mut(child).needs_window_origin = true;
+            self.get_child_state_mut(child).translation_changed = true;
```
Philipp-M, line 873: would `origin_changed` (or `window_origin_changed`) make more sense (as this is the defacto end-result if understand this correctly)?

masonry/src/contexts.rs
```diff
@@ -859,6 +882,19 @@ impl LayoutCtx<'_> {
     }
 }
 
+impl ComposeCtx<'_> {
+    pub fn needs_compose(&self) -> bool {
+        self.widget_state.needs_compose
+    }
+
+    pub fn set_child_translation(&mut self, translation: Vec2) {
```
Philipp-M, line 890:
```suggestion
    /// offsets the child by `translation` relative to the parents `window_origin`
    pub fn set_child_translation(&mut self, translation: Vec2) {
```

masonry/src/passes/compose.rs
```diff
@@ -0,0 +1,103 @@
...
+                widget,
+                state.reborrow_mut(),
+                moved,
+                translation,
+            );
+            parent_state.merge_up(state.item);
+        },
+    );
+
+    // state.item.moved = false;
```
Philipp-M, line 91: Nit: does this still make sense?

masonry/src/tree_arena.rs
```diff
@@ -372,6 +372,38 @@ impl<'a, Item> ArenaMutChildren<'a, Item> {
     }
 }
 
+impl<'a, Item> ArenaRef<'a, Item> {
+    pub fn id(&self) -> u64 {
```
Philipp-M, line 376: As this is `pub` I think it could have a `# Panics` doc-comment, that this could panic under certain conditions (same below for `ArenaMut`).

masonry/src/widget/widget.rs
```diff
@@ -105,6 +106,8 @@ pub trait Widget: AsAny {
     /// The layout strategy is strongly inspired by Flutter.
     fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size;
 
+    fn compose(&mut self, ctx: &mut ComposeCtx) {}
```
Philipp-M, line 109: Since you add this already as default impl, should I make a PR for all the other methods here too? (After merging this, to avoid merge-conflicts)

masonry/src/widget/widget_pod.rs
```diff
@@ -372,8 +372,9 @@ impl<W: Widget> WidgetPod<W> {
                     }
                 }
                 InternalLifeCycle::ParentWindowOrigin { .. } => {
```
Philipp-M, line 374: As this doesn't do anything anymore, can we just remove this?

masonry/src/widget/widget_state.rs
Philipp-M: (I mean it's just a nit, but I agree, it's *slightly* cleaner, same for the other `use` comments here)

masonry/src/tree_arena.rs
```diff
@@ -372,6 +372,38 @@ impl<'a, Item> ArenaMutChildren<'a, Item> {
...
+        self.children.id.unwrap()
+    }
+}
+
+impl<'a, Item> ArenaMut<'a, Item> {
+    pub fn id(&self) -> u64 {
+        self.children.id.unwrap()
+    }
+
+    /// Returns a shared token equivalent to this one.
```
DJMcNab, line 386: `Shared` doesn't seem quite right, given that `ArenaRef` cannot be copied (etc.). That is, `ArenaRef` does not appear to have any additional abilities over `ArenaMut`, except for being passed to methods which accept `ArenaRef`.

This method also doesn't seem to be used?

masonry/src/widget/widget_pod.rs
```diff
@@ -372,8 +372,9 @@ impl<W: Widget> WidgetPod<W> {
...
-                    state.parent_window_origin = parent_ctx.widget_state.window_origin();
-                    state.needs_window_origin = false;
+                    // FIXME
```
DJMcNab, line 375: Does this event need to be removed?

DJMcNab (commented):
I think the old `ParentWindowOrigin` code can be removed now?
I would much prefer that to happen in this PR if so.

masonry/src/contexts.rs
DJMcNab: AIUI, this flag means that "my local translation has changed", and so the "global translation" needs to be updated.

PoignardAzur: Yeah, `translation` and `window_origin` have different semantics.

masonry/src/contexts.rs
```diff
@@ -462,6 +473,17 @@ impl_context_method!(WidgetCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
         self.widget_state.needs_layout = true;
     }
 
+    // TODO - Document better
+    /// Request a compose pass.
+    ///
+    /// The compose pass is often cheaper than the layout pass, because it can only transform individual widgets' position.
+    /// [`compose`]: crate::Widget::compose
```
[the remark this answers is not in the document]
PoignardAzur: Fixed.

masonry/src/passes/compose.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use kurbo::{Point, Vec2};
```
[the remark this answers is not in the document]
PoignardAzur: Won't fix, see other comment.

masonry/src/widget/widget_state.rs
PoignardAzur: Won't fix, see other comment.

masonry/src/widget/widget_state.rs
PoignardAzur: I'm leaving that debate for once we do a complete layout refactor.

masonry/src/tree_arena.rs
PoignardAzur: The id is never None is that context. Added a comment to document it.

masonry/src/passes/compose.rs
PoignardAzur: Nope.

masonry/src/widget/widget.rs
PoignardAzur: I'm adding the default implementations as I convert the passes. I'd rather not add them for passes that still need to recurse on children FWIW.

masonry/src/tree_arena.rs
PoignardAzur: > Shared doesn't seem quite right, given that ArenaRef cannot be copied

Well now it can =P

> This method also doesn't seem to be used?

Yeah, but I'm writing this as though it was a public crate.

PoignardAzur:
I've done some small fixes to address code review. Next step is removing `Lifecycle::ParentWindowOrigin` and associated stuff.

masonry/src/widget/widget_state.rs
Philipp-M: We should probably generally find a unique style in the future, just to be consistent.

waywardmonkeys: We've had an agreement in the past in #223, but I needed to rebase it and it slipped through the cracks.

I've resurrected it.

PoignardAzur:
I believe this is ready to merge.

PoignardAzur:
We should probably merge #223 first, then rebase this on top of it.

masonry/src/contexts.rs
Philipp-M: (Don't want this to be a conflict point, but was this overlooked or do you want some different doc-comment?)

PoignardAzur: Huh, that's odd, I thought I'd added a comment there.

... And, looking again, yes I did. The UI is just absolutely terrible at showing it.

Philipp-M: Ok, yeah indeed weird, I thought it would show `Outdated` or something, thanks :)

PoignardAzur:
> We should probably merge https://github.com/linebender/xilem/pull/223 first, then rebase this on top of it.

Done. I believe this addresses all code review points.

DJMcNab (approved):
This is compelling! It's really nice to see the back of the parent window origin things

masonry/src/contexts.rs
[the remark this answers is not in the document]
Philipp-M: Hmm the Link seems still not to be working using `cargo doc --no-deps`

masonry/src/contexts.rs
```diff
@@ -412,6 +423,17 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
         self.widget_state.needs_layout = true;
     }
 
+    // TODO - Document better
+    /// Request a [`compose`] pass.
+    ///
+    /// The compose pass is often cheaper than the layout pass, because it can only transform individual widgets' position.
```
Philipp-M, line 429:
```suggestion
    /// The compose pass is often cheaper than the layout pass, because it can only transform individual widget's position.
```

(As I just see this)

PoignardAzur: I think the old version is fine?

... you know what, let me look it up.

DJMcNab: I think the original is the correct form for "the position of each of multiple widgets". However,
```suggestion
    /// The compose pass is often cheaper than the layout pass, because it can only change the position of widgets.
```
might be clearer still

PoignardAzur:
... It just now occurs to me, as I start the paint pass, that I completely forgot to hook composition with the Scene transforms. We didn't catch it because widgets don't use the pass yet.

Oh well, I'll hook it up with the Portal widget while I implement the paint pass.

PR #517, opened 2024-08-15
xilem_web: Document a lot and remove `start_<modifier>` as it doesn't do anything in `View::build` and may be misleading

Philipp-M:
I think I've covered every pub item at the top level.

It also renames  (taking `Attributes` as example)  `start_attribute_modifier` -> `rebuild_attribute_modifier`,
and `end_attribute_modifier` -> `mark_end_of_attribute_modifier`, as I think that makes it more clear what these methods do.
`start_attribute_modifier` was a noop in `View::build` and may lead to buggy behavior when `DomNode::apply_props` is called before every parent `View::build` has done modifying the props. This should make `DomNode::apply_props` robust. And `AfterBuild` is fixed now with that.

### DIFF
--- a/xilem_web/src/attribute.rs
+++ b/xilem_web/src/attribute.rs
@@ -11,19 +11,21 @@ use crate::{
 
 type CowStr = std::borrow::Cow<'static, str>;
 
-/// This trait enables having attributes DOM [`Element`](`crate::interfaces::Element`)s. It is used within [`View`]s that modify the attributes of an element.
+/// This trait allows (modifying) HTML/SVG/MathML attributes on DOM [`Element`](`crate::interfaces::Element`)s.
 ///
 /// Modifications have to be done on the up-traversal of [`View::rebuild`], i.e. after [`View::rebuild`] was invoked for descendent views.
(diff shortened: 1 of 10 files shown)


Philipp-M:
(Should probably do that for xilem too...)

xilem_web/src/attribute.rs
```diff
@@ -61,6 +61,15 @@ impl Attributes {
...
 
 fn set_attribute(element: &web_sys::Element, name: &str, value: &str) {
+    debug_assert_ne!(
+        name, "class",
+        "Using `class` as attribute is not supported, use the `el.class()` modifier instead"
+    );
+    debug_assert_ne!(
+        name, "style",
+        "Using `style` as attribute is not supported, use the `el.style()` modifier instead"
+    );
```
flosse, line 71: These are some of the details and possibilities that I love about rust. A very good usecase in my opinion :+1:

flosse (approved):
Very nice! From a developer's perspective, it's sometimes a boring task, but I think the effect for users is very high and valuable.
Thx!

Philipp-M:
Thanks for the kind words. Yeah it was just a bit to see that "100% documented" on the first page :), and I hope makes contributing easier.

PR #519, opened 2024-08-16
xilem_web: Apply props when the element type changes in `replace_inner`

### LINKED ISSUE #518: xilem_web: iframe `src` attribute not set after update
In this example, the `iframe` src attribute isn't set after clicking the button:

```rust
use xilem_web::{document_body, elements::html, interfaces::Element, App, DomView};

fn update(state: &mut bool) -> impl DomView<bool> {
    html::div((
        html::button("click me!").on_click(|state: &mut bool, _| *state = !*state),
        html::br(()),
        if *state {
            html::iframe(())
                .attr("src", "https://www.google.com")
                .boxed()
        } else {
            html::div("div!").boxed()
        },
    ))
}

fn main() {
    App::new(document_body(), false, update).run();
}
```

If the branches are swapped, the iframe `src` attribute is initially set, but unset after clicking the button twice.

Philipp-M:
Fixes #518

### DIFF
--- a/xilem_web/src/lib.rs
+++ b/xilem_web/src/lib.rs
@@ -289,8 +289,9 @@ impl<E: DomNode<P>, P: 'static> AnyElement<Pod<E, P>> for AnyPod {
 impl AnyPod {
     pub(crate) fn replace_inner<E: DomNode<P>, P: 'static>(
         this: &mut PodMut<'_, DynNode, Box<dyn Any>>,
-        node: Pod<E, P>,
+        mut node: Pod<E, P>,
     ) {
+        node.node.apply_props(&mut node.props);
         if let Some(parent) = this.parent {
             parent
                 .replace_child(node.node.as_ref(), this.node.as_ref())


DJMcNab (approved):
Don't really know the context, but optimistic approval for Xilem Web

Philipp-M:
(Attributes such as `src` in `<iframe src="..">` were not applied when changing the inner element (e.g. `<div>` -> `<iframe>`))

Thanks

DJMcNab:
When I say the context here, I meant of why the change fixes that. But thanks, yeah

Philipp-M:
Oh, ok, yeah it's basically, that the `ElementProps` are a virtual DOM version of the actual DOM representation and get updated in the modifier views. To actually change the DOM, `DomNode::apply_props` has to be run. This didn't happen when changing the inner element of an `AnyView`. The attributes are not applied in the type-erased version of the `Pod` (`AnyPod`), as... well the type of the props is erased, this happens before upcasting, or generally when erasing the type of the element.

PR #520, opened 2024-08-16
masonry: Fix a couple of broken doc links.

waywardmonkeys:


PoignardAzur (approved):
LGTM.

I'm really not happy that those methods are documented in `WidgetMut`, but we can't do much about it for now.

waywardmonkeys:
There's a lot of other broken links (which is why we've disabled checking for these in masonry). Probably going to take an axe to things soon.

Philipp-M:
> I'm really not happy that those methods are documented in WidgetMut, but we can't do much about it for now.

This could *probably* be tackled by having Modifier traits I've mentioned some time ago on zulip.

I.e. something like `trait WithFlex` (maybe even with default methods, which would save us code-duplication).
And `impl WithFlex for WidgetMut<Flex>`, `impl WithFlex for WidgetPod<Flex>` and  `impl WithFlex for Flex`.
(This may then also allow a modifier based system as proposed I think).

DJMcNab:
The other unfortunate thing about this is that those methods aren't unique - e.g. the methods which are the same on Label and Prose don't like properly.

Extension traits can't come soon enough...

PR #525, opened 2024-08-17
Implement update passes

PoignardAzur:
Implement update_anim pass.
Implement update_focus pass.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -4,11 +4,12 @@
 use std::collections::HashSet;
 
 use cursor_icon::CursorIcon;
-use tracing::trace;
+use tracing::{info_span, trace};
 
 use crate::passes::merge_state_up;
-use crate::render_root::{RenderRoot, RenderRootSignal};
-use crate::{LifeCycleCtx, StatusChange, Widget, WidgetId, WidgetState};
+use crate::render_root::{RenderRoot, RenderRootSignal, RenderRootState};
+use crate::tree_arena::{ArenaMut, ArenaMutChildren};
+use crate::{LifeCycle, LifeCycleCtx, StatusChange, Widget, WidgetId, WidgetState};
 
 fn get_id_path(root: &RenderRoot, widget_id: Option<WidgetId>) -> Vec<WidgetId> {
     let Some(widget_id) = widget_id else {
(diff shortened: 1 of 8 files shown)


masonry/src/widget/widget_state.rs
```diff
@@ -90,6 +90,7 @@ pub struct WidgetState {
     pub(crate) needs_accessibility_update: bool,
 
     /// Any descendant has requested an animation frame.
+    pub(crate) needs_anim: bool,
     pub(crate) request_anim: bool,
```
jaredoconnell, line 94: I think some differentiating comments for these two fields would be helpful. Looking at the code, it looks like needs_anim allows it to start the update anim pass, and request_anim instructs it to send the lifecycle event during the update anim pass. If so, I think clearer variable names could be helpful.

PoignardAzur: It's not clearly documented yet, but a naming scheme I'm starting to adopt for these flags is that "request_x" means "this widget requested x" and "needs_x" means "this widget or a descendant requested x".

masonry/src/passes/update.rs
```diff
@@ -145,3 +146,86 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
     // Pass root widget state to synthetic state create at beginning of pass
     root_state.merge_up(root.widget_arena.get_state_mut(root.root.id()).item);
 }
+
+// ----------------
+
+fn recurse_on_children(
```
DJMcNab, line 152: This looks extremely similar to the function in `passes/compose.rs`. Please document the differences here

PoignardAzur: It's the same. I'll remove the duplicate once #522 is merged.

masonry/src/widget/widget_state.rs
PoignardAzur: I've added some comments in #522. Will add more here later.

PoignardAzur: Done.

masonry/src/passes/update.rs
PoignardAzur: Done.

PoignardAzur:
I'm closing this PR. Will reopen as two separate PRs.

PR #527, opened 2024-08-19
Fix Portal widget

PoignardAzur:
One of the problems we ran into with the `to_do_list` example was that we had a `Flex` inside a `Portal`: the `Portal` widget is designed to display scrollbars and provide scrolling if the child is larger than its constraints... but then Flex's `layout` method always clipped the returned size so it wasn't ever larger than its constraints. Hence, no scrolling. Removing the `bc.constrain(my_size)` part makes the example have scrolling behavior again.

This PR also tweaks scrolling code and adds `#[track_caller]` attributes to help with debugging.

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -708,6 +708,7 @@ impl LifeCycleCtx<'_> {
 
 // --- MARK: UPDATE LAYOUT ---
 impl LayoutCtx<'_> {
+    #[track_caller]
     fn assert_layout_done(&self, child: &WidgetPod<impl Widget>, method_name: &str) {
         if self.get_child_state(child).needs_layout {
             debug_panic!(
@@ -720,6 +721,7 @@ impl LayoutCtx<'_> {
         }
     }
 
+    #[track_caller]
     fn assert_placed(&self, child: &WidgetPod<impl Widget>, method_name: &str) {
         if self.get_child_state(child).is_expecting_place_child_call {
             debug_panic!(
@@ -760,6 +762,7 @@ impl LayoutCtx<'_> {
     ///
     /// This method will panic if the child's [`layout()`](WidgetPod::layout) method has not been called yet
     /// and if [`LayoutCtx::place_child()`] has not been called for the child.
+    #[track_caller]
     pub fn compute_insets_from_child(
         &mut self,
         child: &WidgetPod<impl Widget>,
@@ -805,17 +808,13 @@ impl LayoutCtx<'_> {
         self.widget_state.mark_as_visited(visited);
     }
 
-    pub(crate) fn mark_child_as_visited(&self, child: &WidgetPod<impl Widget>, visited: bool) {
-        #[cfg(debug_assertions)]
-        self.get_child_state(child).mark_as_visited(visited);
-    }
-
     /// The distance from the bottom of the given widget to the baseline.
     ///
     /// ## Panics
     ///
     /// This method will panic if [`WidgetPod::layout`] has not been called yet for
     /// the child.
+    #[track_caller]
     pub fn child_baseline_offset(&self, child: &WidgetPod<impl Widget>) -> f64 {
         self.assert_layout_done(child, "child_baseline_offset");
         self.get_child_state(child).baseline_offset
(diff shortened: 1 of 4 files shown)


masonry/src/widget/portal.rs
```diff
@@ -245,11 +245,8 @@ impl<W: Widget> Widget for Portal<W> {
 
         match event {
             PointerEvent::MouseWheel(delta, _) => {
-                self.set_viewport_pos_raw(
-                    portal_size,
-                    content_size,
-                    self.viewport_pos + Vec2::new(delta.x, delta.y),
-                );
+                let delta = Vec2::new(delta.x * -10., delta.y * -10.);
```
waywardmonkeys, line 248: What are these magic numbers?

waywardmonkeys (commented):
Overall, I'm not sure what's going on here at all as there's only a single comment (and not related to the temporary changes).

I'm not sure why the track caller stuff is part of a same commit with a fix for something... and no comment explaining why that's there either.

PoignardAzur:
> Overall, I'm not sure what's going on here at all as there's only a single comment (and not related to the temporary changes).

The comment is somewhat related.

Basically, one of the problems we ran into with the to_do_list example was that we had a Flex inside a Portal, and the Portal is design to display scrollbars and provide scrolling if the child is larger than its constraints... but then Flex's layout method always clipped the returned size it wasn't ever larger than its constraints. Hence, no scrolling. Removing the `bc.constrain(my_size)` part makes the example have scrolling behavior again.

I guess I should add this explanation to the commit message.

> I'm not sure why the track caller stuff is part of a same commit with a fix for something... and no comment explaining why that's there either.

I think `#[track_caller]` is standard / researchable enough not to need a comment.

The `#[track_caller]` annotation makes it so that if a function crashes, the crash location reported (file, line number, etc) will be the location of the function's caller, not the part inside the function that crashed. I added a bunch of them because I kept getting crashes and these annotations are helpful to pinpoint their cause.

masonry/src/widget/portal.rs
PoignardAzur: Plugging the default values as deltas makes scrolling way too slow.

You're right that using magic numbers is a poor solution, though. I'll use a constant for now.

masonry/src/widget/flex.rs
```diff
@@ -852,18 +852,11 @@ impl Widget for Flex {
...
-        // if we don't have to fill the main axis, we loosen that axis before constraining
-        let my_size = if !self.fill_major_axis {
-            let max_major = self.direction.major(bc.max());
-            self.direction
-                .constraints(bc, 0.0, max_major)
-                .constrain(my_size)
-        } else {
-            bc.constrain(my_size)
-        };
-
```
PoignardAzur, line 866: Upon re-read, I'm realizing I'm not sure what this block does (specifically the fill_major_axis part), so removing it entirely may be ill-advised.

PoignardAzur: Upom re-re-read, I think it's fine. This block essentially applies both minimum constraints and maximum constraints to the returned size. We specifically don't want to apply maximum constraints (so scrolling works), and minimum constraints are applied inconsistently in Masonry anyway.

jaredoconnell (approved):
Tested. This fixes the lack of scroll bars.
In the future it would be better to move the top textbox and button out of the portal so they stay on top while the list items freeze in the to_do_list example.

masonry/src/contexts.rs
```diff
@@ -851,11 +852,18 @@ impl LayoutCtx<'_> {
     ///
     /// This method will panic if [`WidgetPod::layout`] has not been called yet for
     /// the child.
+    #[track_caller]
     pub fn child_size(&self, child: &WidgetPod<impl Widget>) -> Size {
         self.assert_layout_done(child, "child_size");
         self.get_child_state(child).layout_rect().size()
     }
 
+    pub(crate) fn skip_layout(&mut self, child: &mut WidgetPod<impl Widget>) {
```
DJMcNab, line 861: Why is this `pub(crate)`? It seems possible that this could be usefully exposed externally.

I also think we should move this to between the private and public methods; I'd normally prefer the order to be public, pub(crate) and private, but that's not the extant style, which is fine.

DJMcNab (approved):
Looks good, and works for me. I would really like to see `skip_layout` clarified and documented before landing this.

waywardmonkeys:
Just checking that you will add that to the commit message…

PoignardAzur:
> Just checking that you will add that to the commit message…

Good catch =)

PR #529, opened 2024-08-19
Migrate layout pass

PoignardAzur:


masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
DJMcNab, line 3:
```suggestion

//! The layout pass, which provides the size and position of each widget
//! before any translations applied in [`compose`](super::compose).
//! Most of the logic for this pass happens in [`Widget::layout`] implementations,
//!
//! In the future, this might also support scaling the child widgets.

```

masonry/src/widget/widget_pod.rs
```diff
@@ -504,26 +505,12 @@ impl<W: Widget> WidgetPod<W> {
     ///
     /// [`layout`]: Widget::layout
     pub fn layout(&mut self, parent_ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
```
DJMcNab, line 507: Why is this still here?

masonry/src/contexts.rs
```diff
@@ -856,6 +859,16 @@ impl LayoutCtx<'_> {
         self.get_child_state(child).layout_rect().size()
     }
 
+    /// Compute layout of a child widget.
+    ///
+    /// Container widgets must call this on every child as part of
+    /// their [`layout`] method.
+    ///
+    /// [`layout`]: Widget::layout
+    pub fn run_layout<W: Widget>(&mut self, child: &mut WidgetPod<W>, bc: &BoxConstraints) -> Size {
```
DJMcNab, line 868: We should ensure that this method is clearly pointed to in the docs for `Widget::layout`.

I'd also encourage moving this method higher up in this impl block, because it is among the most important methods on `LayoutCtx`

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,192 @@
...
+    let id = pod.id().to_raw();
+    let state_mut = parent_ctx
+        .widget_state_children
+        .get_child_mut(id)
+        .expect("run_layout_on: inner widget not found in widget tree");
+    state_mut.item.size
+}
+
+pub fn root_layout(
```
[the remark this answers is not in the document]
PoignardAzur: No, that's a typo. It should be `pub(crate)`.

masonry/src/passes/layout.rs
PoignardAzur: I think scaling the child widgets would happen in compose.

DJMcNab: Maybe. It depends on what purpose the scaling has

Because for text, layout can depend on the absolute logical size of the text, due to the `opsz` (optical size) variable font axis. So it depends on whether the scaling is logically scaling the logic size

masonry/src/passes/layout.rs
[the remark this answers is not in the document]
PoignardAzur: Fixed.

masonry/src/passes/layout.rs
PoignardAzur: Added, except for the part about scaling children.

masonry/src/widget/widget_pod.rs
PoignardAzur: To keep the diff small. We can change all the widgets in a follow-up PR. I'll include a comment to that effect.

PoignardAzur: Done.

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,196 @@
...
+    let new_size = {
+        let mut inner_ctx = LayoutCtx {
+            widget_state: state,
+            widget_state_children: state_mut.children.reborrow_mut(),
+            widget_children: widget_mut.children,
+            global_state: parent_ctx.global_state,
+            mouse_pos: parent_ctx.mouse_pos,
+        };
+
+        widget.layout(&mut inner_ctx, bc)
```
DJMcNab, line 72: Would this be the case to evaluate a potentially cached layout result? I guess that would be what `request_layout` would be needed for?

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,196 @@
...
+    };
+
+    // TODO - One we add request_layout, check for that flag.
+    // If it's true, that's probably an error.
+    state.needs_layout = false;
+
+    state.local_paint_rect = state
+        .local_paint_rect
+        .union(new_size.to_rect() + state.paint_insets);
```
DJMcNab, line 81: I'm not a *huge* fan of using `Add` for `Insets`, but that's a Kurbo thing - this is an idiomatic usage thereof

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,196 @@
...
+    state_mut.item.size = new_size;
+
+    log_layout_issues(widget.short_type_name(), new_size);
+
+    true
+}
+
+fn log_layout_issues(type_name: &str, size: Size) {
+    if size.width.is_infinite() {
+        warn!("Widget `{type_name}` has an infinite width.");
```
DJMcNab, line 142: Should we log the id here as well?

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+    if size.height.is_infinite() {
+        warn!("Widget `{type_name}` has an infinite height.");
+    }
+}
+
+pub(crate) fn run_layout_on<W: Widget>(
```
DJMcNab, line 149: We should have a small doc comment here. E.g.:
```suggestion
/// Run [`Widget::layout`] method on the widget contained in `pod`.
/// This will be called by [`LayoutCtx::run_layout`], which is itself called in the parent widget's `layout`.
pub(crate) fn run_layout_on<W: Widget>(
```

masonry/src/widget/widget_pod.rs
DJMcNab: All of the widgets already appear to be translated in this PR?

DJMcNab (approved):

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,196 @@
...
+        && smaller.x1 <= larger.x1
+        && smaller.y0 >= larger.y0
+        && smaller.y1 <= larger.y1
+}
+
+pub(crate) fn run_layout_inner<W: Widget>(
+    parent_ctx: &mut LayoutCtx<'_>,
+    pod: &mut WidgetPod<W>,
+    bc: &BoxConstraints,
+) -> bool {
```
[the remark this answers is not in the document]
PoignardAzur: I think this pattern pulled more weight when every WidgetPod method used it and `call_widget_method_with_checks` was used everywhere. We'll probably refactor it away soon. In the meantime, I added a comment for clarity.

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,196 @@
...
+            id,
+        );
+        state.size = Size::ZERO;
+        return false;
+    }
+
+    state.needs_compose = true;
+    state.is_expecting_place_child_call = true;
+    // TODO - Not everything that has been re-laid out needs to be repainted.
+    state.needs_paint = true;
```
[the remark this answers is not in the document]
PoignardAzur: I think I wrote this code before I renamed those flags. Will fix.

masonry/src/passes/layout.rs
PoignardAzur: True. I'm adding a comment and leaving it for a future PR.

PoignardAzur:
Will merge before this evening unless a blocker comes up.

masonry/src/passes/layout.rs
```diff
@@ -0,0 +1,217 @@
...
+    };
+    if state.request_layout {
+        debug_panic!(
+            "Error in '{}' #{}: layout request flag was set during layout pass",
+            widget.short_type_name(),
+            id,
+        );
+    }
+
+    state.needs_layout = false;
```
DJMcNab, line 89: There's a little bit of knowledge encoded in this line, so it could be worth having a comment here.

Not going to block on that, though

PR #530, opened 2024-08-21
Reduce duplicated logic in  `tracing_backend`, and add some docs

DJMcNab:
The calculation of the levels was previously duplicated for both web and native, and the `WARN` level was not used properly on WASM for tests.

### DIFF
--- a/masonry/src/tracing_backend.rs
+++ b/masonry/src/tracing_backend.rs
@@ -1,6 +1,18 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+#![warn(rustdoc::broken_intra_doc_links, clippy::doc_markdown, missing_docs)]
+
+//! Configures a suitable default [`tracing`] implementation for a Masonry application.
+//!
+//! This uses a custom log format specialised for GUI applications,
+//! and will write all logs to a temporary file in debug mode.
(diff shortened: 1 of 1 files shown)


masonry/src/tracing_backend.rs
```diff
@@ -65,6 +44,7 @@ pub(crate) fn try_init_layered_tracing(
         // We append a `Z` here to indicate clearly that this is a UTC time
         "[hour repr:24]:[minute]:[second].[subsecond digits:3]Z"
     ));
+    // If modifying, also update the module level docs
```
PoignardAzur, line 47:
```suggestion
    // If you change these values, also update the module level docs
```

masonry/src/tracing_backend.rs
```diff
@@ -77,14 +57,21 @@ pub(crate) fn try_init_layered_tracing(
             .unwrap()
             .as_millis();
         let tmp_path = std::env::temp_dir().join(format!("masonry-{id:016}-dense.log"));
-        // TODO - For some reason, `.with_ansi(false)` still leaves some italics in the output.
+        // If modifying, also update the module level docs
```
PoignardAzur, line 60:
```suggestion
        // If you change these values, also update the module level docs
```

PoignardAzur (approved):
LGTM.

PR #531, opened 2024-08-21
Add trace span to AnyView

PoignardAzur:


xilem/src/any_view.rs
```diff
@@ -106,4 +107,8 @@ impl Widget for DynWidget {
     fn children_ids(&self) -> SmallVec<[WidgetId; 16]> {
         smallvec![self.inner.id()]
     }
+
+    fn make_trace_span(&self) -> Span {
+        trace_span!("DynWidget")
```
DJMcNab, line 112: I don't think we expect the name `DynWidget` to be user-facing. Ideally, we'd just have this widget not have a span (or even not exist). But if not, I think `AnyWidgetView` *might* be a better name here.

Not blocking on that suggestion, though

DJMcNab (approved):

xilem/src/any_view.rs
PoignardAzur: I kind of get what you mean, that ideally Dynwidget would be a fully encapsulated abstraction invisible to the users.

In the meantime though, these spans are mainly used for debugging, and the string they give should be usable in a grep search to find the widget's source code.

PR #532, opened 2024-08-21
Fix Flex widget

PoignardAzur:
This fixes a regression in the Mason example.


DJMcNab (approved):
Looks to work as expected. Thanks

PR #535, opened 2024-08-22
Disable hinting whilst a `VariableLabel` animation is ongoing

DJMcNab:
If using hinting during an animation, a shimmering effect can occur.

### DIFF
--- a/masonry/src/text/layout.rs
+++ b/masonry/src/text/layout.rs
@@ -89,6 +89,18 @@ pub enum TextBrush {
     },
 }
 
+impl TextBrush {
+    pub fn set_hinting(&mut self, hinting: Hinting) {
+        match self {
+            TextBrush::Normal(_, should_hint) => *should_hint = hinting,
+            TextBrush::Highlight {
+                hinting: should_hint,
+                ..
+            } => *should_hint = hinting,
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


jaredoconnell (approved):
This fixes the choppiness that I was experiencing earlier. It works perfectly on Windows and Mac.

DJMcNab:
Hmm, I'm surprised. I didn't think this had any impact on example directly... 
Are you sure you're still seeing the choppiness on main?

jaredoconnell:
> Hmm, I'm surprised. I didn't think this had any impact on example directly... Are you sure you're still seeing the choppiness on main?

Ah. You are correct. In which case, I can confirm that I cannot spot any regressions in the example.

DJMcNab:
Grand! No big surprises here, but this is the right behaviour. See also https://github.com/linebender/vello/issues/204 for the explanation of the shimmering.

This is more important once https://github.com/googlefonts/fontations/pull/1083 lands.

(@dfrg would it be useful for you if I hooked that up in Vello for testing?)

dfrg:
> (@dfrg would it be useful for you if I hooked that up in Vello for testing?)

I appreciate the offer but that work is still in flux so it’s probably best to wait for a release.

PR #536, opened 2024-08-24
Remove bloom filter

PoignardAzur:
The filter was only used for RouteFocusChanged which is about to be removed. Since it was a performance feature, we can remove it early.


waywardmonkeys (changes requested):
You will be able to remove the `fnv` dependency from both The masonry Cargo.toml but also the workspace one and Cargo.lock

PoignardAzur:
Done.

xilem/examples/to_do_mvc.rs
```diff
@@ -14,6 +14,7 @@ struct Task {
 
 struct TaskList {
     next_task: String,
+    next_task2: String,
```
waywardmonkeys, line 17: These changes to a xilem example seem unrelated and don’t belong in this PR.

waywardmonkeys (approved):
LGTM once the changes to the. Item example are removed.

xilem/examples/to_do_mvc.rs
```diff
@@ -66,7 +66,7 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<TaskList> {
         })
         .collect::<Vec<_>>();
 
-    flex((first_line, tasks))
+    flex((first_line, input_box2, tasks))
```
waywardmonkeys: You still have this bit in there...

waywardmonkeys: You got it now.

PR #537, opened 2024-08-24
Make `Textbox` focusable and trigger redraw in response to `request_layout`

### LINKED ISSUE #301: Masonry: Request layout does not implicitly start a redraw
If, when responding to an event, you change state which will be updated in layout, the way you have to handle that is:
```
ctx.request_layout();
ctx.request_paint();
```
e.g.
https://github.com/linebender/xilem/blob/aaeeadbbd75a3984e404c2d545fcf7e8abaa72f0/crates/masonry/src/widget/textbox.rs#L182-L184

Otherwise, the update does not actually get drawn. This is not very ergonomic, and is a significant footgun in the masonry layer that this does not happen. We need to find a solution here. That might just be that `request_layout` also requests a paint; this should be at the right level in the passes

PoignardAzur:
Fixes #301.

### DIFF
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -527,7 +527,11 @@ impl RenderRoot {
         // We request a redraw if either the render tree or the accessibility
         // tree needs to be rebuilt. Usually both happen at the same time.
         // A redraw will trigger a rebuild of the accessibility tree.
-        if self.root_state().needs_paint || self.root_state().needs_accessibility {
+        // TODO - We assume that a relayout will trigger a repaint
+        if self.root_state().needs_paint
+            || self.root_state().needs_accessibility
+            || self.root_state().needs_layout
+        {
             self.state
                 .signal_queue
                 .push_back(RenderRootSignal::RequestRedraw);
--- a/masonry/src/widget/textbox.rs
+++ b/masonry/src/widget/textbox.rs
@@ -236,6 +236,7 @@ impl Widget for Textbox {
             }
             StatusChange::FocusChanged(true) => {
                 // TODO: Focus on first link
+                ctx.request_layout();
             }
             _ => {}
         }
--- a/masonry/src/widget/textbox.rs
+++ b/masonry/src/widget/textbox.rs
@@ -235,6 +235,7 @@ impl Widget for Textbox {
                 // TODO: Stop focusing on any links
             }
             StatusChange::FocusChanged(true) => {
+                self.editor.focus_gained();
                 // TODO: Focus on first link
                 ctx.request_layout();
             }
--- a/masonry/examples/two_textboxes.rs
+++ b/masonry/examples/two_textboxes.rs
@@ -1,0 +1,43 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! This is a very small example to demonstrate tab focus.
+//! It will probably be removed in the future.
+
+// On Windows platform, don't show a console when opening the app.
+#![windows_subsystem = "windows"]
+
+use masonry::app_driver::{AppDriver, DriverCtx};
+use masonry::dpi::LogicalSize;
+use masonry::widget::{Flex, RootWidget, Textbox};
+use masonry::{Action, WidgetId};
+use winit::window::Window;
+
+const VERTICAL_WIDGET_SPACING: f64 = 20.0;
+
+struct Driver;
+
+impl AppDriver for Driver {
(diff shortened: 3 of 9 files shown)


masonry/src/render_root.rs
```diff
@@ -576,7 +576,11 @@ impl RenderRoot {
         // We request a redraw if either the render tree or the accessibility
         // tree needs to be rebuilt. Usually both happen at the same time.
         // A redraw will trigger a rebuild of the accessibility tree.
-        if self.root_state().needs_paint || self.root_state().needs_accessibility {
+        // TODO - We assume that a relayout will trigger a repaint
+        if self.root_state().needs_paint
+            || self.root_state().needs_accessibility
+            || self.root_state().needs_layout
```
PoignardAzur, line 582: Quick comment on this: right now layout is implicitly triggered as part of the paint/accessibility pass: you want to paint something, therefore you lay it out first.

A side-effect of this is that previous to this change, calling request_layout wouldn't call a layout pass until *some* widget somewhere requested a paint pass. (Even though the layout pass automatically requests a paint pass.)

Once pass spec is complete, layout will run with all the other passes whenever anything happens, and this line will be removed.

DJMcNab, line 582: We also need to fix all of the widgets which do:
```rust
ctx.request_layout();
ctx.request_paint();
```

DJMcNab (commented):
It's not clear to me why this isn't marked as fixing https://github.com/linebender/xilem/issues/301
We also need to remove all of the other workarounds to that bug.

The other orthogonal change looks fine.

masonry/src/widget/textbox.rs
```diff
@@ -230,6 +230,7 @@ impl Widget for Textbox {
             }
             StatusChange::FocusChanged(true) => {
                 // TODO: Focus on first link
+                ctx.request_layout();
```
DJMcNab, line 233: Why is this needed? Is it because the cursor is now visible, or some other reason?

It's not clear to me why this can't just be `request_paint`, say.

PoignardAzur: The cursor currently changes visibility in layout, not paint.

I think it's fine to keep it that way until we refactor the text code (which will be pretty soon).

masonry/src/render_root.rs
PoignardAzur: Done.

PoignardAzur:
> We also need to remove all of the other workarounds to that bug.

Anything left besides the ones I just patched?

DJMcNab (approved):
These code changes look good to me.

Is there an example where this can be tested? I don't think any other widget can yet be focused, and `TextBox` eats <kbd>Tab</kbd>

We could probably change the title to include a mention of the new layout behaviour. E.g. to:
```
Make `Textbox` focusable and redraw in response to `request_layout`
```

PoignardAzur:
IIRC Textbox doesn't eat Tab. I just made an example with two textboxes to test this. I guess I could push a similar one.

DJMcNab:
The two textbox case isn't working for me...

PoignardAzur:
I've pushed a few fixes for crashes I saw, and I pushed an example with two textboxes.

Tab switching works with that example for me. If it doesn't for you, that's something we'll have to investigate more in-depth.

PoignardAzur:
(Pinging Philipp since IIRC Daniel is on vacation.)

Philipp-M (approved):
LGTM.

(I'm also vacationing, but sitting in a train right now...)

DJMcNab:
What behaviour do you expect me to be seeing when I press <kbd>Tab</kbd> in two_textboxes? I cannot see a cursor after pressing it at the moment, and typing doesn't work either.

PoignardAzur:
> What behaviour do you expect me to be seeing when I press Tab in two_textboxes? I cannot see a cursor after pressing it at the moment, and typing doesn't work either.

For me the cursor doesn't appear in the second box, but typing does work. It weird that it doesn't for you.

We might need to do a live debugging session.

mwcampbell:
There's some overlap between this branch and the work I just did on my textbox-accesskit-focus branch (before I discovered this branch). You may want to look at what I did on that branch, particularly with the new `TextWithSelection::focus_gained` method, for a resolution to the remaining erratic behavior with tabbing and textboxes.

PoignardAzur:
Will do.

mwcampbell:
Is this supposed to work yet when pressing Tab to focus the first (or only) textbox if nothing has focus yet? That doesn't work for me when running the to_do_mvc example from the latest commit on this branch.

PoignardAzur:
No, I fixed that in #593.

PR #539, opened 2024-08-24
Implement update_anim pass

PoignardAzur:


masonry/src/render_root.rs
```diff
@@ -203,9 +203,13 @@ impl RenderRoot {
                 let last = self.last_anim.take();
                 let elapsed_ns = last.map(|t| now.duration_since(t).as_nanos()).unwrap_or(0) as u64;
                 let root_state = self.root_state();
-                if root_state.request_anim {
-                    root_state.request_anim = false;
-                    self.root_lifecycle(LifeCycle::AnimFrame(elapsed_ns));
+                if root_state.needs_anim {
```
DJMcNab, line 206: I wonder if this check should be inside `run_update_anim_pass`. I can understand why it isn't with `post_event_processing`, but my understanding is that that will be removed?

masonry/src/passes/update.rs
```diff
@@ -145,3 +146,52 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
...
+            widget_state_children: state.children.reborrow_mut(),
+            widget_children: widget.children.reborrow_mut(),
+        };
+        widget
+            .item
+            .lifecycle(&mut ctx, &LifeCycle::AnimFrame(elapsed_ns));
+    }
+
+    state.item.needs_anim = false;
+    state.item.request_anim = false;
```
DJMcNab, line 177: It's not clear to me why this code is here.
Currently, this stops any animation after one frame, if it requests a new frame in the handler.

What is the expected behaviour if an animation wants to last more than one frame?

DJMcNab (commented):
Any examples using animation just doesn't work in this PR...

We could also use a link to linebender/rfcs#7 in this PR's description

masonry/src/render_root.rs
PoignardAzur: I need to write down my thoughts on this (and correct the RFC); The short version is that update_anims doesn't really have the semantics of a "rewrite" pass. It's more of a special event pass.

masonry/src/passes/update.rs
PoignardAzur: Yeah, I think I rushed this PR too much.

masonry/src/render_root.rs
PoignardAzur: Re-reading your question again, I realize I misunderstood it. You're right that the check is a bit superfluous. The pass already does it once per widget anyway.

PoignardAzur:
New version, which I've tested with the variable clocks example. Everything works fine now.

masonry/src/contexts.rs
```diff
@@ -443,6 +443,7 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
     /// Request an animation frame.
     pub fn request_anim_frame(&mut self) {
         trace!("request_anim_frame");
+        self.widget_state.needs_anim = true;
         self.widget_state.request_anim = true;
```
Philipp-M, line 447: I guess we want to either rename the `needs_...` variables to be more clear, or document the difference between `request_...` and `needs_...` well.

At a first glance, the naming suggests a little bit redundancy here, at least it wasn't clear to me what the difference exactly should be.
(But I don't think that this should be done in this PR, and it's probably better to be consistent with the rest of the naming).

Philipp-M (approved):
Code looks good to me, I can't really test this on my ultra-old Notebook, which doesn't seem to support Vulkan properly. Though I get a panic in parley which seems weird to me (as unrelated to Vulkan I guess)? Anyhow this is not related to this PR as I also get this on main.

masonry/src/passes/update.rs
```diff
@@ -145,3 +146,51 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
...
+) {
+    let _span = widget.item.make_trace_span().entered();
+
+    if !state.item.needs_anim {
+        return;
+    }
+    state.item.needs_anim = false;
+
+    if state.item.request_anim {
+        state.item.request_anim = false;
```
DJMcNab, line 166: Maybe we should have a comment here that most passes reset their `needs` and `request` after the call to the widget method, but that it's valid and expected for `request_anim` to be called in response to `AnimFrame`.

masonry/src/render_root.rs
```diff
@@ -202,12 +202,14 @@ impl RenderRoot {
...
-                    self.root_lifecycle(LifeCycle::AnimFrame(elapsed_ns));
-                    self.last_anim = Some(now);
-                }
+
+                run_update_anim_pass(self, elapsed_ns);
+
+                let mut root_state = self.widget_arena.get_state_mut(self.root.id()).item.clone();
+                self.post_event_processing(&mut root_state);
+
+                self.last_anim = Some(now);
```
DJMcNab, line 211: I don't think this will have the behaviour we want. Consider for example:
```patch
diff --git a/xilem/examples/variable_clock.rs b/xilem/examples/variable_clock.rs
index 8bead4d80..ea7dd599f 100644
--- a/xilem/examples/variable_clock.rs
+++ b/xilem/examples/variable_clock.rs
@@ -60,7 +60,7 @@ fn app_logic(data: &mut Clocks) -> impl WidgetView<Clocks> {
                     };
                 }
             },
-            |data: &mut Clocks, ()| data.now_utc = OffsetDateTime::now_utc(),
+            |data: &mut Clocks, ()| data.weight = (data.weight + 100.) % 1000.,
         ),
     )
 }
```
This will not animate smoothly, but will instead to jump to the full value, because when the animation frame occurs, the time will be the time since the last frame, rather than `0` as it is before this PR.

DJMcNab (dismissed):
This still doesn't work as expected, unfortunately.

masonry/src/contexts.rs
PoignardAzur: > I guess we want to either rename the `needs_...` variables to be more clear, or document the difference between `request_...` and `needs_...` well.

Right now the fields all have doc comments that follow this pattern:

- **request_xxx:** "This widget explicitly requested XXX"
- **needs_xxx:** "This widget or a descendant explicitly requested XXX"

Once we do a documentation pass I'll explicitly note the pattern as well, at least in WidgetState and `ARCHITECTURE.md`. If you think there's other places we could cover, I'm interested. This is the sort of pattern where you want redundancy in documenting it.

Although note that it's strictly an internal naming convention that library users won't be exposed to, so it limits the places we're likely to document it in.

I do like those names. I think they're concise, which is valuable, and the "request/needs" difference feels to me like it's easy to remember once it's been explained.

masonry/src/passes/update.rs
PoignardAzur: Done.

masonry/src/render_root.rs
PoignardAzur: Would you mind taking responsibility for this PR, then? It seems like this is more in your wheelhouse than mine, if you're currently working on frame pacing.

DJMcNab: Can do. I think this only needs a very small change to be ready to land.

PoignardAzur:
I'm going to do a small tweak on your changes, pushing now.

PoignardAzur:
I think this is ready to merge, it's on the critical path, and Daniel's concerns are essentially addressed.

I think Phillip is still on vacation and isn't going to review this week. Either way, unless a blocker comes up, I'll merge this tomorrow.

masonry/src/widget/widget_pod.rs
```diff
@@ -337,12 +337,10 @@ impl<W: Widget> WidgetPod<W> {
...
                 true
             }
-            LifeCycle::AnimFrame(_) => {
-                state.request_anim = false;
-                true
-            }
             // Routing DisabledChanged has been moved to the update_disabled pass
             LifeCycle::DisabledChanged(_) => false,
+            // Animations have been moved to the update_anim pass
+            LifeCycle::AnimFrame(_) => false,
```
DJMcNab, line 343: Is this `unreachable!`?

DJMcNab (approved):

masonry/src/widget/widget_pod.rs
PoignardAzur: Not quite. A container widget could request anim. If it recurses the lifecycle function over children, then that line of code is reached.

PR #540, opened 2024-08-24
Implement update_disabled pass

PoignardAzur:


PoignardAzur:
Woohoo, one more! At this point the only pass left is update_scrolls. There still a lot of work after that to complete the pass spec rework, though.

masonry/src/contexts.rs
```diff
@@ -309,7 +309,7 @@ impl_context_method!(
         ///
         /// [`set_disabled`]: EventCtx::set_disabled
         pub fn is_disabled(&self) -> bool {
-            self.widget_state.is_disabled()
+            self.widget_state.is_disabled
```
Philipp-M, line 312: I mean that's one of the nittiest nits: Why this change (and at the other places too)?
I feel like an abstraction i.e. function call potentially allows in the future to wrap more stuff in it, while this would need to be refactored again to such a function call when that would be the case, and is otherwise optimized anyways.

masonry/src/widget/widget_state.rs
```diff
@@ -124,12 +119,11 @@ pub struct WidgetState {
     pub(crate) text_registrations: Vec<TextFieldRegistration>,
 
     // --- STATUS ---
-    // `true` if one of our ancestors is disabled (meaning we are also disabled).
-    pub(crate) ancestor_disabled: bool,
 
-    // `true` if this widget has been explicitly disabled.
-    // A widget can be disabled without being *explicitly* disabled if an ancestor is disabled.
+    // This widget has been disabled.
```
Philipp-M, line 123: This should probably be a doc-comment
```suggestion
    /// This widget has been disabled.
```

masonry/src/widget/widget_state.rs
```diff
@@ -124,12 +119,11 @@ pub struct WidgetState {
...
     pub(crate) is_explicitly_disabled: bool,
+    // This widget or an ancestor has been disabled.
```
Philipp-M, line 125:
```suggestion
    /// This widget or an ancestor has been disabled.
```

Philipp-M (approved):
Thanks, looks good apart of nits.

> There still a lot of work after that to complete the pass spec rework

(Yeah like documenting a lot :see_no_evil: )

masonry/src/contexts.rs
PoignardAzur: I try to avoid function calls when not necessary. Even if they're inlined, there's still a cognitive penalty of not knowing how much / which code the function executes.

Someone trying to understand this code and reading the function version needs (1) to jump to the `is_disabled()` function definition, then (2) look for code setting the value of the `is_disabled` field. If you replace the function with its field access, that someone only needs to do part 2.

masonry/src/passes/update.rs
```diff
@@ -145,3 +146,59 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
...
+fn update_disabled_for_widget(
+    global_state: &mut RenderRootState,
+    mut widget: ArenaMut<'_, Box<dyn Widget>>,
+    mut state: ArenaMut<'_, WidgetState>,
+    parent_disabled: bool,
+) {
+    let _span = widget.item.make_trace_span().entered();
+    let id = state.item.id;
+
+    let disabled = state.item.is_explicitly_disabled | parent_disabled;
```
DJMcNab, line 161: I don't like using `BitOr` for bool. This seems an odd thing to try and optimise.

masonry/src/passes/update.rs
```diff
@@ -145,3 +146,59 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
...
+            .lifecycle(&mut ctx, &LifeCycle::DisabledChanged(disabled));
+        state.item.is_disabled = disabled;
+    }
+
+    state.item.needs_update_disabled = false;
+
+    if disabled && global_state.next_focused_widget == Some(id) {
+        // This may get overwritten. That's ok, because either way the
+        // focused widget, if there's one, won't be disabled.
+        global_state.next_focused_widget = None;
```
DJMcNab, line 184: If a widget is forcefully disabled, I'd expect pressing <kbd>Tab</kbd> next to focus what the next widget would be, rather than restarting at the beginning. It's hard to get a feel for this, though - I think this behaviour is OK at the moment, as it will change when we do more careful focus handling.

masonry/src/passes/update.rs
Philipp-M: Oh yeah you're right, booleans should just use `||`. I actually wonder why it's possible at all in rust (on booleans I mean).

Philipp-M: Btw. I think we should use the `needless_bitwise_bool` clippy lint?

DJMcNab: Agreed. Are you happy to PR that?

Philipp-M: Can do

DJMcNab: Discussion continues in https://xi.zulipchat.com/#narrow/stream/419691-linebender/topic/Standard.20Lint.20set

PR #541, opened 2024-08-25
Enable BackgroundBrush in SizedBox view

RagibHasin:
The `SizedBox` widget in Masonry had the option to set a background brush that was not exposed in its corresponding view in Xilem. The `SizedBox` view now exposes the option.

Incidentally, `BackgroundBrush::PainterFn` variant had the wrong type, as such it was not usable. A `scene` parameter is added and the closure now requires to be `Send + Sync`.

### DIFF
--- a/masonry/src/widget/sized_box.rs
+++ b/masonry/src/widget/sized_box.rs
@@ -7,7 +7,7 @@ use accesskit::Role;
 use smallvec::{smallvec, SmallVec};
 use tracing::{trace, trace_span, warn, Span};
 use vello::kurbo::{Affine, RoundedRectRadii};
-use vello::peniko::{BlendMode, Color, Fill, Gradient};
+use vello::peniko::{BlendMode, Brush, Color, Fill};
 use vello::Scene;
 
 use crate::paint_scene_helpers::{fill_color, stroke};
@@ -18,16 +18,6 @@ use crate::{
 };
 
 // FIXME - Improve all doc in this module ASAP.
-
-/// Something that can be used as the background for a widget.
-#[non_exhaustive]
-#[allow(missing_docs)]
-#[allow(clippy::type_complexity)]
-pub enum BackgroundBrush {
-    Color(Color),
-    Gradient(Gradient),
-    PainterFn(Box<dyn FnMut(&mut PaintCtx)>),
-}
 
 /// Something that can be used as the border for a widget.
 struct BorderStyle {
@@ -51,7 +41,7 @@ pub struct SizedBox {
     child: Option<WidgetPod<Box<dyn Widget>>>,
     width: Option<f64>,
     height: Option<f64>,
-    background: Option<BackgroundBrush>,
+    background: Option<Brush>,
     border: Option<BorderStyle>,
     corner_radius: RoundedRectRadii,
 }
@@ -154,9 +144,11 @@ impl SizedBox {
 
     /// Builder-style method for setting the background for this widget.
     ///
-    /// This can be passed anything which can be represented by a [`BackgroundBrush`];
-    /// notably, it can be any [`Color`], any gradient, or a fully custom painter `FnMut`.
-    pub fn background(mut self, brush: impl Into<BackgroundBrush>) -> Self {
+    /// This can be passed anything which can be represented by a [`Brush`];
+    /// notably, it can be any [`Color`], any gradient, or an [`Image`].
+    ///
+    /// [`Image`]: vello::peniko::Image
+    pub fn background(mut self, brush: impl Into<Brush>) -> Self {
         self.background = Some(brush.into());
         self
     }
@@ -234,9 +226,11 @@ impl WidgetMut<'_, SizedBox> {
 
     /// Set the background for this widget.
     ///
-    /// This can be passed anything which can be represented by a [`BackgroundBrush`];
-    /// notably, it can be any [`Color`], any gradient, or a fully custom painter `FnMut`.
-    pub fn set_background(&mut self, brush: impl Into<BackgroundBrush>) {
+    /// This can be passed anything which can be represented by a [`Brush`];
+    /// notably, it can be any [`Color`], any gradient, or an [`Image`].
+    ///
+    /// [`Image`]: vello::peniko::Image
+    pub fn set_background(&mut self, brush: impl Into<Brush>) {
         self.widget.background = Some(brush.into());
         self.ctx.request_paint();
     }
@@ -371,7 +365,7 @@ impl Widget for SizedBox {
 
             trace_span!("paint background").in_scope(|| {
                 scene.push_layer(BlendMode::default(), 1., Affine::IDENTITY, &panel);
-                background.paint(ctx, scene);
+                paint(background, ctx, scene);
                 scene.pop_layer();
             });
         }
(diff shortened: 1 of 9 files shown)


masonry/src/widget/sized_box.rs
```diff
@@ -26,7 +26,7 @@ use crate::{
 pub enum BackgroundBrush {
     Color(Color),
     Gradient(Gradient),
-    PainterFn(Box<dyn FnMut(&mut PaintCtx)>),
+    PainterFn(Box<dyn FnMut(&mut PaintCtx, &mut Scene) + Send + Sync>),
```
PoignardAzur, line 29: I think you could spare yourself from having to write mutex code by making PainterFn cloneable and deriving Clone on BackgroundBrush.

```suggestion
    PainterFn(Arc<dyn Fn(&mut PaintCtx, &mut Scene) + Send + Sync>),
```

Philipp-M: Yeah, I've thought about something similar, not sure if `FnMut` is really necessary here. I'd like to avoid Mutex locking in views as long as possible (I imagine there are few cases though where it's *really* necessary)

masonry/src/widget/sized_box.rs
```diff
@@ -26,7 +26,7 @@ use crate::{
 pub enum BackgroundBrush {
```
DJMcNab, line 26: My inclination would be to remove this type entirely, and just use a `Brush` from Peniko.

DJMcNab (commented):
This is clearly pretty far from idiomatic code. I do think it's important to design our fully-custom drawing, but the way to do that isn't to sneak it in like this.

masonry/src/widget/sized_box.rs
Philipp-M: Yeah, agree, I just see that `PaintCtx` is "leaked" in xilem with the `PainterFn`, which I don't think should be a thing?

RagibHasin: As it stands, I've been using the `PaintFn` variant specifically to show progress behind a nested view, like a watermark.

RagibHasin: So, if we remove the `PaintFn` variant what would be the idiomatic way to show a view over another?

xilem/src/view/sized_box.rs
```diff
@@ -22,13 +25,15 @@ where
...
         width: None,
+        background: Mutex::default(),
     }
 }
 
 pub struct SizedBox<V> {
     inner: V,
     width: Option<f64>,
     height: Option<f64>,
+    background: Mutex<Option<BackgroundBrush>>,
```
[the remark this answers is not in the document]
RagibHasin: I also am not happy with the mutex. But, I've avoided blocking wait anywhere and just used `try_lock`, as I don't think it would ever fail.
(Handled the failing cases though. No panics 😊)

masonry/src/widget/sized_box.rs
Philipp-M: Isn't something like #553 more appropriate for such cases (when we have appropriate lay(er)outing possibilities)? (Or the direction I would rather like to go: having views similar as in the xilem_web svgtoy example)

RagibHasin: Yes, #553 would enable creating an ad-hoc watermark view. I was wondering about laying it out behind another view.

masonry/src/widget/sized_box.rs
RagibHasin: Busy waiting for a lock is not used here. I used `try_lock` and defaulted to existing background if the lock cannot be acquired immediately. But, I think the mutex can't be already locked in practice as we do not build/rebuild concurrently?

RagibHasin:
Is replacing `BackgroundBrush` with `peniko::Brush` is the way forward? If so, then I would take the PR to that route, but it would leave out my use case. In that case, an absolutely positioned `Canvas` container would be the most versatile solution.

DJMcNab:
I'm certainly not likely to approve this PR except for in the `Peniko::Brush` form. We do definitely need to solve the "z-layering" problem, but this isn't the way to do it.

I'd suggest making a PR for an extremely simple widget which layers one widget on top of a base widget, with all sizing of the "stack" widget based on the base widget.
The base widget wouldn't really be able to have interactivity, but that would still meet the case that you are using this for, as far as I can tell

RagibHasin:
Yes, that would be the most minimalistic solution without using BackgroundBrush.

RagibHasin:
This latest commit reworks the widget to replace `BackgroundBrush` with `peniko::Brush` and also exposes the background, border, and corner radius from the view.

I tried adding a test for background and noticed that the test `label_box_with_size` was overwriting the snapshot of the test `label_box_no_size`.

But I couldn't figure out how to populate the snapshot initially, which is the reason of the CI failure.

@PoignardAzur, can you look into the tests?

masonry/src/widget/sized_box.rs
```diff
@@ -416,39 +410,19 @@ impl Widget for SizedBox {
...
+        Brush::Solid(color) => fill_color(scene, &bounds, *color),
+        Brush::Gradient(grad) => scene.fill(
+            Fill::NonZero,
+            Affine::IDENTITY,
+            grad,
+            Some(Affine::IDENTITY),
+            &bounds,
+        ),
+        Brush::Image(image) => scene.draw_image(image, Affine::IDENTITY),
     }
```
DJMcNab, line 426: Why not use?
```suggestion
    scene.fill(
        Fill::NonZero,
        Affine::IDENTITY,
        brush,
        Some(Affine::IDENTITY),
        &bounds,
    );
```
I also don't really think this needs a helper function.

xilem/src/view/sized_box.rs
```diff
@@ -4,13 +4,22 @@
...
     core::{Mut, View, ViewId},
     Pod, ViewCtx, WidgetView,
 };
 
+/// Something that can be used as the border for a widget.
+#[derive(PartialEq)]
+struct BorderStyle {
+    width: f64,
+    color: Color,
+}
```
DJMcNab, line 21: Since this isn't public API, we should move this to below `SizedBox`

DJMcNab (approved):
Other than not applying the new tests, this is looking good. A few additional suggestions

To apply the tests, you're expected to rename the `.new.png` image to not have `.new`.
I do intend to improve the UX there, but haven't gotten around to it yet.

masonry/src/widget/sized_box.rs
RagibHasin: Yeah, wanted to clarify that.

xilem/src/view/sized_box.rs
RagibHasin: Done.

masonry/src/widget/sized_box.rs
RagibHasin: Done.

RagibHasin:
> Other than not applying the new tests, this is looking good. A few additional suggestions
> 
> To apply the tests, you're expected to rename the `.new.png` image to not have `.new`. I do intend to improve the UX there, but haven't gotten around to it yet.

Yeah, documenting it somewhere would be helpful for first-time contributors.

DJMcNab:
My intention is to apply the UX seen in linebender/vello#610, where you can use environment variables to update the tests.

masonry/src/widget/sized_box.rs
```diff
@@ -371,7 +365,13 @@ impl Widget for SizedBox {
 
             trace_span!("paint background").in_scope(|| {
                 scene.push_layer(BlendMode::default(), 1., Affine::IDENTITY, &panel);
```
DJMcNab, line 367:
```suggestion
```

masonry/src/widget/sized_box.rs
```diff
@@ -371,7 +365,13 @@ impl Widget for SizedBox {
...
-                background.paint(ctx, scene);
+                scene.fill(
+                    Fill::NonZero,
+                    Affine::IDENTITY,
+                    &*background,
+                    Some(Affine::IDENTITY),
+                    &ctx.size().to_rect(),
+                );
                 scene.pop_layer();
```
DJMcNab, line 375:
```suggestion
                scene.fill(
                    Fill::NonZero,
                    Affine::IDENTITY,
                    &*background,
                    Some(Affine::IDENTITY),
                    &panel,
                );
```

DJMcNab (commented):
A slight tweak which I think is equivalent, but more efficient.

Once these are actioned and checked, please confirm that you're ready for this to land and I'll merge.

masonry/src/widget/sized_box.rs
RagibHasin: So, we don't need a new layer for the background now?

RagibHasin:
I think it is now good to go.

masonry/src/widget/sized_box.rs
DJMcNab: Yeah - the layer was previously needed because of the paint function, but because it's just a single item, this works as expected.

PR #545, opened 2024-08-25
masonry: Add some missing backticks.

waywardmonkeys:


waywardmonkeys:
I'm aware that some of this is outdated documentation, but I'm cleaning out all of the issues that prevent us from re-enabling `clippy::doc_markdown`.

jplatte (approved):

PR #515, opened 2024-08-15
De-traitify text module

PoignardAzur:

Issue #424, opened 2024-07-04
Xilem's to_do_mvc enters an apparent infinite loop (depends on specifics of IME, blocked on rework of pass order)


matta:
Xilem's to_do_mvc example (`xilem/examples/to_do_mvc.rs`) seems to go into a loop when I click on the text edit.

This is on Debian, running from Xilem's current head (e76cf31258c79d263c5669be052353c3e99c012d).

1. cd `<repo root>`
1. cd xilem
1. cargo run --example to_do_mvc
1. Click on the text input.

CPU usage for the app goes up to about 40%, and the log spews this seemingly forever:

```text
17:55:17.227Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.227Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.244Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.244Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.260Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.260Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.277Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.277Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.294Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.294Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.310Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.310Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.327Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.327Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
17:55:17.343Z DEBUG text_event: Running ON_TEXT_EVENT pass with Ime::Preedit("")
17:55:17.344Z DEBUG text_event: ON_TEXT_EVENT finished focused_widget=1 handled=true
[...]

PR #550, opened 2024-08-25
Implement update_scrolls pass

PoignardAzur:
This is part of the Pass Specification RFC: https://github.com/linebender/rfcs/pull/7

Note: This PRs comes with a lot of new TODO items. Addressing most of these items is difficult without major refactors, because Portal code deals with accessing values across multiple widgets, which is still hard to do elegantly.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -205,6 +205,11 @@ pub(crate) fn run_update_disabled_pass(root: &mut RenderRoot) {
 
 // ----------------
 
+// This pass will update scroll positions in cases where a widget has requested to be
+// scrolled into view (usually a textbox getting text events).
+// Each parent that implements scrolling will update its scroll position to ensure the
+// child is visible. (If the target area is larger than the parent, the parent will try
+// to show the top left of that area.)
 pub(crate) fn run_update_scroll_pass(root: &mut RenderRoot) {
     let _span = info_span!("update_scroll").entered();
 
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -202,3 +202,26 @@ pub(crate) fn run_update_disabled_pass(root: &mut RenderRoot) {
     let (root_widget, root_state) = root.widget_arena.get_pair_mut(root.root.id());
     update_disabled_for_widget(&mut root.state, root_widget, root_state, false);
 }
+
+// ----------------
+
+pub(crate) fn run_update_scroll_pass(root: &mut RenderRoot) {
+    let _span = info_span!("update_scroll").entered();
+
+    let scroll_request_targets = std::mem::take(&mut root.state.scroll_request_targets);
+    for (target, rect) in scroll_request_targets {
+        let mut target_rect = rect;
+
+        run_targeted_update_pass(root, Some(target), |widget, ctx| {
+            let event = LifeCycle::RequestPanToChild(rect);
+            widget.lifecycle(ctx, &event);
+
+            // TODO - We should run the compose method after this, so
+            // translations are updated and the rect passed to parents
+            // is more accurate.
+
+            let state = &ctx.widget_state;
+            target_rect = target_rect + state.translation + state.origin.to_vec2();
+        });
+    }
+}
--- a/masonry/src/widget/portal.rs
+++ b/masonry/src/widget/portal.rs
@@ -339,9 +339,24 @@ impl<W: Widget> Widget for Portal<W> {
                 let portal_size = ctx.size();
                 let content_size = ctx.get_raw_ref(&mut self.child).ctx().layout_rect().size();
 
-                // FIXME - Update scrollbars
                 self.pan_viewport_to_raw(portal_size, content_size, *target);
                 ctx.request_compose();
+
+                // TODO - There's a lot of code here that's duplicated from the `MouseWheel`
+                // event in `on_pointer_event`.
+                // Because this code directly manipulates child widgets, it's hard to factor
+                // it out.
+                let mut scrollbar = ctx.get_raw_mut(&mut self.scrollbar_vertical);
+                scrollbar.widget().cursor_progress =
+                    self.viewport_pos.y / (content_size - portal_size).height;
+                scrollbar.ctx().request_paint();
+
+                std::mem::drop(scrollbar);
+
+                let mut scrollbar = ctx.get_raw_mut(&mut self.scrollbar_horizontal);
+                scrollbar.widget().cursor_progress =
+                    self.viewport_pos.x / (content_size - portal_size).width;
+                scrollbar.ctx().request_paint();
             }
             _ => {}
         }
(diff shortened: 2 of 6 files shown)


PoignardAzur:
And that's the last pass implemented!

Once we've merged all of them, the next step will be to replace `RenderRoot::post_event_processing` with something like `RenderRoot::run_rewrite_passes`, following the logic outlined in the pass spec RFC.

masonry/src/passes/update.rs
```diff
@@ -145,3 +145,26 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
...
+    let _span = info_span!("update_scroll").entered();
+
+    let scroll_request_targets = std::mem::take(&mut root.state.scroll_request_targets);
+    for (target, rect) in scroll_request_targets {
+        let mut target_rect = rect;
+
+        run_targeted_update_pass(root, Some(target), |widget, ctx| {
+            let event = LifeCycle::RequestPanToChild(rect);
+            widget.lifecycle(ctx, &event);
+
```
DJMcNab, line 161: How will smooth scroll would be handled?

The best logic I can see if to "get where the widget *will be*", and all the parents scroll to that.

masonry/src/passes/update.rs
```diff
@@ -145,3 +145,26 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
...
     root_state.merge_up(root.widget_arena.get_state_mut(root.root.id()).item);
 }
+
+// ----------------
+
+pub(crate) fn run_update_scroll_pass(root: &mut RenderRoot) {
+    let _span = info_span!("update_scroll").entered();
+
+    let scroll_request_targets = std::mem::take(&mut root.state.scroll_request_targets);
+    for (target, rect) in scroll_request_targets {
```
DJMcNab, line 155: Does the order matter here? Is it meaningful to document that?

masonry/src/widget/portal.rs
```diff
@@ -141,6 +141,31 @@ impl<W: Widget> Portal<W> {
             false
         }
     }
+
+    // Note - Rect is in child coordinates
+    // TODO - Merge with pan_viewport_to
+    // Right now these functions are just different enough to be a pain to merge.
+    pub fn pan_viewport_to_raw(
```
DJMcNab, line 148: This is pub, so needs a doc comment.

masonry/src/widget/portal.rs
```diff
@@ -141,6 +141,31 @@ impl<W: Widget> Portal<W> {
...
+        let new_pos_x = compute_pan_range(
+            viewport.min_x()..viewport.max_x(),
+            target.min_x()..target.max_x(),
+        )
+        .start;
+        let new_pos_y = compute_pan_range(
+            viewport.min_y()..viewport.max_y(),
+            target.min_y()..target.max_y(),
+        )
+        .start;
```
DJMcNab, line 165: So we prefer the `top-left` of the rectangle to go into view. That seems reasonable, but should probably be documented.

DJMcNab (approved):
Is there any way in the current examples to test this?

Actual code changes look reasonable, presuming that the panning maths is correct

masonry/src/passes/update.rs
PoignardAzur: The order doesn't matter. `scroll_request_target` was initially an option, and I turned it to a Vec just to cover hypothetical cases where someone would want to pan *several* scroll areas at once.

I guess several targets could interfere if they're in the same scroll region. I think this will be rare enough that we won't need to document it.

PoignardAzur:
>Is there any way in the current examples to test this?

The `to_do_list` example.

masonry/src/passes/update.rs
DJMcNab: I don't mean on a touchscreen, I mean scrolling to a position over a certain amount of time.

Right, yeah. This logic would have to change at that point, but it's also deliberately unaddressed at the moment.

DJMcNab:
> The `to_do_list` example.

Thanks, checked it out and I can see that it is now impossible to scroll in that example.

PoignardAzur:
> Thanks, checked it out and I can see that it is now impossible to scroll in that example.

That doesn't sound right, it worked when I tested it. Maybe I forgot to push some changes? Will look into it.

DJMcNab:
Ah, this is #424

masonry/src/widget/portal.rs
```diff
@@ -311,8 +336,14 @@ impl<W: Widget> Widget for Portal<W> {
             LifeCycle::WidgetAdded => {
                 ctx.register_as_portal();
             }
-            //TODO
-            //LifeCycle::RequestPanToChild(target_rect) => {}
+            LifeCycle::RequestPanToChild(target) => {
+                let portal_size = ctx.size();
+                let content_size = ctx.get_raw_ref(&mut self.child).ctx().layout_rect().size();
+
+                // FIXME - Update scrollbars
```
DJMcNab, line 343: We need to do this in this PR, or explain why it's hard. The behaviour without that is quite bad.

masonry/src/widget/textbox.rs
```diff
@@ -207,6 +207,8 @@ impl Widget for Textbox {
         let result = self.editor.text_event(ctx, event);
         // If focused on a link and enter pressed, follow it?
         if result.is_handled() {
+            // TODO - Use request_scroll_to with cursor rect
```
DJMcNab, line 210: I imagine this will happen in #515 ?

DJMcNab (commented):
With the workaround mentioned above, this now works well.

Some remaining concerns are documenting the preferred part of the rectangle which is shown, and fixing the scrollbars properly.

masonry/src/widget/textbox.rs
PoignardAzur: Probably.

masonry/src/widget/portal.rs
PoignardAzur: It's not supposed to be pub. Fixed.

masonry/src/widget/portal.rs
PoignardAzur: Done.

masonry/src/widget/portal.rs
PoignardAzur: I added in a comment on the pass function. I'll do a more thorough documentation pass later.

masonry/src/widget/textbox.rs
```diff
@@ -205,8 +206,13 @@ impl Widget for Textbox {
 
     fn on_text_event(&mut self, ctx: &mut EventCtx, event: &TextEvent) {
         let result = self.editor.text_event(ctx, event);
-        // If focused on a link and enter pressed, follow it?
         if result.is_handled() {
+            // Some platforms will send a lot of spurious Preedit events.
+            // We only want to request a scroll on user input.
+            if !matches!(event, TextEvent::Ime(Ime::Preedit(..))) {
```
DJMcNab, line 212: Ooh, apparently we can do
```suggestion
            if !matches!(event, TextEvent::Ime(Ime::Preedit(preedit, ..)) if preedit.is_empty()) {
```
which in my testing has better behaviour (but is not the easiest to reason about in its current form)

DJMcNab (approved):
One small remaining proposed tweak.

PR #553, opened 2024-08-25
Implement basic Canvas widget

PoignardAzur:


DJMcNab (commented):
Agreed, as mentioned on the Zulip, I think a more SVG-esque canvas would be better.

That probably would want support for a raw scene as a contained item, which would effectively be this widget. But the `Canvas` name isn't really the right name for it - this is more like a `VelloRegion` or something.

jaredoconnell:
There was some debate in Zulip about the naming and what defines a canvas. It sounds like this should be potentially called a VelloCanvas or VelloScene. The thing other people were asking for was a Widget Canvas. We may want to make a widget canvas into two things: a relative or absolute layout, and a portal to allow the zoom and related features.

PoignardAzur:
Closing this for now. It's generating way too much debate for how low a priority it is. We can always re-examine what counts as a canvas and what we want in our widget set later.

PR #554, opened 2024-08-26
Fix inference of wrapper views

RagibHasin:


RagibHasin (commented):
Fixes `sized_box`. I didn't find any other wrapper views. Is there any?

DJMcNab (approved):
This looks as expected, thanks.

Does this not fix the inference in the `state_machine` example?
If not, why not?

DJMcNab:
Thank you!

Philipp-M:
I wonder if this would also be necessary for non-wrapping views, but I don't think it is? IIRC I only had issues yet when views either contain a `ViewSequence` or a  `View`.

PR #555, opened 2024-08-26
masonry: Use `||` instead of `|` in update pass for boolean expression

Philipp-M:
Unfortunately this lint is very conservative in its detection, it hasn't detected the other fix in this PR...
I manually checked, if we had any other bitwise or/and for booleans, and indeed this was the only case.


DJMcNab (approved):
~~What is the other fix you mention? I only see one here~~

Oh I see. This is https://github.com/rust-lang/rust-clippy/pull/7133#issuecomment-827174202

Cargo.toml
```diff
@@ -58,6 +58,7 @@ clippy.large_include_file = "warn"
 clippy.large_stack_arrays = "warn"
 clippy.mismatching_type_param_order = "warn"
 clippy.missing_fields_in_debug = "warn"
+clippy.needless_bitwise_bool = "warn"
```
DJMcNab, line 61:
```suggestion
```
Given that the lint is functionally useless, I'd just do the targeted fix

Philipp-M: I mean I wouldn't say useless, it would catch cases such as:
```rust
if true & false {
}
```

But it could of course be better I think.

DJMcNab: I'm not sure it even will catch that, because those RHSed aren't potentially expensive.

Philipp-M: Oh right, it doesn't, thought I tested this, it works with (which I think is what I tested)
```rust
if true & !false {}
```

So yeah not really useful in those rare cases where it triggers, the last comment in the linked issue with something like:

```rust
fn foo2(x1: i32, x2: i32) {
    if (x1 < x2) & (x1 != 0) {}
}
```
doesn't even trigger this lint. Alright in that case I'm focusing this to the actual fix...

PR #557, opened 2024-08-26
Remove manual recursion to child `paint` and `accessibility`

tomcur:
Recursing is done inside the paint and accessibility passes since ff7635e4c281ff381433728bd070323347638f29. I believe this is the correct continuation of #522, with the removal of these methods "left for later" as mentioned in https://github.com/linebender/xilem/pull/522#issuecomment-2298610203.

One note is that Flex now debug-paints its baseline under its children, rather than over them.

### DIFF
--- a/masonry/src/widget/tests/safety_rails.rs
+++ b/masonry/src/widget/tests/safety_rails.rs
@@ -16,9 +16,6 @@ fn make_parent_widget<W: Widget>(child: W) -> ModularWidget<WidgetPod<W>> {
             ctx.place_child(child, Point::ZERO);
             size
         })
-        .paint_fn(move |child, ctx, scene| {
-            child.paint(ctx, scene);
-        })
         .children_fn(|child| smallvec![child.id()])
 }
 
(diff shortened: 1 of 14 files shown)


masonry/src/widget/tests/safety_rails.rs
```diff
@@ -16,9 +16,6 @@ fn make_parent_widget<W: Widget>(child: W) -> ModularWidget<WidgetPod<W>> {
             ctx.place_child(child, Point::ZERO);
             size
         })
-        .paint_fn(move |child, ctx, scene| {
-            child.paint(ctx, scene);
-        })
```
tomcur, line 21: Should this get at child paint (and maybe accessibility) in some other way?

DJMcNab: I don't think so - all that this paint implementation did was paint the child, which is now implicit/always the behaviour.

I think that this is fine.

DJMcNab (approved):
This looks as expected, thanks.

I've done some thinking on the modular widget thing, and I think this is the correct translation. Thanks

PR #558, opened 2024-08-26
xilem: Add a way to allow `impl WidgetViewSequence` as `<seq>` in `flex(<seq>)`

Philipp-M:
(I'll add doc-tests/comments before (potentially) merging, tests would probably also be useful, but we don't have such infrastructure yet...)


DJMcNab (commented):
I don't have the energy to reason about this at the moment, but it looks reasonable.

We should add a `diagnostic::on_unimplemented` on `FlexSequence` which uses this.

Issue #559, opened 2024-08-27
Extra gap is added at the end of a Flex


Keshyu:
## Preview
<img width="251" alt="image" src="https://github.com/user-attachments/assets/c185d8c7-9bd2-4cd7-88c0-029d8b85e55c">

## Code
```rust
flex(
    sized_box(
        flex((
            sized_box(label("text"))
                .width(300.)
                .border(Color::WHITE, 2)
        ))
        .direction(Axis::Vertical)
        .gap(32.),
    )
    .border(Color::MAGENTA, 2),
)
.main_axis_alignment(MainAxisAlignment::Center)
```

## With `.gap(0.)`
<img width="263" alt="image" src="https://github.com/user-attachments/assets/639f2968-9cb5-4f9a-a99a-23c344e1db61">

PR #560, opened 2024-08-28
Fix the handling of `Flex::gap`

DJMcNab:
Fixes https://github.com/linebender/xilem/issues/559

See comment on the fix for an explanation. Essentially, the `major` value was used for multiple purposes, and this corrects into the expected "direction".


Philipp-M (approved):
LGTM

PR #561, opened 2024-08-28
Add the `Portal` view to Xilem

DJMcNab:
This is also added to `variable_clocks` as needed.

This could also be added to `mason`, but I see a worrying amount of lag if I do...
I could do with some help tracking this down.

### DIFF
--- a/xilem/examples/variable_clock.rs
+++ b/xilem/examples/variable_clock.rs
@@ -13,7 +13,8 @@ use time::{error::IndeterminateOffset, macros::format_description, OffsetDateTim
 use winit::error::EventLoopError;
 use xilem::{
     view::{
-        button, flex, label, prose, sized_box, task, variable_label, Axis, FlexExt, FlexSpacer,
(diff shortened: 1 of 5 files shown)


xilem/examples/variable_clock.rs
```diff
@@ -43,8 +44,11 @@ fn app_logic(data: &mut Clocks) -> impl WidgetView<Clocks> {
         FlexSpacer::Fixed(40.),
         local_time(data),
         controls(),
-        // TODO: When we get responsive layouts, move this into a two-column view.
-        TIMEZONES.iter().map(|it| it.view(data)).collect::<Vec<_>>(),
+        portal(flex(
+            // TODO: When we get responsive layouts, move this into a two-column view on desktop.
+            TIMEZONES.iter().map(|it| it.view(data)).collect::<Vec<_>>(),
+        ))
+        .flex(1.),
```
DJMcNab, line 51: This is unfortunate, but it does make this work properly

xorgy (approved):

PR #563, opened 2024-08-29
masonry: fix updating Portal on scrollbar drag

tomcur:
Fixes scrolling `Portal` on scrollbar drag by recomposing instead of relayouting.

Regression probably caused in 59ee615651266421ffe65dadee6d16663f056c9d (https://github.com/linebender/xilem/pull/488) or ff7635e4c281ff381433728bd070323347638f29 (https://github.com/linebender/xilem/pull/522).

PR #564, opened 2024-08-29
masonry: replace `set_active` and `is_active` with pointer capture

tomcur:
Also improve documentation of pointer capture.

Continuation of 59ee615651266421ffe65dadee6d16663f056c9d (https://github.com/linebender/xilem/pull/488).

Makes `has_pointer_capture` available on all context types except `LayoutCtx`, like `is_active` used to be.

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -600,7 +600,16 @@ impl EventCtx<'_> {
     /// captured the pointer and another widget captures it, the first widget loses the pointer
     /// capture.
     ///
+    /// # Releasing the pointer
+    ///
+    /// Any widget can [`release`] the pointer during any event. The pointer is automatically
+    /// released after handling of a [`PointerUp`] or [`PointerLeave`] event completes. A widget
+    /// holding the pointer capture will be the target of these events.
+    ///
     /// [`PointerDown`]: crate::PointerEvent::PointerDown
+    /// [`PointerUp`]: crate::PointerEvent::PointerUp
+    /// [`PointerLeave`]: crate::PointerEvent::PointerLeave
+    /// [`release`]: Self::release_pointer
     #[track_caller]
     pub fn capture_pointer(&mut self) {
         debug_assert!(
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -232,7 +232,7 @@ impl_context_method!(
 
 // --- MARK: GET STATUS ---
 // Methods on all context types except LayoutCtx
-// Access status information (hot/active/disabled/etc).
+// Access status information (hot/pointer captured/disabled/etc).
 impl_context_method!(
     MutateCtx<'_>,
     EventCtx<'_>,
@@ -252,16 +252,20 @@ impl_context_method!(
         /// mouse position have hot status.
         ///
         /// Discussion: there is currently some confusion about whether a
-        /// widget can be considered hot when some other widget is active (for
-        /// example, when clicking to one widget and dragging to the next).
-        /// The documentation should clearly state the resolution.
+        /// widget can be considered hot when some other widget has captured the
+        /// pointer (for example, when clicking one widget and dragging to the
+        /// next). The documentation should clearly state the resolution.
         pub fn is_hot(&self) -> bool {
             self.widget_state.is_hot
         }
 
-        // TODO - remove
-        pub fn is_active(&self) -> bool {
-            self.global_state.pointer_capture_target == Some(self.widget_id())
+        /// Whether the pointer is captured by this widget.
+        ///
+        /// See [`capture_pointer`] for more information about pointer capture.
+        ///
+        /// [`capture_pointer`]: EventCtx::capture_pointer
+        pub fn has_pointer_capture(&self) -> bool {
+            self.global_state.pointer_capture_target == Some(self.widget_state.id)
         }
 
         /// The focus status of a widget.
@@ -329,14 +333,14 @@ impl_context_method!(EventCtx<'_>, {
     /// Set the cursor icon.
     ///
     /// This setting will be retained until [`clear_cursor`] is called, but it will only take
-    /// effect when this widget is either [`hot`] or [`active`]. If a child widget also sets a
-    /// cursor, the child widget's cursor will take precedence. (If that isn't what you want, use
-    /// [`override_cursor`] instead.)
+    /// effect when this widget is [`hot`] and/or [`has_pointer_capture`]. If a child widget also
+    /// sets a cursor, the child widget's cursor will take precedence. (If that isn't what you
+    /// want, use [`override_cursor`] instead.)
     ///
     /// [`clear_cursor`]: EventCtx::clear_cursor
     /// [`override_cursor`]: EventCtx::override_cursor
     /// [`hot`]: EventCtx::is_hot
-    /// [`active`]: EventCtx::is_active
+    /// [`has_pointer_capture`]: EventCtx::has_pointer_capture
     pub fn set_cursor(&mut self, cursor: &CursorIcon) {
         trace!("set_cursor {:?}", cursor);
         self.widget_state.cursor = Some(*cursor);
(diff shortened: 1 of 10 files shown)


masonry/src/contexts.rs
```diff
@@ -577,9 +583,27 @@ impl_context_method!(
...
+    /// layout box. When the pointer is captured by a widget, that widget will continue receiving
+    /// pointer events when the pointer is outside the widget's layout box.
+    ///
+    /// The pointer cannot be captured by multiple widgets at the same time. If a widget has
+    /// captured the pointer and another widget captures it, the first widget loses the pointer
+    /// capture.
+    ///
+    /// Discussion: there is currently some confusion about whether a widget can be considered hot
+    /// when some other widget has captured the pointer (for example, when clicking one widget and
+    /// dragging to the next). The documentation should clearly state the resolution.
```
PoignardAzur, line 604: I don't think there's any confusion, but maybe I missed a Zulip thread.

The semantics are this: if a widget has pointer capture, either that widget is hot/hovered, or no widget is. This matches web behavior.

masonry/src/widget/scroll_bar.rs
```diff
@@ -162,7 +162,7 @@ impl Widget for ScrollBar {
             }
             PointerEvent::PointerUp(_, _) => {
                 self.grab_anchor = None;
-                ctx.set_active(false);
+                ctx.release_pointer();
```
PoignardAzur, line 165: Oh, this reminds me. PointerUp (and PointerLeave) events should automatically release pointer capture at the pass level. I forgot to implement it initially.

I'd appreciate if you included it in your PR, but it's optional.

PoignardAzur (approved):
Good work.

masonry/src/contexts.rs
tomcur: I duplicated the discussion point from here

https://github.com/linebender/xilem/blob/6c4951635da34cf04632332114e7b04ca343d8c0/masonry/src/contexts.rs#L254-L260

I probably should've referenced it instead of copying. If the question is resolved I can remove the comment from this PR, and create a new PR removing the old discussion point on is_hot and document that behavior.

Somewhat related to this discussion is this Zulip thread about pointer exclusivity: https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Exclusive.20events.20when.20interacting

PoignardAzur: > If the question is resolved I can remove the comment from this PR, and create a new PR removing the old discussion point on is_hot and document that behavior.

Sure.

> Somewhat related to this discussion is this Zulip thread about pointer exclusivity: https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Exclusive.20events.20when.20interacting

Oh, I forgot about that thread. I wonder if we still have an active issue somewhere referencing it. We might want to add a comment with a link to either a github issue or the zulip thread.

masonry/src/widget/scroll_bar.rs
tomcur: I've included this, and added a test for `PointerLeave`. `PointerUp` releasing the pointer was already tested in https://github.com/linebender/xilem/blob/6c4951635da34cf04632332114e7b04ca343d8c0/masonry/src/widget/tests/status_change.rs#L205

masonry/src/contexts.rs
tomcur: I've removed the discussion point. I can update the docs with the behavior of capture/hot on top of this PR. Easiest after it's merged.

tomcur:
@PoignardAzur This is ready for testing.

PoignardAzur (approved):

PoignardAzur:
LGTM.

I especially appreciate the added tests.

masonry/src/contexts.rs
```diff
@@ -252,16 +252,22 @@ impl_context_method!(
...
             self.widget_state.is_hot
         }
 
-        // TODO - remove
-        pub fn is_active(&self) -> bool {
-            self.global_state.pointer_capture_target == Some(self.widget_id())
+        /// Whether the pointer is captured by this widget.
+        ///
+        /// A widget can capture the pointer by calling [`capture_pointer`] during a pointer event.
+        /// It can subsequently release it by calling [`release_pointer`] during a pointer event.
```
DJMcNab, line 265: These docs now don't match the other new docs. Maybe we should just point at `capture_pointer`:
```rust
/// Any widget can [`release`] the pointer during any event. The pointer is automatically
/// released after handling of a [`PointerUp`] or [`PointerLeave`] event completes. A widget
/// holding the pointer capture will be the target of these events.
```

That is, the docs here imply that `release_pointer` is never implicit (and that it can only be called during a pointer event). I believe that neither of those are true.

masonry/src/contexts.rs
```diff
@@ -577,9 +583,32 @@ impl_context_method!(
...
+    /// layout box. When the pointer is captured by a widget, that widget will continue receiving
+    /// pointer events when the pointer is outside the widget's layout box.
+    ///
+    /// The pointer cannot be captured by multiple widgets at the same time. If a widget has
+    /// captured the pointer and another widget captures it, the first widget loses the pointer
+    /// capture.
+    ///
+    /// # Releasing the pointer
+    ///
+    /// Any widget can [`release`] the pointer during any event. The pointer is automatically
```
DJMcNab, line 604: In what context would it be valid for a pointer without the capture to release this capture?

masonry/src/contexts.rs
```diff
@@ -577,9 +583,32 @@ impl_context_method!(
...
     // TODO - Figure out cases where widget should be notified of pointer capture
     // loss
+    /// Capture the pointer in the current widget.
+    ///
+    /// Pointer capture is only allowed during a [`PointerDown`] event. It is a logic error to
+    /// capture the pointer during any other event.
+    ///
+    /// A widget normally only receives pointer events when the pointer is inside the widget's
+    /// layout box. When the pointer is captured by a widget, that widget will continue receiving
+    /// pointer events when the pointer is outside the widget's layout box.
```
DJMcNab, line 596:
```suggestion
    /// pointer events when the pointer is outside the widget's layout box (and other widgets will not receive these events).
```

masonry/src/contexts.rs
```diff
@@ -577,9 +583,32 @@ impl_context_method!(
 pub struct TimerToken;
 
 impl EventCtx<'_> {
-    // TODO - Document
+    // TODO - clearly document all semantics of pointer capture when they've been decided on
```
DJMcNab, line 586: Another question here: Is it valid for a disabled widget to have the pointer capture?

DJMcNab (approved):
A few small suggestions, but otherwise looks good. The same thing as in #563 applies

masonry/src/contexts.rs
tomcur: Good idea. That also means fewer docs to keep in sync.

masonry/src/contexts.rs
tomcur: Done.

tomcur: In addition to this, I've added context about events bubbling to their parents. 

https://github.com/linebender/xilem/blob/09bc929e021734a564980f4e0622dce66a2229ea/masonry/src/contexts.rs#L592-L597

masonry/src/contexts.rs
tomcur: Currently, any widget can release the capture during any event. That's either a target of an event or one of its ancestors. The target can be the widget with the pointer capture for a pointer event, or the widget with keyboard focus for a text event, or can be the target widget of an accessibility event. Widgets are not currently notified when they lose the capture in this way.

There's no hard requirement at the moment that says, e.g., the pointer capture widget (if any) should be equal to the widget with keyboard focus (if any).

masonry/src/contexts.rs
tomcur: Currently, any widget can release the capture during any event. That's either a target of an event or one of its ancestors. The target can be the widget with the pointer capture for a pointer event, or the widget with keyboard focus for a text event, or can be the target widget of an accessibility event. Widgets are not currently notified when they lose the capture in this way.

There's no hard requirement at the moment that says, e.g., the pointer capture widget (if any) should be equal to the widget with keyboard focus (if any).

masonry/src/contexts.rs
tomcur: It is at the moment. If we decide against that, the same should be invalid for keyboard focus. Capturing or requesting focus should not be allowed when disabled, and in `EventCtx::set_disabled` we should traverse the tree up from the widgets with pointer capture and keyboard focus to check whether their subtree is now disabled.

PoignardAzur:
@tomcur Anything you'd like to change before merging?

tomcur:
Happy for this to be merged. I believe I should do the merge because of the author-merges policy, now Daniel added me to the org?

PoignardAzur:
Yup.

PR #565, opened 2024-08-29
masonry: reimplement `Widget::get_child_at_pos`

tomcur:
The `Widget::get_child_at_pos` has a default linear search implementation, upholding some conditions (e.g., picking the last child in "z-order" in case of overlap). Widgets can override this with a custom implementation (e.g., a widget with many children that maintains a quadtree to search through). Custom implementations should uphold the same conditions, we can add some debug assertions to check this in a later PR.

This introduces a `QueryCtx`, allowing widget methods access to the global state, widget state, and widget subtree. `QueryCtx` is similar to the other (mutable) context types, but can be shared.

This also changes `WidgetRef` to consist of a `Widget` and a `QueryCtx`, similar to `WidgetMut` consisting of a `Widget` and `MutateCtx`. This required changing where `WidgetRef` can be constructed.

This resolves code comment https://github.com/linebender/xilem/blob/ac95f2524ac33fccd02f444b8a75ae0daaf98c3c/masonry/src/widget/widget_ref.rs#L192-L199

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -429,14 +429,14 @@ impl<'a> MutateCtx<'a> {
 // Methods to get a child WidgetRef from a parent.
 impl<'w> QueryCtx<'w> {
     /// Return a [`WidgetRef`] to a child widget.
-    pub fn get(&self, child: WidgetId) -> WidgetRef<'_, dyn Widget> {
+    pub fn get(self, child: WidgetId) -> WidgetRef<'w, dyn Widget> {
         let child_state = self
             .widget_state_children
-            .get_child(child.to_raw())
+            .into_child(child.to_raw())
             .expect("get: child not found");
         let child = self
             .widget_children
-            .get_child(child.to_raw())
+            .into_child(child.to_raw())
             .expect("get: child not found");
 
         let ctx = QueryCtx {
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -439,10 +439,15 @@ impl<'w> QueryCtx<'w> {
             .get_child(child.to_raw())
             .expect("get: child not found");
 
-        WidgetRef {
+        let ctx = QueryCtx {
+            global_state: self.global_state,
             widget_state_children: child_state.children,
             widget_children: child.children,
             widget_state: child_state.item,
+        };
+
+        WidgetRef {
+            ctx,
             widget: child.item,
         }
     }
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -429,7 +429,7 @@ impl<'a> MutateCtx<'a> {
 // Methods to get a child WidgetRef from a parent.
 impl<'w> QueryCtx<'w> {
     /// Return a [`WidgetRef`] to a child widget.
-    pub fn get<'c>(&'c self, child: WidgetId) -> WidgetRef<'c, dyn Widget> {
+    pub fn get(&self, child: WidgetId) -> WidgetRef<'_, dyn Widget> {
         let child_state = self
             .widget_state_children
             .get_child(child.to_raw())
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -15,8 +15,8 @@ use crate::passes::layout::run_layout_on;
 use crate::render_root::{MutateCallback, RenderRootSignal, RenderRootState};
 use crate::text::TextBrush;
 use crate::text_helpers::ImeChangeSignal;
-use crate::tree_arena::ArenaMutChildren;
-use crate::widget::{WidgetMut, WidgetState};
+use crate::tree_arena::{ArenaMutChildren, ArenaRefChildren};
+use crate::widget::{WidgetMut, WidgetRef, WidgetState};
 use crate::{
     AllowRawMut, BoxConstraints, CursorIcon, Insets, Point, Rect, Size, Widget, WidgetId, WidgetPod,
 };
@@ -49,6 +49,15 @@ pub struct MutateCtx<'a> {
     pub(crate) widget_state: &'a mut WidgetState,
     pub(crate) widget_state_children: ArenaMutChildren<'a, WidgetState>,
     pub(crate) widget_children: ArenaMutChildren<'a, Box<dyn Widget>>,
+}
+
+/// A context provided to methods of widgets requiring shared, read-only access.
+#[derive(Clone, Copy)]
+pub struct QueryCtx<'a> {
+    pub(crate) global_state: &'a RenderRootState,
+    pub(crate) widget_state: &'a WidgetState,
+    pub(crate) widget_state_children: ArenaRefChildren<'a, WidgetState>,
+    pub(crate) widget_children: ArenaRefChildren<'a, Box<dyn Widget>>,
 }
 
 /// A context provided to event handling methods of widgets.
@@ -124,6 +133,7 @@ pub struct AccessCtx<'a> {
 // Methods for all context types
 impl_context_method!(
     MutateCtx<'_>,
+    QueryCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     LayoutCtx<'_>,
@@ -200,6 +210,7 @@ impl_context_method!(
 // These methods access layout info calculated during the layout pass.
 impl_context_method!(
     MutateCtx<'_>,
+    QueryCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     ComposeCtx<'_>,
@@ -243,6 +254,7 @@ impl_context_method!(
 // Access status information (hot/pointer captured/disabled/etc).
 impl_context_method!(
     MutateCtx<'_>,
+    QueryCtx<'_>,
     EventCtx<'_>,
     LifeCycleCtx<'_>,
     ComposeCtx<'_>,
@@ -409,6 +421,29 @@ impl<'a> MutateCtx<'a> {
             widget_state: self.widget_state,
             widget_state_children: self.widget_state_children.reborrow_mut(),
             widget_children: self.widget_children.reborrow_mut(),
+        }
+    }
+}
+
+// --- MARK: WIDGET_REF ---
+// Methods to get a child WidgetRef from a parent.
+impl<'w> QueryCtx<'w> {
+    /// Return a [`WidgetRef`] to a child widget.
+    pub fn get<'c>(&'c self, child: WidgetId) -> WidgetRef<'c, dyn Widget> {
+        let child_state = self
+            .widget_state_children
+            .get_child(child.to_raw())
+            .expect("get: child not found");
+        let child = self
+            .widget_children
+            .get_child(child.to_raw())
+            .expect("get: child not found");
+
+        WidgetRef {
+            widget_state_children: child_state.children,
+            widget_children: child.children,
+            widget_state: child_state.item,
+            widget: child.item,
         }
     }
 }
--- a/masonry/src/widget/widget.rs
+++ b/masonry/src/widget/widget.rs
@@ -178,9 +178,9 @@ pub trait Widget: AsAny {
 
     // --- Auto-generated implementations ---
 
-    /// Return which child, if any, has the given `pos` in its layout rect and clip path. In case
-    /// of overlapping children, the last child as determined by [`Widget::children_ids`] is
-    /// chosen.
+    /// Return which child, if any, has the given `pos` in its layout rect. In case of overlapping
+    /// children, the last child as determined by [`Widget::children_ids`] is chosen. No child is
+    /// returned if `pos` is outside the widget's clip path.
     ///
     /// The child returned is a direct child, not e.g. a grand-child.
     ///
@@ -194,6 +194,15 @@ pub trait Widget: AsAny {
         ctx: QueryCtx<'c>,
         pos: Point,
     ) -> Option<WidgetRef<'c, dyn Widget>> {
+        let relative_pos = pos - ctx.widget_state.window_origin().to_vec2();
+        if !ctx
+            .widget_state
+            .clip
+            .map_or(true, |clip| clip.contains(relative_pos))
+        {
+            return None;
+        }
+
         // Assumes `Self::children_ids` is in increasing "z-order", picking the last child in case
         // of overlapping children.
         for child_id in self.children_ids().iter().rev() {
@@ -202,13 +211,7 @@ pub trait Widget: AsAny {
             let relative_pos = pos - child.state().window_origin().to_vec2();
             // The position must be inside the child's layout and inside the child's clip path (if
             // any).
-            if !child.widget.skip_pointer()
-                && child.state().window_layout_rect().contains(pos)
-                && child
-                    .state()
-                    .clip
-                    .map_or(true, |clip| clip.contains(relative_pos))
-            {
+            if !child.widget.skip_pointer() && child.state().window_layout_rect().contains(pos) {
                 return Some(child);
             }
         }
--- a/masonry/src/widget/widget.rs
+++ b/masonry/src/widget/widget.rs
@@ -14,9 +14,10 @@ use vello::Scene;
 
 use crate::contexts::ComposeCtx;
 use crate::event::{AccessEvent, PointerEvent, StatusChange, TextEvent};
+use crate::widget::WidgetRef;
 use crate::{
     AccessCtx, AsAny, BoxConstraints, EventCtx, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx,
-    RegisterCtx, Size,
+    Point, RegisterCtx, Size,
 };
 
 /// A unique identifier for a single [`Widget`].
@@ -177,20 +178,29 @@ pub trait Widget: AsAny {
 
     // --- Auto-generated implementations ---
 
-    // FIXME
-    #[cfg(FALSE)]
     /// Return which child, if any, has the given `pos` in its layout rect.
     ///
-    /// The child return is a direct child, not eg a grand-child. The position is in
+    /// The child returned is a direct child, not eg a grand-child. The position is in
     /// relative coordinates. (Eg `(0,0)` is the top-left corner of `self`).
     ///
     /// Has a default implementation, that can be overridden to search children more
     /// efficiently.
-    fn get_child_at_pos(&self, pos: Point) -> Option<WidgetRef<'_, dyn Widget>> {
-        // layout_rect() is in parent coordinate space
-        self.children()
-            .into_iter()
-            .find(|child| child.state().layout_rect().contains(pos))
+    ///
+    /// The child widget references in `children` are in the same order as returned by
+    /// [`Self::children_ids`].
+    fn get_child_at_pos<'w>(
+        &self,
+        children: &[WidgetRef<'w, dyn Widget>],
+        pos: Point,
+    ) -> Option<WidgetRef<'w, dyn Widget>> {
+        children
+            .iter()
+            // TODO: currently assumes `Self::children_ids` is in increasing "z-order". Picks the
+            // last child in case of overlapping children. Is this the desired behavior?
+            .rev()
+            // A child's layout_rect() is in the parent's coordinate space.
+            .find(|child| !child.widget.skip_pointer() && child.state().layout_rect().contains(pos))
+            .copied()
     }
 
     /// Get the (verbose) type name of the widget for debugging purposes.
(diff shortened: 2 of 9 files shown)


masonry/src/widget/widget_ref.rs
```diff
@@ -172,31 +172,32 @@ impl<'w> WidgetRef<'w, dyn Widget> {
     ///
     /// **pos** - the position in local coordinates (zero being the top-left of the
     /// inner widget).
-    pub fn find_widget_at_pos(&self, pos: Point) -> Option<WidgetRef<'w, dyn Widget>> {
+    pub fn find_widget_at_pos(&self, mut pos: Point) -> Option<WidgetRef<'w, dyn Widget>> {
```
waywardmonkeys, line 175: I think it would be better to shadow `pos` within the function explicitly than have this be `mut`

waywardmonkeys (commented):
I don't know about the overall strategy that is desired for this, but I'm definitely interested in seeing this feature be underway.

Would be good to have a couple of tests as well though, including checking `skip_pointer` and some other edge cases.

masonry/src/widget/widget_ref.rs
PoignardAzur: This is non-blocking, but agreed. It's the pattern the rest of the codebase uses.

PoignardAzur:
> The default `Widget::get_child_at_pos` implementation assumes children are in increasing "z-order", picking the last child in case of overlapping children. If that assumption is kept, it should be documented, but perhaps a better solution is needed.

That assumption is likely to remain, since it's how children are painted right now, and the cursor should general pick the widget whose pixels are "visible" under the mouse.

I'm not crazy about the approach of passing an array of WidgetRef, though. I would lean towards passing a nont-mutable LifecycleCtx (soon to be renamed to UpdateCtx) instead. In the cases where that method would be useful (eg a very large container), you want to *avoid* any kind of `O(N)` work.

masonry/src/widget/widget_ref.rs
tomcur: Done

waywardmonkeys:
It would be nice if this supported the sort of use case where one might want an acceleration structure (such as a quad tree) for a widget-canvas with a lot of widgets.

tomcur:
> It would be nice if this supported the sort of use case where one might want an acceleration structure (such as a quad tree) for a widget-canvas with a lot of widgets.

Indeed, that's a good use-case. If, as @PoignardAzur suggested, a cheaply-constructable context type is passed to the widget, the widget can choose to either do a linear search through its children (for example, sticking to the default behavior), or it can use whatever domain knowledge it has (for example, your use-case of a canvas with a quad tree for searching).

PoignardAzur:
> That would mean `WidgetRef::find_widget_at_pos` would need mutable references to construct `LifeCycleCtx`, but it by design holds shared references.

Oh, that's right. That makes things more awkward.

> Perhaps a new context type with shared references is required?

Probably, yes, but I'm not happy about adding yet another context type.

tomcur:
Instead of creating a new context type, `WidgetRef` can be provided to the widget. The signature then becomes:

```rust
fn get_child_at_pos<'w>(
    &self,
    widget_ref: WidgetRef<'w, dyn Widget>,
    pos: Point,
) -> Option<WidgetRef<'w, dyn Widget>>
```

I've pushed a commit with this change to illustrate. It is a bit circular, though: `WidgetRef` carries a reference to the widget it is now calling with a reference to itself.

Going the context route, the context could look as follows. Some of the methods on `WidgetRef` (like `WidgetRef::find_widget_by_id`) could be moved/duplicated to `QueryCtx`.

```rust
/// A context provided to widget methods that need read-only (shared) access.
pub struct QueryCtx<'a> {
    pub(crate) global_state: &'a RenderRootState,
    pub(crate) widget_state: &'a WidgetState,
    pub(crate) widget_state_children: ArenaRefChildren<'a, WidgetState>,
    pub(crate) widget_children: ArenaRefChildren<'a, Box<dyn Widget>>,
}
```

Note `QueryCtx` itself would not be sufficient to iterate over child `WidgetRef`, as a reference to the widget is needed to call `Widget::children_ids`. This is no problem inside `Widget::get_child_at_pos`, of course.

https://github.com/linebender/xilem/blob/6c4951635da34cf04632332114e7b04ca343d8c0/masonry/src/widget/widget_ref.rs#L109-L112

PoignardAzur:
Sounds good. Fair warning, I won't have time to look at this over the weekend, but I'll have a look on Monday.

tomcur:
A context type plus a `Widget` is similar to a `WidgetRef` (a `WidgetRef` can be constructed from those two), with the option of having additional fields. In the example above, `QueryCtx` has a `global_state: &RenderRootState` field like the other context types (though shared, not mutable), so it can have methods like `QueryCtx::has_pointer_capture`.

Unlike `WidgetRef`, context types don't hold references to the `Widget` itself, so calling a `Widget` with a `QueryCtx` doesn't have that somewhat inelegant circular referencing.

Existing `Widget` methods requiring state receive it through mut ref context types, using a shared ref `QueryCtx` is perhaps less surprising than using a `WidgetRef`. It also prevents the `Widget` API from needing to know about `WidgetRef` at all.

masonry/src/widget/widget.rs
```diff
@@ -165,20 +167,30 @@ pub trait Widget: AsAny {
...
     ///
     /// Has a default implementation, that can be overridden to search children more
     /// efficiently.
-    fn get_child_at_pos(&self, pos: Point) -> Option<WidgetRef<'_, dyn Widget>> {
-        // layout_rect() is in parent coordinate space
-        self.children()
-            .into_iter()
-            .find(|child| child.state().layout_rect().contains(pos))
+    ///
+    /// The child widget references in `children` are in the same order as returned by
```
DJMcNab, line 178: What are `children`? And what value should `widget_ref` have?

masonry/src/widget/widget.rs
```diff
@@ -165,20 +167,30 @@ pub trait Widget: AsAny {
...
+        pos: Point,
+    ) -> Option<WidgetRef<'w, dyn Widget>> {
+        widget_ref
+            .children()
+            .iter()
+            // TODO: currently assumes `Self::children_ids` is in increasing "z-order". Picks the
+            // last child in case of overlapping children. Is this the desired behavior?
+            .rev()
+            // A child's layout_rect() is in the parent's coordinate space.
+            .find(|child| !child.widget.skip_pointer() && child.state().layout_rect().contains(pos))
```
DJMcNab, line 192: Checking `skip_pointer` here is a little bit weird...

DJMcNab (commented):
Some "drive-by" thoughts, pending Olivier's review.

PoignardAzur (changes requested):
Something that occurs to me reading this is we probably want to rename `find_widget_at_pos` to `find_widget_under_cursor` across the crate. It's of scope for now though.

I think the WidgetRef based API is too awkward. I'd rather have one based on a new QueryCtx type.

tomcur:
Rebased onto main, and added commits. One introduces `QueryCtx`, replacing `WidgetRef` in the method call, but otherwise introduces minimal changes. 

The commit after explores a more substantial change. `WidgetRef` and `WidgetMut` are conceptually similar. `WidgetMut` is a struct of a `Widget` and a `MutateCtx`. For symmetry, in this second commit `WidgetRef` becomes a struct of a `Widget` and a `QueryCtx`.

Some thoughts:

1. `MutateCtx::get_mut` was already used to get a `WidgetMut`, I've added `QueryCtx::get` to get a `WidgetRef`.
2. `QueryCtx::get` should perhaps take a `&WidgetPod`, like `MutateCtx::get_mut` takes. It currently takes a `WidgetId`, because `Widget::children_ids` returns just that: ids, not pods. Maybe we can construct `WidgetPod` in the default `Widget::get_child_at_pos` method implementation by assuming `WidgetPodInner::Inserted`.
3. `QueryCtx` having access to global state, like all other context types, has some ramifications. For example, to construct a `WidgetRef` containing a `QueryCtx`, having access to a widget's state is no longer enough, and means `WidgetArena` cannot construct a `WidgetRef` by itself anymore. In that commit I've moved the method for getting widget refs to `RenderRoot::get_widget`, which already had a method for getting a `WidgetMut` through `RenderRoot::edit_widget`.

masonry/src/widget/widget.rs
tomcur: This discussion should probably continue in #572

masonry/src/widget/widget_ref.rs
```diff
@@ -175,31 +168,42 @@ impl<'w> WidgetRef<'w, dyn Widget> {
...
-                !child.widget.skip_pointer() && child.state().window_layout_rect().contains(pos)
-            }) {
+
+            if let Some(child) = innermost_widget
+                .widget
+                .get_child_at_pos(&innermost_widget.ctx, pos)
+            {
+                // Get child from the widget arena to bind it to the arena's lifetime. Is there a
+                // way around this?
+                let child = root.get_widget(child.id()).unwrap();
```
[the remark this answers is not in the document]
PoignardAzur: Yeah, this exact kind of lifetime issue (when you want to return a borrow to something pointed to by a local, but rustc thinks you're returning a borrow to the local) has bitten me before.

I'm not sure there's an elegant solution, but finding one is a mid-high priority.

masonry/src/widget/widget.rs
```diff
@@ -165,20 +167,29 @@ pub trait Widget: AsAny {
...
-        self.children()
-            .into_iter()
-            .find(|child| child.state().layout_rect().contains(pos))
+    ///
+    /// The child widget references in `children` are in the same order as returned by
+    /// [`Self::children_ids`].
+    fn get_child_at_pos<'w>(
+        &self,
+        children: &[WidgetRef<'w, dyn Widget>],
+        pos: Point,
```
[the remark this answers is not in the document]
PoignardAzur: > Perhaps `Widget::get_child_at_pos` should mirror that and take `pos` in global coordinates.

Definitely. I'm not a fan of the "remove origin from pos each loop cycle" game. It's too brittle, and we've had bugs because of it before. I'd like to remove it anywhere we can.

PoignardAzur (commented):
LGTM at a skim. Will look more in-depth later.

PoignardAzur (dismissed):

masonry/src/widget/widget_ref.rs
```diff
@@ -175,27 +168,36 @@ impl<'w> WidgetRef<'w, dyn Widget> {
...
-                // If the widget has a clip, the point must be inside
-                // else we don't iterate over children.
-                if !clip.contains(relative_pos.to_point()) {
+                // If the widget has a clip, the point must be inside, else we don't iterate over
+                // children.
+                let pos = pos - innermost_widget.state().window_origin().to_vec2();
+                if !clip.contains(pos) {
                     break;
                 }
             }
```
PoignardAzur, line 192: This part should probably be in `Widget::get_child_at_pos`.

tomcur: Done, `Widget::get_child_at_pos` now checks the clip path of children.

Because `WidgetRef::find_widget_at_pos` does explicitly check the layout rect of `self` before recursing down the tree (else it returns None), I've added a check of the clip path of `self`, too.

masonry/src/widget/widget.rs
```diff
@@ -165,20 +167,41 @@ pub trait Widget: AsAny {
...
-    /// Return which child, if any, has the given `pos` in its layout rect.
+    /// Return which child, if any, has the given `pos` in its layout rect and clip path. In case
+    /// of overlapping children, the last child as determined by [`Widget::children_ids`] is
+    /// chosen.
+    ///
+    /// The child returned is a direct child, not e.g. a grand-child.
     ///
-    /// The child return is a direct child, not eg a grand-child. The position is in
-    /// relative coordinates. (Eg `(0,0)` is the top-left corner of `self`).
+    /// Has a default implementation that can be overridden to search children more efficiently.
```
tomcur, line 176: Should implementations overriding the default uphold certain conditions?

Custom implementations should probably uphold the guarantee that `pos` is actually inside the returned child's layout rect. Must custom custom implementations also uphold the guarantee that `pos` is contained in the clip path, and that in case of overlap the last child in "z-order" is chosen?

PoignardAzur: Yes to all of these.

Ideally we should even implement some best-effort checks in debug_assertions mode to enforce that these conditions are upheld.

tomcur: The low effort version is to check the child contains the position in its layout rect and clip path.

The conditions uniquely identify a child, so we could do the default search alongside `Widget::get_child_at_pos` and check the same children are returned. For widgets with many children—where custom implementations are most likely—this would be slow. Valid options are to accept that cost or to limit this exhaustive check to widgets with N children for some value of N.

PoignardAzur: Right, most of that is beyond the scope of this PR.

We could probably add a TODO comment and keep it to that for now.

> Valid options are to accept that cost or to limit this exhaustive check to widgets with N children for some value of N.

Sure. I was thinking we'd probably need to add a `child_count()` method for cases where building a child array is needlessly expensive (eg a Flex with thousands of children).

masonry/src/contexts.rs
```diff
@@ -403,6 +415,34 @@ impl<'a> MutateCtx<'a> {
...
+    /// Return a [`WidgetRef`] to a child widget.
+    pub fn get(&self, child: WidgetId) -> WidgetRef<'_, dyn Widget> {
+        let child_state = self
+            .widget_state_children
+            .get_child(child.to_raw())
+            .expect("get: child not found");
+        let child = self
+            .widget_children
+            .get_child(child.to_raw())
+            .expect("get: child not found");
```
PoignardAzur, line 430:
```suggestion
    pub fn get(self, child: WidgetId) -> WidgetRef<'w, dyn Widget> {
        let child_state = self
            .widget_state_children
            .into_child(child.to_raw())
            .expect("get: child not found");
        let child = self
            .widget_children
            .into_child(child.to_raw())
            .expect("get: child not found");
```

masonry/src/widget/widget.rs
```diff
@@ -354,6 +378,14 @@ impl Widget for Box<dyn Widget> {
         self.deref().get_cursor()
     }
 
+    fn get_child_at_pos<'c>(
+        &self,
+        ctx: &'c QueryCtx,
```
PoignardAzur, line 383:
```suggestion
        ctx: &QueryCtx<'c>,
```

masonry/src/widget/widget_ref.rs
```diff
@@ -168,38 +161,39 @@ impl<'w> WidgetRef<'w, dyn Widget> {
...
+    /// **pos** - the position in global coordinates (e.g. `(0,0)` is the top-left corner of the
+    /// window).
+    pub fn find_widget_at_pos<'a>(
+        &self,
+        root: &'a RenderRoot,
+        pos: Point,
+    ) -> Option<WidgetRef<'a, dyn Widget>> {
+        // Get self from the widget arena to bind it to the arena's lifetime. Is there a way around
+        // this? Also see the comment inside the loop rebinding child to the arena's lifetime.
+        let mut innermost_widget = root.get_widget(self.id()).unwrap();
```
PoignardAzur, line 177:
```suggestion
    pub fn find_widget_at_pos(&self, pos: Point) -> Option<WidgetRef<'_, dyn Widget>> {
        let mut innermost_widget = *self;
```

PoignardAzur (commented):
This is roughly what a idiomatic version looks like. The key thing is that since QueryCtx is Copy, you can take it by value in `QueryCtx::get`, and that means you can request "smaller" lifetimes.

masonry/src/widget/widget.rs
```diff
@@ -165,20 +167,42 @@ pub trait Widget: AsAny {
...
+
+            let relative_pos = pos - child.state().window_origin().to_vec2();
+            // The position must be inside the child's layout and inside the child's clip path (if
+            // any).
+            if !child.widget.skip_pointer()
+                && child.state().window_layout_rect().contains(pos)
+                && child
+                    .state()
+                    .clip
+                    .map_or(true, |clip| clip.contains(relative_pos))
```
PoignardAzur, line 199: The clip check should be outside the loop. The logic (for now) is that a widget's clip only constrains its children.

(We might change it, because it *is* counter-intuitive.)

masonry/src/widget/widget_ref.rs
```diff
@@ -168,38 +161,39 @@ impl<'w> WidgetRef<'w, dyn Widget> {
...
+
+        let relative_pos = pos - self.state().window_origin().to_vec2();
+        if !self.state().window_layout_rect().contains(pos)
+            || !self
+                .state()
+                .clip
+                .map_or(true, |clip| clip.contains(relative_pos))
+        {
             return None;
         }
```
PoignardAzur, line 187:
```suggestion
        if !self.state().layout_rect().contains(relative_pos) {
            return None;
        }
        if !self.state().clip.unwrap_or_default().contains(relative_pos) {
            return None;
        }
```

masonry/src/contexts.rs
tomcur: Thanks, `ArenaRefChildren::into_child` was the piece I missed when first playing around with these lifetime issues.

masonry/src/widget/widget_ref.rs
tomcur: Clip path as `Rect` defaults to (0,0,0,0), so unwrap_or_default doesn't work as expected. We'd need something like `Rect::INFINITY`.

tomcur: Did split into two statements

masonry/src/widget/widget.rs
tomcur: Ah, right. I indeed thought it affected the widget itself. That means the clip check can be removed from `WidgetRef::find_widget_at_pos`, as it's now done by `Widget::get_child_at_pos`.

masonry/src/widget/widget.rs
```diff
@@ -165,20 +167,45 @@ pub trait Widget: AsAny {
...
-            .into_iter()
-            .find(|child| child.state().layout_rect().contains(pos))
+    /// **pos** - the position in global coordinates (e.g. `(0,0)` is the top-left corner of the
+    /// window).
+    fn get_child_at_pos<'c>(
+        &self,
+        ctx: QueryCtx<'c>,
+        pos: Point,
+    ) -> Option<WidgetRef<'c, dyn Widget>> {
+        let relative_pos = pos - ctx.widget_state.window_origin().to_vec2();
```
tomcur, line 186: `QueryCtx::widget_state` is `pub(crate)`, we can access it here but user code currently can't.

PoignardAzur: Right, same with the `clip` field access.

masonry/src/contexts.rs
```diff
@@ -403,6 +415,34 @@ impl<'a> MutateCtx<'a> {
     }
 }
 
+// --- MARK: WIDGET_REF ---
+// Methods to get a child WidgetRef from a parent.
+impl<'w> QueryCtx<'w> {
+    /// Return a [`WidgetRef`] to a child widget.
+    pub fn get(self, child: WidgetId) -> WidgetRef<'w, dyn Widget> {
```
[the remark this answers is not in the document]
tomcur: In a previous comment I wrote:

> `QueryCtx::get` should perhaps take a `&WidgetPod`, like how `MutateCtx::get_mut` takes a `&mut WidgetPod`. It currently takes a `WidgetId`, because `Widget::children_ids` returns just that: ids, not pods. Maybe we can construct `WidgetPod` in the default `Widget::get_child_at_pos` method implementation by assuming `WidgetPodInner::Inserted`.

Is it fine to assume `WidgetPodInner::Inserted`? In that case, we need to add a pub(crate) method that allows constructing a pod that way.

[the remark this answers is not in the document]
PoignardAzur: Oh, you're right, I forgot about that.

I think taking an id is fine. The other solutions come with their own set of problems.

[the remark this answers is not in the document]
tomcur: See also https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/.60Widget.3A.3Achildren_ids.60.20vs.20.60Widget.3A.3Achildren_pods.60.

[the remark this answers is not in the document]
tomcur: Whoops, at the same time you posted this, I created the Zulip thread. Let's leave it as it is for this PR, and if we decide to, we can do a follow-up changing the signature.

masonry/src/widget/widget.rs
tomcur: Added a todo comment, so we can leave this for a future PR

tomcur:
> You could probably add a few methods to QueryCtx so that the default implementation doesn't need to use private fields, but either way, it's a good first implementation.

I'll do a follow-up PR for that, if that's fine with you? Then we can think about the signatures of `QueryCtx` methods (and whether other context types might want some of those methods, too).

PoignardAzur:
Sure.

PR #567, opened 2024-08-29
Introduces date selector to xilem

giannissc:


giannissc:
I believe this is ready for review! I still haven't evaluated the `state_accessor`alternative we talked about in Zulip but since I don't hate this I am happy to merge and create a new PR for any future changes! @Philipp-M

xilem/src/view/date_picker.rs
DJMcNab, line 1: I wonder if this should live in a `components` folder? From a developer of Xilem perspective, there is a meaningful difference between "raw" views and composed components. But from an end-user perspective, they are the same.

I don't know what the right answer is here.

xilem/src/view/date_picker.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use super::{button, flex, FlexSpacer};
+
+pub enum DatePickerMessage {
+    Select(Date),
+    ChangeView,
+    Nop,
+}
+
+pub struct DatePicker {
```
DJMcNab, line 14: We should have some docs here.

DJMcNab (commented):
Don't have time for a full review, but some initial thoughts.

Could you also include a screenshot in this thread?

xilem/src/view/date_picker.rs
giannissc: I am going to actually create an RFC to discuss this among ither things that i have in mind. For now I would suggest to do both. In the internal folder organisation there should be a primitives folder and a components folder but at the tip the should all be re-exported together. Does this sound reasonable?

PR #568, opened 2024-09-01
add checkbox to widget example

derekdreery:
This PR adds another widget to the widgets example, and also draws boxes around each widget to separate them visually. I wanted to use flex wrap to display them, but I don't think that's implemented yet in Xilem, and the layout I did use isn't bad.

It also provides a simple example of the use of `Adapt`

### DIFF
--- a/xilem/examples/widgets.rs
+++ b/xilem/examples/widgets.rs
@@ -7,50 +7,95 @@ use masonry::dpi::LogicalSize;
 use masonry::event_loop_runner::{EventLoop, EventLoopBuilder};
 use winit::error::EventLoopError;
 use winit::window::Window;
-use xilem::view::{button, checkbox, flex, label, progress_bar, FlexSpacer};
-use xilem::{WidgetView, Xilem};
+use xilem::core::adapt;
+use xilem::view::{button, checkbox, flex, flex_item, progress_bar, sized_box, Axis, FlexSpacer};
+use xilem::{Color, WidgetView, Xilem};
+
+const SPACER_WIDTH: f64 = 10.;
 
 /// The state of the entire application.
 ///
 /// This is owned by Xilem, used to construct the view tree, and updated by event handlers.
 struct WidgetGallery {
     progress: Option<f64>,
+    checked: bool,
 }
 
-fn app_logic(data: &mut WidgetGallery) -> impl WidgetView<WidgetGallery> {
+fn progress_bar_view(data: Option<f64>) -> impl WidgetView<Option<f64>> {
     flex((
-        label("this 'widgets' example currently only has 1 widget"),
-        FlexSpacer::Flex(1.),
-        progress_bar(data.progress),
+        progress_bar(data),
         checkbox(
             "set indeterminate progress",
-            data.progress.is_none(),
-            |state: &mut WidgetGallery, checked| {
+            data.is_none(),
+            |state: &mut Option<f64>, checked| {
                 if checked {
-                    state.progress = None;
+                    *state = None;
                 } else {
-                    state.progress = Some(0.5);
+                    *state = Some(0.5);
                 }
             },
         ),
-        button("change progress", |state: &mut WidgetGallery| {
-            match state.progress {
-                Some(ref mut v) => *v = (*v + 0.1).rem_euclid(1.),
-                None => state.progress = Some(0.5),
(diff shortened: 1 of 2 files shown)


derekdreery:
hmm perhaps this should use `map_state`, or possibly both `adapt` and `map_state` to demonstrate the equivalence.

xilem/examples/widgets.rs
DJMcNab, line 1: Widgets is not a very compelling example name. Could this be renamed to `widget_gallery`?

It seems that you have preferred this name in every other place you are naming the example...

xilem/examples/widgets.rs
```diff
@@ -7,50 +7,95 @@ use masonry::dpi::LogicalSize;
...
 /// The state of the entire application.
 ///
 /// This is owned by Xilem, used to construct the view tree, and updated by event handlers.
 struct WidgetGallery {
     progress: Option<f64>,
+    checked: bool,
 }
 
-fn app_logic(data: &mut WidgetGallery) -> impl WidgetView<WidgetGallery> {
+fn progress_bar_view(data: Option<f64>) -> impl WidgetView<Option<f64>> {
```
DJMcNab, line 24:
```suggestion
fn progress_bar_view(progress: Option<f64>) -> impl WidgetView<Option<f64>> {
```

xilem/examples/widgets.rs
```diff
@@ -7,50 +7,95 @@ use masonry::dpi::LogicalSize;
...
-            }
+        button("change progress", |state: &mut Option<f64>| match state {
+            Some(ref mut v) => *v = (*v + 0.1).rem_euclid(1.),
+            None => *state = Some(0.5),
         }),
-        FlexSpacer::Flex(1.),
     ))
 }
 
+fn checkbox_view(data: bool) -> impl WidgetView<bool> {
```
DJMcNab, line 45:
```suggestion
fn checkbox_view(checked: bool) -> impl WidgetView<bool> {
```
I'm not convinced that this is idiomatic, however. I'd probably just inline this here.

xilem/examples/widgets.rs
```diff
@@ -7,50 +7,95 @@ use masonry::dpi::LogicalSize;
...
+    checkbox("a simple checkbox", data, |data, new_state| {
+        *data = new_state;
+    })
+}
+
+/// Wrap widgets in a box with a border
```
DJMcNab, line 51:
```suggestion
/// Wrap `inner` in a box with a border
```

xilem/src/view/mod.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+//! A collection of useful pre-made views (sometimes called 'widgets' in other UI systems)
```
DJMcNab, line 4: It would be useful to have documentation here, but I suspect this is actively misleading. Maybe something like:
```suggestion
//! Views for the widgets which are built-in to Masonry. These are the primitives your Xilem app's view tree will generally be constructed from.
```

xilem/examples/widgets.rs
```diff
@@ -7,50 +7,95 @@ use masonry::dpi::LogicalSize;
...
+fn app_logic(data: &mut WidgetGallery) -> impl WidgetView<WidgetGallery> {
+    // Use a `sized_box` to pad the window contents
+    sized_box(
+        flex((
+            adapt(
+                flex_item(border_box(progress_bar_view(data.progress)), 1.),
+                |data: &mut WidgetGallery, thunk| thunk.call(&mut data.progress),
+            ),
+            adapt(
+                flex_item(border_box(checkbox_view(data.checked)), 1.),
```
DJMcNab, line 78: Why not use `.flex` here?

DJMcNab (commented):
image

The progress bar overflows in this example as-is.

derekdreery:
> image
> 
> The progress bar overflows in this example as-is. There are a few other tweaks needed.
> 
> Having a place to put new widgets will be useful, though

I'm not sure why the progress bar overflows - it might be a bug in xilem or with the example itself.

DJMcNab (approved):
Even if the pending review comments aren't going to get addressed, I think this is still an improvement as-is.

Thanks!

PR #573, opened 2024-09-03
masonry: add docs for `TreeArena::get_id_path`

tomcur:
beats reading the implementation to know the order :-)

### DIFF
--- a/masonry/src/tree_arena.rs
+++ b/masonry/src/tree_arena.rs
@@ -207,6 +207,10 @@ impl<Item> TreeArena<Item> {
         Some(node.arena_mut(self.parents_map[&id], &mut self.parents_map))
     }
 
+    /// Construct the path of items from the given item to the root of the tree.
+    ///
+    /// The path is in order from the bottom to the top, starting at the given item and ending at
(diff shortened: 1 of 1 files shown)


masonry/src/tree_arena.rs
```diff
@@ -207,6 +207,10 @@ impl<Item> TreeArena<Item> {
         Some(node.arena_mut(self.parents_map[&id], &mut self.parents_map))
     }
 
+    /// Construct the path of items from the given item to the root of the tree.
```
jaredoconnell, line 210: Nit-pick: I would use active voice with "Constructs the path...", rather than an imperative command.

tomcur: Thanks! It looks like much of the codebase currently is imperative, e.g., all methods on `Widget` have imperative docs.

jaredoconnell: If that is the case, we'll stick to it for consistency.

DJMcNab (approved):
Thanks! Getting a first pass of documentation in places is very important.

PR #577, opened 2024-09-04
Request accessibility update when checkbox state changes

### LINKED ISSUE #576: Accessibility: checkbox accessibility node isn't updated
In the mason example, the accessibility node for the checkbox isn't updated when the checkbox's state is toggled. I reproduced this on Windows, using both NVDA and Narrator, as well as on Android using the WIP AccessKit Android adapter. Note that if the accessibility tree is created after checking the checkbox (e.g. when starting the screen reader after clicking the checkbox), then the checkbox accessibility node is in the checked state. So Masonry's accessibility method for the checkbox is setting the correct state. It just appears that the node isn't being rebuilt when the state changes.

mwcampbell:
Fixes #576

Can someone who's more familiar with Masonry please confirm that I'm doing this the right way? I assume that one is supposed to call request_accessibility update along with request_paint whenever the change to the widget state could affect accessibility.

### DIFF
--- a/masonry/src/widget/checkbox.rs
+++ b/masonry/src/widget/checkbox.rs
@@ -47,6 +47,7 @@ impl WidgetMut<'_, Checkbox> {
     pub fn set_checked(&mut self, checked: bool) {
         self.widget.checked = checked;
         self.ctx.request_paint();
+        self.ctx.request_accessibility_update();
     }
 
     /// Set the text.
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
I believe that this is correct for our current state. I wonder if we should make this mostly implicit, i.e. make `request_paint` do both, and have a `request_paint_without_accessibility` method.

But that's not a thing for this version.
Not sure if you're aware, but we have an "author-merges" policy, so please merge this when you're ready for it to land.

mwcampbell:
About making this more implicit, I suggest calling the paint-and-accessibility method `request_render`, where we use "render" in the same sense that, say, React does. Then, we normalize using `request_render` by default, and only using `request_paint` if one is sure that one doesn't need to update accessibility.

DJMcNab:
`request_render` is a better name than paint for that, certainly. I think that not renaming `request_paint` would *probably* be a footgun, and would still suggest giving it a longer name which makes the intentionality clear.

Issue #581, opened 2024-09-06
Documentation: Missing characters in `ARCHITECTURE.md`


failingprovince:
The [`ARCHITECTURE.md`](https://github.com/linebender/xilem/blob/main/ARCHITECTURE.md) file is missing a \` character in the Code Organisation section.

PR #582, opened 2024-09-06
docs(#581): fix missing char in `ARCHITECTURE.md`

failingprovince:
Fix issue #581.


DJMcNab (approved):
Thank you! For future reference, it isn't necessary to make an issue for trivial changes, especially if you're going to make the PR yourself immediately.

PR #588, opened 2024-09-14
Refactor arena code

PoignardAzur:
Rewrite doc comments
Move find methods to ArenaRef/MutChildren
Create ArenaMapref/Mut in anticipation of future work.

### DIFF
--- a/masonry/src/tree_arena.rs
+++ b/masonry/src/tree_arena.rs
@@ -21,35 +21,6 @@ struct TreeNode<Item> {
     children: Vec<TreeNode<Item>>,
 }
 
-impl<Item> TreeNode<Item> {
-    fn arena_ref(&self, parent_id: Option<u64>) -> ArenaRef<'_, Item> {
-        ArenaRef {
-            parent_id,
-            item: &self.item,
-            children: ArenaRefChildren {
-                id: Some(self.id),
-                children: &self.children,
-            },
-        }
-    }
-
-    fn arena_mut<'a>(
-        &'a mut self,
-        parent_id: Option<u64>,
-        parents_map: &'a mut HashMap<u64, Option<u64>>,
-    ) -> ArenaMut<'a, Item> {
-        ArenaMut {
-            parent_id,
-            item: &mut self.item,
-            children: ArenaMutChildren {
-                id: Some(self.id),
-                children: &mut self.children,
-                parents_map,
-            },
-        }
-    }
-}
-
 /// A container type for a tree of items.
 ///
 /// This type is used to store zero, one or many tree of a given item types. It
@@ -62,12 +33,28 @@ pub struct TreeArena<Item> {
     parents_map: HashMap<u64, Option<u64>>,
 }
 
+/// A reference type giving shared access to an item and its children.
+///
+/// When you borrow an item from a [`TreeArena`], it returns an `ArenaRef`.
+/// You can iterate over the children to get access to child `ArenaRef` handles.
 pub struct ArenaRef<'a, Item> {
     pub parent_id: Option<u64>,
     pub item: &'a Item,
     pub children: ArenaRefChildren<'a, Item>,
 }
 
+/// A reference type giving mutable access to an item and its children.
+///
+/// When you borrow an item from a [`TreeArena`], it returns an `ArenaMut`.
+/// This struct holds three fields:
+///  - the id of its parent.
+///  - a reference to the item itself.
+///  - an `ArenaMutChildren` handle to access its children.
+///
+/// Because the latter two are disjoint references, you can mutate the node's value
+/// and its children independently without invalidating the references.
+///
+/// You can iterate over the children to get access to child `ArenaMut` handles.
 pub struct ArenaMut<'a, Item> {
     pub parent_id: Option<u64>,
     pub item: &'a mut Item,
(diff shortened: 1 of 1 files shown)


PoignardAzur:
This is work I initially did as part of the "widget tree changes only in mutate passes" part of the pass spec RFC. I decided to split it out for readability.

masonry/src/tree_arena.rs
```diff
@@ -62,12 +33,28 @@ pub struct TreeArena<Item> {
     parents_map: HashMap<u64, Option<u64>>,
 }
 
+/// A reference type giving shared access to an item and its children.
```
DJMcNab, line 36: Maybe `arena item` here and elsewhere?

masonry/src/tree_arena.rs
```diff
@@ -62,12 +33,28 @@ pub struct TreeArena<Item> {
...
+///
+/// When you borrow an item from a [`TreeArena`], it returns an `ArenaRef`.
```
DJMcNab, line 38: Would it be worth pointing to a concrete method here (`TreeArena::find`?)

masonry/src/tree_arena.rs
```diff
@@ -76,22 +63,28 @@ pub struct ArenaMut<'a, Item> {
...
-/// to access its children.
+/// See [`ArenaMut`] for more information.
 pub struct ArenaMutChildren<'a, Item> {
     id: Option<u64>,
     children: &'a mut Vec<TreeNode<Item>>,
+    parents_map: ArenaMapMut<'a>,
+}
+
+#[derive(Clone, Copy)]
+pub struct ArenaMapRef<'a> {
```
DJMcNab, line 83: Would this be clearer if it was just:
```rust
pub struct ArenaMap(HashMap<u64, Option<u64>>);
```
then using `&` and `&mut` references as applicable.

masonry/src/tree_arena.rs
```diff
@@ -138,92 +134,115 @@ impl<Item> TreeArena<Item> {
...
+            },
         }
     }
 
-    // TODO - Move into TreeArenaToken::find method
     /// Find an item in the tree.
     ///
-    /// Returns a tuple of a shared reference to the item, and a token to access
-    /// its children.
+    /// Returns a shared reference to the item.
```
DJMcNab, line 145:
```suggestion
    /// Returns a shared reference to the item if present.
```

masonry/src/tree_arena.rs
```diff
@@ -138,92 +134,115 @@ impl<Item> TreeArena<Item> {
...
+            children: ArenaMutChildren {
+                id: Some(self.id),
+                children: &mut self.children,
+                parents_map: ArenaMapMut { parents_map },
+            },
+        }
+    }
+}
+
+impl<'a, Item> ArenaRef<'a, Item> {
```
DJMcNab, line 213: I think the usual style is to put impls (especially trivial impls) near to the corresponding struct definition.

masonry/src/tree_arena.rs
```diff
@@ -252,14 +271,47 @@ impl<'a, Item> ArenaRefChildren<'a, Item> {
...
+    }
+
+    /// Find an item in the tree.
+    ///
+    /// Returns a shared reference to the item.
+    ///
+    /// ## Complexity
+    ///
+    /// O(Depth). In future implementations, this will be O(1).
+    pub fn find(self, id: u64) -> Option<ArenaRef<'a, Item>> {
```
DJMcNab, line 293: What's the behaviour here if this is passed `id` of `self`?

DJMcNab, line 293: Why does this consume `self`? `ArenaRefChildren` doesn't seem to be `Copy`

masonry/src/tree_arena.rs
```diff
@@ -359,16 +413,17 @@ impl<'a, Item> ArenaMutChildren<'a, Item> {
...
         let child = self.children.remove(i);
-        remove_children(&child, self.parents_map);
+        remove_children(&child, self.parents_map.parents_map);
 
         Some(child.item)
     }
 
     /// Returns a shared token equivalent to this one.
-    pub fn reborrow(&mut self) -> ArenaRefChildren<'_, Item> {
+    pub fn reborrow(&self) -> ArenaRefChildren<'_, Item> {
```
DJMcNab, line 422: Ooh, nice!

DJMcNab (commented):
`parents_map.parents_map` is quite noisy here.

Otherwise, these changes look good - it's nice to see a step down in computational complexity here. Have you noticed any performance impact from this change?

PoignardAzur:
> `parents_map.parents_map` is quite noisy here.

It's fine, it's temporary code. The next step is for parents_map to be passed as a function argument.

> Otherwise, these changes look good - it's nice to see a step down in computational complexity here. Have you noticed any performance impact from this change?

Nope. This PR doesn't change the complexity. That already happened a while ago. This only updates the documentation to reflect the performance change.

And for that matter, we don't keep track of performance *or* have pathological test cases where this would show up. Until we add that, we're mostly blind.

My current proxy for performance is how quickly the unit tests run. There hasn't been a dramatic change this PR. I don't think there was when I introduced the parents map either, because unit tests usually manipulate a small number of widgets.

masonry/src/tree_arena.rs
PoignardAzur: It does. You can grep for `Copy for ArenaRefChildren` to find it.

It's not a derive, because derive creates an implementation conditional on `Item: Copy`, which isn't what we want.

PoignardAzur: It returns None. By construction, ArenaRefChildren can't ever give a reference to the current borrow: they're disjoint. I guess I could document that.

masonry/src/tree_arena.rs
PoignardAzur: I like having all the type declarations together. It's the style the rest of the crate follows.

masonry/src/tree_arena.rs
PoignardAzur: That's not a bad idea in principle, but the idea is that ArenaMapRef represents a capability to the "structure" of the arena. In future PRs it'll be passed as arguments, and in a future arena implementation, it'll be a stateless token.

masonry/src/tree_arena.rs
DJMcNab: I was reviewing this in GitHub's UI, so that wasn't visible.

My expectation was that such a fundamental impl would not be hiding somewhere else...

masonry/src/tree_arena.rs
PoignardAzur: I'm not sure about this, because `TreeArena::find` is more of a syntactic sugar than something your architecture is supposed to rest on.

DJMcNab (approved):
Approving to unblock reliant work

PR #589, opened 2024-09-14
Add temporary passes and remove most of lifecycle method

PoignardAzur:
Add update_focus_chain pass.
Add update_new_widgets pass.
Remove RenderRoot::root_lifecycle.
Move call_widget_method_with_checks out of Widgetpod.

These new passes aren't intended to stay long-term, but are meant to make future refactors easier and more concise.
The other goal is to remove almost all the remaining code in the lifecycle method.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -404,6 +404,84 @@ pub(crate) fn run_update_anim_pass(root: &mut RenderRoot, elapsed_ns: u64) {
 
 // ----------------
 
+fn update_new_widgets(
+    global_state: &mut RenderRootState,
+    mut widget: ArenaMut<'_, Box<dyn Widget>>,
+    mut state: ArenaMut<'_, WidgetState>,
+) {
+    let _span = widget.item.make_trace_span().entered();
+
+    if !state.item.children_changed {
+        return;
+    }
+    state.item.children_changed = false;
+
+    // This will recursively call WidgetPod::lifecycle for all children of this widget,
+    // which will add the new widgets to the arena.
+    {
+        let mut ctx = LifeCycleCtx {
+            global_state,
+            widget_state: state.item,
+            widget_state_children: state.children.reborrow_mut(),
+            widget_children: widget.children.reborrow_mut(),
+        };
+        let event = LifeCycle::Internal(crate::InternalLifeCycle::RouteWidgetAdded);
+        widget.item.lifecycle(&mut ctx, &event);
+    }
+
+    if state.item.is_new {
+        let mut ctx = LifeCycleCtx {
+            global_state,
+            widget_state: state.item,
+            widget_state_children: state.children.reborrow_mut(),
+            widget_children: widget.children.reborrow_mut(),
+        };
+        widget.item.lifecycle(&mut ctx, &LifeCycle::WidgetAdded);
+        trace!(
+            "{} received LifeCycle::WidgetAdded",
+            widget.item.short_type_name()
+        );
+    }
+    state.item.is_new = false;
+
+    // We can recurse on this widget's children, because they have already been added
+    // to the arena above.
+    let id = state.item.id;
+    let parent_state = state.item;
+    recurse_on_children(
+        id,
+        widget.reborrow_mut(),
+        state.children,
+        |widget, mut state| {
+            update_new_widgets(global_state, widget, state.reborrow_mut());
+            parent_state.merge_up(state.item);
+        },
+    );
+}
+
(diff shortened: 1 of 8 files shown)


PoignardAzur:
(FWIW I wrote these commits so they'd be easier to review in isolation. I didn't make two PRs because they'd be annoying to merge.)

masonry/src/passes/update.rs
```diff
@@ -401,3 +401,160 @@ pub(crate) fn run_update_anim_pass(root: &mut RenderRoot, elapsed_ns: u64) {
...
+    let _span = widget.item.make_trace_span().entered();
+    let id = state.item.id;
+
+    if !state.item.update_focus_chain {
+        return;
+    }
+
+    // Replace has_focus to check if the value changed in the meantime
+    state.item.has_focus = global_state.focused_widget == Some(id);
+    let had_focus = state.item.has_focus;
```
DJMcNab, line 503:
```suggestion
    let had_focus = state.item.has_focus;
    state.item.has_focus = had_focus;
```
is equivalent and shows more intentionality.

masonry/src/passes/update.rs
```diff
@@ -401,3 +401,160 @@ pub(crate) fn run_update_anim_pass(root: &mut RenderRoot, elapsed_ns: u64) {
...
+    let _span = info_span!("update_new_widgets").entered();
+
+    if root.root.incomplete() {
+        let mut ctx = LifeCycleCtx {
+            global_state: &mut root.state,
+            widget_state: synthetic_root_state,
+            widget_state_children: root.widget_arena.widget_states.root_token_mut(),
+            widget_children: root.widget_arena.widgets.root_token_mut(),
+        };
+        let event = LifeCycle::Internal(crate::InternalLifeCycle::RouteWidgetAdded);
```
DJMcNab, line 475: Can we rename this to `MaintainArena` or something?

masonry/src/passes/update.rs
```diff
@@ -401,3 +401,160 @@ pub(crate) fn run_update_anim_pass(root: &mut RenderRoot, elapsed_ns: u64) {
...
+    mut state: ArenaMut<'_, WidgetState>,
+) {
+    let _span = widget.item.make_trace_span().entered();
+
+    if !state.item.children_changed {
+        return;
+    }
+    state.item.children_changed = false;
+
+    // This will recursively call WidgetPod::lifecycle for all children of this widget,
```
DJMcNab, line 419: Is this not now only recursive for the "new" children?

DJMcNab (approved):
Some of these changes are a bit hard to understand, but I think this is an improvement in clarity about the role of Lifecycle.

masonry/src/passes/update.rs
PoignardAzur: I think you made a typo, because the code isn't equivalent.

Either way, I think writing `had_focus` second is better in terms of intentionality.

masonry/src/passes/update.rs
PoignardAzur: This code is going away next PR, so I'm not super motivated.

masonry/src/passes/update.rs
PoignardAzur: No, the point is that `Widget::lifecycle` is a "middleman" for calling `WidgetPod::lifecycle` on all children. This is also code that's going away soon, so it's fine if it's a little counter-intuitive.

masonry/src/passes/update.rs
DJMcNab: :laughing: that's a very bad typo. I was aiming to suggest:
```rust
let had_focus = global_state.focused_widget == Some(id);
state.item.has_focus = had_focus;
```
because the current code doesn't *look* like correct code, given resetting a variable, and making a variable called `had_`, but based on the new value.

masonry/src/passes/update.rs
DJMcNab: Oh, so the comment is just not right to say "recursive" at all?

PoignardAzur: I mean, I guess that depends by what you mean by "recursively"? It's a single level of recursion.

DJMcNab: https://github.com/linebender/xilem/pull/589#discussion_r1761183070

---

If the callstack is:
```
update_new_widgets
Widget::lifecycle
WidgetPod::lifecycle
```
and never gets any deeper, that seems pretty non-recursive to me. Maybe I'm misunderstanding the code flows - sometimes things get confusing reviewing code with the +/- lines interleaved.

PR #591, opened 2024-09-16
Add Board widget for absolute positioning

RagibHasin:
Adds a Board widget, view and a corresponding Xilem example.
Also, adds a Shape widget for drawing all built-in Kurbo shapes.


RagibHasin:
Thanks! Didn't know about `OrphanView`. I would definitely look into it.

You're right, it is heavily inspired by Xilem Web's SVG views. But I cannot see how I can make strokes and fills into a wrapper view without a large boilerplate which would be needed anyway. Although I never used `frozen` from Xilem Web, so have to look into it.

Philipp-M:
the `Frozen` view is a `xilem_core` view so both available in xilem (it's e.g. shown in the `memoization` example) as well as xilem_web. It's basically `Memoize` without data (or data == `()`).

Yeah in case of multiple view impls, that itself is somewhat boilerplate indeed, but I do believe trivial enough to live with it or a case for (proc-)macros (but I don't think that's worth it for now at least, I guess it depends on how similar that boilerplate is).

RagibHasin:
After mulling it over, I think I now understand how stroke and fill can be attribute-like views. I will keep you posted on how it goes.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+pub struct Shape {
+    shape: KurboShape,
+    transform: Affine,
+    fill_style: Fill,
+    fill_brush: Brush,
+    fill_brush_transform: Option<Affine>,
+    stroke_style: Stroke,
+    stroke_brush: Brush,
+    stroke_brush_transform: Option<Affine>,
+}
```
PoignardAzur, line 45: This (and the impl blocks) could probably be moved to a separate file.

I suggest `StaticShape` or `DrawnShape` as a name to differentiate from the `Shape` trait.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+            origin: rect.origin(),
+            size: rect.size(),
+        }
+    }
+}
+
+enum Child {
+    Widget(WidgetPod<Box<dyn Widget>>, BoardParams),
+    Shape(WidgetPod<Box<dyn Widget>>),
+}
```
PoignardAzur, line 432: Given that `Shape` implements `Widget`, that enum seems redundant.

Philipp-M (commented):
I wonder, whether we should have a separate `View` derived trait constraining the context.

Something like (I renamed `BoardElement` to a more general `VectorElement`, maybe `SvgElement`/`SvgView` makes also sense):

```rust
trait VectorView<State, Action>: View<State, Action, ViewCtx, Element = Pod<Self::VectorElement> + Send + Sync {
    type VectorElement: VectorElement;
}
```

I'm also not yet sure if an enum will suffice in the future, I would think separate types make sense here, to be more flexible.

For the widget container roughly either something like `struct PositionedElement(impl Widget, Size, Position)` or `struct ForeignElement(impl Widget, Size, Position)`. And then have separate types for the other vector elements, like `Shape` if it's reasonably possible to cover all kurbo shapes with this type, and something like a `VectorGroup` etc.

Having a separate types for vector elements have the advantage to allow specialized views on top of some of these with the type system and be more extensible also outside of xilem/masonry.

regarding the attributes something like this would be possible then:

```rust
trait VectorExt<State, Action>: VectorView<State, Action> {
    fn stroke(self,...) -> Stroke<Self, State, Action>
    where
        Self::Element: WithStroke
    {
    }
}
```

where relevant elements would have this trait implemented (in this case `impl WithStroke for Shape`).
(This is also hinting the extension traits which I think are planned in masonry.)

We could have a separate trait for the vector elements/context, something like:

```rust
trait VectorElement: Widget {
    fn origin(&self) -> Point; // relative to parents transform
    fn size(&self) -> Size;
    // maybe also setters
}
```

to constrain elements, with that we don't need the enum `Child` in masonry, and all the duplicated logic there.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    /// Builder-style method to add a positioned child to the container.
+    pub fn with_child_pod(
+        mut self,
+        widget: WidgetPod<Box<dyn Widget>>,
+        params: impl Into<BoardParams>,
+    ) -> Self {
+        // TODO - dedup?
+        self.children.push(Child::Widget(widget, params.into()));
+        self
```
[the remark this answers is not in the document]
RagibHasin: Yes, it is a copy-pasta. Should I remove this comment or add that it should follow the lead of `Flex` whenever `Flex` changes?

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+// --- MARK: OTHER IMPLS---
+impl BoardParams {
+    /// Create a `BoardParams` with a specific `origin` and `size`.
+    pub fn new(origin: impl Into<Point>, size: impl Into<Size>) -> Self {
+        BoardParams {
+            origin: origin.into(),
+            size: size.into(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
RagibHasin: Will do on the final pass, I think?

masonry/src/widget/board.rs
RagibHasin: It acts as a type-tag for downcasting in https://github.com/RagibHasin/xilem/blob/991f3e5a90e2ad6d927772d942b6939e5b075c47/masonry/src/widget/board.rs#L316 .

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+        match self {
+            Child::Widget(widget, _) | Child::Shape(widget) => Some(widget),
+        }
+    }
+    fn widget(&self) -> Option<&WidgetPod<Box<dyn Widget>>> {
+        match self {
+            Child::Widget(widget, _) | Child::Shape(widget) => Some(widget),
+        }
+    }
+}
```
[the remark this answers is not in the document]
RagibHasin: Yeah, copy-pasta...

RagibHasin:
@PoignardAzur is it possible to get a mutable reference of a widget from a `Pod<W` in `View::build`?

masonry/src/widget/board.rs
RagibHasin: KurboShape the widget seems more reasonable.

PoignardAzur:
I don't think so, but I'm a bit hazy on Xilem's logic.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+    BezPath(BezPath),
+    Circle(Circle),
+    CircleSegment(CircleSegment),
+    CubicBez(CubicBez),
+    Ellipse(Ellipse),
+    Line(Line),
+    QuadBez(QuadBez),
+    Rect(Rect),
+    RoundedRect(RoundedRect),
+}
```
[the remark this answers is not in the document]
PoignardAzur: Let's not block on a new kurbo release for now.

masonry/src/widget/board.rs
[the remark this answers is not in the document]
PoignardAzur: No need for a comment. We'll remove it from Flex in a separate PR.

masonry/src/widget/board.rs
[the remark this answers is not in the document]
PoignardAzur: It's non-blocking. Worst case, I'll re-arrange it in a follow-up PR.

RagibHasin:
> I don't think so, but I'm a bit hazy on Xilem's logic.

Is there a reason for it? Because I think a view should be able to modify its children upon build. Otherwise, things are getting very complicated.

Kurbo shape views and their attribute views would then need a custom pod type, which somehow does not satisfy `Board: WidgetView`.

Anyway, I think it would help to add a method in `xilem::Pod` for getting a mutable reference to the underlying widget with reference to a `ViewCtx`. Would that be acceptable? It would need additional plumbing in Masonry too.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+        bc.debug_check("Board");
+
+        for child in &mut self.children {
+            match child {
+                Child::Widget(widget, params) => {
+                    if ctx.child_needs_layout(widget) {
+                        ctx.run_layout(widget, &BoxConstraints::tight(params.size));
+                    } else {
+                        ctx.skip_layout(widget);
+                    };
```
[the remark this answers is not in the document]
RagibHasin: Will do.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+                    let params = BoardParams::from(
+                        raw_widget
+                            .widget()
+                            .as_ref()
+                            .as_dyn_any()
+                            .downcast_ref::<Shape>()
+                            .unwrap()
+                            .shape()
+                            .bounding_box(),
+                    );
```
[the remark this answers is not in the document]
RagibHasin: I would look into it.

Philipp-M:
> Because I think a view should be able to modify its children upon build. Otherwise, things are getting very complicated.

Another case for the extension traits? AFAIK only `WidgetRef<T>` should not be able to mutate the widget.

I.e. something like `trait ShapeMut` (or separate traits for single properties as described in my previous comment) with all the methods currently in `impl<'a> WidgetMut<'a, Shape> {}` which is implemented for `WidgetMut<'a, Shape>`, `WidgetPod<Shape>` and `Shape` itself. Maybe we could start with this here? (And do a pass for the other widgets/properties later).

> Anyway, I think it would help to add a method in `xilem::Pod` for getting a mutable reference to the underlying widget with reference to a `ViewCtx`.

A `DerefMut` and `Deref` to `WidgetPod` makes sense, I don't see though why it needs `ViewCtx`? It's not dependent on the `ViewCtx` AFAICS?

DJMcNab:
This is essentially running into pass specification being half-completed. Our current plan is to make `WidgetView::build` return an `OrphanPod`, where getting exclusive access would hopefully be relatively trivial.

At a type-system level, Masonry doesn't know that the `WidgetPod` actually still contains the `Widget`. A short-term hacked-in method to allow that access would make sense to me.

RagibHasin:
> A `DerefMut` and `Deref` to `WidgetPod` makes sense, I don't see though why it needs `ViewCtx`? It's not dependent on the `ViewCtx` AFAICS?

An `Inserted` `WidgetPod` requires additional context to resolve into its target widget as essentially it becomes a key to a slot in an arena (Olivier's language, not mine).

But anyway, in a `View::build` the `WidgetPod` has yet to be inserted and thus contains the widget inlined, so in thar case it would be possible to do without any `ViewCtx` and return a `Option<&mut impl Widget>`.

RagibHasin:
> A short-term hacked-in method to allow that access would make sense to me.

I would go that route then.

Philipp-M:
> An Inserted WidgetPod requires additional context to resolve into its target widget as essentially it becomes a key to a slot in an arena (Olivier's language, not mine).

Yes, but not with the `ViewCtx` that doesn't have the widget arena to access the widget in case it's `Inserted` as described by Daniel, so yeah probably just `panic!`, when it's `Inserted`?

RagibHasin:
> > An Inserted WidgetPod requires additional context to resolve into its target widget as essentially it becomes a key to a slot in an arena (Olivier's language, not mine).
> 
> Yes, but not with the `ViewCtx` that doesn't have the widget arena to access the widget in case it's `Inserted` as described by Daniel, so yeah probably just `panic!`, when it's `Inserted`?

I want to leave panicking to the caller, which if used right, should never happen. Otherwise, some unwitting caller may call get mut on `WidgetPod`, and crash the program in some other place.

I actually wanted a reference to a `ViewCtx` as a sanity check for that purpose, but now think that we would still need to check for availability as a `ViewCtx` is available in all view methods.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,633 @@
...
+            self.fill_brush_transform,
+            &self.shape,
+        );
+        scene.stroke(
+            &self.stroke_style,
+            self.transform,
+            &self.stroke_brush,
+            self.stroke_brush_transform,
+            &self.shape,
+        );
```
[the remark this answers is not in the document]
RagibHasin: I'm not very familiar with Vello encoding. Would it bloat the scene too much if most of the shapes use only one of stroke/fill? If it does I would make it optional. cc @DJMcNab .

masonry/src/widget/board.rs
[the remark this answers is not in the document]
RagibHasin: Gone, now.

masonry/src/widget/board.rs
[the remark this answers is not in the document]
RagibHasin: I think it should be kept as is for now. One linebender/kurbo#331 merges, that can be adopted.

masonry/src/widget/board.rs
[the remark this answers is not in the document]
RagibHasin: Okay.

RagibHasin:
> I wonder, whether we should have a separate `View` derived trait constraining the context.

Thanks, @Philipp-M . This solution is far more elegant.

xilem/src/view/board/kurbo_shape.rs
```diff
@@ -0,0 +1,103 @@
...
+
+//! Implementation of the View trait for various kurbo shapes.
+
+use masonry::widget::KurboShape;
+use vello::kurbo;
+use xilem_core::{AnyElement, AnyView, DynMessage, MessageResult, Mut, OrphanView, SuperElement};
+
+use crate::{Pod, ViewCtx, WidgetView};
+
+pub trait GraphicsView<State, Action = ()>: WidgetView<State, Action, Widget = KurboShape> {}
```
Philipp-M, line 12: I would actually really add a new trait as Element bound here as described in my previous comment instead of enforcing only `KurboShape` like a `SvgElement`, `VectorElement` or `GraphicsElement` trait (though I'm not sure if the name "graphics" is a little bit too general for this kind of use).

So that a `PositionedView` (btw. is the name sufficient, as it also sets the size?) is covered as well with this trait, and potentially in the future something like a `SvgGroup` (`<g>` in SVG) etc.

This would simplify the code in `xilem::view::board`, by not needing `AnyBoardChild` and in masonry as well by avoiding this `Child` enum. This needs an additional (trivial) widget though, but I think that's cleaner and more extensible non-the-less.

Here's a *dirty* sketch of what I mean on the masonry side: https://github.com/Philipp-M/xilem/commit/6571b0aaff59e75c3a0a52eb6ee4b442ee5d0ffe

Feel free to use that code as you like.

xilem/src/view/board/style_modifier.rs
```diff
@@ -0,0 +1,342 @@
...
+    fn rebuild<'el>(
+        &self,
+        prev: &Self,
+        child_state: &mut Self::ViewState,
+        ctx: &mut ViewCtx,
+        element: Mut<'el, Self::Element>,
+    ) -> Mut<'el, Self::Element> {
+        let mut element = self.child.rebuild(&prev.child, child_state, ctx, element);
+        if self.transform != prev.transform {
+            element.set_transform(self.transform);
```
Philipp-M, line 165: I believe we want to check if the child views have changed the transform (and similar for other properties) as well, and overwrite it in case.
That's one reason why there's the rather complicated modifier logic in xilem_web so that this does not lead to surprising results (i.e. the most outer modifier always overwrites previous modifiers of the same type).

I think the simplest way is just something like:

```suggestion
        if self.transform != element.transform {
            element.set_transform(self.transform);
```
But that may have the disadvantage of producing unnecessary repaints, when there's other modifiers that transform that property.

We could also use dirty flags etc. to avoid that issue, something like `if self.transform != prev.transform || element.transform_dirty() {}`, or something similar.

Philipp-M (commented):
Yeah that's much closer to what I've imagined for a xilem_svg API :)

Added a few suggestions.

xilem/src/view/board/style_modifier.rs
RagibHasin: How can the properties be changed other than by the parent?

RagibHasin:
I think it is now reasonably complete for a full review.

masonry/src/widget/board.rs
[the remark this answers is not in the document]
RagibHasin: Made it optional anyway.

xilem/src/view/board/kurbo_shape.rs
RagibHasin: Co-opted you code. 😄

xilem/src/view/board.rs
```diff
@@ -0,0 +1,384 @@
...
+    ) -> MessageResult<Action> {
+        self.sequence
+            .seq_message(view_state, id_path, message, app_state)
+    }
+}
+
+pub type AnyBoardView<State, Action = ()> =
+    dyn AnyView<State, Action, ViewCtx, BoardElement> + Send + Sync;
+
+pub struct BoardElement {
```
Philipp-M, line 101: I don't think this extra type is necessary.
I think you can just use `Pod<W: SvgElement>`.
And the `AnyView` could be (see below for more info)

```rust
pub type AnySvgView<State, Action = ()> =
    dyn AnyView<State, Action, ViewCtx, Pod<DynWidget<Box<dyn SvgElement>>>> + Send + Sync;

File: xilem/src/view/board.rs [hunk 13]
```diff
@@ -0,0 +1,384 @@
...
+    element: Pod<Box<dyn SvgElement>>,
+}
+
+pub struct BoardElementMut<'w> {
+    parent: WidgetMut<'w, widget::Board>,
+    idx: usize,
+}
```
Philipp-M, line 108: And I think this could be removed as well, since I think we can just use the `ViewElement` implementation in the top of `xilem` of `Pod<W>`.

I believe we need something like: `impl<W: SvgElement> SuperElement<Pod<W>> for Pod<Box<dyn SvgElement>>`  then instead.

(It could well be that this needs some plumbing on the masonry side, in particular something like `WidgetPod<W: SvgElement>::svg_boxed()`)

The index here is already in the `BoardSplice`, the only other place I see where this is necessary, is for the `AnyElement` impl. But I think we can reuse the slightly modified `DynWidget`, roughly like this:

```rust
pub struct DynWidget<W = Box<dyn Widget>> {
    inner: WidgetPod<W>,
}
```

with `impl SvgElement for DynWidget<Box<dyn SvgElement>>`, which would make all of this more generally useful.

This would make a lot of code below unnecessary too.

xilem/src/view/board.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    fn skip(&mut self, n: usize) {
+        self.idx += n;
+    }
+}
+
+/// An ordered sequence of views for a [`Board`] view.
+/// See [`ViewSequence`] for more technical details.
+pub trait BoardSequence<State, Action = ()>:
+    ViewSequence<State, Action, ViewCtx, BoardElement>
```
Philipp-M, line 258: Same here, this could be with the suggestions above:
```suggestion
pub trait SvgSequence<State, Action = ()>:
    ViewSequence<State, Action, ViewCtx, Pod<Box<dyn SvgElement>>>
```

xilem/src/view/board/style_modifier.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+impl<V, State, Action> Stroke<V, State, Action> {
+    pub fn brush_transform(mut self, brush_transform: Affine) -> Self {
+        self.brush_transform = Some(brush_transform);
+        self
+    }
+}
+
+pub trait GraphicsExt<State, Action>: GraphicsView<State, Action> + Sized {
```
Philipp-M, line 90: I think this is not accurate anymore, since e.g. a `PositionedView` is a `GraphicsView` as well.

So I think there are two ways to solve this and I don't have a strong opinion here either, IME the first option plays better with rust-analyzer though:

1. Separate modifier traits with a super element bound on a class of elements or one element, so something like this (similar as `xilem_web::interfaces::*`):
```rust
pub trait KurboShapeExt<State, Action>: GraphicsView<State, Action, Widget = KurboShape> + Sized {...}
```
2. As Olivier called it "Warehouse" traits, so something like:
```rust
pub trait SvgExt<State, Action>: GraphicsView<State, Action> + Sized {
    fn transform(self, affine: Affine) -> Transform<Self, State, Action>
    where
         // potentially in the future some kind of transform trait bound, maybe it's also general enough to be directly in the `SvgElement` trait
        Self::Element = KurboShape
    {
        transform(self, affine)
    }
    ...
}
```

xilem/src/view/board/style_modifier.rs
Philipp-M: Consider the following:

```rust
Rect::new().transform(inner).transform(outer)
```

when `inner` changes, the new value of this will be the new element value unless `outer` changes as well at the same time, while I would expect, that `outer` overwrites this value regardless.

Philipp-M (commented):
Mostly looks good, there's a few suggestions on the xilem side.

RagibHasin:
Thanks for the guidance in bringing this so far @Philipp-M , but I have to step away from this for a few days.
In the meantime, feel free to push into this branch if you would like to drive this PR forward.

I would be back in two weeks in-sha-Allah.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,464 @@
...
+    // Origin of the widget relative to its parent
+    fn origin(&self) -> Point;
+    // Size of the widget
+    fn size(&self) -> Size;
+
+    // Sets the origin of the widget relative to its parent
+    fn set_origin(&mut self, origin: Point);
+    // Sets the size of the widget
+    fn set_size(&mut self, size: Size);
+}
```
PoignardAzur, line 48: This trait strikes me as a symptom of trying way too hard to be clever in your design. Especially given that some implementations of the trait panic in some methods.

I think you could go pretty far by going a simpler route:

- No SvgElement trait.
- Both widgets and shapes are stored inside PositionedElement.
- PositionedElement doesn't implement Widget.
- The default way to add a shape is to add a PositionedElement with a KurboShape widget, an origin of `shape.bounding_box().origin()` and a size of  `shape.bounding_box().size()`.

masonry/src/widget/board.rs
```diff
@@ -0,0 +1,464 @@
...
+        }
+
+        bc.max()
+    }
+
+    fn paint(&mut self, _ctx: &mut PaintCtx, _scene: &mut Scene) {}
+
+    fn accessibility_role(&self) -> Role {
+        Role::GenericContainer
+    }
```
PoignardAzur, line 188: Open question: is this the right role for this widget, or is there a more specific role for it? I guess "Canvas" doesn't really apply.

masonry/src/widget/kurbo.rs
```diff
@@ -0,0 +1,406 @@
...
+        self.shape.bounding_box().size()
+    }
+
+    fn set_origin(&mut self, _: Point) {
+        panic!("a shape does not support setting its origin after creation")
+    }
+
+    fn set_size(&mut self, _: Size) {
+        panic!("a shape does not support setting its size after creation")
+    }
```
PoignardAzur, line 233: It bears mentioning again: the fact that these trait methods panic is a code smell. To be clear, I'm not saying they should be rewritten to not panic. Rather, this code is a hint that the broader design should be reconsidered.

masonry/src/widget/kurbo.rs
```diff
@@ -0,0 +1,406 @@
...
+    fn on_status_change(&mut self, _ctx: &mut LifeCycleCtx, _event: &StatusChange) {}
+    fn lifecycle(&mut self, _ctx: &mut LifeCycleCtx, _event: &LifeCycle) {}
+
+    fn layout(&mut self, _ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
+        let size = self.shape.bounding_box().size();
+        if !bc.contains(size) {
+            warn!("The shape is oversized");
+        }
+        size
+    }
```
PoignardAzur, line 179: You probably shouldn't warn every time a widget overflows its parent.

Also, this code is valid with the Board as a parent, but I'm not sure what results it would produce with any widget type as a parent. It's not a blocking problem, though.

masonry/src/widget/kurbo.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    fn paint(&mut self, _ctx: &mut PaintCtx, scene: &mut Scene) {
+        let transform = self
+            .transform
+            .then_translate(-self.shape.bounding_box().origin().to_vec2());
```
PoignardAzur, line 184: Same as above.

PoignardAzur (changes requested):
Overall, I'm worried about the amount of code in this PR which does nothing but delegate to an internal implementation. It's a major code smell, and a hint that you're using too many layers of abstractions in your design.

masonry/src/widget/board.rs
Philipp-M: > This trait strikes me as a symptom of trying way too hard to be clever in your design.

That's mostly on me, the reason why I was pushing towards this, is to have strongly typed "border", to enforce only the "specialized" class of "svg" widgets within this kind of context (i.e. have a tree-hierarchy in near/mid-future).

I'd agree here, when this would only be one layer (similar as `Flex` etc.), but my plan with all of this is to have a completely separate (specialized) widget tree for the svg implementation.

I'll take a closer look in the following week and may use this PR as basis for a (xilem_)svg implementation.

> PositionedElement doesn't implement Widget.

The way I see `PositionedElement` is [`<foreignObject>`](https://developer.mozilla.org/en-US/docs/Web/API/SVGForeignObjectElement), and with all the previously said, I think it makes sense to have this implement widget as well.

xilem/src/view/board/style_modifier.rs
Philipp-M: Hmm this is definitely not trivial to get right (i.e. idiomatic, robust and as little extra-code as possible).

[This](https://github.com/linebender/xilem/commit/66cedfe9cc28116d94bcafced74caa140b36e2f2) is a dirty/hacky sketch of how to make this work, but I'm not really happy about it, as it requires boolean dirty flags for every property, and is not optimal when reading values (`View::State` can be used though for this issue, to cache previously set values), when considering a potential future API like:

```rust
Rect::new().rotate().translate()
```
I.e. that views read previously set values (in this case the transform from `rotate` in `translate`).

In xilem_web this is less of an issue, as a minimal changeset needs to be computed anyway (accessing DOM via js is expensive), so that justifies the modifier API there, which among other things accomplishes this.

So in that regard a big struct is more convenient (but requires wrapper types). I'm not sure how future-proof that is either, when considering a lot of attributes per widget (i.e. unnecessary memory usage when using the view), and to my thinking, this general pattern (composing views with new types based on trait-bounds on `View::Element`) *should* be possible.

@PoignardAzur and @DJMcNab do you have any suggestions here?

RagibHasin:
So, I'm back and after some idle thoughts about this, I think we should reach a consensus about design first.

@PoignardAzur 's concern about unnecessary complexity involving `SvgElement` trait comes from @Philipp-M 's wish to replicate the paradigm of SVG more-or-less completely.
@Philipp-M also mentioned the SVG `<g>` element as a motivating example.
But I think a corresponding `group` view would result in more than one `SvgElement` in masonry and we should treat it as a `ViewSequence`. We can then have the attribute methods directly on that `group` view. But I think that may create problems with nested views.

I propose that we explicitly support [SVG shape elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Element#shape_elements) only along with any absolutely placed regular widget (equivalent to `<foregnElement>`) and design our code accordingly. Most else are emualable in Xilem reactively I think. Also, it would allow us to remove the `SvgElement` machinery.

Philipp-M:
See also [Add `Affine` transform to `Widget` trait?](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Add.20.60Affine.60.20transform.20to.20.60Widget.60.20trait.3F/near/472076600).

Or what I mean with it, is that we may get away without introducing an extra svg context, by making this board view the `zstack` (in SwiftUI terms). I.e. that widgets just take the full width/height by default of the `zstack`, but can be sized individually by something like `SizedBox` and positioned by a transform. That may simplify everything. I'm not sure if I missed something where a dedicated SVG context would make sense otherwise.

But I agree, reaching consensus is a good idea.

RagibHasin:
Yeah, I've seen it.

My understanding is that to enable arbitrary transform for any widget, we would need to modify how paint works now. Does Masonry currently allows a parent to draw its children manually, in stead of doing in in the framework?

RagibHasin:
And the code as it sits, kind of does what you described with getting the entirety of the board to draw, translates, sized-boxing just not the other affine transforms, as I think it would require the parent to draw the child.

Philipp-M:
> My understanding is that to enable arbitrary transform for any widget, we would need to modify how paint works now. 

Not just that, it also needs to transform pointer events, filtering of it etc. When supporting a z-value in the transform (probably not in the first iteration though), it gets even more complicated. This'll likely be a more complicated change to get it right, but I think worth it.

RagibHasin:
I think these features have been sitting in all our personal wishlists 😄.

> I think worth it.

I, too, think so.

But for the MVP here, which in my vision includes only sized-boxing and translation, I think we should remove the `SvgElement` abstraction but keep type erased `Child` in the `Board` widget.

How can a consensus be reached here?

cc @DJMcNab @PoignardAzur

RagibHasin:
#737 and #753 fulfills the concerns of this PR except for making Kurbo shapes into `WidgetView`s.

However that has been deemed low priority and thus this I'm closing this PR.

PR #592, opened 2024-09-16
xilem_web checkbox checked attr

burrbull:
Only `.attr("checked", None)` is valid value for unchecked.

### DIFF
--- a/xilem_web/web_examples/fetch/src/main.rs
+++ b/xilem_web/web_examples/fetch/src/main.rs
@@ -188,7 +188,7 @@ fn cat_fetch_controls(state: &AppState) -> impl Element<AppState> {
                 td(input(())
                     .id("reset-debounce-update")
                     .attr("type", "checkbox")
-                    .attr("checked", state.reset_debounce_on_update)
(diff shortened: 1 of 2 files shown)


Philipp-M:
Hmm does something not work as expected?

There's:

```rust
impl IntoAttributeValue for bool {
    fn into_attr_value(self) -> Option<AttributeValue> {
        self.then_some(AttributeValue::True)
    }
}
```

So I believe this should result in the same?

Philipp-M:
Generally we want to support the DOM attribute `HtmlInputElement::checked` at some time, to avoid hacks such as [this](https://github.com/linebender/xilem/blob/01cca4610c58761c06619ac5a8ff2d0fbdb36275/xilem_web/src/attribute.rs#L83).

burrbull:
The issue is how browser processes "checked" attribute. Any value including "false" it is considering as "enabled".
Only absent attribute is "disabled".

Philipp-M:
Yes I know that's why `AttributeValue::True` results in the attribute value `""` which is evaluated to true, while the boolean `false` removes that attribute alltogether. But this is a half-baked way to achieve this (as described in my previous comment)

PR #593, opened 2024-09-16
Fix tab focus when no widget is currently focused

PoignardAzur:


Philipp-M (approved):
Makes sense and LGTM

masonry/src/render_root.rs
```diff
@@ -544,31 +544,33 @@ impl RenderRoot {
...
+            let len = self.focus_chain().len();
+            let new_idx = if forward {
+                (idx + 1) % len
+            } else {
+                (idx + len - 1) % len
+            };
+            Some(self.focus_chain()[new_idx])
+        } else {
+            // If no widget is currently focused
+            // If the currently focused widget isn't in the focus chain,
```
DJMcNab, line 566: Or?

PoignardAzur, line 566:
```suggestion
            // If no widget is currently focused or the
            // currently focused widget isn't in the focus chain,
```

DJMcNab:
I'm giving this another scan as part of https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Focus.20chain.20for.20tabbing, and it's not actually clear to me how this logic is any different - I agree that this new code is clearer, but I'm finding it hard to see where the supposed difference in behaviour comes from

PoignardAzur:
> I agree that this new code is clearer, but I'm finding it hard to see where the supposed difference in behaviour comes from

Basically the old code was "If there is not focused widget return None. Otherwise, look for the next item in the focus chain, or the first/last if it can't be found". This PR rerouted the first case (return None) into the last (return first/last).

DJMcNab:
Thank you for confirming. I agree that that is the change which was made.

I missed that the chain was `a.and_then(|a| b(a).or_else(||c))`

PR #594, opened 2024-09-16
xilem_web: Optimize element casting, and intern often used strings (e.g. "div")

Philipp-M:
Some micro-optimizations, e.g. avoids a js call `instanceof` (via `dyn_into`/`dyn_ref`) when constructing elements.

Leads to a small speed increase when constructing/visiting elements (roughly 2%) and more importantly a leaner wasm binary (around 5 %), respectively tested with the js-framework-benchmark suite.

### DIFF
--- a/xilem_web/src/class.rs
+++ b/xilem_web/src/class.rs
@@ -149,7 +149,7 @@ impl Classes {
             // Svg elements do have issues with className, see https://developer.mozilla.org/en-US/docs/Web/API/Element/className
             if element.dyn_ref::<web_sys::SvgElement>().is_some() {
                 element
-                    .set_attribute("class", &self.class_name)
+                    .set_attribute(wasm_bindgen::intern("class"), &self.class_name)
                     .unwrap_throw();
             } else {
                 element.set_class_name(&self.class_name);
--- a/xilem_web/Cargo.toml
+++ b/xilem_web/Cargo.toml
@@ -179,3 +179,5 @@ features = [
 [features]
 default = ["hydration"]
 hydration = []
+# This interns some often used strings, such as element tags ("div" etc.), which slightly improves performance when creating elements at the cost of a bigger wasm binary
+intern_strings = ["wasm-bindgen/enable-interning"]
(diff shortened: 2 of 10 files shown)


xilem_web/src/class.rs
```diff
@@ -149,7 +149,7 @@ impl Classes {
             // Svg elements do have issues with className, see https://developer.mozilla.org/en-US/docs/Web/API/Element/className
             if element.dyn_ref::<web_sys::SvgElement>().is_some() {
                 element
-                    .set_attribute("class", &self.class_name)
+                    .set_attribute(wasm_bindgen::intern("class"), &self.class_name)
```
flosse, line 152: That's interesting, I didn't know that this exists!
It sound's great :relaxed: 

> When you send that cached string to JS, it will look it up in the cache, which completely avoids the O(n) copy and encoding. This has a significant speed boost (as high as 783%)!

flosse (approved):

xilem_web/src/class.rs
flosse: BTW:
You did not enable the "enable-interning" feature, so maybe you should add a comment somewhere so that someone who want to have this optimization has a hint?

Philipp-M: > You did not enable the `"enable-interning"` feature

Oh lol, I guess i missed that in the docs, so my bench was just testing the unchecked casting.

I also just learned from that feature, and have hoped, that it would make something like sledgehammer less attractive (not so much it seems, or we have more room for improvement somewhere else (my guess are allocations)).

I ran this again, with the feature enabled, and the advantages are marginal, and only really visible, when not using templates in the benchmark (which makes sense, since the elements are created only once with `templated` and then cloned via `clone_node_with_deep`). With templates disabled, it's roughly 5% faster in the create 1k/10k bench...

So since this doesn't really improve the performance significantly, I'll just leave the feature disabled by default, as in this case this slightly increases the wasm bloat (compressed roughly 2kb in that bench...), but when it's enabled, by the user, it doesn't hurt, or at least makes everything *slightly* faster, so I'll leave it in.

I've added an optional feature, which just enables it in wasm-bindgen, with a comment on top of it. I'll try to document the features (like `hydration` etc.) at some point in a more public fashion, so that they're more discoverable.

Philipp-M:
Thanks for the review!

PR #596, opened 2024-09-17
Support the AccessKit focus and blur actions 

mwcampbell:
I believe this is necessary to activate and deactivate input focus with TalkBack on Android. Assistive technologies on other platforms can also request the focus action, though none of the AccessKit backends support the blur action yet.

### DIFF
--- a/masonry/src/passes/accessibility.rs
+++ b/masonry/src/passes/accessibility.rs
@@ -46,6 +46,7 @@ fn build_accessibility_tree(
             rebuild_all,
             scale_factor,
         };
+        set_common_properties(&mut ctx);
         widget.item.accessibility(&mut ctx);
 
         let id: NodeId = ctx.widget_state.id.into();
@@ -99,24 +100,28 @@ fn build_access_node(widget: &dyn Widget, state: &WidgetState, scale_factor: f64
             .collect::<Vec<NodeId>>(),
     );
 
-    if state.is_hot {
-        node.set_hovered();
-    }
-    if state.is_disabled {
-        node.set_disabled();
-    }
-    if state.is_stashed {
-        node.set_hidden();
-    }
-    if state.clip.is_some() {
-        node.set_clips_children();
-    }
-    if state.in_focus_chain {
-        node.add_action(accesskit::Action::Focus);
-        node.add_action(accesskit::Action::Blur);
-    }
-
     node
+}
+
+fn set_common_properties(ctx: &mut AccessCtx) {
+    if ctx.is_hot() {
+        ctx.current_node().set_hovered();
+    }
+    if ctx.is_disabled() {
+        ctx.current_node().set_disabled();
+    }
+    if ctx.is_stashed() {
+        ctx.current_node().set_hidden();
+    }
+    if ctx.widget_state.clip.is_some() {
+        ctx.current_node().set_clips_children();
+    }
+    if ctx.is_in_focus_chain() && !ctx.is_disabled() {
+        ctx.current_node().add_action(accesskit::Action::Focus);
+    }
+    if ctx.is_focused() {
+        ctx.current_node().add_action(accesskit::Action::Blur);
+    }
 }
 
 fn to_accesskit_rect(r: Rect, scale_factor: f64) -> accesskit::Rect {
--- a/masonry/src/passes/accessibility.rs
+++ b/masonry/src/passes/accessibility.rs
@@ -111,6 +111,10 @@ fn build_access_node(widget: &dyn Widget, state: &WidgetState, scale_factor: f64
     if state.clip.is_some() {
         node.set_clips_children();
     }
+    if state.in_focus_chain {
+        node.add_action(accesskit::Action::Focus);
+        node.add_action(accesskit::Action::Blur);
+    }
 
     node
 }
--- a/masonry/src/passes/event.rs
+++ b/masonry/src/passes/event.rs
@@ -195,6 +195,7 @@ pub(crate) fn root_on_access_event(
         event,
         false,
         |widget, ctx, event| {
+            // TODO - Split into "access_event_focus" pass or something similar.
             if event.target == ctx.widget_id() {
                 match event.action {
                     accesskit::Action::Focus => {
--- a/masonry/src/passes/event.rs
+++ b/masonry/src/passes/event.rs
@@ -195,6 +195,23 @@ pub(crate) fn root_on_access_event(
         event,
         false,
         |widget, ctx, event| {
+            if event.target == ctx.widget_id() {
+                match event.action {
+                    accesskit::Action::Focus => {
+                        if ctx.is_in_focus_chain() && !ctx.is_disabled() && !ctx.is_focused() {
+                            ctx.request_focus();
+                        }
+                        return;
+                    }
+                    accesskit::Action::Blur => {
+                        if ctx.is_focused() {
+                            ctx.resign_focus();
+                        }
+                        return;
+                    }
+                    _ => {}
+                }
+            }
             widget.on_access_event(ctx, event);
         },
     );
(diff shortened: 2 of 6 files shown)


masonry/src/widget/textbox.rs
```diff
@@ -220,7 +220,22 @@ impl Widget for Textbox {
         }
     }
 
-    fn on_access_event(&mut self, _ctx: &mut EventCtx, _event: &AccessEvent) {
+    fn on_access_event(&mut self, ctx: &mut EventCtx, event: &AccessEvent) {
+        if event.target == ctx.widget_id() {
```
DJMcNab, line 224: I don't think this check should be necessary any more? Is this because of bubbling?

DJMcNab (commented):
Should all of this behaviour be implicit for an item being in the focus chain?

mwcampbell:
It certainly would be better for the framework to support this automatically for any focusable widget. I'll see if there's an easy way to do that in the current architecture.

mwcampbell:
To let the framework automatically handle the AccessKit focus and blur actions, it seems to me that we should add a flag to `WidgetState` to indicate whether the widget is in the focus chain, so `build_access_node` can check that flag. Does anyone have a different idea?

DJMcNab:
Yeah, I think that makes sense. Don't forget to reset that flag in the `update_focus` pass.

I know that our logic for focus has not been fully reasoned about, so a short-term change to unlock this functionality should be fine. (In particular, I don't think we know how setting a custom 'tab-index' will work)

masonry/src/widget/textbox.rs
mwcampbell: I copied that check from the implementation of `on_access_event` in `Button`.

mwcampbell: Having said that, if this event does bubble in Masonry, then I believe the check is still needed.

mwcampbell:
OK, now the framework generically supports this for any widget in the focus chain.

masonry/src/widget/textbox.rs
DJMcNab: Yeah. I think it isn't needed in this case specifically because TextBox has no child widgets, but in general it would be needed.

masonry/src/passes/event.rs
```diff
@@ -195,6 +195,23 @@ pub(crate) fn root_on_access_event(
         event,
         false,
         |widget, ctx, event| {
+            if event.target == ctx.widget_id() {
+                match event.action {
+                    accesskit::Action::Focus => {
+                        if ctx.is_in_focus_chain() && !ctx.is_disabled() && !ctx.is_focused() {
+                            ctx.request_focus();
+                        }
+                        return;
```
DJMcNab, line 204: There's an argument to be made that this return should be conditional only on `ctx.is_in_focus_chain()`, because that's when the corresponding implicit focus action was added.

However, I think that's probably YAGNI

masonry/src/passes/event.rs
```diff
@@ -195,6 +195,23 @@ pub(crate) fn root_on_access_event(
         event,
         false,
         |widget, ctx, event| {
+            if event.target == ctx.widget_id() {
+                match event.action {
+                    accesskit::Action::Focus => {
+                        if ctx.is_in_focus_chain() && !ctx.is_disabled() && !ctx.is_focused() {
+                            ctx.request_focus();
```
DJMcNab, line 202: I think we should probably also call `ctx.set_handled()` here. See [#masonry>Missing event `set_handled` calls](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Missing.20event.20.60set_handled.60.20calls)

masonry/src/passes/accessibility.rs
```diff
@@ -111,6 +111,10 @@ fn build_access_node(widget: &dyn Widget, state: &WidgetState, scale_factor: f64
     if state.clip.is_some() {
         node.set_clips_children();
     }
+    if state.in_focus_chain {
+        node.add_action(accesskit::Action::Focus);
+        node.add_action(accesskit::Action::Blur);
```
DJMcNab, line 116: I presume it's meaningful to have the blur action available on an already unfocused widget?

DJMcNab (approved):
Thanks, this is looking good.

I've not tested this myself

masonry/src/passes/accessibility.rs
mwcampbell: No, it's not. But can I check whether the widget is focused without passing global state into this function?

DJMcNab: As far as I can see, the global state is available in the calling context, so passing it in is fine.
I'd probably shift this code around so that the functionality to set all of these properties is passed the `AccessCtx`.

mwcampbell: OK, I think I just did what you suggested.

mwcampbell:
OK, I think my last push addresses all remaining review comments.

mwcampbell:
@DJMcNab @PoignardAzur Can you please give this one more look before I merge?

PoignardAzur:
Will do tomorrow.

masonry/src/passes/accessibility.rs
```diff
@@ -99,20 +100,28 @@ fn build_access_node(widget: &dyn Widget, state: &WidgetState, scale_factor: f64
             .collect::<Vec<NodeId>>(),
     );
 
-    if state.is_hot {
-        node.set_hovered();
```
PoignardAzur, line 103: Why remove this one?

masonry/src/passes/accessibility.rs
```diff
@@ -99,20 +100,28 @@ fn build_access_node(widget: &dyn Widget, state: &WidgetState, scale_factor: f64
...
-        node.set_clips_children();
+    if ctx.widget_state.clip.is_some() {
+        ctx.current_node().set_clips_children();
+    }
+    if ctx.is_in_focus_chain() && !ctx.is_disabled() {
+        ctx.current_node().add_action(accesskit::Action::Focus);
+    }
+    if ctx.is_focused() {
+        ctx.current_node().add_action(accesskit::Action::Blur);
     }
```
PoignardAzur, line 124: Is there a reason this is split out from `build_access_node`?

masonry/src/passes/event.rs
```diff
@@ -195,6 +195,25 @@ pub(crate) fn root_on_access_event(
         event,
         false,
         |widget, ctx, event| {
+            if event.target == ctx.widget_id() {
+                match event.action {
+                    accesskit::Action::Focus => {
```
PoignardAzur, line 200: Ideally, I'd like this whole block to be split out into a separate pass, to avoid having logic branches inside pass branch like we did with WidgetPod.

It's not a blocking concern. This code is fine for now, though I'd want to add a TODO comment.
```suggestion
            // TODO - Split into "access_event_focus" pass or something similar.
            if event.target == ctx.widget_id() {
                match event.action {
                    accesskit::Action::Focus => {
```

PoignardAzur (approved):
Looks good to me.

masonry/src/passes/accessibility.rs
DJMcNab: ... https://github.com/linebender/xilem/pull/596#discussion_r1765187247

Essentially, build_access_node doesn't have the AccessCtx, which means it would either need to duplicate the logic of the context methods or not have this functionality implemented properly.

masonry/src/passes/event.rs
DJMcNab: In its current state, I agree. This behaviour makes it really hard to have custom handling of the Focus/Blur events anyway, so a hyper-targetted pass makes sense.
That would also be `O(1)` in the blur case, as it could just check if the target of the blur is focused (although in the focus case you need to re-confirm that it is focusable (in case of race conditions?))

See also https://github.com/linebender/xilem/pull/596#discussion_r1764998676 - if we had that, a separate pass would be invalid.

DJMcNab: To be clear, that's emphatically not needed to land this pr, and I'd prefer to land it with just this comment as-is

PoignardAzur (approved):
LGTM.

PR #597, opened 2024-09-18
Reserve ability to create Pod to ViewCtx

PoignardAzur:
This is another intermediary PR for the "restrict widget creation to Mutate pass" feature.

The basic idea is that, in the near future, it will be impossible to create a WidgetPod without a handle to the arena. In my current WIP code, that handle is passed through ViewCtx. Therefore, this PR changes all the sites in xilem that create a Pod and has them use a ViewCtx method instead.

I've tested most xilem examples and they all worked (except for variable_clock, which currently panics in the last main commit).

### DIFF
--- a/xilem/src/any_view.rs
+++ b/xilem/src/any_view.rs
@@ -25,12 +25,12 @@ use crate::{Pod, ViewCtx};
 pub type AnyWidgetView<State, Action = ()> =
     dyn AnyView<State, Action, ViewCtx, Pod<DynWidget>> + Send + Sync;
 
-impl<W: Widget> SuperElement<Pod<W>> for Pod<DynWidget> {
-    fn upcast(child: Pod<W>) -> Self {
-        WidgetPod::new(DynWidget {
-            inner: child.inner.boxed(),
+impl<W: Widget> SuperElement<Pod<W>, ViewCtx> for Pod<DynWidget> {
+    fn upcast(ctx: &mut ViewCtx, child: Pod<W>) -> Self {
+        let boxed_pod = ctx.boxed_pod(child);
+        ctx.new_pod(DynWidget {
+            inner: boxed_pod.inner.boxed(),
         })
-        .into()
     }
 
     fn with_downcast_val<R>(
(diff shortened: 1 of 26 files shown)


PoignardAzur:
Note for reviewers: the most interesting changes are in:

- [xilem_core/tests/one_of.rs](https://github.com/linebender/xilem/pull/597/files#diff-b8e7fd9cb7aa0fbfb7e0e97abeb8e9fd65df529f9b4ea78f21d4be876ae787be)
- [xilem_core/src/element.rs](https://github.com/linebender/xilem/pull/597/files#diff-cfb546ba9441497c293a47d75ecad469cfa85063e6d78d66a981f19cfaf87c68)
- [xilem/src/lib.rs](https://github.com/linebender/xilem/pull/597/files#diff-496fbee6e50f4ac35bad3dcd267d1c367cdb255f9ea9de7b42719f52c9802529)

Everything else is plumbing those API changes into method implementations.

Philipp-M:
Oh, well, I wanted to start a (zulip-)topic already about having the `ViewCtx` "everywhere", good to see that we're converging on similar ideas I guess.
I have started something very similar here already (I haven't access to my PC where I've implemented it, I forgot to push it, I'll link to it when I have access later), it's basically about:

Have access to `ViewCtx` everywhere where the element is forwarded/created in Xilem (that means in SuperElement as done here, and as well in `ElementSplice` which I think will be useful for keyed viewsequences as well as more efficient tree updates in xilem_web)

Philipp-M (approved):
So this is my [branch](https://github.com/linebender/xilem/compare/main...Philipp-M:xilem:ctx-everywhere), looks quite similar, different ordering of the parameters, and access to the `ViewCtx` in `SuperElement::with_downcast_val`, I guess that can be merged somewhat.
But since your PR is more targeted, probably yours first, I'll do that afterwards.
I think generally it makes sense to also have access to the context in `SuperElement::with_downcast_val`, but it introduces a little bit complexity, though I expect this to be useful (I'm experimenting with this currently in xilem_web).

> except for variable_clock, which currently panics in the last main commit

I think you should open an issue then, it works for me (on this branch as well as main).


Otherwise looks good to me.

PoignardAzur:
> I think you should open an issue then, it works for me (on this branch as well as main).

It's fixed now.

PR #598, opened 2024-09-18
Fix accessibility crash with stashed children

PoignardAzur:


DJMcNab (approved):
For reference, this fixes [#xilem>Portal examples currently get AccessKit panic](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Portal.20examples.20currently.20get.20AccessKit.20panic)

PR #599, opened 2024-09-19
Clean up IME handling code

PoignardAzur:
Should fix the way repaints were requested every paint on some platforms.

This PR unfortunately comes with a lack of tests. TestHarness currently doesn't have good support for testing signals, but even if it did, this is an inherently visual feature that we don't really have a way to test visually.

I don't know how to trigger IME on my machine. (I'm using PopOS, a Debian variant.) If someone wants to either test this or help me get to speed, I'd appreciate it.

### DIFF
--- a/masonry/src/widget/widget_state.rs
+++ b/masonry/src/widget/widget_state.rs
@@ -65,7 +65,7 @@ pub struct WidgetState {
     // TODO - Remove
     pub(crate) is_portal: bool,
 
-    /// Tracks whether widget is elligible for IME events.
+    /// Tracks whether widget is eligible for IME events.
     /// Should be immutable after `WidgetAdded` event.
     pub(crate) is_text_input: bool,
     /// The area of the widget that is being edited by
--- a/masonry/src/widget/widget_state.rs
+++ b/masonry/src/widget/widget_state.rs
@@ -6,7 +6,6 @@
 use std::sync::atomic::{AtomicBool, Ordering};
 use vello::kurbo::{Insets, Point, Rect, Size, Vec2};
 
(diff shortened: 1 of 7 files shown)


PoignardAzur:
I just realized Textbox never actually calls register_as_text_input. Oops.

This is yet one more thing I'd like to test, but expressing it in a test is kinda awkward.

DJMcNab:
I'll review when that is fixed.

PoignardAzur:
Should be fixed, but I haven't tested it locally.

masonry/src/widget/widget_state.rs
```diff
@@ -63,9 +62,16 @@ pub struct WidgetState {
     /// the baseline. Widgets that contain text or controls that expect to be
     /// laid out alongside text can set this as appropriate.
     pub(crate) baseline_offset: f64,
-    // TODO - Document
+    // TODO - Remove
     pub(crate) is_portal: bool,
 
+    /// Tracks whether widget is eligible for IME events.
+    /// Should be immutable after `WidgetAdded` event.
```
DJMcNab, line 69: Why? I mean I realise that there is no way to unset this, but are there scenarios where we'd want this to change?

DJMcNab (approved):
I'm seeing this work. One thing I am running into is that `to_do_list` is extremely laggy, but that's still the case before this PR.

masonry/src/widget/widget_state.rs
PoignardAzur: Basically I'd like to be able to write our logic with the assumption that things like "is this focusable", "is this text-editable", etc, are immutable properties of a widget and don't need to be re-checked after creation.

PR #600, opened 2024-09-19
Re-arrange `flex.rs` file to match format of other widgets

PoignardAzur:


DJMcNab (commented):
Github's diff view makes it extremely hard to determine if there's any change in behaviour here, so I'm choosing to defer reviewing this.

I also disagree with the primary change of not having the `Flex` and `WidgetMut<Flex>`  impls be next to each other. The other re-orderings are probably an improvement.

PoignardAzur:
For what it's worth I did strict cut-and paste, there really shouldn't be behavior changes.

DJMcNab (approved):

PR #601, opened 2024-09-19
Implement stashed state

PoignardAzur:
Add update_stashed pass
Add is_still_interactive method

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -77,6 +77,14 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
     let pointer_pos = root.last_mouse_pos.map(|pos| (pos.x, pos.y).into());
 
     // -- UPDATE HOVERED WIDGETS --
+    // If the capture target is disabled, stashed or removed, we set
+    // the capture target id to None.
+    if let Some(id) = root.state.pointer_capture_target {
+        if !root.is_still_interactive(id) {
+            // TODO - Send PointerLeave event
+            root.state.pointer_capture_target = None;
+        }
+    }
 
     let mut next_hovered_widget = if let Some(pos) = pointer_pos {
         // TODO - Apply scale?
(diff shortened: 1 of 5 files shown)


masonry/src/passes/update.rs
```diff
@@ -77,6 +77,14 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot, root_state: &mut Wi
     let pointer_pos = root.last_mouse_pos.map(|pos| (pos.x, pos.y).into());
 
     // -- UPDATE HOVERED WIDGETS --
+    // If the capture target is disabled, stashed or removed, we set
+    // the capture target id to None.
```
DJMcNab, line 81:
```suggestion
    // Release the pointer capture if it no longer makes sense.
```
The previous comment was just exactly restating the code.

DJMcNab (approved):
I am finding it quite hard to work out what concrete use case stashing fulfills - is it for when a widget scrolls out of the viewport? I see it's currently only used for scrollbars.

That being said, this change itself looks like a reasonable implementation of the concept.

masonry/src/passes/update.rs
```diff
@@ -319,6 +322,50 @@ pub(crate) fn run_update_disabled_pass(root: &mut RenderRoot) {
...
+
+    let stashed = state.item.is_explicitly_stashed || parent_stashed;
+    if !state.item.needs_update_stashed && stashed == state.item.is_stashed {
+        return;
+    }
+
+    if stashed != state.item.is_stashed {
+        // TODO - Send update event
+        state.item.is_stashed = stashed;
+        state.item.update_focus_chain = true;
```
DJMcNab, line 343: Does this need to also request a repaint?

PoignardAzur:
Right, it's not documented yet.

Stashed is for widgets that are no longer "part of the graph". So they can't get keyboard events, don't get painted, etc, but should keep some state.

The stereotypical use case would be the contents of hidden tabs in a "tab group" widget.

Scrolled-out widgets are *not* stashed, which we should document. I'll probably update the pass RFC too.

jaredoconnell:
What optimizations are planned for stashed widgets?

> Scrolled-out widgets are _not_ stashed, which we should document. I'll probably update the pass RFC too.

Would this also not apply to widgets in a virtual list, where the widgets are otherwise removed if scrolled out of view? I'm assuming that it'd be more efficient for some widgets to leave them cached instead of immediately deleting them, since the layout could be retained if cached.

PoignardAzur:
It would *probably* not apply. The idea of a scrolled widget list is that it cheats somewhat, so we'll see when we implement it.

The rule of thumb for whether a widget should be stashed would be something like "in a browser, would you expect it to not be reached by Ctrl+F?"

jaredoconnell (approved):
Looks good, aside from Daniel's comments.

masonry/src/passes/update.rs
PoignardAzur: Good catch.

PR #602, opened 2024-09-19
Replace RouteWidgetAdded event with new register_children method

PoignardAzur:
Make lifecycle method optional.
Remove InternalLifecycle.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -9,7 +9,7 @@ use tracing::{info_span, trace};
 use crate::passes::{merge_state_up, recurse_on_children};
 use crate::render_root::{RenderRoot, RenderRootSignal, RenderRootState};
 use crate::tree_arena::ArenaMut;
-use crate::{LifeCycle, LifeCycleCtx, StatusChange, Widget, WidgetId, WidgetState};
+use crate::{LifeCycle, LifeCycleCtx, RegisterCtx, StatusChange, Widget, WidgetId, WidgetState};
 
 fn get_id_path(root: &RenderRoot, widget_id: Option<WidgetId>) -> Vec<WidgetId> {
     let Some(widget_id) = widget_id else {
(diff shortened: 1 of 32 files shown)


masonry/src/passes/update.rs
```diff
@@ -416,17 +416,14 @@ fn update_new_widgets(
...
             widget_children: widget.children.reborrow_mut(),
         };
-        let event = LifeCycle::Internal(crate::InternalLifeCycle::RouteWidgetAdded);
-        widget.item.lifecycle(&mut ctx, &event);
+        // The widget will call `RegisterCtx::register_child` on all its children,
+        // which will add the new widgets to the arena.
+        widget.item.register_children(&mut ctx);
     }
 
     if state.item.is_new {
```
DJMcNab, line 429: Just scanning here, does this mean that parents get their `WidgetAdded` before their children?

I'm not sure which way around is best here...

masonry/src/widget/image.rs
```diff
@@ -77,6 +77,8 @@ impl Widget for Image {
 
     fn on_access_event(&mut self, _ctx: &mut EventCtx, _event: &AccessEvent) {}
 
+    fn register_children(&mut self, _ctx: &mut RegisterCtx) {}
+
     fn on_status_change(&mut self, _ctx: &mut LifeCycleCtx, _event: &StatusChange) {}
 
     fn lifecycle(&mut self, _ctx: &mut LifeCycleCtx, _event: &LifeCycle) {}
```
DJMcNab, line 84:
```suggestion
```
(and a more general question of whether this should be done for the other methods/across other widgets)

masonry/src/passes/update.rs
```diff
@@ -416,17 +416,14 @@ fn update_new_widgets(
...
+        let mut ctx = RegisterCtx {
             widget_state_children: state.children.reborrow_mut(),
             widget_children: widget.children.reborrow_mut(),
         };
-        let event = LifeCycle::Internal(crate::InternalLifeCycle::RouteWidgetAdded);
-        widget.item.lifecycle(&mut ctx, &event);
+        // The widget will call `RegisterCtx::register_child` on all its children,
+        // which will add the new widgets to the arena.
+        widget.item.register_children(&mut ctx);
     }
```
DJMcNab, line 427: We should mention this method at:
https://github.com/linebender/xilem/blob/12ddfa35a556fa426f0bbe9c58fe9a7792bbae9d/masonry/src/passes/mod.rs#L31

DJMcNab (approved):

masonry/src/widget/image.rs
PoignardAzur: It could be done in a separate PR. I'm going to merge lifecycle and on_status_change anyway.

masonry/src/passes/update.rs
PoignardAzur: It does. We could switch it up by changing the order of paragraphs in that function. I'm adding this to the list of things I need to document.

masonry/src/passes/update.rs
PoignardAzur: I'm leaning more towards adding a safety check in the update_new_widgets pass. If the error you mention still happens after that safety check, then the problem likely comes from the widget's children_ids implementation, not register_children.

PR #605, opened 2024-09-22
feat(#574): reimplementation of `Image` widget `layout` function

### LINKED ISSUE #574: Masonry: Reason about `Image` widget layout implementation
This is the issue from the TODO comment at:
https://github.com/linebender/xilem/blob/7be4dd2a2f8bf820fc19a3af2a00600d05e4d4c7/masonry/src/widget/image.rs#L94-L97

In particular, the current layout logic entirely ignores the [`FillStrat`](https://github.com/linebender/xilem/blob/7be4dd2a2f8bf820fc19a3af2a00600d05e4d4c7/masonry/src/widget/mod.rs#L62). Effectively, the [`aspect-ratio`](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) is always `auto`, so the only troublesome case is if we have a minimum/maximum set which never matches the given aspect ratio.

This logic has recently (#571) been changed from the original [Druid version](https://github.com/linebender/druid/blob/4fe393927f8e59baacb53527a665980a157a0ad4/druid/src/widget/image.rs#L194-L202). 

The steps for this issue are:
1) Work out the correct layout logic. It might be that the current version is correct, at least whilst we can't set `aspect-ratio`.
2) Write a few test cases for the layout logic, and integrate them into the tests module. These could probably use screenshot tests.
3) Remove the original TODO comment.

failingprovince:
Reimplement calculation of image size in `Image` widget `layout` function.
Makes it match against `self.fill` before setting size.

Tests still missing.

Fixes #574

### DIFF
--- a/masonry/src/widget/image.rs
+++ b/masonry/src/widget/image.rs
@@ -91,10 +91,26 @@ impl Widget for Image {
             trace!("Computed size: {}", size);
             return size;
         }
-        // This size logic has NOT been carefully considered, in particular with regards to self.fill.
-        // TODO: Carefully consider it
-        let size =
-            bc.constrain_aspect_ratio(image_size.height / image_size.width, image_size.width);
+        let image_aspect_ratio = image_size.height / image_size.width;
+        let size = match self.fill {
+            FillStrat::Contain => bc.constrain_aspect_ratio(image_aspect_ratio, image_size.width),
+            FillStrat::Cover => Size::new(bc.max().width, bc.max().width * image_aspect_ratio),
+            FillStrat::Fill => bc.max(),
+            FillStrat::FitHeight => {
+                Size::new(bc.max().height / image_aspect_ratio, bc.max().height)
+            }
+            FillStrat::FitWidth => Size::new(bc.max().width, bc.max().width * image_aspect_ratio),
+            FillStrat::None => image_size,
+            FillStrat::ScaleDown => {
+                let mut size = image_size;
+
+                if !bc.contains(size) {
+                    size = bc.constrain_aspect_ratio(image_aspect_ratio, size.width);
+                }
+
+                size
+            }
+        };
         trace!("Computed size: {}", size);
         size
     }
--- a/masonry/src/widget/image.rs
+++ b/masonry/src/widget/image.rs
@@ -28,7 +28,7 @@ use crate::{
 /// than the image size).
 pub struct Image {
     image_data: ImageBuf,
-    fill: ObjectFit,
+    object_fit: ObjectFit,
 }
 
 // --- MARK: BUILDERS ---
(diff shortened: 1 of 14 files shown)


DJMcNab:
I've edited your comment to include the text "fixes #574", as the version in the title doesn't count.

failingprovince:
Ready for review.
Test looks good to me, but still I'm not sure if they are written correctly.

failingprovince:
There is still one point open: should `FillStrat` be renamed?
If yes, to what?

jaredoconnell (approved):
I can confirm that this PR does not introduce a regression in the default `Contain` case in the examples. And the tests look good.
Regarding renaming `FillStrat`, I'll defer to the maintainers. I don't love the current name, but it works. So I'm open to either renaming or keeping it.

jaredoconnell:
One comment I have is that there is code in the implementation of `FillStrat` that is not used with this code. What do we want to do about that?

failingprovince:
I think the code of `FillStrat::affine_to_fill` isn't of any use anymore.

Positioning of an image is a work for the `Image::layout` function, plus the current implementation is much simplier than calculating an affine matrix for then defining positioning/layout inside a `BoxConstraint`.

But maybe I'm missing something.

DJMcNab:
Next steps here are removing any unused code, if there is some. I think renaming `FillStrat` to `ObjectFit` is uncontroversial, and I'd recommend doing so in this PR.  We can always rename it again if needed.

waywardmonkeys:
If you rename it, we can remove it from the typos config too!

failingprovince:
> Next steps here are removing any unused code, if there is some. I think renaming `FillStrat` to `ObjectFit` is uncontroversial, and I'd recommend doing so in this PR. We can always rename it again if needed.

Renamed the struct but couldn't find any unused code to remove.

> If you rename it, we can remove it from the typos config too!

Done.

masonry/src/widget/image.rs
```diff
@@ -28,7 +28,7 @@ use crate::{
 /// than the image size).
 pub struct Image {
     image_data: ImageBuf,
-    fill: FillStrat,
+    fill: ObjectFit,
```
DJMcNab, line 31: Should we rename this field? I'd suggest either `fit` or `object_fit`?

masonry/src/widget/mod.rs
```diff
@@ -61,7 +61,7 @@ use crate::{Affine, Size};
 // These are based on https://api.flutter.dev/flutter/painting/BoxFit-class.html
 /// Strategies for inscribing a rectangle inside another rectangle.
```
DJMcNab, line 62: Can we link to https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit in these docs?

masonry/src/widget/image.rs
failingprovince: Yes we should rename it. 
`object_fit` is ok for the moment, anyway we can always rename it in the future

xilem/src/view/image.rs
```diff
@@ -24,7 +24,7 @@ pub fn image(image: &vello::peniko::Image) -> Image {
         // We take by reference as we expect all users of this API will need to clone, and it's
         // easier than documenting that cloning is cheap.
         image: image.clone(),
-        fill: FillStrat::default(),
+        fill: ObjectFit::default(),
```
DJMcNab, line 27: Same thing applies here

masonry/src/widget/image.rs
```diff
@@ -28,26 +28,26 @@ use crate::{
...
-            fill: FillStrat::default(),
+            object_fit: ObjectFit::default(),
         }
     }
 
     /// Builder-style method for specifying the fill strategy.
     #[inline]
-    pub fn fill_mode(mut self, mode: FillStrat) -> Self {
-        self.fill = mode;
+    pub fn fill_mode(mut self, mode: ObjectFit) -> Self {
```
DJMcNab, line 49: And here

masonry/src/widget/image.rs
```diff
@@ -56,8 +56,8 @@ impl Image {
 impl<'a> WidgetMut<'a, Image> {
     /// Modify the widget's fill strategy.
     #[inline]
-    pub fn set_fill_mode(&mut self, newfil: FillStrat) {
-        self.widget.fill = newfil;
+    pub fn set_fill_mode(&mut self, newfil: ObjectFit) {
```
DJMcNab, line 59: And here

DJMcNab:
Thanks!

PR #608, opened 2024-09-23
xilem_core: Add a `WithoutElements` `ViewSequence`

Philipp-M:
This is similar as the `fork` view, but it allows using `View`s/`ViewSequence`s without elements within a `ViewSequence` context that contains elements, see https://github.com/linebender/xilem/issues/461#issuecomment-2282857369 for the original motivation.

Taking `xilem_web` as example, this should be possible (as well as something similar as in the linked comment):

```rust
fn run_once_fragment() -> impl DomFragment<()> {
    without_elements(run_once(|| {}))
}
```

The `WidgetViewSequence` is "fixed" as well, by using the "right" widget type (`DynWidget` is only used in `AnyView`, whereas I think `Box<dyn Widget>` is more general and correct).


DJMcNab:
I haven't had chance to review this yet. One thing I'm wondering is whether we can make an alternative for `ViewMarker` for these kinds of sequences?

Philipp-M:
Good point.

You mean to avoid the extra wrapper, and directly use the no element views in sequences?

At least something like this:
```rust
impl<V: View<State, Action, Context, Message, Element = NoElement> + NoElementViewMarker, ...>
    ViewSequence<State, Action, Context, Element, Message> for V {}
```
is not possible, because of ambiguity again... (We would need something like negative trait bounds?)

The other possible idea I tried, is directly implementing the `ViewSequence` for the concrete `NoElement` views, but this has another issue because of the super-trait bound `View: ViewMarker` which leads to ambiguity again.

I'm not sure if it's worth it, but we *could* remove the super trait requirement `View: ViewMarker` and put it to the concrete `View`s (like `WidgetView`), not sure about the exact implications, but it's probably not optimal, that the compiler doesn't scream when implementing `View` without also implementing `ViewMarker`...

Or do you mean something else?

DJMcNab:
I mean, I was especially saying that we should do that investigation.

xilem_core/src/sequence.rs
```diff
@@ -785,3 +786,112 @@ impl_view_tuple!(M0, Seq0, 0; M1, Seq1, 1; M2, Seq2, 2; M3, Seq3, 3; M4, Seq4, 4
...
+    }
+}
+
+/// The [`ViewSequence`] for [`without_elements`], see its documentation for more context.
+pub struct WithoutElements<Seq, State, Action, Context, Message> {
+    seq: Seq,
+    phantom: PhantomData<fn() -> (State, Action, Context, Message)>,
+}
+
+/// A [`ViewSequence`] that doesn't contain any elements, it can be used everywhere where a view sequence (with or without elements) is expected.
```
DJMcNab, line 821: This documentation is very oblique; I know why it exists, but to someone stumbling upon this, it would likely be extremely mysterious.

Perhaps we should say something like:
> An adapter which turns a [`ViewSequence`]  containing any number of views with side effects, into a `ViewSequence` with any element type.
>
> This can be used to embed side-effects naturally into the flow of your program.
> This can be used as an alternative to [`fork`] , which avoids adding extra nesting at the cost of only being usable in places where there is already an existing sequence.

DJMcNab (approved):
A docs nits, but otherwise I think it's better to land this than not.

Sorry it's taken so long to review; this hasn't been on my critical path.

Philipp-M:
> Sorry it's taken so long to review; this hasn't been on my critical path.

Hmm now I almost forgot about this myself. I merged main into here and fixed the conflicts, as some time went by, maybe give it a quick review again?

xilem_core/src/sequence.rs
Philipp-M: I used your comment, that's indeed a little bit more descriptive...

xilem/src/widget_view.rs
```diff
@@ -72,11 +71,11 @@ where
 /// }
 /// ```
 pub trait WidgetViewSequence<State, Action = ()>:
-    ViewSequence<State, Action, ViewCtx, Pod<DynWidget>>
+    ViewSequence<State, Action, ViewCtx, Pod<dyn Widget>>
```
DJMcNab, line 74: I'm surprised that this change has no other impacts; it seems a bit random to have in here

xilem_core/src/sequence.rs
```diff
@@ -854,3 +855,125 @@ impl_view_tuple!(M0, Seq0, 0; M1, Seq1, 1; M2, Seq2, 2; M3, Seq3, 3; M4, Seq4, 4
...
+}
+
+/// The [`ViewSequence`] for [`without_elements`], see its documentation for more context.
+#[derive(Debug)]
+pub struct WithoutElements<Seq, State, Action, Context, Message> {
+    seq: Seq,
+    phantom: PhantomData<fn() -> (State, Action, Context, Message)>,
+}
+
+/// An adapter which turns a [`ViewSequence`] containing any number of views with side effects, into a `ViewSequence` with any element type.
```
DJMcNab, line 891:
```suggestion
/// An adapter which turns a [`ViewSequence`] containing any number of views with side effects, into a `ViewSequence` with any element type.
///
/// The returned `ViewSequence` will not contain any elements, and will instead run the side effects
/// of the wrapped `ViewSequence` when built and/or rebuilt.
```
Or something; I'm trying to give some sense of how this works.

xilem_core/src/sequence.rs
```diff
@@ -854,3 +855,125 @@ impl_view_tuple!(M0, Seq0, 0; M1, Seq1, 1; M2, Seq2, 2; M3, Seq3, 3; M4, Seq4, 4
...
+///
+/// ```
+/// # use xilem_core::docs::{DocsViewSequence as WidgetViewSequence, some_component_generic as component};
+/// use xilem_core::{without_elements, run_once};
+///
+/// fn app_logic(state: &mut AppState) -> impl WidgetViewSequence<AppState> {
+///     (component(state), without_elements(run_once(|| {})))
+/// }
+///
+/// struct AppState;
```
DJMcNab:
```suggestion
/// # struct AppState;
```
maybe?

xilem_core/src/sequence.rs
```diff
@@ -854,3 +855,125 @@ impl_view_tuple!(M0, Seq0, 0; M1, Seq1, 1; M2, Seq2, 2; M3, Seq3, 3; M4, Seq4, 4
...
+/// This can be used to embed side-effects naturally into the flow of your program.
+/// This can be used as an alternative to [`fork`](crate::fork) , which avoids adding extra nesting at the cost of only being usable in places where there is already an existing sequence.
+///
+/// # Examples
+///
+/// ```
+/// # use xilem_core::docs::{DocsViewSequence as WidgetViewSequence, some_component_generic as component};
+/// use xilem_core::{without_elements, run_once};
+///
+/// fn app_logic(state: &mut AppState) -> impl WidgetViewSequence<AppState> {
```
DJMcNab: Nit: `app_logic` normally refers to the "main" view of the app, which cannot return a sequence in current Xilem. It might be better to use an example which doesn't suggest that this is supported.

```suggestion
/// fn isolated_child(state: &mut AppState) -> impl WidgetViewSequence<AppState> {
```
or some better name

DJMcNab (approved):
Mostly looks good. I've not run any manual tests

xilem/src/widget_view.rs
Philipp-M: Yeah I've included it here, as I was testing it with returning a `WidgetViewSequence` but that didn't work IIRC, as sequences are using `dyn Widget` instead of `DynWidget` which is used only within `AnyWidgetView`. I don't exactly remember anymore what did happen, but `DynWidget` cause(d/s) problems.

Currently `WidgetViewSequence` is not really used, which is why it doesn't seem to have an effect, and unfortunately it's not possible to mix with other view sequences like `FlexSequence`.
I guess it could also be removed, as we're only using "specialized" view sequences.
I think I'll leave it in here for now, and delay further changes (removal?) to follow-up PRs, I think in this form it's slightly more correct.

PR #609, opened 2024-09-24
Add comment about make_trace_span

PoignardAzur:


jaredoconnell (approved):
Seems fine. Is there a tracking issue on GitHub?

DJMcNab:
We should at the very least point at https://github.com/rust-lang/rust/issues/63084. I don't think there's that much value in making an issue for this, given that it would be unactionable.
But if anyone runs into this in future, they can subscribe to that issue to get the relevant reminder.

masonry/src/widget/widget.rs
```diff
@@ -147,6 +147,8 @@ pub trait Widget: AsAny {
         false
     }
 
+    // TODO - Write a generic default implementation once
+    // `const std::any::type_name` is stable.
```
PoignardAzur, line 151:
```suggestion
    // TODO - Write a generic default implementation once
    // `const std::any::type_name` is stable.
    // See https://github.com/rust-lang/rust/issues/63084
```

PR #612, opened 2024-09-24
Connect to tracy if the `tracy` feature is enabled

DJMcNab:
This is very useful for debugging performance issues. I've already used it to debug [#masonry>`to_do_list`: Horrendous performance](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/to_do_list.3A.20Horrendous.20performance).

Setting a feature works on a workspace-wide basis. I.e `cargo run --example mason --features tracy` works.

Some weirdness:
- Tracy seems to assume we know when a frame will "start", but we only know if we'll repaint once processing for an event has finished.
- Something seems to be starting a default "Frame", which isn't really right. We don't have any continuous frames. Maybe we should always request a redraw if tracy is enabled.

Either way, I think this feature is useful  to land *now*, so am not planning on resolving these in this PR unless someone provides a solution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1334,6 +1334,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "generator"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dbb949699c3e4df3a183b1d2142cb24277057055ed23c68ed58894f76c517223"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "log",
+ "rustversion",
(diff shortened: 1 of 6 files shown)


PoignardAzur (approved):
The code isn't too clear if you don't know wgpu_profiler or tracy, but it's mostly plumbing anyway.

I would like a short comment in the code (not just the Cargo manifest) explaining what tracy is, and giving a short explanation of how to use it.

Ideally we should have a full "How to debug Masonry perf problems" tutorial somewhere in our documentation, but that's a future item.

PR #613, opened 2024-09-24
Propagate the `stashed` state correctly

DJMcNab:
When the scrollbars are not being set, the calculations get NaNs. This breaks rendering, causing extreme slowdowns and artifacts.

This is discussed in [#masonry>`to_do_list`: Horrendous performance](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/to_do_list.3A.20Horrendous.20performance)

I've made this work now using `is_explicitly_stashed` in layout. It does not currently cause issues there.

The underlying issue of `NaN`s being created (and not detected) have not been resolved.


PoignardAzur:
Can do.

My immediate thought is we probably want `layout` to process `is_explicitly_stashed` rather than `is_stashed` for that safety check.

PoignardAzur (approved):

PoignardAzur:
Everything looks correct to me.

The correct flags are being handled in the right places.

PR #614, opened 2024-09-26
Remove `current_node` from `AccessCtx`

mwcampbell:
It makes more sense to pass the `NodeBuilder` reference separately to the `accessibility` method. In particular, this avoids the need to keep calling a method on `AccessCtx` to get that mutable reference.

### DIFF
--- a/masonry/src/passes/accessibility.rs
+++ b/masonry/src/passes/accessibility.rs
@@ -34,7 +34,7 @@ fn build_accessibility_tree(
             widget.item.short_type_name(),
             id.to_raw(),
         );
-        let current_node = build_access_node(widget.item, state.item, scale_factor);
+        let mut builder = build_access_node(widget.item, state.item, scale_factor);
 
         let mut ctx = AccessCtx {
             global_state,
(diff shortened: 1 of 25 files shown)


mwcampbell:
There was a TODO comment in `masonry/src/contexts.rs` suggesting this. I suspect that the previous approach was a vestige of how I originally integrated AccessKit into Druid.

I'm doing this now because I'd like to get this fix in before we start implementing accessibility in more widgets, including the full textbox accessibility implementation.

masonry/src/passes/accessibility.rs
```diff
@@ -34,29 +34,28 @@ fn build_accessibility_tree(
...
             widget_state_children: state.children.reborrow_mut(),
             widget_children: widget.children.reborrow_mut(),
             tree_update,
-            current_node,
             rebuild_all,
             scale_factor,
         };
-        set_common_properties(&mut ctx);
-        widget.item.accessibility(&mut ctx);
+        set_common_properties(&mut ctx, &mut builder);
```
DJMcNab, line 48: I think that `set_common_properties` and `build_access_node` can now be recombined into one function:
```
fn build_access_node(widget: &dyn Widget, ctx: &mut AccessCtx) -> NodeBuilder {...}
```

masonry/src/widget/widget.rs
```diff
@@ -126,7 +126,7 @@ pub trait Widget: AsAny {
 
     fn accessibility_role(&self) -> Role;
 
-    fn accessibility(&mut self, ctx: &mut AccessCtx);
+    fn accessibility(&mut self, ctx: &mut AccessCtx, builder: &mut NodeBuilder);
```
DJMcNab, line 129: Reading through the rest of this code, my preference would be for this variable to called `node` instead of the extremely generic `builder`.

However, if the idiomatic variable name used for this type in accesskit is `builder`, then I guess that's fine.

DJMcNab (approved):
Two potential small tweaks. I'm not blocking on either of them, but would slightly prefer different forms.

masonry/src/widget/widget.rs
mwcampbell: `builder` has become the idiomatic variable name by default. I've used it in egui and in AccessKit's own tests and examples. So far, no one else has pushed back on that. But I'm willing to reconsider it.

DJMcNab (approved):

PR #615, opened 2024-09-27
Accessibility for text

mwcampbell:
This adds the necessary AccessKit properties and child nodes to expose detailed information about text widgets. It also supports the `SetTextSelection` action.

### DIFF
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -352,14 +352,10 @@ impl<T: Selectable> TextWithSelection<T> {
         } else {
             Affinity::Downstream
         };
-        if let Some(anchor) =
-            self.access_position_from_offset(self.selection.anchor, anchor_affinity)
-        {
-            if let Some(focus) = self
-                .access_position_from_offset(self.selection.active, self.selection.active_affinity)
-            {
-                parent_node.set_text_selection(TextSelection { anchor, focus });
-            }
+        let anchor = self.access_position_from_offset(self.selection.anchor, anchor_affinity);
+        let focus = self.access_position_from_offset(self.selection.active, self.selection.active_affinity);
+        if let (Some(anchor), Some(focus)) = (anchor, focus) {
+            parent_node.set_text_selection(TextSelection { anchor, focus });
         }
         parent_node.add_action(accesskit::Action::SetTextSelection);
     }
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -385,6 +385,7 @@ impl<T: Selectable> TextWithSelection<T> {
             return false;
         };
         self.selection = Selection::new(anchor, active, active_affinity);
+        self.selection_visible = true;
         self.needs_selection_update = true;
         true
     }
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -285,7 +285,11 @@ impl<T: Selectable> TextWithSelection<T> {
         self.layout.draw(scene, point);
     }
 
-    fn access_position_from_offset(&self, offset: usize, affinity: Affinity) -> TextPosition {
+    fn access_position_from_offset(
+        &self,
+        offset: usize,
+        affinity: Affinity,
+    ) -> Option<TextPosition> {
         let text = self.text().as_ref();
         debug_assert!(offset <= text.len(), "offset out of range");
 
@@ -318,21 +322,26 @@ impl<T: Selectable> TextWithSelection<T> {
                 let mut length_sum = 0_usize;
                 for (character_index, length) in character_lengths.iter().copied().enumerate() {
                     if run_offset == length_sum {
-                        return TextPosition {
+                        return Some(TextPosition {
                             node: id,
                             character_index,
-                        };
+                        });
                     }
                     length_sum += length as usize;
                 }
-                return TextPosition {
+                return Some(TextPosition {
                     node: id,
                     character_index: character_lengths.len(),
-                };
+                });
             }
         }
 
-        panic!("offset not within the range of any run");
+        debug_panic!(
+            "offset {} not within the range of any run; text length: {}",
+            offset,
+            text.len()
+        );
+        None
     }
 
     pub fn accessibility(&mut self, update: &mut TreeUpdate, parent_node: &mut NodeBuilder) {
@@ -342,10 +351,15 @@ impl<T: Selectable> TextWithSelection<T> {
         } else {
             Affinity::Downstream
         };
-        let anchor = self.access_position_from_offset(self.selection.anchor, anchor_affinity);
-        let focus =
-            self.access_position_from_offset(self.selection.active, self.selection.active_affinity);
-        parent_node.set_text_selection(TextSelection { anchor, focus });
+        if let Some(anchor) =
+            self.access_position_from_offset(self.selection.anchor, anchor_affinity)
+        {
+            if let Some(focus) = self
+                .access_position_from_offset(self.selection.active, self.selection.active_affinity)
+            {
+                parent_node.set_text_selection(TextSelection { anchor, focus });
+            }
+        }
         parent_node.add_action(accesskit::Action::SetTextSelection);
     }
 
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -293,7 +293,8 @@ impl<T: Selectable> TextWithSelection<T> {
             let range = line.text_range();
             if !(range.contains(&offset)
                 || (offset == range.end
-                    && (affinity == Affinity::Upstream || line_index == self.layout.layout.len())))
+                    && (affinity == Affinity::Upstream
+                        || line_index == self.layout.layout.len() - 1)))
             {
                 continue;
             }
@@ -303,8 +304,8 @@ impl<T: Selectable> TextWithSelection<T> {
                 if !(range.contains(&offset)
                     || (offset == range.end
                         && (affinity == Affinity::Upstream
-                            || (line_index == self.layout.layout.len()
-                                && run_index == line.len()))))
+                            || (line_index == self.layout.layout.len() - 1
+                                && run_index == line.len() - 1))))
                 {
                     continue;
                 }
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -285,31 +285,18 @@ impl<T: Selectable> TextWithSelection<T> {
         self.layout.draw(scene, point);
     }
 
-    fn access_position_from_offset(
-        &self,
-        parent_node: &NodeBuilder,
-        offset: usize,
-        affinity: Affinity,
-    ) -> TextPosition {
+    fn access_position_from_offset(&self, offset: usize, affinity: Affinity) -> TextPosition {
         let text = self.text().as_ref();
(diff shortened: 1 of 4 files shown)


mwcampbell:
I'm doing this as a draft for now because it's a bit quick and dirty, doesn't yet expose the bounding rectangles of runs and the relative positions and widths of characters (though AccessKit is designed to degrade gracefully in the absence of this information), isn't optimized, and might violate layering in a way we don't want.

mwcampbell:
Forgot to mention: This is known to work on Windows, when editing or moving in a textbox and when using NVDA's review commands within a prose widget like the one in the mason example.

masonry/src/text/selection.rs
```diff
@@ -283,6 +284,120 @@ impl<T: Selectable> TextWithSelection<T> {
...
+                .layout
+                .character_lengths_by_access_id
+                .get(&last_id)
+                .unwrap();
+            return TextPosition {
+                node: last_id,
+                character_index: character_lengths.len(),
+            };
+        }
+        for (line_index, line) in self.layout.layout.lines().enumerate() {
```
DJMcNab, line 308: I think we can early-continue the search if the line can't contain the given offset.

masonry/src/widget/prose.rs
```diff
@@ -281,11 +281,12 @@ impl Widget for Prose {
...
 
     fn accessibility_role(&self) -> Role {
-        Role::Paragraph
+        Role::Document
     }
 
-    fn accessibility(&mut self, _ctx: &mut AccessCtx, node: &mut NodeBuilder) {
-        node.set_name(self.text().as_ref().to_string());
+    fn accessibility(&mut self, ctx: &mut AccessCtx, node: &mut NodeBuilder) {
+        node.set_read_only();
```
DJMcNab, line 288: :+1: great to see the proper node setup here!

masonry/src/widget/prose.rs
```diff
@@ -281,11 +281,12 @@ impl Widget for Prose {
...
+        self.text_layout.accessibility(ctx, node);
     }
```
DJMcNab, line 291: Should this also handle `SetTextSelection`?

masonry/src/text/layout.rs
```diff
@@ -501,6 +516,89 @@ impl<T: AsRef<str> + Eq> TextLayout<T> {
             &self.layout,
         );
     }
+
+    pub fn accessibility(&mut self, ctx: &mut AccessCtx, parent_node: &mut NodeBuilder) {
```
DJMcNab, line 520: As you predicted, this using `AccessCtx` is a slight layering break. I think for expedience this is fine, although I think if we can pass in the individual fields (I think only `tree_update` is used?)

DJMcNab (commented):
I've not run this, but at first glance this code looks really good!

One thing which does concern me is that any accessibility events targeted at the individual text runs won't get routed correctly. I think it is reasonable to not have a one-to-one widget to accessibility node mapping, but I don't know how to represent that in our architecture (without a side-channel mapping of accessibility ids to widgets).

mwcampbell:
FOrtunately, AccessKit never sends action requests to the individual runs.

I'll address the rest of the comments later.

mwcampbell:
Also to clarify a previous comment, this code should be as functional on macOS and Linux as it is on Windows. I just haven't tested on those platforms yet.

mwcampbell:
Also, I left in a TODO comment speculating about adding a trailing newline to the run node if it's at the end of a line that's not the last line in the layout. But if the prose in the mason example is a reliable indication, the last (or only) run in the last line of a paragraph that's not the last paragraph in the text already does include the trailing `\n`, and blank paragraphs have a single run with nothing but the `\n`. Are there any caveats I need to watch out for here?

mwcampbell:
Taking this out of draft because we agreed we'd land this PR without bounding boxes for runs and positions/widths of characters.

masonry/src/text/layout.rs
```diff
@@ -3,17 +3,24 @@
 
 //! A type for laying out, drawing, and interacting with text.
 
-use std::rc::Rc;
+use std::{
+    collections::{HashMap, HashSet},
+    rc::Rc,
+};
```
PoignardAzur, line 9:
```suggestion
use std::collections::{HashMap, HashSet};
use std::rc::Rc;
```

masonry/src/text/layout.rs
```diff
@@ -501,6 +516,89 @@ impl<T: AsRef<str> + Eq> TextLayout<T> {
...
+                        id
+                    });
+                ids.insert(id);
+                let mut node = NodeBuilder::new(Role::InlineTextBox);
+
+                if let Some((last_id, mut last_node)) = last_node.take() {
+                    last_node.set_next_on_line(id);
+                    node.set_previous_on_line(last_id);
+                    ctx.tree_update.nodes.push((last_id, last_node.build()));
+                    parent_node.push_child(last_id);
```
PoignardAzur, line 548: This could probably be done in the caller. `TextLayout::Accessibility` could return a `Vec<NodeBuilder>`.

masonry/src/text/selection.rs
```diff
@@ -283,6 +284,121 @@ impl<T: Selectable> TextWithSelection<T> {
...
     }
+
+    fn access_position_from_offset(
+        &self,
+        parent_node: &NodeBuilder,
+        offset: usize,
+        affinity: Affinity,
+    ) -> TextPosition {
+        let text = self.text().as_ref();
+        assert!(offset <= text.len());
```
PoignardAzur, line 295: This should probably be a debug panic, or at least have an assert message.

PoignardAzur (commented):
Overall, I'm not very comfortable with this code being added as-is. Aside from the layering break, which shouldn't be too hard to address, this is very logic-intensive code, so to speak, with what seems like a lot of implicit invariants and no comments documenting them.

I'm especially uncomfortable with the three hashmap fields. They seem like the could be prone to cache invalidation errors if we're not careful.

I'm not sure if there's another solution, though. As I understand it, this code performs a one-to-may mapping of "text widget" to "accesskit nodes" that can't really be avoided. Although `character_lengths_by_access_id` seems like information that accesskit should already store for us?

masonry/src/widget/prose.rs
mwcampbell: Done.

masonry/src/text/layout.rs
mwcampbell: Manually applied.

masonry/src/text/selection.rs
mwcampbell: I believe the new version addresses this feedback.

masonry/src/text/layout.rs
mwcampbell: I don't like returning a `Vec` because that requires an intermediate allocation. I'll implement Daniel's suggestion of passing in the `TreeUpdate` rather than the `AccessCtx`.

mwcampbell:
I believe I've now adequately addressed all review feedback.

jaredoconnell (commented):
I'm not going to leave an approval since I'll let the prior reviewers determine if their concerns were addressed. VoiceOver on mac seems to work fine with this branch, but I am not an expert on screen readers. Test selection seems to be properly handled by VoiceOver.

mwcampbell:
@PoignardAzur Please review again when you have time. Thanks.

PoignardAzur:
Sorry, I was low on energy these last few days. Will review tomorrow.

masonry/src/text/selection.rs
```diff
@@ -265,6 +266,111 @@ impl<T: Selectable> TextWithSelection<T> {
...
+    fn access_position_from_offset(&self, offset: usize, affinity: Affinity) -> TextPosition {
+        let text = self.text().as_ref();
+        debug_assert!(offset <= text.len(), "offset out of range");
+
+        for (line_index, line) in self.layout.layout.lines().enumerate() {
+            let range = line.text_range();
+            if !(range.contains(&offset)
+                || (offset == range.end
+                    && (affinity == Affinity::Upstream
+                        || line_index == self.layout.layout.len() - 1)))
```
PoignardAzur, line 279: I wonder if we could add a function in Parley which does this check.

PoignardAzur, line 279: I wonder if we could add a function in Parley which does this check.

masonry/src/text/selection.rs
```diff
@@ -265,6 +266,111 @@ impl<T: Selectable> TextWithSelection<T> {
...
+                    length_sum += length as usize;
+                }
+                return TextPosition {
+                    node: id,
+                    character_index: character_lengths.len(),
+                };
+            }
+        }
+
+        panic!("offset not within the range of any run");
```
PoignardAzur, line 317: This should probably be a warning or a debug assert? We don't want to panic in release builds because the user found a corner case in the selection code.

masonry/src/widget/prose.rs
```diff
@@ -190,8 +190,17 @@ impl Widget for Prose {
         }
     }
 
-    fn on_access_event(&mut self, _ctx: &mut EventCtx, _event: &AccessEvent) {
-        // TODO - Handle accesskit::Action::SetTextSelection
+    fn on_access_event(&mut self, ctx: &mut EventCtx, event: &AccessEvent) {
+        if event.target == ctx.widget_id() {
```
PoignardAzur, line 194: You can remove that condition. Both Prose and Textbox are leaf widgets: if they get an accessibility event, we know for sure they're the target.

(Also `AccessEvent::target` was removed in a recent PR, so you'll need to anyway to merge this.)

masonry/src/widget/prose.rs
```diff
@@ -275,11 +284,12 @@ impl Widget for Prose {
...
     fn accessibility_role(&self) -> Role {
-        Role::Paragraph
+        Role::Document
     }
 
-    fn accessibility(&mut self, _ctx: &mut AccessCtx, node: &mut NodeBuilder) {
-        node.set_name(self.text().as_ref().to_string());
+    fn accessibility(&mut self, ctx: &mut AccessCtx, node: &mut NodeBuilder) {
+        node.set_read_only();
+        self.text_layout.accessibility(ctx.tree_update, node);
```
PoignardAzur, line 292: Any reason the same code isn't used in `Label::accessibility`?

PoignardAzur (approved):
Overall, I'm still uncomfortable with the architecture and the number of invariants it seems to maintain, but I don't have any alternatives.

I don't see any blockers. This can be merged as soon as the code is rebased.

masonry/src/text/selection.rs
mwcampbell: OK, I had to modify `access_position_from_offset` to return an `Option` so it would have something to return in that case, then modify the caller accordingly. But I think you're right. If I were using an application that had a bug which triggered this code path, I would want it to run with degraded functionality rather than crashing because the developer didn't test accessibility enough (something that's very likely).

masonry/src/widget/prose.rs
mwcampbell: Yes, we ought to do this for labels too. I'd like to wait and do that after I do an AccessKit refactor so we don't have to also set the name property on the label. I don't think that should block merging this PR.

PoignardAzur (approved):
LGTM.

PR #616, opened 2024-09-27
Update to Vello 0.3.0 (git)

DJMcNab:
Note that this does *not* update to Parley 0.2.0 (git), because that requires a lot more work.

The current purpose of this PR is to validate Vello 0.3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,9 +209,9 @@ checksum = "9d151e35f61089500b617991b791fc8bfd237ae50cd5950803758a179b41e67a"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "as-raw-xcb-connection"
@@ -221,11 +221,11 @@ checksum = "175571dd1d178ced59193a6fc02dde1b972eb0bc56c892cde9beeceac5bf0f6b"
 
 [[package]]
 name = "ash"
-version = "0.37.3+1.3.251"
+version = "0.38.0+1.3.281"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "39e9c3835d686b0a6084ab4234fcd1b07dbf6e4767dce60874b12356a25ecd4a"
+checksum = "0bb44936d800fea8f016d7f2311c6a4f97aebd5dc86f09906139ec848cf3a46f"
 dependencies = [
- "libloading 0.7.4",
+ "libloading",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


jaredoconnell:
I tested a few examples in Windows with Nvidia and MacOS with M1 Pro and I cannot see any differences.

jaredoconnell:
I experienced a panic while testing http_cats, but it doesn't look related to these changes, so I created an issue. But it might be. #750

jaredoconnell:
There appears to be a problem with the cursor position in Prose (and potentially others) when not left aligned. When clicking on a middle aligned block of text, the cursor ends up far to the right. In the case of the Prose for HTTP Status Code, at the default window size the cursor is placed 16 characters too far to the right.
To confirm that this has something to do with the center alignment, I narrowed the window so that the text was centered without any padding around it, which resulted in the cursor properly aligning.

jaredoconnell:
Other regressions I found include the text within the progress bar not updating in the `widgets` example, and the textbox size shrinking greatly when no text is present in the `two_textboxes` example and the `to_do_mvc` example. For that it sounds like we need logic for finding what the default vertical height of the text would be without text being present.

So there are no hard regressions, but these problems should likely be fixed before merging.

masonry/src/widget/textbox.rs
```diff
@@ -27,278 +31,644 @@ const TEXTBOX_PADDING: f64 = 3.0;
...
+                            txn.backdelete();
+                        }
+                    }),
+                    Key::Named(NamedKey::Enter) => {
+                        ctx.submit_action(crate::Action::TextEntered(self.text().to_string()));
+                        return;
+                        // let (fctx, lctx) = ctx.text_contexts();
+                        // self.editor
+                        //     .transact(fctx, lctx, |txn| txn.insert_or_replace_selection("\n"));
+                    }
```
jaredoconnell, line 511: At some point in the future we need to allow configuration of enter behavior. There are definitely instances where enter should submit. Others where it should add a new line. And in some applications, shift-enter should be new line.

masonry/src/text/editor.rs
```diff
@@ -0,0 +1,672 @@
...
+    pub fn select_word_right(&mut self) {
+        self.editor
+            .set_selection(self.editor.selection.next_word(&self.editor.layout, true));
+    }
+
+    /// Select the word at the point.
+    pub fn select_word_at_point(&mut self, x: f32, y: f32) {
+        self.refresh_layout();
+        self.editor
+            .set_selection(Selection::word_from_point(&self.editor.layout, x, y));
```
jaredoconnell, line 438: The behavior of this doesn't seem correct when I use the editor. It selects the work then something messes it up so that it only selects the word left of the pointer. I don't know if the problem is here, where this function is called, or in Parley.
If this isn't addressed in this PR it would make sense to file an issue to fix it later. This isn't a blocker because there was no double click to select in the prior code.

jaredoconnell (approved):
It's not perfect, but with the exception of the shrinking empty textboxes, all bugs I've found are new and do not regress existing functionality.

I left a few comments about small issues I found. Another small issue is using an accessibility reader along with shift left/right to select results in accessibility only listing the newly selected character. I'm not sure if that's correct behavior.

Before this is merged it likely makes sense to merge the parley PR into main and depend on that instead of the branch.

masonry/src/text/editor.rs
DJMcNab: This should reproduce in Parley's `vello_editor` example, I think

masonry/src/contexts.rs
```diff
@@ -199,6 +199,26 @@ impl_context_method!(
...
+    ComposeCtx<'_>,
+    EventCtx<'_>,
+    LayoutCtx<'_>,
+    MutateCtx<'_>,
+    PaintCtx<'_>,
+    UpdateCtx<'_>,
+    {
+        /// Get the contexts needed to build and paint text sections.
+        ///
+        /// Note that in many cases, these contexts are.
```
tomcur, line 214: Text fragment

DJMcNab: I have no idea what I wanted to say here...

PR #619, opened 2024-09-28
xilem_web: Refactor unnecessary generic type parameter `Props` to associated type

Philipp-M:
This was the initial plan of #403, I can't remember anymore why I didn't made it to work back then. Anyway this should simplify the type signature of `Pod` and `PodMut` a little bit.

This also includes a little bit of refactoring otherwise (mostly aesthetic + removal of unnecessary wrapper  `DynNode`)

### DIFF
--- a/xilem_web/src/one_of.rs
+++ b/xilem_web/src/one_of.rs
@@ -14,53 +14,34 @@ use crate::{
 
 type CowStr = std::borrow::Cow<'static, str>;
 
-impl<P1, P2, P3, P4, P5, P6, P7, P8, P9, N1, N2, N3, N4, N5, N6, N7, N8, N9>
-    OneOfCtx<
-        Pod<N1, P1>,
-        Pod<N2, P2>,
-        Pod<N3, P3>,
-        Pod<N4, P4>,
-        Pod<N5, P5>,
-        Pod<N6, P6>,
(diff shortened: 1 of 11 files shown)


xilem_web/src/one_of.rs
```diff
@@ -137,90 +118,63 @@ where
         };
     }
 
-    fn with_downcast_a(
-        elem: &mut Mut<'_, Self::OneOfElement>,
-        f: impl FnOnce(Mut<'_, Pod<N1, P1>>),
-    ) {
+    fn with_downcast_a(elem: &mut Mut<'_, Self::OneOfElement>, f: impl FnOnce(Mut<'_, Pod<N1>>)) {
```
flosse, line 121: Usually I try to avoid macros, but in that case it might be an idea to reduce boilerplate?

flosse (approved):
LGTM! :relaxed:

xilem_web/src/one_of.rs
Philipp-M: When I initially added this, I thought about it, but I don't think it's really worth the added mental complexity, it's only a little bit duplicated boilerplate. When we would have more than 9 variants, the situation may be different.

Philipp-M:
Thanks for the review, merging.

PR #621, opened 2024-09-30
xilem_web: Add `Rotate` and `Scale` (CSS) transform views

Philipp-M:
This sketches/implements (parts of) the transform modifier API described in [Add `Affine` transform to `Widget` trait?](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Add.20.60Affine.60.20transform.20to.20.60Widget.60.20trait.3F/near/472076600) for xilem_web.

This currently includes `rotate` and `scale`, because there are less cases to handle for these [CSS transform functions](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function).
In `rotate` I think we can reduce this to just radians (there are more units for `<angle>`: `turn`, `grad`, `deg` and `rad` (used here) but they can all be derived from radians), and percent in scale is kinda redundant (as `200%` == `2.0`)), for example `translate` has more cases as it includes all kinds of units (like `px`, `em`, `%` etc.), so this may require more thought (and is thus probably better for a future PR).

This can be combined with untyped `transform: ...` as style, these modifier views extend the `transform` style (while the untyped `style(..)` overwrites previous set values).

The `svgtoy` example is updated to include these views.

### DIFF
--- a/xilem_web/src/interfaces.rs
+++ b/xilem_web/src/interfaces.rs
@@ -160,9 +160,9 @@ pub trait Element<State, Action = ()>:
     /// # Examples
     ///
     /// ```
-    /// use xilem_web::{style as s, interfaces::SvgElement, svg::kurbo::Rect};
+    /// use xilem_web::{style as s, interfaces::Element, svg::kurbo::Rect};
     ///
-    /// # fn component() -> impl SvgElement<()> {
+    /// # fn component() -> impl Element<()> {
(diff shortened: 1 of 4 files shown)


flosse (approved):
LGTM!


Just some random thoughts:
Maybe in the future we can add types that handle different units like degrees, etc, but that's probably not where we're at here (users can use an external crate for proper unit handling). Or we could add a `Radians` type so that the user at least knows what is being used here, although this might be an unnecessary thing that just makes things cumbersome.

PR #622, opened 2024-10-01
Add public methods to access clip path

tomcur:
This adds public getters for clip path to non-`LayoutCtx` contexts, and also to `WidgetState` as it seems relevant for painting.

The motivation is to remove private field accesses from `Widget::get_child_at_pos` as mentioned in
https://github.com/linebender/xilem/pull/565#issuecomment-2374162416.

This leaves one private field access, namely `WidgetState::is_stashed`. As per the docs of `WidgetState` ("widgets will generally not interact with it directly"), that field perhaps should not be publicly accessible from `WidgetState`, but then it should be accessible from `WidgetRef`.

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -239,6 +239,14 @@ impl_context_method!(
             self.widget_state.window_origin()
         }
 
+        /// The clip path of the widget, if any was set.
+        ///
+        /// For more information, see
+        /// [`LayoutCtx::set_clip_path`](crate::LayoutCtx::set_clip_path).
+        pub fn clip_path(&self) -> Option<Rect> {
+            self.widget_state.clip_path()
+        }
+
         /// Convert a point from the widget's coordinate space to the window's.
         ///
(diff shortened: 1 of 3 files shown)


tomcur:
Private `WidgetState::is_stashed` is accessed here

https://github.com/linebender/xilem/blob/67f0043e7529cab33c71c973571d62257a6047f7/masonry/src/widget/widget.rs#L214

`child` is a `WidgetRef`. Perhaps `WidgetRef` can allow access to the underlying `QueryCtx`, and we add the stashed getter on `QueryCtx`.

Philipp-M (approved):
Yeah probably doesn't hurt to expose this, LGTM.
I expect though that this is very much in flux, a clip path just being `Rect` seems... limited.

tomcur:
I'll leave the question of how to expose stashed state for a later PR.

> I expect though that this is very much in flux, a clip path just being `Rect` seems... limited.

Right, iirc clip is intended to become generalized to arbitrary shapes, but it requires an efficient way in Kurbo to do so (maybe just an enum for the most common shapes?).

PR #624, opened 2024-10-03
Rearrange text-handling code

PoignardAzur:
Remove obsolete types.
Remove text_helpers module.


masonry/src/text/mod.rs
```diff
@@ -10,16 +10,22 @@
...
+pub use backspace::offset_for_delete_backwards;
 pub use edit::TextEditor;
+pub use render_text::render_text;
+pub use selection::{len_utf8_from_first_byte, Selectable, StringCursor, TextWithSelection};
+pub use text_layout::{Hinting, LayoutMetrics, TextBrush, TextLayout};
 
-mod backspace;
-pub use backspace::offset_for_delete_backwards;
+// ---
+
```
DJMcNab, line 26:
```suggestion
```

DJMcNab (approved):

masonry/src/text/mod.rs
DJMcNab: Hmm, could you please clarify the purpose this comment has? I read it as a mistaken addition

PoignardAzur: Just to separate imports from code. I could have removed it, but I didn't want to wait for CI again.

waywardmonkeys: For the future, you can just push the change and queue the merge and not wait… as long as CI passes, it will merge, otherwise it won’t. With that, there is no need to wait.

PoignardAzur: Yeah but sometimes rustfmt complains that the code-review change breaks the preferred indentation, so you have to pull the code and run cargo fmt locally (or do another review to tweak the change in a way you think passes rustfmt) and honestly, I didn't want to bother for something like this.

waywardmonkeys: Fixed in #624.

PR #628, opened 2024-10-05
Make selection non-optional in TextWithSelection

PoignardAzur:
There was no good reason for selection to be optional, except to not display it when focus was lost. This is now handled with a boolean flag. A side-effect of this is that TextWithSelection now remembers its selection when focus is gained back.

Selection was used with a default value of `(0, 0)` in cases where it was `None`, so its default value in similar cases is still `(0, 0)`.

### DIFF
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -16,9 +16,8 @@ use vello::Scene;
 use winit::keyboard::NamedKey;
 
 use crate::event::{PointerButton, PointerState};
+use crate::text::{TextBrush, TextLayout};
 use crate::{Handled, TextEvent};
-
-use super::{TextBrush, TextLayout};
 
 pub struct TextWithSelection<T: Selectable> {
     pub layout: TextLayout<T>,
--- a/masonry/src/text/selection.rs
+++ b/masonry/src/text/selection.rs
@@ -23,7 +23,8 @@ pub struct TextWithSelection<T: Selectable> {
     pub layout: TextLayout<T>,
     /// The current selection within this widget
     // TODO: Allow multiple selections (i.e. by holding down control)
-    pub selection: Option<Selection>,
+    pub selection: Selection,
+    pub selection_visible: bool,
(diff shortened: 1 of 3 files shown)


jaredoconnell:
What are the effects of making section non-optional?
Are there any behavioral changes to text edit or selection?

PoignardAzur:
The biggest effect is that Textbox "remembers" its selection when focus is moved away then back.

waywardmonkeys:
Sounds like a good commit message would help here. Or an actual comment.

PoignardAzur:
I've updated the commit message. I'm not sure a comment is needed, given this make the code simpler than the un-commented code it was replacing.

jaredoconnell (approved):
I can confirm that the selection is retained when I tab between textboxes, and I can't find any regressions. Nothing stands out as problematic in the source code, but I am not the most familiar with this code.

masonry/src/text/selection.rs
```diff
@@ -206,16 +188,13 @@ impl<T: Selectable> TextWithSelection<T> {
...
-            self.selection = Some(Selection::caret(self.text().len(), Affinity::Downstream));
-        }
+        self.selection_visible = true;
         self.needs_selection_update = true;
     }
 
     /// Call when another widget becomes focused
     pub fn focus_lost(&mut self) {
-        self.selection = None;
+        self.selection_visible = false;
```
DJMcNab, line 197: Is there ever a time that this is different from the parent widget being focused? I see that it is used in both `rebuild_with_attributes` and `draw`, so passing it in from the parent would be messy.

masonry/src/text/selection.rs
```diff
@@ -66,6 +67,7 @@ impl<T: Selectable> TextWithSelection<T> {
     ) -> bool {
         // TODO: work out which button is the primary button?
         if button == PointerButton::Primary {
+            self.selection_visible = true;
```
DJMcNab, line 70: That is, does this do anything?

Issue #389, opened 2024-06-12
Rewrite Masonry documentation


PoignardAzur:
Masonry's documentation has only received piecemeal updates since the port to Winit.

Even when the documentation is still accurate, it lacks overall cohesion. There isn't an ordered document explaining the major concepts users need to know to understand Masonry's flow.

Once other major changes are implemented, the documentation should be brought up to date. We should avoid procrastinating on documentation too much just because the architecture is still in progress.

PoignardAzur:
Fixed by #809, #811, #813, #815, etc.

Issue #501, opened 2024-08-09
Automating screenshots for documentation


DJMcNab:
Our documentation currently does not include screenshots, which can be instrumental in understanding the purpose of different widgets.

Since #475 (which landed as part of #233), we are able to store screenshots in this repository. We should find a way to include some of these screenshots in our documentation. For example, see [`TextInput`](https://docs.rs/iced/latest/iced/widget/struct.TextInput.html) in `iced`.

Ideally, this would be natively supported by rustdoc. This is tracked in https://github.com/rust-lang/rust/issues/32104
However, we would like to have something working before this is resolved, as that issue has been open for a very long time.

The implementation pathway I would recommend is to host a GitHub pages site from a folder in this repository (https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow). We would then use URLs to this site in our documentation.

I do foresee some concerns with this approach:
1) This does not gracefully handle cases where the screenshot changes between versions of Xilem. This is because the website would show the updated version, even for old versions of the documentation. This is primarily a concern for when the screenshot would change, or if a specific screenshot would get removed. I think the easiest way to mitigate this would be by versioning the screenshots.
2) This does not work for new screenshots in local documentation. I can't see a good way to work around this with the abilities provided by rustdoc, and so. For reviewing, the screenshots would be shown in the PR's diff view, and would become available as soon as the PR was merged.

However, this approach would still be a significant improvement from our current state.

Philipp-M:
I wonder if that should also include snapshot testing?
I don't think we get around a custom solution anyway.

We could use the git commit versioned link I think like this: https://media.githubusercontent.com/media/linebender/vello/5aa5937d78dd1ec3b5b067dc173781858c38550e/vello_tests/snapshots/splash.png (after a quick skim of the response headers, I *think* it should work embedded in docs.rs).

So when a screenshot changes, "breaking" the (doc-)tests, then a local link to the screenshot has to be made to resolve the issue in CI.
And needs to be updated explicitly with a new commit (via a bot in CI, that scans for such local-changes to avoid git-commit-self-references?). I'm not sure about the details, and how that would work out in practice (e.g. does that count as LFS bandwidth? (needs some testing I guess)).

But it would at least make sure that all the screenshots are up to date and valid with the current (or following) commit.

DJMcNab:
> I wonder if that should also include snapshot testing?

[...]

PR #632, opened 2024-10-06
Create skeleton of documentation

### LINKED ISSUE #376: Add tutorial to documentation
Masonry introduces a lot of complex concepts compared to other GUI frameworks. Many of these concepts are intended to reduce future complexity, but crate users don't get those benefits if they can't use the API in the first place.

Masonry should have a tutorial integrated directly into the documentation with rustdoc.

The tutorial should have the following chapters:

- **Getting started.** The steps to creating a "TODO" app.
- **Creating a Widget.** Present the Widget trait, the different methods, the `create_widget` macro, how to do tests.
- **Creating a Container Widget.** Present WidgetPod, container invariants, etc.
- **Widget statuses.** Hot, active, focused, stashed.

Other parts of the documentation should refer to these tutorials.

PoignardAzur:
This is a very messy, very basic skeleton of what Masonry documentation will eventually look like.

Main points are:

- Dedicated documentation modules.
- Re-using most of the language from the RFCs.

Next steps are:

- Flesh out the Widget documentation.
- Rewrite all those docs in a less placeholder-y way.
- Add chapter about the widget arena.
- Spread out that pass documentation to the respective pass files.
- Rewrite ARCHITECTURE.md.
- Add screenshots.

Fixes #376 and #389.


masonry/src/lib.rs
```diff
@@ -126,6 +126,7 @@ pub mod widget;
 pub mod app_driver;
 pub mod debug_logger;
 pub mod debug_values;
+pub mod doc;
```
DJMcNab, line 129:
```suggestion
#[cfg(doc)]
pub mod doc;
```

masonry/src/doc/01_creating_app.md
```diff
@@ -0,0 +1,99 @@
+**TODO - Add screenshots**
```
DJMcNab, line 1:
```suggestion
**TODO - Add screenshots - see [#501](https://github.com/linebender/xilem/issues/501)**
```

https://github.com/linebender/xilem/issues/501

PoignardAzur:
I'm marking this as ready for review despite the TODOs because I think this is a stark improvement over the previous state of our documentation.

The next be steps would be:

- Writing the "Testing widgets" chapters.
- Rewriting doc in the rest of the crate to reference this tutorial (and especially the "Concepts and definitions" chapter).

masonry/src/doc/01_creating_app.md
```diff
@@ -0,0 +1,238 @@
...
+This tutorial explains how to build a simple Masonry app, step by step.
+Though it isn't representative of how we expect Masonry to be used, it does cover the basic architecture.
+
+The app we'll create is identical to the to-do-list example shown in the README.
+
+## The Widget tree
+
+Let's start with the `main()` function.
+
+```ignore
```
Philipp-M, line 14:
```suggestion
```rust,ignore
```

I'm not sure if the Github markdown renderer (or others) handles this better (i.e. syntax highlighting), but at least it has a chance to know that this is rust code.

(Same for the other `ignore` here)

I also wonder, if we should design these inline examples such that they compile, so that they're always in sync and checked via CI.
But I see this may increase the (not so interesting) code in these examples.
That could *probably* be avoided by `# code that should not be rendered` but that on the other hand has the disadvantage, to be shown in other markdown renderers (though I'd rather accept that TBH and be sure that everything is in sync).

PoignardAzur: > I also wonder, if we should design these inline examples such that they compile, so that they're always in sync and checked via CI.

You're welcome to try. It's *tough* to do it in a way that doesn't make the examples tedious and unreadable. (Unless you use a lot of `#` marker to hide code from the reader.)

DJMcNab: This also interacts with #501. That is, we definitely should make these proper examples, but that work can reasonably be deferred.

Yes, that would involve using hidden harness code.

masonry/src/doc/02_implementing_widget.md
```diff
@@ -0,0 +1,306 @@
...
+        SmallVec::new()
+    }
+
+    fn make_trace_span(&self) -> Span {
+        trace_span!("ColorRectangle")
+    }
+}
+```
+
+Don't worry about what they mean for now.
```
DJMcNab, line 274: I think here is the right place for `make_trace_span` to be described.

I'd probably also say something like:
```md
The children methods are described in [the next chapter](crate::doc::doc_03_implementing_container_widget).
Our widget can use the null implementations as it has no children:
```

masonry/src/doc/02_implementing_widget.md
```diff
@@ -0,0 +1,306 @@
...
+
+<!-- TODO - Rewrite once we've decided how WidgetMut should be implemented. -->
+
+```ignore
+struct ColorRectangle {
+    size: Size,
+    color: Color,
+}
+
+impl WidgetMut<'_, ColorRectangle> {
```
DJMcNab, line 286: I'm surprised that this works - I thought it would be a coherence break

masonry/src/doc/02_implementing_widget.md
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    pub fn set_size(&mut self, size: Size) {
+        self.widget.size = size;
+        self.ctx.request_layout();
+    }
+}
+```
+
+By making ColorRectangle's fields private, and making it so the only way to mutate them is through a WidgetMut, we make our users never find themselves in a situation where they forget to propagate invalidation flags and the displayed widget tree ends up with hard-to-debug visual artifacts as a result.
```
DJMcNab, line 299: This is quite a long sentence.

masonry/src/doc/03_implementing_container_widget.md
DJMcNab, line 1: Looks good!

masonry/src/doc/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+pub mod doc_04_testing_widget {}
+
+#[doc = include_str!("./05_pass_system.md")]
+pub mod doc_05_pass_system {}
+
+#[doc = include_str!("./06_masonry_concepts.md")]
+pub mod doc_06_masonry_concepts {}
+
+/*
+> ls masonry/src/doc -1
```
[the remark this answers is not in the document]
PoignardAzur: Nope, that's a leftover.

masonry/src/doc/02_implementing_widget.md
PoignardAzur: It doesn't work, hence the TODO comment. The final doc will use the free-function syntax.

masonry/src/doc/06_masonry_concepts.md
```diff
@@ -0,0 +1,91 @@
...
+## Widget status
+
+The notion of widget status is somewhat vague, but you can think of it as similar to [CSS pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes).
+
+Widget statuses are "things" managed by Masonry that affect how widgets are presented.
+Statuses include:
+
+- Being hovered.
+- Having pointer capture.
+- Having local text focus.
```
runiq, line 14: Below, you describe 'active text focus' and 'inactive text focus'. What's 'local text focus', though?

PoignardAzur: I think it's the same. Basically there's "the widget is focused, but the window isn't", and "the widget and the window are both focused". I need to take another crack at explaining this.

masonry/src/doc/01_creating_app.md
```diff
@@ -0,0 +1,238 @@
+# Building a "To-Do List" app
+
```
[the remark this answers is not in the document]
DJMcNab: This still needs the newlines around the opening/closing tags, because [the rendering](https://github.com/linebender/xilem/blob/837339cca57a832547a916bd87ba44f3d25b6209/masonry/src/doc/01_creating_app.md) is broken.

DJMcNab (approved):
Other than https://github.com/linebender/xilem/pull/632#discussion_r1810631186, I think this is ready to land.

I've not carefully re-reviewed since my last round, but overall it's better to get this in and allow updates to be collaborative.

runiq:
This is a very welcome addition. I've been watching from the peanut gallery for the longest time, waiting for something like this. Thanks a lot for taking the time, I know how hard writing documentation is. <3

PoignardAzur:
> Thanks a lot for taking the time, I know how hard writing documentation is. <3

It's so much harder than it sounds.

You have to put your brain in a certain mode and pull things from it that it really doesn't want to give you by default.

PR #635, opened 2024-10-07
Add safety rail for register_children method

PoignardAzur:
This gives a beginner-friendly error in cases where Masonry detects that `register_children()` was improperly implemented.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -499,10 +499,29 @@ fn update_new_widgets(
         let mut ctx = RegisterCtx {
             widget_state_children: state.children.reborrow_mut(),
             widget_children: widget.children.reborrow_mut(),
+            #[cfg(debug_assertions)]
+            registered_ids: Vec::new(),
         };
         // The widget will call `RegisterCtx::register_child` on all its children,
         // which will add the new widgets to the arena.
         widget.item.register_children(&mut ctx);
+
+        #[cfg(debug_assertions)]
+        {
+            let children_ids = widget.item.children_ids();
+            for child_id in ctx.registered_ids {
+                if !children_ids.contains(&child_id) {
+                    panic!(
+                        "Error in '{}' #{}: method register_children() called \
+                        RegisterCtx::register_child() on child #{}, which isn't \
+                        in the list returned by children_ids()",
+                        widget.item.short_type_name(),
+                        id.to_raw(),
+                        child_id.to_raw()
+                    );
+                }
+            }
+        }
 
         #[cfg(debug_assertions)]
         for child_id in widget.item.children_ids() {
@@ -554,6 +573,8 @@ pub(crate) fn run_update_new_widgets_pass(root: &mut RenderRoot) {
         let mut ctx = RegisterCtx {
             widget_state_children: root.widget_arena.widget_states.root_token_mut(),
             widget_children: root.widget_arena.widgets.root_token_mut(),
+            #[cfg(debug_assertions)]
+            registered_ids: Vec::new(),
         };
         ctx.register_child(&mut root.root);
     }
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -508,7 +508,8 @@ fn update_new_widgets(
         for child_id in widget.item.children_ids() {
             if widget.children.get_child(child_id.to_raw()).is_none() {
                 panic!(
-                    "Error in '{}' #{}: method register_children() did not call RegisterCtx::register_child() on child #{} returned by children_ids()",
+                    "Error in '{}' #{}: method register_children() did not call \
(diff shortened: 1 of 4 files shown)


masonry/src/passes/update.rs
```diff
@@ -492,6 +493,18 @@ fn update_new_widgets(
         // The widget will call `RegisterCtx::register_child` on all its children,
         // which will add the new widgets to the arena.
         widget.item.register_children(&mut ctx);
+
+        #[cfg(debug_assertions)]
+        for child_id in widget.item.children_ids() {
+            if widget.children.get_child(child_id.to_raw()).is_none() {
+                panic!(
+                    "Error in '{}' #{}: method register_children() did not call RegisterCtx::register_child() on child #{} returned by children_ids()",
```
waywardmonkeys, line 501: Is this string long enough that it might break `cargo fmt`?

waywardmonkeys (commented):
When would someone want the implementation of register_children to do anything other than register each of the things in children_ids?

Curious why there are essentially two sources of truth here. This seems like a hidden invariant that isn't yet otherwise documented clearly.

Maybe an addition to the doc comment on `Widget::register_children` or `Widget::children_ids` would be good to have here.

PoignardAzur:
> When would someone want the implementation of register_children to do anything other than register each of the things in children_ids?

Never, hence why we have a safety rail which checks that they're the same.

I'm still thinking about how to document this, but basically `children_ids` acts as the source of truth, and safety rails will give error messages that assume that the other method that doesn't match `children_ids` is wrong. (In my experience, `children_ids` methods are very easy to write and people never get them wrong. On the other hand, it's easy to forget to iterate over children in layout or register_children.)

> Curious why there are essentially two sources of truth here.

Basically the problem is that children_ids returns, well, ids, but when we want to add new widgets to the tree (the purpose of `register_children`), we need access to WidgetPods. In principle we could have a `children()` method that returns `Vec<WidgetPod<dyn Widget>>` or something, but that adds a lot of complications.

masonry/src/passes/update.rs
PoignardAzur: It passes CI, so presumably not.

waywardmonkeys: Nah, `cargo fmt` silently fails on that.

PoignardAzur: What do you mean by silently fails? As in, fail to break the line?

waywardmonkeys: Usually it will also fail to format some of the surrounding code too!

You can detect this by adding this to `rustfmt.toml`, but they're not stable, so have to use `cargo +nightly fmt`:

```
error_on_line_overflow = true
error_on_unformatted = true
```

waywardmonkeys:
If there's no rush, might as well see about the comment improvements before approving rather than have to remember that this needs further work again.

PoignardAzur:
Oh, I thought you were talking about the formatting thing.

PoignardAzur:
I've updated the documentation to make it a bit clearer.

waywardmonkeys (approved):
Much better with the doc comments.

You could link to the methods, but either way...

masonry/src/passes/update.rs
```diff
@@ -498,10 +499,42 @@ fn update_new_widgets(
...
         };
         // The widget will call `RegisterCtx::register_child` on all its children,
         // which will add the new widgets to the arena.
         widget.item.register_children(&mut ctx);
+
+        #[cfg(debug_assertions)]
+        {
+            let children_ids = widget.item.children_ids();
+            for child_id in ctx.registered_ids {
+                if !children_ids.contains(&child_id) {
```
DJMcNab, line 513: I think that if we used `find` and `swap_remove` here, we would get the second check for free (just any items remaining in `children_ids`).

masonry/src/widget/widget.rs
```diff
@@ -110,6 +111,9 @@ pub trait Widget: AsAny {
     /// the size of non-flex widgets first, to determine the amount of space
     /// available for the flex widgets.
     ///
+    /// Forgetting to visit children is a logic error and may lead to debug panics.
```
DJMcNab, line 114: I don't really like the term "debug panics" - it's a bit oblique

masonry/src/passes/update.rs
PoignardAzur: I'd rather have code be maximally clear.

What I'd like to do is add an automatic escape hatch for cases where `children_ids` is large, so we can skip checking those (under the assumption that if there's a bug in child handling, it probably shows up with a small number of children too).

PoignardAzur:
"Panics in dev mode", then?

DJMcNab:
Maybe "trigger debug assertions"

PR #636, opened 2024-10-07
Remove register methods

PoignardAzur:
Replace with `Widget::accepts_xxx()` methods
Tweak accessibility code
Remove `BuildFocusChain` event

Remove `WidgetState::is_portal`, `register_as_portal()`, and overall remove the idea of a first-class "portal" concept.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -253,7 +253,7 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot, root_state: &mut Widg
     if prev_focused != next_focused {
         let was_ime_active = root.state.is_ime_active;
         let is_ime_active = if let Some(id) = next_focused {
-            root.widget_arena.get_state(id).item.is_text_input
+            root.widget_arena.get_state(id).item.accepts_text_input
         } else {
             false
         };
@@ -559,8 +559,11 @@ fn update_new_widgets(
             "{} received LifeCycle::WidgetAdded",
             widget.item.short_type_name()
         );
+        state.item.accepts_pointer_interaction = widget.item.accepts_pointer_interaction();
+        state.item.accepts_focus = widget.item.accepts_focus();
+        state.item.accepts_text_input = widget.item.accepts_text_input();
+        state.item.is_new = false;
     }
-    state.item.is_new = false;
 
     // We can recurse on this widget's children, because they have already been added
     // to the arena above.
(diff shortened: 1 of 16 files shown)


PoignardAzur:
Man.

I just spend half an hour debugging an error I would have found immediately if I'd looked at the clippy error log. I feel like an idiot.

jaredoconnell (approved):
I tested a lot of examples for this one and cannot find any regressions.

masonry/src/passes/update.rs
```diff
@@ -620,16 +623,9 @@ fn update_focus_chain_for_widget(
...
-    {
-        let mut ctx = LifeCycleCtx {
-            global_state,
-            widget_state: state.item,
-            widget_state_children: state.children.reborrow_mut(),
-            widget_children: widget.children.reborrow_mut(),
-        };
-        widget.item.lifecycle(&mut ctx, &LifeCycle::BuildFocusChain);
+    if state.item.accepts_focus {
+        state.item.focus_chain.push(id);
```
DJMcNab, line 628: Why don't we just extend the global focus chain here? That is, what does the per-widget focus chain achive, now that we have an explicit state for it?

masonry/src/widget/label.rs
```diff
@@ -55,14 +55,15 @@ impl Label {
...
-            skip_pointer: false,
+            interactive: true,
         }
     }
 
-    // TODO - Rename
-    // TODO - Document
-    pub fn with_skip_pointer(mut self, skip_pointer: bool) -> Self {
-        self.skip_pointer = skip_pointer;
+    /// Sets the value returned by [`accepts_pointer_interaction`].
```
DJMcNab, line 62: Hmm, I'm surprised that this doc resolves? Is there a rustdoc feature I didn't know about (automatic handling of trait methods?)

It isn't shown as a valid link by rust-analyzer. I've not tried it in rustdoc

masonry/src/passes/update.rs
PoignardAzur: It does some caching.

It's not ideal, though. I think the whole "focus chain" concept should be refactored.

masonry/src/widget/label.rs
PoignardAzur: Nope, rustdoc doesn't resolve it either.

masonry/src/passes/update.rs
DJMcNab: Oh of course, some children won't actually have this pass run on them?

PoignardAzur: Yeah, the pass doesn't run on the entire tree.

PR #638, opened 2024-10-07
Add Display impl to WidgetId

PoignardAzur:


waywardmonkeys (approved):
I'm unreasonably happy to see this.

Only thing I'd ask if it should be a `Debug` impl instead.

PoignardAzur:
WidgetId already derives Debug.

For what it's worth, I'm not a huge fan of Debug impls that introduce custom formatting.

Issue #590, opened 2024-09-14
White flash on app startup on Windows 11 dark mode


gfxenjoyer:
Xilem apps have a white flash on Windows even on dark mode. This does not happen on macOS. I suspect this has something to do with winit.

Hardware:
OS: Windows 11 23H2 OS Build 22631.4169
CPU: Intel 12600k
GPU: Nvidia RTX 3080 Driver 560.70

This also happens in [Vello examples](https://github.com/linebender/vello/tree/main/examples/with_winit).

Reference [Zulipchat](https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/Startup.20white.20background.20flash.20on.20Windows.20dark.20mode)

[xilem_calc_white_flash_windows.webm](https://github.com/user-attachments/assets/5af55497-c882-41ef-acfc-6dfc7263204f)

Edit: 
Calc build in video shown is a release build.

DJMcNab:
Thanks for the report. I believe that there are a few things going on here:
- Windows will set a fully white buffer as the window contents initially. I don't know where this behaviour is documented
- Graphics initialisation is taking a while on your machine. This doesn't really make sense, because your GPU is quite powerful; the biggest issue should have been resolved in linebender/vello#575
- We might not `present` "soon enough" after making the window visible - see https://github.com/linebender/xilem/blob/b54266d605b597d967e4b13dcf04d2e43ce43691/masonry/src/event_loop_runner.rs#L261, but GPU initialisation hasn't happened yet.

Ultimately, I can't afford the time to dig into this at the moment, but hopefully this gives you some good directions to look towards when investigating this. In particular, you should try and find some documentation expressing a correct startup sequence for an app on Windows, or else perform some experiments (such as only making the window visible immediately before we call `present` for the first time - I am slightly concerned that rendering wouldn't be scheduled in that case...)

DJMcNab:
A pr which experimentally implements that would be welcome!

PR #639, opened 2024-10-07
Refactor rewrite passes

PoignardAzur:
Create run_rewrite_passes function.
Remove redundant uses of synthetic WidgetState.
Small bugfixes.
Move RenderRoot::root_layout code into the layout pass. Rename update_new_widgets pass to update_widget_tree.

### DIFF
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -466,7 +466,7 @@ impl RenderRoot {
         // Note: this code doesn't do any short-circuiting, because each pass is
         // expected to have its own early exits.
         // Calling a run_xxx_pass (or root_xxx) should always be very fast if
-        // the passes don't need to do anything.
+        // the pass doesn't need to do anything.
 
         run_mutate_pass(self);
         run_update_widget_tree_pass(self);
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -27,16 +27,14 @@ use crate::passes::paint::root_paint;
 use crate::passes::recurse_on_children;
 use crate::passes::update::{
     run_update_anim_pass, run_update_disabled_pass, run_update_focus_chain_pass,
-    run_update_focus_pass, run_update_new_widgets_pass, run_update_pointer_pass,
-    run_update_scroll_pass, run_update_stashed_pass,
+    run_update_focus_pass, run_update_pointer_pass, run_update_scroll_pass,
+    run_update_stashed_pass, run_update_widget_tree_pass,
 };
 use crate::text::TextBrush;
 use crate::tree_arena::{ArenaMut, TreeArena};
 use crate::widget::WidgetArena;
 use crate::widget::{WidgetMut, WidgetRef, WidgetState};
-use crate::{
-    AccessEvent, Action, BoxConstraints, CursorIcon, Handled, QueryCtx, Widget, WidgetId, WidgetPod,
-};
+use crate::{AccessEvent, Action, CursorIcon, Handled, QueryCtx, Widget, WidgetId, WidgetPod};
 
 // --- MARK: STRUCTS ---
 
@@ -160,15 +158,7 @@ impl RenderRoot {
         };
 
         // We run a set of passes to initialize the widget tree
-        run_update_new_widgets_pass(&mut root);
-        // TODO - Remove this line
-        let mut dummy_state = WidgetState::synthetic(root.root.id(), root.get_kurbo_size());
-        root.post_event_processing(&mut dummy_state);
-
-        // We run a layout pass right away to have a SetSize signal ready
-        if size_policy == WindowSizePolicy::Content {
-            root.root_layout();
-        }
+        root.run_rewrite_passes();
 
         root
     }
@@ -192,6 +182,7 @@ impl RenderRoot {
             }
             WindowEvent::Resize(size) => {
                 self.size = size;
+                self.root_state().request_layout = true;
                 self.root_state().needs_layout = true;
                 self.state.emit_signal(RenderRootSignal::RequestRedraw);
                 Handled::Yes
@@ -205,7 +196,8 @@ impl RenderRoot {
                 let last = self.last_anim.take();
                 let elapsed_ns = last.map(|t| now.duration_since(t).as_nanos()).unwrap_or(0) as u64;
 
-                self.root_anim_frame(elapsed_ns);
+                run_update_anim_pass(self, elapsed_ns);
+                self.run_rewrite_passes();
 
                 // If this animation will continue, store the time.
                 // If a new animation starts, then it will have zero reported elapsed time.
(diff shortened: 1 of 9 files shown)


PoignardAzur:
This goes 99% of the way towards implementing the Pass Spec RFC. The big missing element that remains is running the rewrite passes in a loop until no flags are set. I'm leaving that off for a future PR.

PoignardAzur:
I'm getting a bunch of test failures from Parley. Not sure why that happens, I don't think we've bumped dependencies?

PoignardAzur:
Turns out the problem was that fonts weren't loaded when I ran the layout pass.

I'll probably open an issue in Parley, because I really shouldn't get an "index out of range" panic for this kind of problem.

masonry/src/render_root.rs
```diff
@@ -497,44 +453,22 @@ impl RenderRoot {
...
-        // Always do this before sending focus change, since this event updates the focus chain.
-        if self.root_state().update_focus_chain {
-            run_update_focus_chain_pass(self);
-        }
-
-        run_update_focus_pass(self, widget_state);
-
-        if self.root_state().request_anim {
+    // --- MARK: REWRITE PASSES ---
+    pub(crate) fn run_rewrite_passes(&mut self) {
```
Philipp-M, line 457: Could use a doc comment, explaining what is meant with "rewrite pass"

masonry/src/render_root.rs
```diff
@@ -497,44 +453,22 @@ impl RenderRoot {
...
+        // TODO - Rerun passes if invalidation flags are still set
+
+        run_mutate_pass(self);
+        run_update_widget_tree_pass(self);
+        run_update_disabled_pass(self);
+        run_update_stashed_pass(self);
+        run_update_focus_chain_pass(self);
+        run_update_focus_pass(self);
+        root_layout(self);
```
Philipp-M, line 466: I think this runs a full layout pass even if it's not requested/needed doesn't it? (I haven't checked the other passes in detail, might be worth checking this as well)

masonry/src/render_root.rs
```diff
@@ -497,44 +453,22 @@ impl RenderRoot {
...
+        run_update_scroll_pass(self);
+        root_compose(self);
+        run_update_pointer_pass(self);
```
Philipp-M, line 469: Why not the method syntax here? (in general for passes and all the functions there, seems (almost) every function in the passes has `&mut RenderRoot` as parameter)

PoignardAzur: I'm a little hesitant to add impl blocks for RenderRoot outside of `render_root.rs`. Doing so seems just a little too surprising.

Calling free functions with self is a little surprising, but it's not actually more verbose, so I'd say it's fine.

masonry/src/render_root.rs
PoignardAzur: Oh, good catch.

Yeah, right now I think layout is the only pass with no "short circuits" whatsoever.

(Though update_pointer and update_focus could have more explicit short circuits too.)

masonry/src/render_root.rs
```diff
@@ -98,6 +96,15 @@ pub struct RenderRootOptions {
...
     pub scale_factor: f64,
+
+    /// Add a font from its raw data for use in tests.
+    /// The font is added to the fallback chain for Latin scripts.
+    /// This is expected to be used with `use_system_fonts = false`
+    /// to ensure rendering is consistent cross-platform.
+    ///
+    /// We expect to develop a much more fully-featured font API in the future, but
+    /// this is necessary for our testing of Masonry.
+    pub test_font: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
PoignardAzur: I'm fine with leaving that as-is for now.

masonry/src/render_root.rs
PoignardAzur: Added an early exist to root_layout.

masonry/src/render_root.rs
PoignardAzur: I added a short comment. Eventually I'll probably link to the passes documentation.

masonry/src/render_root.rs
```diff
@@ -501,44 +453,28 @@ impl RenderRoot {
...
-            run_update_scroll_pass(self);
-        }
-
-        if self.root_state().needs_compose && !self.root_state().needs_layout {
-            root_compose(self, widget_state);
-        }
-
-        // Update the disabled and stashed state if necessary
-        // Always do this before updating the focus-chain
-        if self.root_state().needs_update_disabled {
```
Philipp-M, line 526: I think these should also be either in the pass function below this line, or in `run_rewrite_passes`? (same for the other  `self.root_state().needs_...` below)

PoignardAzur: Each pass handles its own early-exit. I think that's cleaner than having a lot of sort-circuit code in run_rewrite_passes, for a few reasons:

- It makes the run_rewrite_passes function shorter and very easy to read.
- In a lot of cases the early exit is part of the recursive algorithm, so doing an early-exit in run_rewrite_passes ends up duplicating code.
- I think a general rule of "all passes do their own early exit" is easier to remember than "these passes do their own early exit, these passes don't and need to be called within a short-circuit block".

I should probably document this in the code though.

Philipp-M: Ok, I guess I missed that.

Philipp-M (approved):
I hope I haven't missed anything significant, but I think this is certainly a good improvement otherwise.

masonry/src/render_root.rs
```diff
@@ -159,21 +167,23 @@ impl RenderRoot {
...
+            // Make sure that all of these fonts are in the fallback chain for the Latin script.
+            // <https://en.wikipedia.org/wiki/Script_(Unicode)#Latn>
+            root.state
+                .font_context
+                .collection
+                .append_fallbacks(*b"Latn", families.iter().map(|(family, _)| *family));
         }
 
+        // We run a set of passes to initialize the widget tree
+        root.run_rewrite_passes();
```
DJMcNab, line 181: It's a bit weird to be doing a layout pass here, since the size of the window doesn't meaningfully exist yet.

I suspect some of this lifecycle/ordering things might need to be thought through very carefully.
See also #590. OTOH, this isn't urgent to fix, so this version is fine.

masonry/src/widget/tests/snapshots/masonry__widget__tests__lifecycle_basic__app_creation.snap
```diff
@@ -11,6 +11,10 @@ expression: record
     L(
         BuildFocusChain,
     ),
+    Layout(
+        0.0W×0.0H,
```
DJMcNab, line 15: Ha, indeed.

masonry/src/render_root.rs
PoignardAzur: I think it acts as a decent smoke test for layout bugs that lead to crashes when size constraints are zero. And we do want to detect those bugs early.

PR #641, opened 2024-10-07
Add marker comments to passes

PoignardAzur:


waywardmonkeys (approved):
I guess you're finding these useful then...

PoignardAzur:
Oh, absolutely. They're helpful for navigation, especially in very large files.

PR #643, opened 2024-10-07
Refactor Blur case of on_access_event pass

PoignardAzur:
Match how tab focus works in on_text_event pass.


mwcampbell (approved):

DJMcNab (commented):
There is some weirdness here - you can now get AccessKit actions you never registered for if you're focusable. Maybe that's always been the case - see that button never explicitly requests the.
The semantics of some of this stuff on the AccessKit side isn't immediately obvious to me.

PR #644, opened 2024-10-07
Refactor safety rails for layout pass

PoignardAzur:
Merge call_widget_method_with_checks, run_layout_inner and run_layout_on.
Remove needs_visit flag.
Add checks for invalid values in place_child, set_child_translation and run_layout.

### DIFF
--- a/masonry/src/passes/layout.rs
+++ b/masonry/src/passes/layout.rs
@@ -110,6 +110,7 @@ pub(crate) fn run_layout_on<W: Widget>(
 
     #[cfg(debug_assertions)]
     {
+        let name = widget.short_type_name();
         for child_id in widget.children_ids() {
             let child_id = child_id.to_raw();
             let child_state = state_mut.children.get_child_mut(child_id).unwrap().item;
@@ -121,7 +122,7 @@ pub(crate) fn run_layout_on<W: Widget>(
             if child_state.request_layout {
                 debug_panic!(
                     "Error in '{}' {}: LayoutCtx::run_layout() was not called with child widget '{}' {}.",
-                    widget.short_type_name(),
+                    name,
                     pod.id(),
                     child_state.widget_name,
                     child_state.id,
@@ -131,7 +132,7 @@ pub(crate) fn run_layout_on<W: Widget>(
             if child_state.is_expecting_place_child_call {
                 debug_panic!(
                     "Error in '{}' {}: LayoutCtx::place_child() was not called with child widget '{}' {}.",
-                    widget.short_type_name(),
+                    name,
                     pod.id(),
                     child_state.widget_name,
                     child_state.id,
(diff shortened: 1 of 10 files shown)


jaredoconnell:
Interestingly in the `http_cats` example layout is being called repeatedly somewhere. There is a warning constantly printing to the logs:
```
20:38:38.073Z  WARN Widget requested layout during layout pass
20:38:38.091Z  WARN Widget requested layout during layout pass
```
However, the re-painting is "fixing" the perceived lagginess of the http_cats example, confirming that it was a problem with paint not getting called.

jaredoconnell:
Also, when clicking activate from the first time in the mason example, it's crashing on a NaN value. I'm guessing that this is just detecting a bug elsewhere, and that it isn't a problem with the changes, which improve invalid number detection.

PoignardAzur:
> I'm guessing that this is just detecting a bug elsewhere, and that it isn't a problem with the changes, which improve invalid number detection.

Yup, that's the system working as intended.

> Interestingly in the http_cats example layout is being called repeatedly somewhere. There is a warning constantly printing to the logs

That one is odd though.

PoignardAzur:
> Also, when clicking activate from the first time in the mason example, it's crashing on a NaN value.

I've fixed it. I don't really like my solution, but it'll do for now.

PoignardAzur:
> Interestingly in the http_cats example layout is being called repeatedly somewhere. There is a warning constantly printing to the logs

Should be fixed now. I don't know if that means the lagginess is back.

masonry/src/passes/layout.rs
```diff
@@ -25,207 +25,139 @@ fn rect_contains(larger: &Rect, smaller: &Rect) -> bool {
...
+        pod.id(),
+        widget.reborrow_mut(),
+        state.children.reborrow_mut(),
+        |_, state| {
+            if state.item.is_stashed {
+                state.item.needs_layout = false;
+                state.item.request_layout = false;
+            }
+        },
+    );
```
PoignardAzur, line 99: Five minutes after pushing this code, it occurs to me it basically doesn't work as a safety net, because it doesn't cover grandchildren.

I really need to sit down and write down how layout and stashing interact at some point.

jaredoconnell (commented):
It appears that this broke the portal widget for me. No scrolling happens, and the scroll bar doesn't show up. Tested on http_cats.

The Xilem calc example doesn't start and outputs the error:
```
thread 'main' panicked at masonry/src/box_constraints.rs:145:13:
Bad BoxConstraints passed to Flex: BoxConstraints { min: 0.0W×-2.0H, max: 0.0W×-2.0H }
```

PoignardAzur:
Okay, looks like the tweak I did to flex code is responsible.

Interestingly, while investigating this bug and checking out previous commits in this PR, I bumped into the "http_cats is lagging" problem for the first time. I don't think I had it last time I worked on this commit. Weird.

PoignardAzur:
Fix the flex bug, scrollbar is visible again.

The http_cats example is still laggy, but I'm starting to think that's a non-deterministic bug. I don't think it's a regression with this specific PR.

jaredoconnell (approved):
I had to make one fix to Grid so that it passed in a valid BoxConstraint. The problem was when the window wasn't rendered yet, resulting in a zero-sized BoxConstraint on first layout on MacOS. So be aware that I committed to this branch.

With that fix I do not see any regressions.

PR #647, opened 2024-10-08
Send synthetic PointerLeave event when pointer capture is lost

PoignardAzur:


jaredoconnell (approved):
I can't find any regressions while testing the examples.

PR #648, opened 2024-10-08
Add doc in layout about not adding/removing children

PoignardAzur:


Philipp-M (approved):
LGTM. I wonder, if it's possible to make this impossible via the type-system

PoignardAzur:
It's possible if you only allow WidgetPod creation from a context method. That's a feature that's officially out of scope for now, because implementing it requires some fairly heavy changes to Xilem and Masonry.

But even then, users can always drop a WidgetPod and not call `remove_child`. You can emit warnings or panics when that happens, but you can't really make it impossible in the type system.

PR #649, opened 2024-10-08
Add StashedChanged event

PoignardAzur:


jaredoconnell (approved):
I don't know if there's any functionality to test for this, but the code looks fine. Ideally there would be a test that would validate that the lifecycle event is received.

PR #651, opened 2024-10-08
Remove `BoxConstraints::min`

PoignardAzur:
I'll probably break this into multiple PRs at some point. This creates a *lot* of snapshot test errors.

### DIFF
--- a/masonry/src/box_constraints.rs
+++ b/masonry/src/box_constraints.rs
@@ -8,9 +8,10 @@ use vello::kurbo::Size;
 /// The layout strategy for Masonry is strongly inspired by Flutter,
 /// and this struct is similar to the [Flutter BoxConstraints] class.
 ///
-/// At the moment, it represents simply a minimum and maximum size.
+/// At the moment, it represents only a maximum size.
 /// A widget's [`layout`] method should choose an appropriate size that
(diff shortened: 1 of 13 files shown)


jaredoconnell:
This is what happened to the calc example:
image

With min removed, we're going to need to add the RequestedSize enum to the layout calls. That will allow us to account for the case where min is currently used. Basically just tell it to fill the space instead of use the min size.

When I get a chance I'll experiment with that route and if I come up with something useful I'll submit a PR.

PoignardAzur:
I'd rather have an "expand" attribute on the button itself, if we go that route. I still don't like the idea of telling a widget "you need to be at least / exactly this size".

But yeah, these examples need to get un-broken before this can get out of draft.

PoignardAzur:
For the record, I think the way to go is to split this PR into multiple chunks. So first we might call `loosen()` in a bunch of places where we currently pass tight constraints, and see which errors pop up. Then once roughly all constraints we pass are loose, we can remove the `mix` parameter without disruption.

jaredoconnell:
I'm currently experimenting with replacing BoxConstraints itself.

PoignardAzur:
This is superseded by #812 and Jared's own work.

Issue #346, opened 2023-12-04
Where do "hot" and "active" come from?


PoignardAzur:
Masonry borrows Druid's "hot" and "active" nomenclature for widgets being hovered and widgets being clicked, respectively.

But where does this nomenclature come from? The earliest example I was able to find was Casey Muratori's legendary [Immediate-Mode Graphical User Interfaces](https://www.youtube.com/watch?v=Z1qyvQsjK5Y) video from 2005. (Yes, that's the Handmade Hero guy.)

I'm extremely curious to know where *he* got it from, if anywhere.

PR #654, opened 2024-10-08
Remove last mentions of hot state from code

PoignardAzur:


masonry/src/contexts.rs
```diff
@@ -362,13 +362,13 @@ impl_context_method!(EventCtx<'_>, {
     /// Set the cursor icon.
     ///
     /// This setting will be retained until [`clear_cursor`] is called, but it will only take
-    /// effect when this widget is [`hot`] and/or [`has_pointer_capture`]. If a child widget also
+    /// effect when this widget is [`hovered`] and/or [`has_pointer_capture`]. If a child widget also
```
waywardmonkeys, line 365: I understand that it sort of reads poorly, but I think this link should be `is_hovered` since that's the actual name. If it is just "hovered", then it probably should lose the backticks to avoid letting people think it is a function name.

waywardmonkeys (approved):
One nit about something in the comments.

I think this is solely a terminology change, so that's okay. But it would be nice if I didn't have to figure that out by reading the whole PR to look for something else happening or not. :)

Philipp-M:
I'm not sure if the name hovered is really correct, *unless* it's planned to later check, if it really intersects the widget (thinking about custom shapes etc.).
IIRC, it was an initial check if the bounding-box of that widget, or its children are intersected. I think that's where the name came from instead of hovered?

PoignardAzur:
> IIRC, it was an initial check if the bounding-box of that widget, or its children are intersected. I think that's where the name came from instead of hovered?

For where the name came from, see https://github.com/linebender/xilem/issues/346

It was meant to be synonymous with "hovered".

> I'm not sure if the name hovered is really correct, unless it's planned to later check, if it really intersects the widget (thinking about custom shapes etc.).

I do plan to eventually check that. (After the next release.)

PR #655, opened 2024-10-08
Make WidgetState private

PoignardAzur:
This enforces a pattern in the codebase where that WidgetState can only be accessed through pass contexts, which prevents invalidation bugs.

### DIFF
--- a/masonry/src/widget/widget_ref.rs
+++ b/masonry/src/widget/widget_ref.rs
@@ -6,7 +6,7 @@ use std::ops::Deref;
 use smallvec::SmallVec;
 use vello::kurbo::Point;
 
-use crate::{QueryCtx, Widget, WidgetId, WidgetState};
+use crate::{QueryCtx, Widget, WidgetId};
 
 /// A rich reference to a [`Widget`].
 ///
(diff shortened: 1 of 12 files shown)


masonry/src/widget/widget_ref.rs
```diff
@@ -169,7 +168,7 @@ impl<'w> WidgetRef<'w, dyn Widget> {
     pub fn find_widget_at_pos(&self, pos: Point) -> Option<WidgetRef<'_, dyn Widget>> {
         let mut innermost_widget = *self;
 
-        if !self.state().window_layout_rect().contains(pos) {
+        if !self.ctx.widget_state.window_layout_rect().contains(pos) {
```
jaredoconnell, line 171: Is there a reason you're calling `window_layout_rect` on the state instead of the context? It would seem better abstracted to call it on the context.

jaredoconnell (approved):
Looks good. I just have one minor comment.

masonry/src/widget/widget_ref.rs
PoignardAzur: Good catch.

PR #657, opened 2024-10-08
Add target field to EventCtx

PoignardAzur:
Remove target field from AccessEvent

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -68,6 +68,7 @@ pub struct EventCtx<'a> {
     pub(crate) widget_state: &'a mut WidgetState,
     pub(crate) widget_state_children: ArenaMutChildren<'a, WidgetState>,
     pub(crate) widget_children: ArenaMutChildren<'a, Box<dyn Widget>>,
+    pub(crate) target: WidgetId,
     pub(crate) allow_pointer_capture: bool,
     pub(crate) is_handled: bool,
 }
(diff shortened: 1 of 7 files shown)


jaredoconnell (approved):
Works with no visible regressions in accessibility when testing calc_masonry. I learned how to use VoiceOver for the review.

masonry/src/passes/event.rs
```diff
@@ -51,6 +52,7 @@ fn run_event_pass<E>(
             widget_state: state_mut.item,
             widget_state_children: state_mut.children,
             widget_children: widget_mut.children,
+            target: original_target.unwrap(),
```
DJMcNab, line 55: Why is the `unwrap` here rather than at the definition of `original_target`?

masonry/src/contexts.rs
```diff
@@ -716,6 +717,13 @@ impl EventCtx<'_> {
         self.is_handled
     }
 
+    /// The widget originally targeted by the event.
+    ///
+    /// This will be different from [`widget_id`](Self::widget_id) during event bubbling.
```
DJMcNab, line 722: Is it actually meaningful for accessibility events to bubble?

PoignardAzur: I *think* it is? Though we don't really use bubbling anywhere for now.

> Every implementation seems to check if the event targets self

I think most of those checks are vestigial, from the initial implementation where accessibility events were sent to the entire tree (with the bloom filter to cut some subtrees off).

masonry/src/passes/event.rs
PoignardAzur: Because a value of None is valid at that point.

If `original_target` is None, then the loop is never entered.

DJMcNab: Thanks. Sorry for the confusion

PR #658, opened 2024-10-08
Add needs_rewrite_passes and content_changed methods

PoignardAzur:
These methods will be used with #639 to implement a fixpoint loop for rewrite passes.

This also lets us remove `mark_changed`, since we can use `content_changed` for the same purpose.

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -131,7 +131,6 @@ where
         let mut ctx = ViewCtx {
             widget_map: WidgetMap::default(),
             id_path: Vec::new(),
-            view_tree_changed: false,
             proxy,
             runtime: self.runtime,
         };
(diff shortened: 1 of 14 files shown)


xilem/src/lib.rs
```diff
@@ -273,12 +271,6 @@ impl ViewCtx {
         }
     }
 
-    pub fn mark_changed(&mut self) {
-        if cfg!(debug_assertions) {
-            self.view_tree_changed = true;
-        }
-    }
-
```
Philipp-M, line 281: I mean I agree with that change, but it doesn't seem related to the title of this PR?

PoignardAzur: The idea is Xilem now uses `needs_rewrite_passes` to check the same thing.

Philipp-M (approved):
LGTM.

But I wonder whether we need this (previously `mark_changed` now `content_changed`) at all? I think it's kind of redundant?

PoignardAzur:
Eh, maybe. It'll be useful on the Masonry side. We might remove it on the Xilem side.

jaredoconnell:
Nice. This simplifies the code and prevents accidentally leaving a widget stale.

PR #659, opened 2024-10-08
Add RenderRoot::request_render_all method

PoignardAzur:
Add a method which brute-force resets the paint and accessibility flags for all widgets.
This is useful for rescale, and for testing purposes where you want a quick and dirty way to repaint everything.

### DIFF
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -559,7 +559,6 @@ impl RenderRoot {
 
     pub(crate) fn request_render_all(&mut self) {
         fn request_render_all_in(
-            global_state: &mut RenderRootState,
             mut widget: ArenaMut<'_, Box<dyn Widget>>,
             state: ArenaMut<'_, WidgetState>,
         ) {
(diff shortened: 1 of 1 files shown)


waywardmonkeys (commented):
This needs a commit message.

On the one hand, this looks like it solving an issue in window resizing, however, it also seems like it might be related to discussions with @frankvgompel about theming and color changes (apart from not being publicly accessible)?

PoignardAzur:
It's about rescaling, not resizing, for which layout works fine.

And yeah, it's also mean to support  frankvgompel's use case. Since they're working on a fork, the method doesn't need to be public.

I added a commit message, but I'm mostly of the opinion that a commit title alone is enough for self-explanatory changes like this?

waywardmonkeys:
Oops, yes, rescale vs resize ... I was wondering in my head that resize sounded wrong (why needed here).

For the commit message, I'd meant something that just stated why it was there, so in this case, that it forces a repaint of everything when layout and other stuff hasn't changed to otherwise cause a repaint, and that changing the DPI scale is the current usage.

But no matter ... this one is pretty clear.

waywardmonkeys (approved):

jaredoconnell:
We need to be careful to make sure people don't over-use this function.

masonry/src/render_root.rs
```diff
@@ -559,6 +557,31 @@ impl RenderRoot {
...
     }
 
+    pub(crate) fn request_render_all(&mut self) {
+        fn request_render_all_in(
+            mut widget: ArenaMut<'_, Box<dyn Widget>>,
+            state: ArenaMut<'_, WidgetState>,
+        ) {
+            state.item.needs_paint = true;
+            state.item.needs_accessibility = true;
+            state.item.request_paint = true;
```
DJMcNab, line 567: With the current state of Masonry, I'm less sure that we need a repaint on a rescale. The logical size of everything is still the same, and there is no way for a widget to access the UI scale in its painting method.

(Accessibility does need it, as the accesskit rectangles work on physical coordinates)

PoignardAzur: I'm still hazy on how scale even works in Masonry. One of my todo items before next release is to go all scale-adjacent code and document out semantics.

PR #660, opened 2024-10-08
Add request_render context method

### LINKED ISSUE #585: Add `request_render` (and rename `request_paint`)
In many cases, if you update state which impacts how the element is painted, the accessibility state for the widget  also needs to be updated. The only time this isn't the case is in response to an animation change.

Currently, this means that you need to call `request_paint` *and* `request_accessibility_update` in these places.
We propose adding a new method `request_render` which calls both `request_paint` and `request_accessibility_update`, and updating most uses of `request_paint` to use this method.

See original discussion in https://github.com/linebender/xilem/pull/577

Once this is done, we would also like rename `request_paint` to `request_paint_only`, to steer users towards `request_render`.

PoignardAzur:
Remove some redundant request_paint calls

Fixes #585.

### DIFF
--- a/masonry/examples/calc_masonry.rs
+++ b/masonry/examples/calc_masonry.rs
@@ -176,7 +176,6 @@ impl Widget for CalcButton {
             match event.action {
                 accesskit::Action::Default => {
                     ctx.submit_action(Action::Other(Box::new(self.action)));
-                    ctx.request_paint_only();
                 }
                 _ => {}
             }
--- a/masonry/examples/calc_masonry.rs
+++ b/masonry/examples/calc_masonry.rs
@@ -176,7 +176,7 @@ impl Widget for CalcButton {
             match event.action {
                 accesskit::Action::Default => {
                     ctx.submit_action(Action::Other(Box::new(self.action)));
-                    ctx.request_paint();
+                    ctx.request_paint_only();
                 }
                 _ => {}
             }
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -482,11 +482,19 @@ impl<'w> QueryCtx<'w> {
 // --- MARK: UPDATE FLAGS ---
 // Methods on MutateCtx, EventCtx, and LifeCycleCtx
 impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
+    /// Request a [`paint`](crate::Widget::paint) and an [`accessibility`](crate::Widget::accessibility) pass.
+    pub fn request_render(&mut self) {
+        trace!("request_render");
+        self.widget_state.request_paint = true;
+        self.widget_state.needs_paint = true;
+        self.widget_state.needs_accessibility = true;
+        self.widget_state.request_accessibility = true;
+    }
+
     /// Request a [`paint`](crate::Widget::paint) pass.
     ///
-    /// This doesn't request an [`accessibility`](crate::Widget::accessibility) pass,
-    /// which usually isn't what you want.
-    /// See [`request_render`](Self::request_render).
+    /// Unlike [`request_render`](Self::request_render), this does not request an [`accessibility`](crate::Widget::accessibility) pass.
+    /// Use request_render unless you're sure an accessibility pass is not needed.
     pub fn request_paint_only(&mut self) {
         trace!("request_paint");
         self.widget_state.request_paint = true;
@@ -495,20 +503,10 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
 
     /// Request an [`accessibility`](crate::Widget::accessibility) pass.
     ///
-    /// This doesn't request a [`paint`](crate::Widget::paint) pass,
-    /// which usually isn't what you want.
-    /// See [`request_render`](Self::request_render).
+    /// This doesn't request a [`paint`](crate::Widget::paint) pass.
+    /// If you want to request both an accessibility pass and a paint pass, use [`request_render`](Self::request_render).
     pub fn request_accessibility_update(&mut self) {
         trace!("request_accessibility_update");
-        self.widget_state.needs_accessibility = true;
-        self.widget_state.request_accessibility = true;
-    }
-
-    /// Request a [`paint`](crate::Widget::paint) and an [`accessibility`](crate::Widget::accessibility) pass.
-    pub fn request_render(&mut self) {
-        trace!("request_render");
-        self.widget_state.request_paint = true;
-        self.widget_state.needs_paint = true;
         self.widget_state.needs_accessibility = true;
         self.widget_state.request_accessibility = true;
     }
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -483,10 +483,37 @@ impl<'w> QueryCtx<'w> {
 // Methods on MutateCtx, EventCtx, and LifeCycleCtx
 impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
     /// Request a [`paint`](crate::Widget::paint) pass.
-    pub fn request_paint(&mut self) {
+    ///
+    /// This doesn't request an [`accessibility`](crate::Widget::accessibility) pass,
+    /// which usually isn't what you want.
+    /// See [`request_render`](Self::request_render).
+    pub fn request_paint_only(&mut self) {
         trace!("request_paint");
         self.widget_state.request_paint = true;
         self.widget_state.needs_paint = true;
+    }
+
+    /// Request an [`accessibility`](crate::Widget::accessibility) pass.
+    ///
+    /// This doesn't request a [`paint`](crate::Widget::paint) pass,
+    /// which usually isn't what you want.
+    /// See [`request_render`](Self::request_render).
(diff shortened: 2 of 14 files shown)


waywardmonkeys (commented):
This is difficult to review without having more knowledge about what each interaction is expected to do / permitted to do, so that you know which things need to be updated.

It sort of sucks, but it would almost be useful to have a one line comment at each spot saying what's getting updated or not (to give an indication of why).

PoignardAzur:
Yeah. Ideally we'd want some way to test that the request flags are set (eg that we don't have a change that affected layout but didn't set the request_layout flag), but it would probably be pretty painful to implement.

DJMcNab:
Should this be marked as fixing #585?

PoignardAzur:
Oh, right, knew I forgot something.

waywardmonkeys:
> Yeah. Ideally we'd want some way to test that the request flags are set (eg that we don't have a change that affected layout but didn't set the request_layout flag), but it would probably be pretty painful to implement.

What I mean is that in each spot, you've made a decision as to which updates are appropriate and there were some criteria that you used in making that decision. It might be good to note that at the various callsites in a comment so that future-us will know why each decision was made and what the intent was.

masonry/src/widget/spinner.rs
```diff
@@ -93,15 +93,13 @@ impl Widget for Spinner {
...
                 ctx.request_anim_frame();
-                ctx.request_paint();
             }
             LifeCycle::AnimFrame(interval) => {
                 self.t += (*interval as f64) * 1e-9;
                 if self.t >= 1.0 {
                     self.t = 0.0;
                 }
                 ctx.request_anim_frame();
-                ctx.request_paint();
```
DJMcNab, line 104: This strikes me as strange - does `AnimFrame` implicitly request a repaint? It seems to me that there are reasonable animations where you wouldn't want that to be the case.

masonry/examples/calc_masonry.rs
```diff
@@ -176,7 +176,7 @@ impl Widget for CalcButton {
             match event.action {
                 accesskit::Action::Default => {
                     ctx.submit_action(Action::Other(Box::new(self.action)));
-                    ctx.request_paint();
+                    ctx.request_paint_only();
```
DJMcNab, line 179: I'm surprised this line is necessary at all. Nothing has changed which would require a repaint to happen.

masonry/src/contexts.rs
```diff
@@ -462,12 +462,39 @@ impl<'w> QueryCtx<'w> {
...
+    /// See [`request_render`](Self::request_render).
+    pub fn request_accessibility_update(&mut self) {
+        trace!("request_accessibility_update");
+        self.widget_state.needs_accessibility = true;
+        self.widget_state.request_accessibility = true;
+    }
+
+    /// Request a [`paint`](crate::Widget::paint) pass.
+    ///
+    /// This doesn't request an accessibility pass, which usually isn't what you want.
```
DJMcNab, line 488: ? Yes it does?

masonry/src/widget/button.rs
```diff
@@ -104,15 +104,15 @@ impl Widget for Button {
             match event.action {
                 accesskit::Action::Default => {
                     ctx.submit_action(Action::ButtonPressed(PointerButton::Primary));
-                    ctx.request_paint();
+                    ctx.request_paint_only();
```
DJMcNab, line 107: Again, here. How would you expect this paint to be different from the previous one?

masonry/src/widget/checkbox.rs
```diff
@@ -94,16 +92,15 @@ impl Widget for Checkbox {
...
+                    ctx.request_render();
                 }
                 _ => {}
             }
         }
     }
 
     fn on_status_change(&mut self, ctx: &mut LifeCycleCtx, _event: &StatusChange) {
-        ctx.request_paint();
+        ctx.request_render();
```
DJMcNab, line 103: Hmm, should this be the default implementation of `on_status_changed`?

masonry/src/contexts.rs
PoignardAzur: That's what I get for copy-pasting.

masonry/examples/calc_masonry.rs
PoignardAzur: Right. Now that you mention it, the paint method of that widget is empty, so `request_paint` should never be called anyway.

masonry/src/widget/checkbox.rs
PoignardAzur: Maybe. I'd rather not have a default implementation at all for now.

masonry/src/widget/spinner.rs
PoignardAzur: Nope, you're right, the new test for the animated spinner was failing because of that.

(Hurray for screenshot tests!)

PoignardAzur:
@waywardmonkeys I've added some comments to some of the call sites.

I don't want to comment *every* call site; not just because that would be redundant, but because I think redundant comments actually make code *less* understandable (they make the reader wonder if they missed something).

So for something like this:

```rust
    pub fn set_line_break_mode(&mut self, line_break_mode: LineBreaking) {
        self.widget.line_break_mode = line_break_mode;
        self.ctx.request_layout();
    }
```

I think it's self-explanatory why changing the line breaking mode of a label impacts its layout.

Some other notes:

- Some of the changes in the PR were `request_paint` -> `request_render` where the widget didn't actually benefit from changing the accessibility node, just because I did a project-wide rename. I think it's fine, request_render is a fine default, and it's okay if it schedules some redundant work. Still, I've reverted some of these changes to `request_paint_only`.
- The Portal/ScrollBar code is kind of cursed and I tried to touch the request flags as little as possible there and keep behavior unchanged. In the future we might change requests to be a little more narrow (eg compose instead of paint).

masonry/src/widget/portal.rs
```diff
@@ -20,6 +20,7 @@ use crate::{
 
 // TODO - refactor - see https://github.com/linebender/xilem/issues/366
 // TODO - rename "Portal" to "ScrollPortal"?
+// TODO - Document which cases need request_layout, request_compose and request_render
```
jaredoconnell, line 23: Is there a reason this isn't being done in this PR?

jaredoconnell (approved):
This is definitely an improvement. Should help reduce stale accessibility nodes.
I tested general (non-screen-reader) use, and I don't see any regressions.

masonry/src/contexts.rs
```diff
@@ -483,12 +483,36 @@ impl<'w> QueryCtx<'w> {
 // Methods on MutateCtx, EventCtx, and LifeCycleCtx
 impl_context_method!(MutateCtx<'_>, EventCtx<'_>, LifeCycleCtx<'_>, {
     /// Request a [`paint`](crate::Widget::paint) pass.
-    pub fn request_paint(&mut self) {
+    ///
+    /// This doesn't request an [`accessibility`](crate::Widget::accessibility) pass,
+    /// which usually isn't what you want.
```
waywardmonkeys, line 488: I find this wording confusing with the multiple negatives. Same for the function below.

masonry/src/widget/button.rs
```diff
@@ -82,7 +82,8 @@ impl Widget for Button {
             PointerEvent::PointerDown(_, _) => {
                 if !ctx.is_disabled() {
                     ctx.capture_pointer();
-                    ctx.request_paint();
+                    // Changes in pointer capture impact appearance, but not accessibility node
```
waywardmonkeys, line 85: This is great and exactly what I was hoping for when I asked about adding some comments. Thanks!

masonry/src/widget/portal.rs
PoignardAzur: Because it's more work than the rest of the PR combined.

masonry/src/widget/button.rs
```diff
@@ -104,15 +106,15 @@ impl Widget for Button {
...
                 }
                 _ => {}
             }
         }
     }
 
     fn on_status_change(&mut self, ctx: &mut LifeCycleCtx, _event: &StatusChange) {
-        ctx.request_paint();
+        // Changes in hovered/focused status impact appearance, but not accessibility node
+        ctx.request_paint_only();
```
DJMcNab, line 117: Changing the focus status *does* change the accessibility node - https://github.com/linebender/xilem/blob/3ba0df6064a48a92a3efab51455efd378776d33e/masonry/src/passes/accessibility.rs#L119.

However, requesting the accessibility update in that case is not the responsibility of this code, so this is still true.
One thing I'm not sure of - does this update actually happen?

masonry/src/widget/split.rs
```diff
@@ -355,15 +355,18 @@ impl WidgetMut<'_, Split> {
     /// Set whether the split point can be changed by dragging.
     pub fn set_draggable(&mut self, draggable: bool) {
         self.widget.draggable = draggable;
-        self.ctx.request_paint();
+        // Bar mutability impacts appearance, but not accessibility node
+        // TODO - This might change in a future implementation
```
DJMcNab, line 359: I agree that this is suspicious

masonry/src/widget/button.rs
PoignardAzur: Good catch. It would have to happen in `run_update_focus_pass`, and it currently doesn't.

PR #661, opened 2024-10-08
Add TestHarness::animate_ms method

PoignardAzur:


masonry/src/widget/spinner.rs
```diff
@@ -183,9 +182,8 @@ mod tests {
         let mut harness = TestHarness::create(spinner);
         assert_render_snapshot!(harness, "spinner_init");
 
-        // TODO - See https://github.com/linebender/xilem/issues/369
-        //harness.move_timers_forward(Duration::from_millis(700));
-        //assert_render_snapshot!(harness, "spinner_700ms");
+        harness.animate_ms(700);
+        assert_render_snapshot!(harness, "spinner_700ms");
```
jaredoconnell, line 186: Would it make sense to do this test with a second render snapshot after another time period?

jaredoconnell (approved):
It's always good to move code out of large functions. Tested locally. Just one comment.

masonry/src/widget/spinner.rs
PoignardAzur: I guess it could be worth doing to check the cycle-ness of it.

jaredoconnell (approved):

PR #663, opened 2024-10-09
Improve WidgetMut documentation

PoignardAzur:
This PR introduces an `ImplMut` macro and a `SelfMut` macro which resolve to different types depending on whether we're generating doc. The doc version uses the unstable receiver trait.

For details, see [this zulip thread](https://xi.zulipchat.com/#narrow/stream/317477-masonry/topic/Improving.20docs.20for.20WidgetMut).

> As far I know, the doc is generated using a nightly compiler, so that should be fine. It means we are using unstable features to generate our docs. And not only that, unstable features that are scheduled to be broken fairly soon (the receiver trait syntax above works with a current nightly compiler, but it's not the one promoted by the latest RFC).
>
> But since this can only cause breakage when uploading a new version and not, like, when people download masonry as a dependency, I think it's a worthwhile tradeoff.

I'm not quite sure how to document this yet. It's a pretty far-ranging change.

Also, because I want to avoid merge conflicts with other PRs in the pipeline, imports in this PR are still messy.

Leaving this PR as a draft for now.

### DIFF
--- a/masonry/src/lib.rs
+++ b/masonry/src/lib.rs
@@ -100,6 +100,8 @@
 #![cfg_attr(not(debug_assertions), allow(unused))]
 // False-positive with dev-dependencies only used in examples
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
+// Needed because of the SelfMut macro
+#![allow(clippy::needless_arbitrary_self_type)]
 
 // TODO - Add logo
 
--- a/masonry/src/widget/button.rs
+++ b/masonry/src/widget/button.rs
@@ -14,8 +14,8 @@ use crate::paint_scene_helpers::{fill_lin_gradient, stroke, UnitPoint};
 use crate::widget::{Label, WidgetMut, WidgetPod};
 
 use crate::{
-    theme, AccessCtx, AccessEvent, ArcStr, BoxConstraints, EventCtx, Insets, LayoutCtx,
-    LifeCycleCtx, PaintCtx, PointerEvent, Size, StatusChange, TextEvent, Widget, WidgetId,
+    theme, AccessCtx, AccessEvent, ArcStr, BoxConstraints, EventCtx, ImplMut, Insets, LayoutCtx,
+    LifeCycleCtx, PaintCtx, PointerEvent, SelfMut, Size, StatusChange, TextEvent, Widget, WidgetId,
 };
 
 // the minimum padding added to a button.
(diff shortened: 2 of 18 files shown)


masonry/src/lib.rs
```diff
@@ -100,6 +100,8 @@
 #![cfg_attr(not(debug_assertions), allow(unused))]
 // False-positive with dev-dependencies only used in examples
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
+// Needed because of the SelfMut macro
+#![allow(clippy::needless_arbitrary_self_type)]
```
DJMcNab, line 105: It's not clear how this can work without a `#![feature(arbitrary_self_types)]` (probably in a `cfg_attr`)

DJMcNab (commented):
This is a very neat hack. I do think that it is necessary to improve the documentation here substantially.

However, this makes Masonry's code significantly harder to understand, and it also makes implementing external widgets harder (as, for example, there is no longer a reference to `WidgetMut` in the implementations, to even allow the docs about extension traits to be found).

My order of preferred options are:
1) Move all of the widgets to extension traits, ensuring a uniform but bad experience for all users.
2) Keep things as-is until arbitrary self types is stable. Re-evaluate on (approx) 2025-01-09 if new blockers for arbitrary self types v2 appear
3) This solution

masonry/src/lib.rs
PoignardAzur: To be clear, this lint doesn't need `#![feature(arbitrary_self_types)]`. It applies to cases like:

```rust
fn foobar(self: Self) {}
```

which have a more concise version.

DJMcNab: I wasn't commented on the added line - I was commenting on "the top of `masonry/src/lib.rs`", because it isn't clear to me how the new docs can work without enabling the feature for `std::ops::Receiver`.
Like, they do seem to work based on local testing, but I'm really struggling to work out how.

I think it's a rustdoc bug.

PoignardAzur:
> Move all of the widgets to extension traits, ensuring a uniform but bad experience for all users.

If we go for a uniform bad experience, I'd rather convert all methods to free functions, the other solution we discussed.

I still don't like the Ext trait solution. I think it produces worse documentation, and I suspect it would produce worse error messages and a worse learning curve as well.

DJMcNab:
Associated functions wouldn't be world-endingly terrible, yeah. So basically, each block is:
```rust
impl Button {
    pub fn label_mut(this: &mut WidgetMut<Self>) -> WidgetMut<Label>;
}
```
or whatever. Replacing `this` with `self` would also be backwards compatible, which is neat.

PoignardAzur:
Closing this, since we've agreed we won't pursue this doc trick.

PR #665, opened 2024-10-09
Remove masonry/src/text/reference

PoignardAzur:
I'm pretty sure no-one has consulted these files in months. They're dead code, and there's always source control if someone really needs them.


waywardmonkeys (approved):
Emphatically approve.

PR #625, opened 2024-10-03
Text handling refactor

PoignardAzur:

PR #666, opened 2024-10-09
Stop storing text in TextLayout

PoignardAzur:
Remove type parameter from TextLayout.
Tweak error message in text widgets.

This makes the TextLayout code slightly simpler (though with more invariants) in anticipation of splitting it into its own crate.

### DIFF
--- a/masonry/src/text/text_layout.rs
+++ b/masonry/src/text/text_layout.rs
@@ -8,7 +8,7 @@ use std::rc::Rc;
 use parley::context::RangedBuilder;
 use parley::fontique::{Style, Weight};
 use parley::layout::{Alignment, Cursor};
-use parley::style::{Brush as BrushTrait, FontFamily, FontStack, GenericFamily, StyleProperty};
+use parley::style::{FontFamily, FontStack, GenericFamily, StyleProperty};
 use parley::{FontContext, Layout, LayoutContext};
 use vello::kurbo::{Affine, Line, Point, Rect, Size};
 use vello::peniko::{self, Color, Gradient};
@@ -37,8 +37,7 @@ use crate::text::render_text;
 ///
 /// TODO: Update docs to mentionParley
 #[derive(Clone)]
-pub struct TextLayout<T> {
-    text: T,
+pub struct TextLayout {
     // TODO: Find a way to let this use borrowed data
     scale: f32,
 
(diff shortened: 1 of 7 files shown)


PoignardAzur:
This is another chunk of work split off from #625.

masonry/src/text/text_layout.rs
```diff
@@ -37,8 +37,7 @@ use crate::text::render_text;
 ///
 /// TODO: Update docs to mentionParley
 #[derive(Clone)]
-pub struct TextLayout<T> {
-    text: T,
+pub struct TextLayout {
     // TODO: Find a way to let this use borrowed data
```
waywardmonkeys, line 41: Is this comment still accurate?

waywardmonkeys (commented):
Didn't go into detail yet ...

waywardmonkeys:
I'm curious if this has any bearing on changing from just displaying text of a single run of size + color to being able to do richer text via a builder or something.

PoignardAzur:
> I'm curious if this has any bearing on changing from just displaying text of a single run of size + color to being able to do richer text via a builder or something.

I don't think it does.

masonry/src/text/text_layout.rs
PoignardAzur: If you mean the "use borrowed data" comment, yes, it's still accurate.

It's from this commit: https://github.com/linebender/xilem/commit/732cfa8376d5e19fbdd98b4b73806f0c522e740e#diff-da6cde61a550125d612ef93c9aecd16c1210e2cfbfb664d3ba0a9ab00ddc2caf

Not sure what it means either. @DJMcNab ?

jaredoconnell (commented):
The percent label in progress_bar is not updating in the widgets example.

Other than that regression, I don't see any other issues.

PoignardAzur:
> The percent label in progress_bar is not updating in the widgets example.

Fixed.

jaredoconnell (approved):
Tested. It works now.

masonry/src/text/text_layout.rs
DJMcNab: I have no idea. It seems you understand this comment better than me - because I assume that it applied to the `text` field and was misplaced.

DJMcNab (commented):
I find the motivaion for this change a little bit confusing, as the same bookkeeping is still happening, just in more places.

masonry/src/text/text_layout.rs
PoignardAzur: We'll remove it then.

PR #669, opened 2024-10-12
Expose font weight parameter in label widget

melix99:


xilem/src/view/label.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use masonry::{text::TextBrush, widget, ArcStr};
+use masonry::{parley::fontique::Weight, text::TextBrush, widget, ArcStr};
```
Philipp-M, line 4: I think this should probably also be exported at the root in xilem/src/lib, similar as e.g. `TextAlignment`. I'm not sure, maybe as `TextWeight` (and that export could be used here instead then)?

Philipp-M (approved):
Looks good mostly, just some nits.

You could probably add the `.weight()` to our catch-all mason example, so that it's used somewhere (for basic testing).

Philipp-M (approved):
Thanks!

PR #670, opened 2024-10-14
Add basic window control

melix99:
This is the first merge-able part of #606.

It allows to implement custom CSDs via window management signals and a new `WindowHandle` widget which is useful to implement a custom titlebar.

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -6,9 +6,11 @@
 use std::time::Duration;
 
 use accesskit::TreeUpdate;
+use dpi::LogicalPosition;
 use parley::{FontContext, LayoutContext};
 use tracing::{trace, warn};
 use vello::kurbo::Vec2;
+use winit::window::ResizeDirection;
 
 use crate::action::Action;
 use crate::passes::layout::run_layout_on;
@@ -604,6 +606,58 @@ impl_context_method!(
             trace!("submit_action");
             self.global_state
                 .emit_signal(RenderRootSignal::Action(action, self.widget_state.id));
+        }
+
+        /// Start a window drag.
+        ///
+        /// Moves the window with the left mouse button until the button is released.
+        pub fn drag_window(&mut self) {
+            trace!("drag_window");
+            self.global_state
+                .signal_queue
+                .push_back(RenderRootSignal::DragWindow);
+        }
+
+        /// Start a window resize.
+        ///
+        /// Resizes the window with the left mouse button until the button is released.
+        pub fn drag_resize_window(&mut self, direction: ResizeDirection) {
+            trace!("drag_resize_window");
+            self.global_state
+                .signal_queue
+                .push_back(RenderRootSignal::DragResizeWindow(direction));
+        }
+
+        /// Toggle the maximized state of the window.
+        pub fn toggle_maximized(&mut self) {
+            trace!("toggle_maximized");
+            self.global_state
+                .signal_queue
+                .push_back(RenderRootSignal::ToggleMaximized);
+        }
+
+        /// Minimize the window.
+        pub fn minimize(&mut self) {
+            trace!("minimize");
+            self.global_state
+                .signal_queue
+                .push_back(RenderRootSignal::Minimize);
+        }
+
+        /// Exit the application.
+        pub fn exit(&mut self) {
+            trace!("exit");
+            self.global_state
+                .signal_queue
+                .push_back(RenderRootSignal::Exit);
+        }
+
+        /// Show the window menu at a specified position.
+        pub fn show_window_menu(&mut self, position: LogicalPosition<f64>) {
+            trace!("show_window_menu");
+            self.global_state
+                .signal_queue
+                .push_back(RenderRootSignal::ShowWindowMenu(position));
         }
 
         /// Request a timer event.
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -649,7 +649,7 @@ impl MasonryState<'_> {
     pub fn handle_memory_warning(&mut self, _: &ActiveEventLoop) {}
 
     // --- MARK: SIGNALS ---
-    fn handle_signals(&mut self, _event_loop: &ActiveEventLoop, app_driver: &mut dyn AppDriver) {
+    fn handle_signals(&mut self, event_loop: &ActiveEventLoop, app_driver: &mut dyn AppDriver) {
         let WindowState::Rendering { window, .. } = &mut self.window else {
             tracing::warn!("Tried to handle a signal whilst suspended or before window created");
             return;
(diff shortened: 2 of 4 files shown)


masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+// --- MARK: BUILDERS ---
+impl<W: Widget> WindowHandle<W> {
+    pub fn new(child: W) -> Self {
+        Self {
+            child: Some(WidgetPod::new(child)),
+            last_pos: None,
+        }
+    }
+
+    #[doc(alias = "null")]
```
DJMcNab, line 40:
```suggestion
    #[doc(alias = "null")]
```
I don't think this is helpful

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+    AccessCtx, AccessEvent, BoxConstraints, EventCtx, LayoutCtx, LifeCycleCtx, PaintCtx,
+    PointerEvent, Size, StatusChange, TextEvent, Widget, WidgetId,
+};
+
+// TODO: Maybe make this configurable somehow
+const DRAG_THRESHOLD: f64 = 8.0;
+
+/// A titlebar area widget.
+///
+/// An area that can be dragged to move the window.
```
DJMcNab, line 25: This needs some more documentation about how it's expected to be used.

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
+        match self.child.as_mut() {
+            Some(child) => {
+                let size = ctx.run_layout(child, bc);
+                ctx.place_child(child, Point::ORIGIN);
+                bc.constrain(size)
+            }
+            None => bc.max(),
```
DJMcNab, line 119: I'd expect this to impose a max height by default.

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+    fn register_children(&mut self, ctx: &mut crate::RegisterCtx) {
+        if let Some(ref mut child) = self.child {
+            ctx.register_child(child);
+        }
+    }
+
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
+        match self.child.as_mut() {
+            Some(child) => {
+                let size = ctx.run_layout(child, bc);
```
DJMcNab, line 115: Maybe similarly here, although I can see arguments for not doing so

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+///
+/// An area that can be dragged to move the window.
+pub struct WindowHandle<W: Widget> {
+    child: Option<WidgetPod<W>>,
+    last_pos: Option<LogicalPosition<f64>>,
+}
+
+// --- MARK: BUILDERS ---
+impl<W: Widget> WindowHandle<W> {
+    pub fn new(child: W) -> Self {
```
DJMcNab, line 33: So the idea is that you would put a `Flex` (or w/e) containing your titlebar buttons/icon (/maybe window menu) in here, and the non-interactive areas would be treated as a draggable area?

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+
+// TODO: Maybe make this configurable somehow
+const DRAG_THRESHOLD: f64 = 8.0;
+
+/// A titlebar area widget.
+///
+/// An area that can be dragged to move the window.
+pub struct WindowHandle<W: Widget> {
+    child: Option<WidgetPod<W>>,
+    last_pos: Option<LogicalPosition<f64>>,
```
DJMcNab, line 28: Can we rename this to `pointer_down_pos` or similar?

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+                bc.constrain(size)
+            }
+            None => bc.max(),
+        }
+    }
+
+    fn paint(&mut self, _ctx: &mut PaintCtx, _scene: &mut Scene) {}
+
+    fn accessibility_role(&self) -> Role {
+        Role::GenericContainer
```
DJMcNab, line 126: I suspect this is right, but I'm not certain

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+                        PointerButton::Secondary => ctx.show_window_menu(state.position),
+                        _ => (),
+                    }
+                }
+            }
+            PointerEvent::PointerMove(state) => {
+                if let Some(last_pos) = self.last_pos {
+                    let distance = ((state.position.x - last_pos.x).powi(2)
+                        + (state.position.y - last_pos.y).powi(2))
+                    .sqrt();
```
DJMcNab, line 85: No need to sqrt here, since we're just doing a comparison - we can have `DRAG_THRESHOLD_SQUARED`

masonry/src/widget/window_handle.rs
```diff
@@ -0,0 +1,142 @@
...
+            Some(child) => {
+                let size = ctx.run_layout(child, bc);
+                ctx.place_child(child, Point::ORIGIN);
+                bc.constrain(size)
+            }
+            None => bc.max(),
+        }
+    }
+
+    fn paint(&mut self, _ctx: &mut PaintCtx, _scene: &mut Scene) {}
```
DJMcNab, line 123: Nothing drawn at all here feels a bit weird

DJMcNab (commented):
I'm happy with the additional signals; less sure about the new widget

masonry/src/widget/window_handle.rs
melix99: Me neither, I just copied it from [SizedBox::empty()](https://github.com/linebender/xilem/blob/2832766eff557f0d739f3fe430c90bf914caeaaf/masonry/src/widget/sized_box.rs#L91) to be consistent.

masonry/src/widget/window_handle.rs
melix99: Basically, yes.

melix99:
Well, I guess I should've included more information on how the `WindowHandle` is supposed to work, so that it's clearer what my intentions are with this PR.

The `WindowHandle` is a widget area that allows you to move the window by dragging with your mouse. It's also a single-child container, so you are expected to put the content of the titlebar in there (title label, buttons, menus, etc.) . It's important to know that it's not supposed to be a titlebar by itself, you are still supposed to implement your own titlebar widget that wraps the `WindowHandle`. This way this widget is completely flexible on the titlebar design you want to implement.

masonry/src/widget/window_handle.rs
melix99: See my reply below. This is supposed to be just a container, so users of this widget should wrap it in their titlebar custom widget. So, I think that just filling the constrain is the right thing to do.

masonry/src/widget/window_handle.rs
melix99: I think it's right, since this is a container. The actual titlebar should have the `Role::TitleBar`.

masonry/src/widget/window_handle.rs
melix99: Done, I hope it's clear enough.

melix99: Just discovered that floem has something similar, if that can help: https://github.com/lapce/floem/blob/main/src/views/drag_window_area.rs

DJMcNab:
I still feel that `WindowHandle` as in this PR is poorly motivated; it seems hard to use correctly (given how anemic our other layout control is, and how little thought has been put into event bubbling; see [#masonry > Missing event `set_handled` calls](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Missing.20event.20.60set_handled.60.20calls)). Additionally, it's not used at all in this repository; I'd want to at least see an example.

My suggested next step here is to make this PR *only* the changes required to allow creating this `WindowHandle` externally (i.e. the changes except for the new widget), which we can easily land. I'd then like to see `WindowHandle` added in a more complete form.

DJMcNab (approved):
Apologies, this slipped out of my queue. This looks great, thanks.

Luckily, no merge conflicts seem to have arisen in the intervening period. Let's hope that's also true for CI :)

PR #672, opened 2024-10-15
docs: Fix some missing backticks.

waywardmonkeys:


xilem_core/src/sequence.rs
```diff
@@ -33,7 +33,7 @@ impl<T> AppendVec<T> {
     pub fn push(&mut self, item: T) {
         self.inner.push(item);
     }
-    /// [Drain](Vec::drain) all items from this `AppendVec`.
+    /// [`Drain`](Vec::drain) all items from this `AppendVec`.
```
DJMcNab, line 36: This was intentionally non-graved - it's using the verb "to drain" as used in the function `drain`, rather than a reference to the function `drain` itself.

xilem_core/src/view.rs
```diff
@@ -113,9 +113,9 @@ pub trait View<State, Action, Context: ViewPathTracker, Message = DynMessage>:
 }
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-/// An identifier used to differentiation between the direct children of a [View].
+/// An identifier used to differentiation between the direct children of a [`View`].
```
DJMcNab, line 116: IIRC, this one was also intentional, but is more marginal. I think with the capitalisation as-is, it's correct to change it.

xilem_core/src/view.rs
```diff
@@ -113,9 +113,9 @@ pub trait View<State, Action, Context: ViewPathTracker, Message = DynMessage>:
...
 ///
-/// These are [u64] backed identifiers, which will be added to the "view path" in
+/// These are [`u64`] backed identifiers, which will be added to the "view path" in
```
DJMcNab, line 118: :+1: looks like std does use things this way

DJMcNab (approved):
I disagree with one of these changes, but I'm not going to block on it

xilem_core/src/sequence.rs
waywardmonkeys: Fixed.

xilem_core/src/view.rs
waywardmonkeys: Nothing to do here!

xilem_core/src/view.rs
waywardmonkeys: As long as it is capitalized here, I think it should have backticks, so I've not updated my PR.

waywardmonkeys:
This PR happened because I noticed the `Scene` reference in the `render_text` docs and then did a `git grep` to find other similar things.

PR #674, opened 2024-10-16
Reorder update passes

PoignardAzur:
Reorder update passes so their code is in the same order as they're run.
Split out anim pass to a separate file (since it's not run with the other passes).
This is strictly cut-and-paste, no code has been changed.


jaredoconnell (approved):
I didn't give the code the most detailed look over since it's all strictly moving code. I can confirm that the animation code still works in the examples.

PR #675, opened 2024-10-16
Rename pass functions from root_xxx to run_xxx_pass

PoignardAzur:
This means all the pass functions now have the same naming convention, which makes the code more grep-able.


Philipp-M (approved):

waywardmonkeys (commented):
But it does mean they now differ from the `root_on_*_event` functions....

PoignardAzur:
> But it does mean they now differ from the root_on_*_event functions....

Yeah but they're almost empty by now, we'll probably remove them.

PR #677, opened 2024-10-16
Rename lifecycle to update

PoignardAzur:


PoignardAzur:
I recommend reading the first commit (before cargo fmt) for a more readable diff.

docs/assets/xilem-architecture.svg
Philipp-M, line 1: There's `update()` twice now, in the "Widget" box.

Also this file seems to be quite out of date, I wonder how difficult it would be to bring this up-to-date, to be more accurate (but this is stuff for a follow-up PR I think).

Philipp-M (approved):
Apart from the now reverted change of the svg, this looks reasonable to me.

jaredoconnell (approved):
Other than some issues with the svg, it looks good.

PR #678, opened 2024-10-16
Create on_anim method

PoignardAzur:
This marks animation as something that's somewhere between an event (it's triggered externally, it can't be requested synchronously as part of rewrite passes) and an update (it takes the same context type, and isn't tied to a user interaction).

### DIFF
--- a/masonry/src/testing/helper_widgets.rs
+++ b/masonry/src/testing/helper_widgets.rs
@@ -30,6 +30,7 @@ use crate::*;
 pub type PointerEventFn<S> = dyn FnMut(&mut S, &mut EventCtx, &PointerEvent);
 pub type TextEventFn<S> = dyn FnMut(&mut S, &mut EventCtx, &TextEvent);
 pub type AccessEventFn<S> = dyn FnMut(&mut S, &mut EventCtx, &AccessEvent);
+pub type AnimFrameFn<S> = dyn FnMut(&mut S, &mut UpdateCtx, u64);
 pub type RegisterChildrenFn<S> = dyn FnMut(&mut S, &mut RegisterCtx);
 pub type StatusChangeFn<S> = dyn FnMut(&mut S, &mut UpdateCtx, &StatusChange);
 pub type UpdateFn<S> = dyn FnMut(&mut S, &mut UpdateCtx, &Update);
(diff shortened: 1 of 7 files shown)


jaredoconnell (approved):
I think this change makes sense. I'm not sure what the other contributors think.
Tested and it works as intended.

masonry/src/testing/helper_widgets.rs
```diff
@@ -105,9 +107,10 @@ pub enum Record {
     PE(PointerEvent),
     TE(TextEvent),
     AE(AccessEvent),
-    RegisterChildren,
+    AF(u64),
+    RC,
```
DJMcNab, line 111: Why are there no doc comments here? I can only work out what this refers to *because I have this diff open*

I'm presuming the added terseness is to decrease the size of the created snapshots. But it looks to me like there's only one snapshot that uses this event, looking at the diff?

masonry/src/widget/widget.rs
```diff
@@ -79,6 +79,18 @@ pub trait Widget: AsAny {
     /// Handle an event from the platform's accessibility API.
     fn on_access_event(&mut self, ctx: &mut EventCtx, event: &AccessEvent) {}
 
+    /// Called at the beginning of a new animation frame.
+    ///
+    /// On the first frame when transitioning from idle to animating, `interval`
+    /// will be 0. (This logic is presently per-window but might change to
+    /// per-widget to make it more consistent). Otherwise it is in nanoseconds.
+    ///
+    /// The `paint` method will be called shortly after this event is finished.
```
DJMcNab, line 88: Isn't this only necessarily true if you request a paint?

DJMcNab (commented):
Making this its own method is good, but there's some weirdness in this PR

masonry/src/testing/helper_widgets.rs
PoignardAzur: > But it looks to me like there's only one snapshot that uses this event, looking at the diff?

A fair number of tests have been commented out through refactors. I'd like to progressively re-add them. I *think* some of them will use this event?

> Why are there no doc comments here?

Yeah, there probably should be.

I'm not sure the terse names are really needed either. The snapshots all use the "one-line per nesting level" format anyway.

masonry/src/widget/widget.rs
PoignardAzur: True.

PR #681, opened 2024-10-16
xilem: Change `View::rebuild` to not return the element, and cleanup imports and `Mut`

Philipp-M:
It's possible to avoid returning the `ViewElement::Mut` in `View::rebuild`, while having access to the element *after* a child view was rebuilt.

This PR removes that again making the signature a little bit cleaner, and also cleans up a lot of other stuff:
* `Mut<'_, Self::Element>` can be shortened to `Mut<Self::Element>`
* `use xilem::core::*` instead of `use xilem_core::*`, this is especially useful in the examples, to avoid newcomers giving the impression, that an additional dependency xilem_core is needed.
* Various cosmetic stuff, like one typo fix, empty line between doc-comment (as (nightly) clippy currently bleats for me)

For more context see https://xi.zulipchat.com/#narrow/stream/354396-xilem/topic/.E2.9C.94.20.60.26mut.20ViewElement.3A.3AMut.3C'_.3E.60.20in.20.60View.3A.3Arebuild.60.3F

### DIFF
--- a/xilem/src/view/worker.rs
+++ b/xilem/src/view/worker.rs
@@ -1,24 +1,26 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::{
+    core::{
+        DynMessage, Message, MessageProxy, MessageResult, Mut, NoElement, View, ViewId, ViewMarker,
+        ViewPathTracker,
+    },
+    ViewCtx,
+};
 use std::{future::Future, marker::PhantomData, sync::Arc};
-
 use tokio::{
     sync::mpsc::{UnboundedReceiver, UnboundedSender},
     task::JoinHandle,
 };
-use xilem_core::{
-    DynMessage, Message, MessageProxy, NoElement, View, ViewId, ViewMarker, ViewPathTracker,
-};
-
-use crate::ViewCtx;
 
 /// Launch a task which will run until the view is no longer in the tree.
+///
 /// `init_future` is given a [`MessageProxy`], which it will store in the future it returns.
 /// This `MessageProxy` can be used to send a message to `on_event`, which can then update
 /// the app's state.
 ///
-/// For exampe, this can be used with the time functions in [`crate::tokio::time`].
+/// For example, this can be used with the time functions in [`crate::tokio::time`].
 ///
 /// Note that this task will not be updated if the view is rebuilt, so `init_future`
 /// cannot capture.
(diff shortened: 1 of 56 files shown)


xilem/src/view/worker.rs
```diff
@@ -1,24 +1,26 @@
...
-use crate::ViewCtx;
 
 /// Launch a task which will run until the view is no longer in the tree.
+///
 /// `init_future` is given a [`MessageProxy`], which it will store in the future it returns.
 /// This `MessageProxy` can be used to send a message to `on_event`, which can then update
 /// the app's state.
 ///
-/// For exampe, this can be used with the time functions in [`crate::tokio::time`].
+/// For example, this can be used with the time functions in [`crate::tokio::time`].
```
Philipp-M, line 23: I wonder why our CI didn't caught that before...

xilem_core/src/view.rs
```diff
@@ -78,13 +78,13 @@ pub trait View<State, Action, Context: ViewPathTracker, Message = DynMessage>:
     /// This returns `element`, to allow parent views to modify the element after this `rebuild` has
     /// completed. This returning is needed as some reference types do not allow reborrowing,
     /// without unwieldy boilerplate.
```
PoignardAzur, line 80: Should probably remove this comment.

PoignardAzur (approved):
LGTM.

I'm surprised to see there's almost no new code. In particular, there's no new reborrows on the xilem side. Sounds like we were already doing most of the work and not taking advantage of it.

Philipp-M:
> In particular, there's no new reborrows on the xilem side. Sounds like we were already doing most of the work and not taking advantage of it.

The changes to make this possible are mostly on the xilem_web side, as there's currently (AFAIK) no access on the element after it has been modified by the most inner composed view in xilem (native) (i.e. the view that represents the widget).
But it was indeed a trivial change to also implement the `reborrow_mut` method in xilem_web. 

But since we already have `reborrow_mut` on xilem (native), it could be done there too.
Although as noticed in the Board PR, it's tricky to not produce unwanted behavior (e.g. with `Memoize`), without further changes on `xilem::Pod` and its `ViewElement::Mut` (which is currently `WidgetMut`).


I'm going to merge this, to avoid possible merge-conflicts, as it touches almost every file in all xilem* crates and I think it's otherwise uncontroversial (and I tested all the examples in xilem_web where the more interesting changes were).

DJMcNab (commented):
The choice to change all of the `Mut<'_, ...>` to `Mut<...>` seems a strange one, given that `elided_lifetimes_in_paths` is one of our canonical lints.

Philipp-M:
> given that `elided_lifetimes_in_paths` is one of our canonical lints.

I guess I've missed that. Since it looks syntactically cleaner, I prefer omitting them (no strong opinion though), but I see the argument for not doing this...

PR #683, opened 2024-10-17
Remove potential interactivity from `Label`

DJMcNab:
The whole reason the Label widget exists is as a non-interactive text holder for higher-level widgets. Therefore, it isn't meaningful for it to have interactivity.

More concretely, there is no interactivity currently implemented

### DIFF
--- a/masonry/src/widget/button.rs
+++ b/masonry/src/widget/button.rs
@@ -58,7 +58,7 @@ impl Button {
     /// ```
     pub fn from_label(label: Label) -> Button {
         Button {
-            label: WidgetPod::new(label.with_pointer_interaction(false)),
+            label: WidgetPod::new(label),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


xorgy (approved):

xorgy:
And maybe there's a place for a version with interaction as a "value" or some such.

DJMcNab:
I'm not sure that's meaningful. What interactivity would there be? I guess link handling? In my mind:

If you've got enough text that it contains multiple links, that should be a text which users can select from
If you've got just one link, that is just a button.

waywardmonkeys:
> I'm not sure that's meaningful. What interactivity would there be? I guess link handling? In my mind:
> 
> If you've got enough text that it contains multiple links, that should be a text which users can select from If you've got just one link, that is just a button.

I feel like that's way too much policy to be dictating in a foundational library / framework.

DJMcNab:
Yeah, that's fair. I can see that widget being useful for some users. Something like `DisplayText` or something.

What I don't want us to end up supporting is a button with links in its inner text, because that's a recipe for jank.

xorgy:
> I'm not sure that's meaningful. What interactivity would there be? I guess link handling? In my mind:

It's a common thing, where you have a list or table of selectable/copyable fields, like the connection info/details view in a wifi menu or whatever.

PR #684, opened 2024-10-17
Only send ImeMoved signal for focused text input

PoignardAzur:


PoignardAzur:
I still get a weird amount of lag in http_cats, so I suspect other footguns are being fired in the compose pass.

DJMcNab:
It looks to me like the source of this lag is that we're not coalescing multiple mouse movement inputs. So every single mouse movement runs a whole new compose pass; and we're also handling these *before* a `request_render`. I don't know of an easy solution without frame pacing here.

I think that this is an improvement, though

DJMcNab (approved):
This is definitely much more sensible

PR #686, opened 2024-10-18
Remove the dead `links` field from TextLayout

DJMcNab:
cc for awareness @mwcampbell.

That field was dead code left behind when we removed previous links support. On that struct is probably not the right field for links to live, since `Label` also uses `TextLayout`, but can't meaningfully support links as part of a different interactive widget.


PoignardAzur (approved):
LGTM.

It does bring a question: how do we expect to handle links in the future? Maybe we just pass a list of byte spans and get a list of rects in return?

PR #687, opened 2024-10-18
Don't use full tracing spans in full-tree passes by default

DJMcNab:
This can be toggled using the `MASONRY_TRACE_PASSES` environment variable.

This is as-discussed in [#xilem > Scrolling is insanely laggy](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Scrolling.20is.20insanely.20laggy). There are a few overlapping issues, but this *significantly* improves performance in debug mode.

The main pass which was problematic was the compose pass, however this also had a significant impact on the time taken by the accessibility/paint passes.

This only applies to the passes which traverse the entire tree, so not the targeted passes. I also chose to exclude update_disabled and update_stashed, as those will not necessarily happen to all widgets.

This also significantly reduces the size of the created log files - see [#masonry > Heavy amounts of logs with large app](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Heavy.20amounts.20of.20logs.20with.20large.20app). In most cases, if you're using the log file, you will be in development, which means that you can hopefully recreate the issue with the logging for the passes you need enabled.

### DIFF
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -979,11 +979,7 @@ impl LayoutCtx<'_> {
         // We intentionally always log this because clip paths are:
         // 1) Relatively rare in the tree
         // 2) An easy potential source of items not being visible when expected
-        trace!(
-            "{} ({}): set_clip_path {path:?}",
-            self.widget_state.widget_name,
-            self.widget_state.id
-        );
+        trace!("set_clip_path {path:?}");
         self.widget_state.clip = Some(path);
         // TODO - Updating the clip path may have
         // other knock-on effects we'd need to document.
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -982,6 +982,9 @@ impl LayoutCtx<'_> {
     /// - It serves as a mask for painting operations of the widget's children (*not* the widget itself).
     /// - Pointer events must be inside that path to reach the widget's children.
     pub fn set_clip_path(&mut self, path: Rect) {
+        // This is intentionally not-gated because clip paths are:
+        // 1) Relatively rare in the tree
+        // 2) An easy potential source of items not being visible when expected
         trace!("set_clip_path {:?}", path);
         self.widget_state.clip = Some(path);
         // TODO - Updating the clip path may have
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -858,9 +858,6 @@ impl LayoutCtx<'_> {
     /// [`WidgetPod::paint_insets`]: crate::widget::WidgetPod::paint_insets
     pub fn set_paint_insets(&mut self, insets: impl Into<Insets>) {
         let insets = insets.into();
-        if self.global_state.trace.layout {
-            trace!("set_paint_insets {:?}", insets);
-        }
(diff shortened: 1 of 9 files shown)


masonry/src/contexts.rs
```diff
@@ -858,7 +858,9 @@ impl LayoutCtx<'_> {
     /// [`WidgetPod::paint_insets`]: crate::widget::WidgetPod::paint_insets
     pub fn set_paint_insets(&mut self, insets: impl Into<Insets>) {
         let insets = insets.into();
-        trace!("set_paint_insets {:?}", insets);
+        if self.global_state.trace.layout {
+            trace!("set_paint_insets {:?}", insets);
+        }
```
PoignardAzur, line 863: I think it's okay to outright remove those logs rather than gating them. They're redundant with the log in the layout pass the gives the final layout.

masonry/src/contexts.rs
```diff
@@ -900,7 +902,9 @@ impl LayoutCtx<'_> {
     /// The provided value should be the distance from the *bottom* of the
     /// widget to the baseline.
     pub fn set_baseline_offset(&mut self, baseline: f64) {
-        trace!("set_baseline_offset {}", baseline);
+        if self.global_state.trace.layout {
+            trace!("set_baseline_offset {}", baseline);
+        }
```
PoignardAzur, line 907: Same.

masonry/src/passes/accessibility.rs
```diff
@@ -22,19 +22,24 @@ fn build_accessibility_tree(
     rebuild_all: bool,
     scale_factor: f64,
 ) {
-    let _span = widget.item.make_trace_span().entered();
+    let _span = global_state
+        .trace
+        .access
+        .then(|| widget.item.make_trace_span().entered());
```
PoignardAzur, line 28: I'm surprised just creating the spans has such a heavy performance impact. Are you sure we're not conflating them with something else?

PoignardAzur (approved):

masonry/src/passes/accessibility.rs
DJMcNab: I believe that this is correct. Creating a filtered out span has very little performance impact, obviously, but the log file layer has no filtering.

A large amount of the work in using a span is done on creation - see https://github.com/tokio-rs/tracing/blob/bdbaf8007364ed2a766cca851d63de31b7c47e72/tracing-subscriber/src/registry/sharded.rs#L237-L269.

An alternative approach we could take would be to cache the span for each widget in its state. That approach would be likely to get most of the same advantages, but requires more fundamental changes. That is also done in https://github.com/bevyengine/bevy/pull/9390. I think I'll do that as a follow-up to this PR; I still think this is useful, because the current logs are extraordinarily spammy; but if we can maintain a proper tracing stack at all times, that would be neat.

PoignardAzur: Yeah, I'm in favor of removing spammy logs, but removing spans (and/or making span code more verbose) bothers me a bit.

If we do caching, we could simply create the cached span during the WidgetAdded event. To avoid using an optional span we could create WidgetState with a default placeholder span.

PR #688, opened 2024-10-18
Refactor tracing handling for events

DJMcNab:
Inspired by #687, I thought some more tidying up of the log file was in order.

The main differences are:
1) Higher-scale spans exist for key operations
2) Something is always logged for each event, but less for high-density eventskey 
3) Less is logged for high-density events
4) Key repeats are no longer treated as high density
5) We no longer do extra work if the hover and focus paths haven't changed (which also means less logging)

This PR *does not* depend on #687

### DIFF
--- a/masonry/src/widget/widget.rs
+++ b/masonry/src/widget/widget.rs
@@ -196,6 +196,7 @@ pub trait Widget: AsAny {
     /// As methods recurse through the widget tree, trace spans are added for each child
     /// widget visited, and popped when control flow goes back to the parent. This method
     /// returns a static span (that you can use to filter traces and logs).
+    // TODO: Make include the widget's id?
     fn make_trace_span(&self) -> Span {
         trace_span!("Widget", r#type = self.short_type_name())
     }
--- a/masonry/src/event.rs
+++ b/masonry/src/event.rs
@@ -384,6 +384,7 @@ impl PointerEvent {
 impl TextEvent {
     pub fn short_name(&self) -> &'static str {
         match self {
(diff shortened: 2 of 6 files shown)


masonry/src/widget/widget.rs
```diff
@@ -199,6 +199,7 @@ pub trait Widget: AsAny {
     /// As methods recurse through the widget tree, trace spans are added for each child
     /// widget visited, and popped when control flow goes back to the parent. This method
     /// returns a static span (that you can use to filter traces and logs).
+    // TODO: Make include the widget's id?
```
Philipp-M, line 202: Agree, could probably be done in this PR.

Philipp-M (approved):
It's really interesting, that tracing seems to have such performance impact...

Otherwise looks good.

masonry/src/widget/widget.rs
DJMcNab: That's an extremely significant change, because the widget's id isn't already in scope. I don't want to do this in this PR.

DJMcNab:
Re-requested review as I've not actioned the comment for making the trace spans include the widget's id.

masonry/src/widget/widget.rs
Philipp-M: Oh, right, we're in the widget trait, I guess I missed that. Probably needs the id as function parameter or something like that. Sure let's keep that for a follow-up PR

Philipp-M (approved):
Yeah that comment wasn't too important, I missed that this requires more significant changes (which obviously should be done in a follow-up PR).

PR #637, opened 2024-10-07
Reorganize module exports

PoignardAzur:
Stop publicly exporting public modules with no public symbols, reorder some code.

PR #690, opened 2024-10-20
Reorder imports across project

PoignardAzur:
Split off from #637.


PoignardAzur:
This should ideally get merged soon, to avoid merge conflicts with new PRs. All changes are strictly changes in the order of imports, almost all them the result of running `cargo +nightly fmt`.

xilem/src/view/label.rs
```diff
@@ -1,11 +1,11 @@
...
-use crate::{
-    core::{DynMessage, Mut, ViewMarker},
-    Color, MessageResult, Pod, TextAlignment, TextWeight, View, ViewCtx, ViewId,
-};
-use masonry::{text::TextBrush, widget, ArcStr};
+use masonry::text::TextBrush;
+use masonry::{widget, ArcStr};
+
+use crate::core::{DynMessage, Mut, ViewMarker};
+use crate::{Color, MessageResult, Pod, TextAlignment, TextWeight, View, ViewCtx, ViewId};
```
Philipp-M, line 8: I just did the opposite in #681, I think we should be consistent in all crates and avoid this back-and-forth, is there some kind of lint for this to enforce a unique style? (That said, I have no strong opinion of either way)

PoignardAzur: There's `rustfmt.toml` which is our implicit formatting guide.

It includes this:

```toml
# Subjectively, this combo improves readability on imports. The full prefix is always
# visible right after `use`, and different types of imports are grouped together in
# a predictable way (first std, then other crates, then local imports).
# (These options are still unstable and thus commented out)
# imports_granularity = "Module"
# group_imports = "StdExternalCrate"
```

So basically what I did was uncomment those last two lines and run `cargo +nightly fmt`.

> I just did the opposite in https://github.com/linebender/xilem/pull/681

Well you didn't *quite* do the opposite, you just created a `core` re-exports in xilem and folded it in. This PR spreads the import out, but `core` is still re-exported.

Philipp-M: Ok, nice, good to know, I missed that.

Philipp-M (approved):
LGTM.

PR #692, opened 2024-10-20
Make passes that affect accesskit nodes request an accessibility pass

PoignardAzur:


Philipp-M (approved):
LGTM.

DJMcNab (commented):
Would it be worth having a mirror comment in the accessibility pass, so that changes there can be recognised?

PoignardAzur:
Sure. Something along the line of "If a pass changes these attributes it should request accessibility"?

DJMcNab:
Yeah, that sounds good.

PR #694, opened 2024-10-20
Refactor away some redundant methods in RenderRoot

PoignardAzur:


jaredoconnell (approved):
Tested. It works. I don't see anything wrong with it.

PR #696, opened 2024-10-20
Implement fixpoint loop in run_rewrite_passes

PoignardAzur:
This implements the last feature from the Pass Specification RFC :tada:

### DIFF
--- a/masonry/src/passes/compose.rs
+++ b/masonry/src/passes/compose.rs
@@ -75,6 +75,10 @@ fn compose_widget(
 pub(crate) fn run_compose_pass(root: &mut RenderRoot) {
     let _span = info_span!("compose").entered();
 
+    if root.root_state().needs_compose {
+        root.global_state.pointer_changed = true;
+    }
+
     let (root_widget, root_state) = root.widget_arena.get_pair_mut(root.root.id());
     compose_widget(
         &mut root.global_state,
--- a/masonry/src/passes/compose.rs
+++ b/masonry/src/passes/compose.rs
@@ -75,8 +75,10 @@ fn compose_widget(
 pub(crate) fn run_compose_pass(root: &mut RenderRoot) {
     let _span = info_span!("compose").entered();
 
+    // If widgets are moved, pointer-related info may be stale.
+    // For instance, the "hovered" widget may have moved and no longer be under the pointer.
     if root.root_state().needs_compose {
-        root.global_state.pointer_changed = true;
+        root.global_state.needs_pointer_pass = true;
     }
 
     let (root_widget, root_state) = root.widget_arena.get_pair_mut(root.root.id());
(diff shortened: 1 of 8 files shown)


masonry/src/passes/compose.rs
```diff
@@ -75,6 +75,10 @@ fn compose_widget(
 pub(crate) fn run_compose_pass(root: &mut RenderRoot) {
     let _span = info_span!("compose").entered();
 
+    if root.root_state().needs_compose {
+        root.global_state.pointer_changed = true;
```
jaredoconnell, line 79: A comment may be useful here. Is this because the composition changed, so the relative location of the pointer may have changed?

masonry/src/render_root.rs
```diff
@@ -74,6 +74,7 @@ pub(crate) struct RenderRootState {
     pub(crate) mutate_callbacks: Vec<MutateCallback>,
     pub(crate) is_ime_active: bool,
     pub(crate) scenes: HashMap<WidgetId, Scene>,
+    pub(crate) pointer_changed: bool,
```
jaredoconnell, line 77: It may be helpful document this field.

jaredoconnell (approved):
Everything works as intended. This definitely looks to theoretically significantly improve efficiency in some scenarios. I just have some minor comments.

masonry/src/render_root.rs
Philipp-M, line 77: Hmm, I'm not sure if that name is slightly misleading. `needs_pointer` doesn't sound too great though, and `needs_pointer_pass` is also not 100% optimal as it is not consistent to the other `needs_` flags, but it's probably better in describing what this is for.

masonry/src/render_root.rs
```diff
@@ -492,10 +509,7 @@ impl RenderRoot {
         // tree needs to be rebuilt. Usually both happen at the same time.
         // A redraw will trigger a rebuild of the accessibility tree.
         // TODO - We assume that a relayout will trigger a repaint
-        if self.root_state().needs_paint
-            || self.root_state().needs_accessibility
-            || self.root_state().needs_layout
-        {
+        if self.root_state().needs_paint || self.root_state().needs_accessibility {
```
Philipp-M, line 512: I guess `needs_layout` is not necessary because `needs_paint` (and `needs_accessibility`) is implicitly set in `run_layout_on` (and it's planned to detect whether a layout pass really needs a rerender)?

Philipp-M (commented):
In general looks good, just some minor naming stuff.

masonry/src/render_root.rs
PoignardAzur: `needs_layout` is not necessary anymore because it's one of the rewrite passes. So if `needs_layout` is true, the check above will have triggered and requested a new frame.

masonry/src/render_root.rs
Philipp-M: Maybe `needs_pointer_update`?

PR #698, opened 2024-10-20
Update test-related documentation

PoignardAzur:


masonry/src/testing/helper_widgets.rs
```diff
@@ -155,7 +162,13 @@ impl<S> ModularWidget<S> {
             children: None,
         }
     }
+}
 
+/// Builder methods.
+///
+/// Each method takes a flag which is then returned by the matching Widget method.
+#[allow(missing_docs)]
```
Philipp-M, line 170: I wonder if we should add the reasons (as comment) to all these `#[allow]` , to avoid creeping up more of the unpleasant work of fixing `allow_attributes_without_reason` later.

PoignardAzur: Well, in any case they'll be replaced with `#[expect]` annotations, which don't need reasons.

But I guess in that case it's worth documenting.

Philipp-M (approved):
Nice, more (or improved) documentation is always good. LGTM.

PR #702, opened 2024-10-21
Give actionable suggestions for implementing `on_anim_frame`

DJMcNab:
Users could easily run into difficulties where animation does not work as expected, with no obvious hints for how to debug this. Documenting the correct usage of `on_anim_frame` should help them.


Philipp-M (approved):
Nice, more docs :)

PR #708, opened 2024-10-21
Rename RenderRoot to WindowRoot

PoignardAzur:


DJMcNab (commented):
I'm not a huge fan of this change. I don't think that `WindowRoot` is that much clearer than the existing `RenderRoot`; this is going to need to change significantly for multi-window, but this rename more strongly mostly just moves multi-window to be out-of-scope.

(I'm not going to block it though)

PoignardAzur:
I think the implication that multi-window is out of scope is correct.

The code we've written so far in RenderRoot and in our passes works a lot better if you assume there's a single window, and multi-window is handled at a level "above" RenderRoot.

Code handling text focus, pointer icons, the focus chain, IME area, etc, are all written under the assumption that there's a single window. If we later write multi-window code under the assumption "one WindowRoot equals one Window", that code can stay unchanged. If we write multi-window-in-one-RenderRoot, that's a lot of code we'll need to refactor.

PoignardAzur:
@Philipp-M @jaredoconnell Thoughts about renaming to WindowRoot? Opinions for or against?

Philipp-M:
What about just `Root` or `WidgetRoot`? I see it similar as Daniel, but your reasoning makes sense. I think we want to share the `WidgetArena` across windows? So that would require a bigger refactor then anyways (resulting in new types, probably making naming more straight-forward)?

jaredoconnell:
I agree. WidgetRoot, if it doesn't conflict with another type, makes the most sense to me.

PoignardAzur:
> What about just `Root` or `WidgetRoot`?

I'm not super comfortable with those when we already have a RootWidget. Event if we remove or rename it, I think it would be inherently confusing for people that WidgetRoot does *not* implement the Widget trait.

PoignardAzur:
Closing this for now. I don't want to move forward without consensus, but also we have much more pressing priorities than debating this.

PR #710, opened 2024-10-21
Remove `WidgetState::Cursor`

PoignardAzur:


masonry/src/widget/widget.rs
```diff
@@ -220,9 +220,11 @@ pub trait Widget: AsAny {
...
     }
 
-    // TODO - Document
-    // TODO - Add &UpdateCtx argument
-    fn get_cursor(&self) -> CursorIcon {
+    /// Return the cursor icon for this widget.
+    ///
+    /// **pos** - the position in global coordinates (e.g. `(0,0)` is the top-left corner of the
+    /// window).
+    fn get_cursor(&self, ctx: &QueryCtx, pos: Point) -> CursorIcon {
```
DJMcNab, line 227: The combination of the `cursor_icon_changed` mechanism and this method also taking a position is quite weird.

If you know the cursor has changed, you know that it has entered a specific region.

I can't think of a cleaner solution at the moment, though

masonry/src/widget/widget.rs
```diff
@@ -220,9 +220,11 @@ pub trait Widget: AsAny {
         None
     }
 
-    // TODO - Document
-    // TODO - Add &UpdateCtx argument
-    fn get_cursor(&self) -> CursorIcon {
+    /// Return the cursor icon for this widget.
+    ///
+    /// **pos** - the position in global coordinates (e.g. `(0,0)` is the top-left corner of the
```
DJMcNab, line 225: the position of what? I presume the mouse?

masonry/src/widget/widget.rs
```diff
@@ -220,9 +220,11 @@ pub trait Widget: AsAny {
         None
     }
 
-    // TODO - Document
-    // TODO - Add &UpdateCtx argument
-    fn get_cursor(&self) -> CursorIcon {
+    /// Return the cursor icon for this widget.
+    ///
```
DJMcNab, line 224: Should this mention `cursor_icon_changed` here?

DJMcNab (approved):
Some docs improvements are needed here.

masonry/src/widget/widget.rs
PoignardAzur: `cursor_icon_changed` is more for situations where the cursor doesn't move, but the icon does.

I guess I should document that.

masonry/src/widget/widget.rs
```diff
@@ -220,9 +220,11 @@ pub trait Widget: AsAny {
         None
     }
 
-    // TODO - Document
-    // TODO - Add &UpdateCtx argument
-    fn get_cursor(&self) -> CursorIcon {
+    /// Return the cursor icon for this widget.
```
DJMcNab, line 223: We should document when this will be called here as well, i.e. something like `when the mouse cursor moves. It may also be called due to cursor_icon_changed, for cases when the state under the cursor might have changed (consider for example a link moving under the cursor due to a keypress in a rich text box)`

DJMcNab (approved):
Some slightly expanded documentation to make the semantics crystal clear would be good, otherwise this is looking good. Thanks!

PR #711, opened 2024-10-22
Add unit tests to passes

PoignardAzur:


masonry/src/tests/accessibility.rs
```diff
@@ -0,0 +1,33 @@
...
+    harness.edit_widget_with_tag(target_tag, |mut widget| {
+        widget.ctx.request_accessibility_update();
+    });
+    let _ = harness.render();
+
+    let records = harness.get_records_of(target_tag);
+    assert!(records.iter().any(|r| matches!(r, Record::Accessibility)));
+
+    let records = harness.get_records_of(parent_tag);
+    assert!(records.iter().all(|r| !matches!(r, Record::Accessibility)));
```
DJMcNab, line 35: Do we also need to test that if we do another render, the "target" also doesn't get accessibility ran on it? Or do we assume that whatever snapshot tests we have will catch this?

masonry/src/tests/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! We test core Masonry features and passes here instead of in `masonry_testing`,
+//! both to centralize tests in a single crateand to have access to the `masonry`
```
DJMcNab:
```suggestion
//! both to centralize tests in a single crate, and to have access to the `masonry`
```

masonry_testing/src/harness.rs
```diff
@@ -709,7 +710,7 @@ impl<W: Widget> TestHarness<W> {
     ///
     /// # Panics
     ///
-    /// Panics if no Widget with this tag can be found.
+    /// Panics if no widget with this tag can be found.
```
DJMcNab, line 713: I did wonder

masonry_testing/src/lib.rs
```diff
@@ -23,10 +23,19 @@ use masonry_core::core::{Widget, WidgetId};
 ///
 /// Implements helper methods useful for unit testing.
 pub trait TestWidgetExt: Widget + Sized + 'static {
+    // TODO - Remove, keep only record()
```
DJMcNab: This TODO comment probably need an explanation of how tests which use the one form are migrated/why they haven't been already

DJMcNab (approved):
Seems reasonable enough. Some comments which need addressing.

masonry/src/tests/accessibility.rs
PoignardAzur: What do you mean?

DJMcNab: So this test doesn't rule out that the heuristic for "is accessibility recorded" is "is this a leaf node", for example.
It's not very important.

PR #712, opened 2024-10-22
Use PresentMode::Mailbox when available.

xorgy:
This provides much better latency by default on X11 and Wayland.


DJMcNab (commented):
This is listed as a `NoVsync` mode in wgpu's docs. We kind of do want vsync behaviour for our use case, because power efficiency is something we want to at least aim for.
I presume you've confirmed that this doesn't lead to more frames than are displayed being rendered?

xorgy:
> I presume you've confirmed that this doesn't lead to more frames than are displayed being rendered?

No, I meant this to show what latency is left on the table for interactions by not rendering immediately. By default Fifo and FifoRelaxed introduce a lot of latency even on systems with plenty of throughput. In practice, when we have animations in the mix, we have to be limiting somehow.

xorgy:
Probably there's a middle ground where we configure the FIFO. Floem always uses FIFO, but asks wgpu to try to limit the length to two frames.

PR #713, opened 2024-10-23
Tweak get_cursor code

PoignardAzur:


masonry/src/widget/widget.rs
```diff
@@ -224,6 +224,8 @@ pub trait Widget: AsAny {
 
     /// Return the cursor icon for this widget.
     ///
+    /// This will be called when the mouse moves or [`MutateCtx::cursor_icon_changed`] is called.
```
DJMcNab, line 227:
```suggestion
    /// This will be called when the mouse moves or [`cursor_icon_changed`](MutateCtx::cursor_icon_changed) is called.
```
Given that it is also on the other context types. It would be nice to find a clearer way to document this, because the status quo is quite unfortunate.

DJMcNab (approved):

PR #715, opened 2024-10-25
Fix kurbo::BezPath d attribute for xilem_web::svg::path element

RagibHasin:
Regressed in #699

cc @Philipp-M 🦆

### DIFF
--- a/xilem_web/src/svg/kurbo_shape.rs
+++ b/xilem_web/src/svg/kurbo_shape.rs
@@ -211,7 +211,7 @@ impl<State: 'static, Action: 'static> OrphanView<BezPath, State, Action, DynMess
         Attributes::rebuild(element, 1, |mut element| {
             let attrs: &mut Attributes = element.modifier();
             if attrs.was_created() {
-                attrs.push(("path", new.to_svg()));
+                attrs.push(("d", new.to_svg()));
             } else if new != prev {
                 attrs.mutate(|m| *m = ("d", new.to_svg()).into());
             } else {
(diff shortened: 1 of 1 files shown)


xorgy (approved):

xilem_web/src/svg/kurbo_shape.rs
```diff
@@ -213,7 +213,7 @@ impl<State: 'static, Action: 'static> OrphanView<BezPath, State, Action, DynMess
             if attrs.was_created() {
                 attrs.push(("path", new.to_svg()));
```
xorgy, line 214: Doesn't this also need changed?

xorgy (changes requested):

Philipp-M:
Whoops, not sure how that did happen...

Nice catch, I thought I tested this, but this is the only kurbo view that's not in the svgtoy example... Bad luck...

xilem_web/src/svg/kurbo_shape.rs
RagibHasin: Yeah, I think so. BTW, it was this way before #699 too.

Philipp-M (approved):
Thanks!

xilem_web/src/svg/kurbo_shape.rs
RagibHasin: Or maybe it hasn't been. I am too sleepy to think coherently.

DJMcNab:
Should we have an example which uses this?

PR #717, opened 2024-10-26
xilem_web: Add a few boolean attribute modifiers for `HtmlInputElement` and factor element state flags out of these modifiers into `ElementFlags`

### LINKED ISSUE #716: xilem_web svg regression: Missing stroke attribute
Yet another regression after #699 

```rust
Circle::new((0.0, 0.0), 5.0)
            .fill(Color::rgb8(0x99, 0x99, 0x99))
            .stroke(Color::BLACK, Stroke::new(0.5))
```

Before:
Screenshot_20241026_132228

After:
Screenshot_20241026_132513

Philipp-M:
This PR starts with supporting specialized elements (i.e. more than just the `Element` DOM interface), starting with boolean attributes in `HtmlInputElement`  (`checked`, `default_checked`, `disabled`, `required`, `multiple`).
With a general `OverwriteBool` modifier which is optimized by encoding boolean flags into a `u32`, thus not requiring allocations (compared to the other modifiers we currently have), but is limited to max 32 overwrites, which I don't think should be a real-world limitation (i.e. `input_el.checked().checked()...checked()` 32 times makes no sense), I also started adding tests for this, as it juggles around with bit-operations (and we should generally start writing more tests).
I have originally planned to feature-flag this (i.e. have a feature like `HtmlInputElement`).
But I'd like to see how far we can go without this, I haven't yet noticed significant (binary-size) regressions in the todomvc example (which uses `input` elements) that justifies the worse DX that additional features introduce.
Having features for this is also not optimal for another reason: It changes the API (e.g. `DomNode::Props` is `props::HtmlInputElement` instead of `props::Element`.

It also factors the element state flags (`was_created`, `in_hydration`) into a separate `ElementFlags` which is shared within a `Modifier<M>` struct (similar as `PodMut` or `WidgetMut` in masonry), so that we don't have to duplicate that state in every modifier. Additionally a new flag `needs_update` is introduced, which indicates that the element in general needs to update any modifier, and is entirely an optimization to avoid checking every modifier whether it has changed (not yet that important, but when we have a lot of modifiers per element, having to check every modifier is less efficient, it's also already *slightly* visible in the js-framework-benchmark).
For this, it unfortunately suffers similar as #705 from the free-form syntax by being a little bit more verbose (which may be reverted after `arbitrary_self_types` are stable).

It should also fix #716

### DIFF
--- a/xilem_web/src/interfaces.rs
+++ b/xilem_web/src/interfaces.rs
@@ -830,7 +830,7 @@ pub trait HtmlInputElement<State, Action = ()>:
     /// use xilem_web::{interfaces::{Element, HtmlInputElement}, elements::html::input};
     ///
     /// # fn component() -> impl HtmlInputElement<()> {
-    /// input(()).required(true) // results in <input required></input>
+    /// input(()).multiple(true) // results in <input multiple></input>
     /// # }
     /// ```
     fn multiple(self, required: bool) -> html_input_element::view::Multiple<Self, State, Action> {
--- a/xilem_web/src/interfaces.rs
+++ b/xilem_web/src/interfaces.rs
@@ -833,8 +833,8 @@ pub trait HtmlInputElement<State, Action = ()>:
     /// input(()).multiple(true) // results in <input multiple></input>
     /// # }
     /// ```
-    fn multiple(self, required: bool) -> html_input_element::view::Multiple<Self, State, Action> {
-        html_input_element::view::Multiple::new(self, required)
+    fn multiple(self, multiple: bool) -> html_input_element::view::Multiple<Self, State, Action> {
+        html_input_element::view::Multiple::new(self, multiple)
     }
 }
 
--- a/xilem_web/src/interfaces.rs
+++ b/xilem_web/src/interfaces.rs
@@ -766,6 +766,7 @@ pub trait HtmlInputElement<State, Action = ()>:
     Action,
     DomNode: DomNode<
         Props: With<html_input_element::Checked>
+                   + With<html_input_element::DefaultChecked>
                    + With<html_input_element::Disabled>
                    + With<html_input_element::Required>
                    + With<html_input_element::Multiple>,
@@ -857,6 +858,7 @@ where
     T: HtmlElement<State, Action>,
     T::DomNode: AsRef<web_sys::HtmlInputElement>,
     <T::DomNode as DomNode>::Props: With<html_input_element::Checked>
+        + With<html_input_element::DefaultChecked>
         + With<html_input_element::Disabled>
         + With<html_input_element::Required>
         + With<html_input_element::Multiple>,
--- a/xilem_web/src/interfaces.rs
+++ b/xilem_web/src/interfaces.rs
@@ -16,10 +16,8 @@ use std::borrow::Cow;
 
 use crate::{
     events,
-    modifiers::{
-        Attr, Attributes, Class, ClassIter, Classes, Rotate, Scale, ScaleValue, Style, StyleIter,
-        Styles, With,
-    },
+    modifiers::{Attr, Class, ClassIter, Rotate, Scale, ScaleValue, Style, StyleIter},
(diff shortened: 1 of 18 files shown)


xilem_web/src/interfaces.rs
```diff
@@ -758,17 +758,110 @@ where
...
 // #[cfg(feature = "HtmlInputElement")]
 impl<State, Action, T> HtmlInputElement<State, Action> for T
 where
     T: HtmlElement<State, Action>,
     T::DomNode: AsRef<web_sys::HtmlInputElement>,
+    <T::DomNode as DomNode>::Props: With<html_input_element::Checked>
+        + With<html_input_element::DefaultChecked>
+        + With<html_input_element::Disabled>
+        + With<html_input_element::Required>
+        + With<html_input_element::Multiple>,
```
Philipp-M, line 864: We could also sum these together into "alias traits", or in case of the "last" inheriting interface (i.e. no other trait having this as super trait), force this to:
```suggestion
    <T::DomNode as DomNode>::Props = props::HtmlInputElement
```

To avoid this being unnecesseraily verbose...

But this would have the disadvantage to not easily allow custom elements (i.e. with custom props) that inherit this interface, so alias traits are probably better.

Philipp-M: Well, while writing this, I've already decided, that the traits make more sense (as more generally applicable to all the interfaces). So I did this in the latest commit.

xilem_web/src/interfaces.rs
```diff
@@ -758,17 +750,100 @@ where
...
+    /// ```
+    /// use xilem_web::{interfaces::{Element, HtmlInputElement}, elements::html::input};
+    ///
+    /// # fn component() -> impl HtmlInputElement<()> {
+    /// input(()).required(true) // results in <input required></input>
+    /// # }
+    /// ```
+    fn multiple(self, required: bool) -> html_input_element::view::Multiple<Self, State, Action> {
+        html_input_element::view::Multiple::new(self, required)
+    }
```
RagibHasin, line 838: I think it is a copy-paste artifact.

Philipp-M: Thanks for catching this, I went over the rest of the doc-comments and have found a few inconsistencies as well.
I hope I have now found and fixed all...
I've also renamed `With` to `WithModifier` as it seems more appropriate, and it's likely summed up in these alias traits anyway.

flosse (approved):
I have only skimmed the code, but at least nothing obviously problematic caught my eye (as far as i can judge at all).

xilem_web/src/interfaces.rs
flosse: > I've also renamed `With` to `WithModifier` as it seems more appropriate

:+1: It looks like a small change, but from my point of view it is enormously helpful for readability! :+1:

RagibHasin (changes requested):

xilem_web/src/interfaces.rs
```diff
@@ -830,7 +830,7 @@ pub trait HtmlInputElement<State, Action = ()>:
     /// use xilem_web::{interfaces::{Element, HtmlInputElement}, elements::html::input};
     ///
     /// # fn component() -> impl HtmlInputElement<()> {
-    /// input(()).required(true) // results in <input required></input>
+    /// input(()).multiple(true) // results in <input multiple></input>
     /// # }
     /// ```
     fn multiple(self, required: bool) -> html_input_element::view::Multiple<Self, State, Action> {
```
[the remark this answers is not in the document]
Philipp-M: Whoops you're right, I guess I was too focused on the doc-comments, I've also removed the size-hints for boolean modifiers in the latest commit, as it's kind of useless boilerplate there. (32-limit boundary assert check happens on length expanding methods anyway).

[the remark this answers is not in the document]
RagibHasin: Thanks. It looks good to me now.

Philipp-M:
Thanks for the reviews :)

PR #718, opened 2024-10-29
masonry: Export `WindowSizePolicy`

waywardmonkeys:
This is a pub enum within `render_root` and is required for creating a `RenderRootOptions` which is already exported. This is needed for people embedding masonry in a non-Xilem way.

### DIFF
--- a/masonry/src/lib.rs
+++ b/masonry/src/lib.rs
@@ -156,7 +156,7 @@ pub use event::{
     AccessEvent, PointerButton, PointerEvent, PointerState, TextEvent, Update, WindowEvent,
     WindowTheme,
 };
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
This is why `unreachable_pub` is in https://linebender.org/wiki/canonical-lints/ - looks good to me.

jaredoconnell:
This is not the most well tested feature. And many of our layouts are not designed for infinite bounds. This should be improved with my upcoming measure function PR.

DJMcNab:
Possibly a better fix here would be for the calling location to just use `Default::default`. Nevertheless, I don't think it makes things meaningfully worse to export this for the interim.

PR #719, opened 2024-10-29
masonry example: Fix reference to `on_status_changed`

waywardmonkeys:
This is now just `update`.


waywardmonkeys:
There are other eferences to `on_status_changed`, but someone else can resolve them when fixing docs and tests.

DJMcNab (approved):

PR #720, opened 2024-10-29
masonry: Export `UnitPoint`

waywardmonkeys:
This is needed for `Align::horizontal`, `Align::vertical`, etc.


DJMcNab (approved):
Given that this type is effectively already public (making it `pub(crate)` gives `private_interfaces` errors), I think this is fine.

PR #721, opened 2024-10-29
xilem_web: Fix event views (when the underlying element was recreated) and factor flags and `Pod` into its own module

Philipp-M:
This renames the `ElementFlags` into `PodFlags` as these are general enough to keep them directly in the `Pod(Mut)`.
This reduces duplicated code (and thus a potential of making errors) in the props (that logic is now in `PodMut::drop`, `SuperElement::upcast` and `AnyElement::replace_inner`).

The `WithModifier` trait is adjusted and only implemented on the `View::Element` with a blanket impl of `AsMut<T>` instead (which should also reduce boilerplate), this was mostly a necessity to avoid `impl WithModifier` for either variants `Pod` and `PodMut` on every props (i.e. more boilerplate).

Further all kinds of event views were fixed, as they would break currently when the underlying element changes (e.g. via a variant change of `OneOf`). This was actually the original planned change, but I guess that escalated a little bit...

### DIFF
--- a/xilem_web/src/events.rs
+++ b/xilem_web/src/events.rs
@@ -2,8 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    core::{MessageResult, Mut, View, ViewElement, ViewId, ViewMarker, ViewPathTracker},
-    modifiers::WithModifier,
+    core::{MessageResult, Mut, View, ViewId, ViewMarker, ViewPathTracker},
     DomView, DynMessage, ElementAsRef, OptionalAction, ViewCtx,
 };
 use std::{borrow::Cow, marker::PhantomData};
@@ -135,8 +134,7 @@ fn build_event_listener<State, Action, V, Event>(
 where
     State: 'static,
     Action: 'static,
-    V: View<State, Action, ViewCtx, DynMessage>,
-    V::Element: ElementAsRef<web_sys::EventTarget>,
+    V: DomView<State, Action>,
     Event: JsCast + 'static + crate::Message,
 {
     // we use a placeholder id here, the id can never change, so we don't need to store it anywhere
@@ -167,9 +165,7 @@ fn rebuild_event_listener<State, Action, V, Event>(
 ) where
     State: 'static,
     Action: 'static,
-    V: DomView<State, Action, Element: AsRef<web_sys::EventTarget> + WithModifier<Events>>,
-    for<'a> <V::Element as ViewElement>::Mut<'a>:
-        AsRef<web_sys::EventTarget> + WithModifier<Events>,
+    V: DomView<State, Action>,
     Event: JsCast + 'static + crate::Message,
 {
     ctx.with_id(ON_EVENT_VIEW_ID, |ctx| {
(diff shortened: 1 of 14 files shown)


xilem_web/src/events.rs
```diff
@@ -163,18 +162,25 @@ fn rebuild_event_listener<State, Action, V, Event>(
...
+            prev_element_view,
+            &mut state.child_state,
+            ctx,
+            element.reborrow_mut(),
+        );
+        let was_created = element.flags.was_created();
+        if prev_capture != capture || prev_passive != passive || was_created {
+            if !was_created {
+                remove_event_listener(element.as_ref(), event, &state.callback, prev_capture);
+            }
```
flosse, line 179: :thinking: I don't know why, but somehow i get a knot in my head when i read it
```rust
let was_created = // ..
if /*... */ || was_created {
   if !was_created {}
}
```

xilem_web/src/pod.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub node: N,
+    pub flags: PodFlags,
+    pub props: N::Props,
+}
+
+/// Type-erased [`Pod`], it's used for example as intermediate representation for children of a DOM node
+pub type AnyPod = Pod<Box<dyn AnyNode>>;
+
+impl<N: DomNode> Pod<N> {
+    pub fn new(node: N, props: N::Props, flags: PodFlags) -> Self {
```
flosse, line 23:
```suggestion
    pub const fn new(node: N, props: N::Props, flags: PodFlags) -> Self {
```

xilem_web/src/pod.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub node: &'a mut N,
+    pub props: &'a mut N::Props,
+    pub flags: &'a mut PodFlags,
+    pub parent: Option<&'a web_sys::Node>,
+    pub was_removed: bool,
+    pub is_reborrow: bool,
+}
+
+impl<'a, N: DomNode> PodMut<'a, N> {
+    pub fn new(
```
flosse, line 110:
```suggestion
    pub const fn new(
```

xilem_web/src/pod.rs
```diff
@@ -0,0 +1,219 @@
...
+            false,
+        )
+    }
+}
+
+impl<N: DomNode> Drop for PodMut<'_, N> {
+    fn drop(&mut self) {
+        if !self.is_reborrow && !self.was_removed {
+            self.apply_changes();
+        }
```
flosse, line 162: maybe avoid `!` for better readability?
```suggestion
        if self.is_reborrow || self.was_removed {
            return;
        }
        self.apply_changes();
```

flosse (approved):

xilem_web/src/events.rs
Philipp-M: Do you have alternative suggestions? I mean it doesn't look super beautiful, but does the job?

flosse: Yes, it's definitely doing its job, and I only noticed it when my concentration was poor,
I don't want to waste your time here, sorry.

It's not significantly better, but it may still be a small thing: naming the condition + early return:

```rust
let was_created = element.flags.was_created();
let needs_update = prev_capture != capture || prev_passive != passive || was_created;
if !needs_update {
    return;
}
if !was_created {
    remove_event_listener(element.as_ref(), event, &state.callback, prev_capture);
}
state.callback =
    create_event_listener::<Event>(element.as_ref(), event, capture, passive, ctx);
```

Philipp-M: > I don't want to waste your time here, sorry.

All good, yeah your version is probably slightly more clear, thanks, I'll use it.

PR #723, opened 2024-11-01
Eliminate needless lifetimes

Philipp-M:
Result from `__CARGO_FIX_YOLO=1 cargo clippy --fix`.
As newest nightly clippy bleats about all of this.

### DIFF
--- a/masonry/src/app_driver.rs
+++ b/masonry/src/app_driver.rs
@@ -26,7 +26,7 @@ pub trait AppDriver {
     fn on_start(&mut self, state: &mut MasonryState) {}
 }
 
-impl<'a> DriverCtx<'a> {
+impl DriverCtx<'_> {
     // TODO - Add method to create timer
 
     /// Return a [`WidgetMut`] to the root widget.
--- a/masonry/src/contexts.rs
+++ b/masonry/src/contexts.rs
@@ -409,7 +409,7 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, UpdateCtx<'_>, {
 
 // --- MARK: WIDGET_MUT ---
 // Methods to get a child WidgetMut from a parent.
-impl<'a> MutateCtx<'a> {
+impl MutateCtx<'_> {
     /// Return a [`WidgetMut`] to a child widget.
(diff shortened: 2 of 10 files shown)


waywardmonkeys:
I've been avoiding doing this to any repo as I think it is worse and loses useful information.

I'd much rather just suppress it. There's an open issue in clippy about separating this or changing it as well.

waywardmonkeys:
https://github.com/rust-lang/rust-clippy/issues/13514 is that issue.

Philipp-M:
I see your point, though I think it's fine for all the cases in this PR though? But I have no strong opinion here...

DJMcNab:
I don't really want to suppress this lint, because it does give useful results for e.g. functions.

That being said, these changes are also worse... https://github.com/rust-lang/rust-clippy/pull/13286 seems to just be an unfortunate breaking change.

flosse (approved):
I prefer to be as close to the default as possible, but am still in agreement with other styles :smile:

Philipp-M:
> That being said, these changes are also worse...

You mean those in this PR?

So I think it comes down to sooner or later disable that lint or subdue to it (or add alternatives to clippy as suggested in this issue).

Should I merge this or close it?

waywardmonkeys:
Will leave it up to Daniel but in the other repos, my inclination is to suppress this rather than do these sorts of changes. Hoping that it gets fixed upstream.

DJMcNab:
If you're going to be running nightly, you should probably supress it locally until the story on the clippy side becomes clearer.

Philipp-M:
Sure that's not the issue, I can suppress it locally. So rather close this PR for now? As at least two people *seem* to be opposed, and I don't *really* care, but I like less syntax if possible (without too much loss of information).

DJMcNab:
I think so; we should see what decision is made on the clippy side first.

DJMcNab (approved):
I've changed my mind, I think these changes specifically are an improvement

Philipp-M:
So I guess in anticipation to updating the Rust version I should merge this?

DJMcNab:
Yeah, that would be good.

PR #730, opened 2024-11-08
Scope exceptions to the Linebender lint set

DJMcNab:
This allows exceptions to be burned down on a per-crate basis, rather than needing to be addressed across the whole codebase at once.

Additionally, this now follows completely the Linebender lint set, except for the unexpected-cfgs. I don't think I've introduced any behaviour changes in this PR.


waywardmonkeys (approved):
Skimmed it. Looks good. Probably best to land quickly to avoid having to rebase.

PR #731, opened 2024-11-08
xilem_web: Add `svgdraw` example

Philipp-M:
Add a simple example showing, that svg nodes can also be used similarly as a `CanvasRenderingContext2D` to draw some lines. This takes advantage of a `kurbo::QuadSpline` to avoid sharp edges  when the pointer moves fast.

See [this](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/web.3A.20Canvas.20options.20set.20with.20.60after_build.60.20doesn't.20persist) zulip topic for more context.

This could potentially be optimized further (don't clone/recalculate all the lines every reconciliation), but I think in its current state it's also a good test to see how a naive implementation performs, and so far it's not too bad. 

Btw. as noted [here](https://github.com/linebender/xilem/pull/715#issuecomment-2438408323) this implicitly also adds the `BezPath` as the example (for more manual testing opportunities).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3461,6 +3461,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "20e16a0f46cf5fd675563ef54f26e83e20f2366bcf027bcb3cc3ed2b98aaf2ca"
 
 [[package]]
+name = "svgdraw"
+version = "0.1.0"
+dependencies = [
+ "console_error_panic_hook",
+ "wasm-bindgen",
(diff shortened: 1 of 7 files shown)


flosse (approved):
nice :+1:

burrbull:
As I understand `pointer` is related to https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events

Is this recommended way for interactivity? Cuold you somehow emphasize this in the documentation for people who are not familiar with web development?

PR #732, opened 2024-11-09
xilem_core: implement `View` for `Rc<impl View>`

Philipp-M:
I needed this in xilem_web, I could've used `Arc` as well, but I think it's fine to add this for consistency when an `Arc` is not needed.

It's basically copy-pasta of the `Arc`. (I don't think a macro is worth it here?). Had to fix some resulting issues in the `Templated` view (due to ambiguity?).

### DIFF
--- a/xilem_core/src/message.rs
+++ b/xilem_core/src/message.rs
@@ -17,9 +17,10 @@ pub enum MessageResult<Action, Message = DynMessage> {
     ///
     /// This allows for sub-sections of your app to use an elm-like architecture
     Action(Action),
-    // TODO: What does this mean?
-    /// This message's handler needs a rebuild to happen.
-    /// The exact semantics of this method haven't been determined.
+    /// A view has requested a full rebuild, even if it wasn't changed by the user directly.
+    ///
+    /// This can happen for example by some kind of async action.
+    /// An example would be an async virtualized list, which fetches new entries, and requires a rebuild for the new entries.
     RequestRebuild,
     #[default]
     /// This event had no impact on the app state, or the impact it did have
--- a/xilem_core/src/message.rs
+++ b/xilem_core/src/message.rs
@@ -17,7 +17,7 @@ pub enum MessageResult<Action, Message = DynMessage> {
     ///
     /// This allows for sub-sections of your app to use an elm-like architecture
     Action(Action),
-    /// A view has requested a full rebuild, even if it wasn't changed by the user directly.
+    /// A view has requested a rebuild, even though its value hasn't changed.
     ///
     /// This can happen for example by some kind of async action.
     /// An example would be an async virtualized list, which fetches new entries, and requires a rebuild for the new entries.
--- a/xilem_core/src/view.rs
+++ b/xilem_core/src/view.rs
@@ -4,6 +4,7 @@
 //! The primary view trait and associated trivial implementations.
 
 use alloc::boxed::Box;
+use alloc::rc::Rc;
 use alloc::sync::Arc;
 use core::ops::Deref;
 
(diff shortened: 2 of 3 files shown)


xilem_core/src/view.rs
```diff
@@ -271,3 +272,65 @@ where
...
+    V: View<State, Action, Context, Message> + ?Sized,
+{
+    type Element = V::Element;
+    type ViewState = ArcState<V::ViewState>;
+
+    fn build(&self, ctx: &mut Context) -> (Self::Element, Self::ViewState) {
+        let (element, view_state) = self.deref().build(ctx);
+        (
+            element,
+            ArcState {
```
jplatte, line 290: If this type can be used for both `Rc` and `Arc`, I'd argue the "atomic" part is irrelevant and shouldn't be part of its name, i.e. it should be renamed `RcState`.

xilem_core/src/view.rs
```diff
@@ -271,3 +272,65 @@ where
...
+    }
+
+    fn rebuild(
+        &self,
+        prev: &Self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut Context,
+        element: Mut<Self::Element>,
+    ) {
+        // If this is the same value, or no rebuild was forced, there's no need to rebuild
```
DJMcNab, line 304: As a minimal change, this should be `and`. 
I think the main interesting part of this comment is the "rebuild was forced" statement.

This could have a longer comment here, but I'm not sure how much value that adds. I'd probably like to see some minimal documentation added to at least `RcState::dirty`; having the same comment repeated twice isn't great.

xilem_web/src/templated.rs
```diff
@@ -70,7 +71,6 @@ where
         // If this is the same value, or no rebuild was forced, there's no need to rebuild
         if core::mem::take(&mut view_state.dirty) || !Rc::ptr_eq(&self.0, &prev.0) {
```
DJMcNab, line 72: Why is this still needed?

DJMcNab (approved):

xilem_core/src/view.rs
Philipp-M: I can readd the comment if it makes sense, I've added docs to either `MessageResult::RequestRebuild` and to the dirty flag.

xilem_web/src/templated.rs
Philipp-M: Good point/catch. It now uses `Rc<impl View>` as inner view, so yeah not needed anymore.

xilem_core/src/message.rs
```diff
@@ -17,9 +17,10 @@ pub enum MessageResult<Action, Message = DynMessage> {
     ///
     /// This allows for sub-sections of your app to use an elm-like architecture
     Action(Action),
-    // TODO: What does this mean?
-    /// This message's handler needs a rebuild to happen.
-    /// The exact semantics of this method haven't been determined.
+    /// A view has requested a full rebuild, even if it wasn't changed by the user directly.
```
DJMcNab, line 20:
```suggestion
    /// A view has requested a rebuild, even though its value hasn't changed.
```

xilem_core/src/view.rs
```diff
@@ -208,6 +208,9 @@ where
 #[allow(unnameable_types)] // reason: Implementation detail, public because of trait visibility rules
 pub struct RcState<ViewState> {
     view_state: ViewState,
+    /// This is a flag that is set, when an inner view signifies that it requires a rebuild (via [`MessageResult::RequestRebuild`]).
+    /// This can happen, e.g. when an inner view wasn't changed by the user directly, but e.g. some kind of async action.
```
DJMcNab, line 212: It's not really the user, it's effectively the app developer, right?

I guess the semantics of who is the user isn't that clear - is it the end user, or the user of Xilem?

DJMcNab (approved):

PR #733, opened 2024-11-09
xilem_web: Add pointer events and fix bounds on event listeners and correct event type for mouse events.

Philipp-M:
As [this Firefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1675847) is finally fixed, we can be more close to the spec with the `click/auxclick/contextmenu` events by using a `PointerEvent` instead of a `MouseEvent`.

I've also noticed that the DOM interface bounds/types were out of sync with the event types (bounds), this was corrected as well.

Additionally all the `pointer...` events were added.
And a simple `PointerMsg::position() -> Point` accessor.

### DIFF
--- a/xilem_web/src/events.rs
+++ b/xilem_web/src/events.rs
@@ -392,7 +392,7 @@ macro_rules! event_definitions {
             State: 'static,
             Action: 'static,
             V: DomView<State, Action>,
-            OA: OptionalAction<Action>,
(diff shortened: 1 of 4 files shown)


Philipp-M:
Added some documentation for the `Pointer` view as well (with an example).

flosse (approved):
👍

PR #734, opened 2024-11-09
xilem_web: Extend `svgdraw` example to be an infinite zoomable canvas

Philipp-M:
This allows zooming with the mousewheel and panning with the mid/right mouse button.
I think it would be nice to add touch support with a pinch gesture handler as well.

It also replaces the random color with a color selector. And makes the stroke width range logarithmic.
And adds `type_` and `name` attributes to `HtmlInputElement`.

It was also optimized by using an `Rc<AnyDomView>` to memoize all previously drawn lines. This way I think the bottleneck now certainly is the browser.


flosse (approved):
LGTM :+1:

PR #735, opened 2024-11-10
Add getters to masonry widget contexts

### LINKED ISSUE #728: Make Masonry Ctx attributes public?
All the struct attributes [`masonry/contexts.rs`](https://github.com/linebender/xilem/blob/main/masonry/src/contexts.rs) are `pub(crate)`, but why are these not accessible for other crates?

Is there an alternative way to access these fields?

Example of context structs:
https://github.com/linebender/xilem/blob/ee3126139fb162c1f051d8af227f3e88fa615931/masonry/src/contexts.rs#L101-L132

Some of the internal widgets make use of these fields, eg. in the Grid and Flex widgets:
https://github.com/linebender/xilem/blob/ee3126139fb162c1f051d8af227f3e88fa615931/masonry/src/widget/grid.rs#L288-L290

https://github.com/linebender/xilem/blob/ee3126139fb162c1f051d8af227f3e88fa615931/masonry/src/widget/flex.rs#L1179-L1183

https://github.com/linebender/xilem/blob/ee3126139fb162c1f051d8af227f3e88fa615931/masonry/src/widget/flex.rs#L959

viktorstrate:
This fixes #728.

And update `flex` and `grid` examples to use them.
This allows external crates to access these fields for custom widgets.


masonry/src/contexts.rs
```diff
@@ -1069,6 +1074,12 @@ impl_context_method!(LayoutCtx<'_>, PaintCtx<'_>, {
     }
 });
 
+impl PaintCtx<'_> {
+    pub fn debug_paint(&self) -> bool {
```
Philipp-M, line 1078: I think when making this property public, ideally we have documentation about what it does roughly, or what this is for, as I don't think it's immediately obvious.

masonry/src/widget/flex.rs
```diff
@@ -1178,9 +1178,9 @@ impl Widget for Flex {
 
     fn paint(&mut self, ctx: &mut PaintCtx, scene: &mut Scene) {
         // paint the baseline if we're debugging layout
-        if ctx.debug_paint && ctx.widget_state.baseline_offset != 0.0 {
+        if ctx.debug_paint() && ctx.baseline_offset() != 0.0 {
```
Philipp-M, line 1181: I'm fine with method getters (`ctx.debug_paint()`) here, but IIRC Olivier had arguments against it (i.e. abstraction on a simple property)?

Philipp-M (approved):
In general LGTM apart of nits. I like to give others a chance to take a look over it though before merging.

masonry/src/widget/flex.rs
viktorstrate: The reason I chose to make `debug_paint` a getter rather than just making the attribute public, was to disallow mutation of the value.

Philipp-M: Yeah I get the motivation, that's why I think it's fine for me, also to serve as documentation for external custom widgets (which can't use these properties directly).

masonry/src/contexts.rs
viktorstrate: I have added documentation in [7b21dda](https://github.com/linebender/xilem/pull/735/commits/7b21ddaceb9d2b0d9e9a46f54a656b6faff98623) 👍

masonry/src/contexts.rs
```diff
@@ -1069,6 +1074,16 @@ impl_context_method!(LayoutCtx<'_>, PaintCtx<'_>, {
     }
 });
 
+impl PaintCtx<'_> {
+    /// Whether debug paint is enabled.
+    ///
+    /// Debug paint can be enabled using the environment variable `MASONRY_DEBUG_PAINT`.
+    /// When enabled widget boundaries are outlined.
+    pub fn debug_paint(&self) -> bool {
```
DJMcNab, line 1082: I'd like to brainstorm a different name here - to me, `debug_paint` sounds like an "action" rather than a property read. That is, this method name suggests that it performs some kind of debug operation itself.

This needn't block this from landing.

masonry/src/contexts.rs
```diff
@@ -1069,6 +1074,16 @@ impl_context_method!(LayoutCtx<'_>, PaintCtx<'_>, {
     }
 });
 
+impl PaintCtx<'_> {
+    /// Whether debug paint is enabled.
+    ///
+    /// Debug paint can be enabled using the environment variable `MASONRY_DEBUG_PAINT`.
+    /// When enabled widget boundaries are outlined.
```
DJMcNab, line 1081:
```suggestion
    /// If this property is set, your widget may draw additional debug information
    /// (such as the position of the text baseline).
    /// These should normally use the [debug color][Self::debug_color] for this widget.
    /// Please note that when debug painting is enabled, each widget's layout boundaries are
    /// outlined by Masonry, so you should avoid duplicating that.
    ///
    /// Debug paint can be enabled by setting the environment variable `MASONRY_DEBUG_PAINT`.
```

masonry/src/theme.rs
```diff
@@ -79,6 +79,9 @@ static DEBUG_COLOR: &[Color] = &[
     Color::rgb8(0, 0, 0),
 ];
 
+/// A color used for debug painting.
+///
+/// The same color is always returned given the same id, usually the id of a widget.
 pub fn get_debug_color(id: u64) -> Color {
```
DJMcNab, line 85: Can we add a method to `PaintCtx` for this? As a standalone function, this is a bit undiscoverable.

Keeping this as a raw method is still correct, but since the vast majority of potential users will be calling it with `ctx.widget_id().to_raw()`, we can improve ergonomics and discoverability.

DJMcNab (approved):
Looks good! I've suggested a few small tweaks which could help make this slightly easier to understand for future developers.

DJMcNab:
Thank you! Once this is rebased, I'll merge it

viktorstrate:
Thank you, I've rebased it now.

PR #736, opened 2024-11-10
Add padding attribute to `sized_box` view

viktorstrate:
### Masonry
- Added new padding attribute to sized_box Masonry widget.
- 3 unit tests demonstrating how padding is used.

### Xilem
- Added `Padding` struct with convenience methods for working with paddings.
- Updated `http_cats` example to use padding when more convenient and fixed hack.

## Examples

```rs
sized_box(label("hello world")).padding(10.) // Equal padding on all edges
sized_box(label("hello world")).padding(Padding::top(10.)) // Padding only on top edge
sized_box(label("hello world")).padding((10., 20., 30., 40.)) // Different padding for each edge
```

## HTTP Cats

Added padding on the left in the `http_cats` example, to make it more balanced with the right side.
Screenshot 2024-11-10 at 16 22 52

### Rename `sized_box` to `frame`?

In swiftUI the view modifier [`.frame()`](https://developer.apple.com/documentation/swiftui/view/frame(width:height:alignment:)) is used to change the size of a view. I think the name `frame` better describes the purpose of `sized_box`, since it also does backgrounds, borders (and now padding). So I wanted to suggest that `sized_box` be renamed to `frame`.

### Add `SizedBoxExt` for better ergonomics?

Similar to [`FlexExt`](https://github.com/linebender/xilem/blob/62588565692584e16197fb6d19cd3b41c104b675/xilem/src/view/flex.rs#L340C11-L340C18) and [`GridExt`](https://github.com/linebender/xilem/blob/62588565692584e16197fb6d19cd3b41c104b675/xilem/src/view/grid.rs#L248), I was thinking that a new `SizedBoxExt` could be introduced to easily wrap any view in a `sized_box`, something like the following.

```rust
pub trait SizedBoxExt<State, Action>: WidgetView<State, Action> {
  fn sized_box(self) -> SizedBox<Self, State, Action> {
    sized_box(self)
  }
}
```

This would allow for chaining modifiers like this:

```rust
label("Hello world")
  .text_size(20.)
  .sized_box() // After this padding, background, width, height can be added
  .padding(20.)
```

Or even somthing more advanced like this:

```rust
label("Hello world")
  .sized_box()
  .padding(20.)
  .background(Color::Teal)
  .border(Color::Blue, 4.)
  .sized_box() // By wrapping in another sized box we add another border rather than overwriting the previous one
  .border(Color::Magenta, 4.)
```

### DIFF
--- a/masonry/src/widget/sized_box.rs
+++ b/masonry/src/widget/sized_box.rs
@@ -25,6 +25,23 @@ struct BorderStyle {
     color: Color,
 }
 
+/// Padding specifies the spacing between the edges of the box and the child view.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Padding {
+    /// The amount of padding in logical pixels for the top edge.
+    pub top: f64,
+    /// The amount of padding in logical pixels for the trailing edge.
+    ///
+    /// For LTR contexts this is the right edge, for RTL it is the left edge.
+    pub trailing: f64,
+    /// The amount of padding in logical pixels for the bottom edge.
+    pub bottom: f64,
+    /// The amount of padding in logical pixels for the leading edge.
+    ///
+    /// For LTR contexts this is the left edge, for RTL it is the right edge.
+    pub leading: f64,
+}
+
 // TODO - Have Widget type as generic argument
 
 /// A widget with predefined size.
@@ -44,6 +61,83 @@ pub struct SizedBox {
     border: Option<BorderStyle>,
     corner_radius: RoundedRectRadii,
     padding: Padding,
+}
+
+// --- MARK: IMPL PADDING ---
+
+impl Padding {
+    /// Constructs a new `Padding` by specifying the amount of padding for each edge.
+    pub const fn new(top: f64, trailing: f64, bottom: f64, leading: f64) -> Self {
+        Self {
+            top,
+            trailing,
+            bottom,
+            leading,
+        }
+    }
+
+    /// A padding of zero for all edges.
+    pub const ZERO: Padding = Padding::all(0.);
+
+    /// Constructs a new `Padding` with equal amount of padding for all edges.
+    pub const fn all(padding: f64) -> Self {
+        Self::new(padding, padding, padding, padding)
+    }
+
+    /// Constructs a new `Padding` with the same amount of padding for the horizontal edges,
+    /// and zero padding for the vertical edges.
+    pub const fn horizontal(padding: f64) -> Self {
+        Self::new(0., padding, 0., padding)
+    }
+
+    /// Constructs a new `Padding` with the same amount of padding for the vertical edges,
+    /// and zero padding for the horizontal edges.
+    pub const fn vertical(padding: f64) -> Self {
+        Self::new(padding, 0., padding, 0.)
+    }
+
+    /// Constructs a new `Padding` with padding only at the top edge and zero padding for all other edges.
+    pub const fn top(padding: f64) -> Self {
+        Self::new(padding, 0., 0., 0.)
+    }
+
+    /// Constructs a new `Padding` with padding only at the trailing edge and zero padding for all other edges.
+    pub const fn trailing(padding: f64) -> Self {
+        Self::new(0., padding, 0., 0.)
+    }
+
+    /// Constructs a new `Padding` with padding only at the bottom edge and zero padding for all other edges.
+    pub const fn bottom(padding: f64) -> Self {
+        Self::new(0., 0., padding, 0.)
+    }
+
+    /// Constructs a new `Padding` with padding only at the leading edge and zero padding for all other edges.
+    pub const fn leading(padding: f64) -> Self {
+        Self::new(0., 0., 0., padding)
+    }
+}
+
+impl From<f64> for Padding {
+    /// Converts the value to a `Padding` object with that amount of padding on all edges.
+    fn from(value: f64) -> Self {
+        Self::all(value)
+    }
+}
+
+impl From<(f64, f64, f64, f64)> for Padding {
+    /// Converts the tuple to a `Padding` object,
+    /// following CSS padding order for 4 values (top, trailing, bottom, leading).
+    fn from(value: (f64, f64, f64, f64)) -> Self {
+        Self::new(value.0, value.1, value.2, value.3)
+    }
+}
+
+impl From<(f64, f64)> for Padding {
+    /// Converts the tuple to a `Padding` object,
+    /// following CSS padding order for 2 values (vertical, horizontal)
+    fn from(value: (f64, f64)) -> Self {
+        Self::new(value.0, value.1, value.0, value.1)
+    }
 }
 
 // --- MARK: BUILDERS ---
@@ -184,7 +278,7 @@ impl SizedBox {
         self
     }
 
-    /// Builder style method for specifying the padding between the box and the child widget.
+    /// Builder style method for specifying the padding added by the box.
     pub fn padding(mut self, padding: impl Into<Padding>) -> Self {
         self.padding = padding.into();
         self
@@ -587,98 +681,4 @@ mod tests {
     }
 
     // TODO - add screenshot tests for different brush types
-}
-
-// --- MARK: PADDING STRUCT ---
-
-/// Padding specifies the spacing between the edges of the box and the child view.
-#[derive(Debug, Clone, Copy, PartialEq)]
-pub struct Padding {
-    /// The amount of padding for the top edge.
-    pub top: f64,
-    /// The amount of padding for the trailing edge.
-    ///
-    /// For LTR contexts this is the right edge, for RTL it is the left edge.
-    pub trailing: f64,
-    /// The amount of padding for the bottom edge.
-    pub bottom: f64,
-    /// The amount of padding for the leading edge.
-    ///
-    /// For LTR contexts this is the left edge, for RTL it is the right edge.
-    pub leading: f64,
-}
-
-impl Padding {
-    /// Creates a new [Padding] object by specifying the amount of padding for each edge.
-    pub const fn new(top: f64, trailing: f64, bottom: f64, leading: f64) -> Self {
-        Self {
-            top,
-            trailing,
-            bottom,
-            leading,
-        }
-    }
-
-    /// A padding of zero for all edges.
-    pub const ZERO: Padding = Padding::all(0.);
-
-    /// Creates a new [Padding] object with equal amount of padding for all edges.
-    pub const fn all(padding: f64) -> Self {
-        Self::new(padding, padding, padding, padding)
-    }
-
-    /// Creates a new [Padding] object with the same amount of padding for the horizontal edges,
-    /// and zero padding for the vertical edges.
-    pub const fn horizontal(padding: f64) -> Self {
-        Self::new(0., padding, 0., padding)
-    }
-
-    /// Creates a new [Padding] object with the same amount of padding for the vertical edges,
-    /// and zero padding for the horizontal edges.
-    pub const fn vertical(padding: f64) -> Self {
-        Self::new(padding, 0., padding, 0.)
-    }
-
-    /// Creates a new [Padding] object with padding only at the top edge and zero padding for all other edges.
-    pub const fn top(padding: f64) -> Self {
-        Self::new(padding, 0., 0., 0.)
-    }
-
-    /// Creates a new [Padding] object with padding only at the trailing edge and zero padding for all other edges.
-    pub const fn trailing(padding: f64) -> Self {
-        Self::new(0., padding, 0., 0.)
-    }
-
-    /// Creates a new [Padding] object with padding only at the bottom edge and zero padding for all other edges.
-    pub const fn bottom(padding: f64) -> Self {
-        Self::new(0., 0., padding, 0.)
-    }
-
-    /// Creates a new [Padding] object with padding only at the leading edge and zero padding for all other edges.
-    pub const fn leading(padding: f64) -> Self {
-        Self::new(0., 0., 0., padding)
-    }
-}
-
-impl From<f64> for Padding {
-    /// Converts the value to a [Padding] object with that amount of padding on all edges.
-    fn from(value: f64) -> Self {
-        Self::all(value)
-    }
-}
-
-impl From<(f64, f64, f64, f64)> for Padding {
-    /// Converts the tuple to a [Padding] object,
-    /// following CSS padding order for 4 values (top, trailing, bottom, leading).
-    fn from(value: (f64, f64, f64, f64)) -> Self {
-        Self::new(value.0, value.1, value.2, value.3)
-    }
-}
-
-impl From<(f64, f64)> for Padding {
-    /// Converts the tuple to a [Padding] object,
-    /// following CSS padding order for 2 values (vertical, horizontal)
-    fn from(value: (f64, f64)) -> Self {
-        Self::new(value.0, value.1, value.0, value.1)
-    }
 }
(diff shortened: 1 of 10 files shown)


masonry/src/widget/sized_box.rs
```diff
@@ -476,6 +505,19 @@ mod tests {
...
+    #[test]
+    fn label_box_with_padding() {
+        let widget = SizedBox::new(Label::new("hello"))
+            .border(Color::BLUE, 5.0)
+            .rounded(5.0)
+            .padding((60., 40.));
+
+        let mut harness = TestHarness::create(widget);
+
+        assert_debug_snapshot!(harness.root_widget());
```
DJMcNab, line 517: I'm not sure these debug snapshot tests are needed here.

masonry/src/widget/sized_box.rs
```diff
@@ -180,6 +184,11 @@ impl SizedBox {
         self
     }
 
+    pub fn padding(mut self, padding: impl Into<Insets>) -> Self {
```
DJMcNab, line 187: This should be documented.

masonry/src/widget/sized_box.rs
```diff
@@ -266,6 +275,18 @@ impl SizedBox {
         this.ctx.request_paint_only();
     }
 
+    /// Clears padding.
+    pub fn clear_padding(this: &mut WidgetMut<'_, Self>) {
+        this.widget.padding = Insets::ZERO;
+        this.ctx.request_paint_only();
```
DJMcNab, line 281: Maybe? This should be a wrapper around `self.set_padding(Insets::ZERO)`.

masonry/src/widget/sized_box.rs
```diff
@@ -266,6 +275,18 @@ impl SizedBox {
...
+    }
+
+    /// Set the padding around this widget.
+    pub fn set_padding(this: &mut WidgetMut<'_, Self>, padding: impl Into<Insets>) {
+        this.widget.padding = padding.into();
+        this.ctx.request_paint_only();
```
DJMcNab, line 287: This is surprising to me. How does this work? The padding field is used in `layout`, so I'd expect changing it to have an impact on layout.

Could we have a test which changes the padding at runtime and demonstrates that the correct result is reached?

xilem/examples/http_cats.rs
```diff
@@ -197,8 +199,8 @@ impl Status {
             FlexSpacer::Fixed(10.),
             image,
             // TODO: Overlay on top of the image?
-            // HACK: Trailing spaces workaround scrollbar covering content
-            prose("Copyright ©️ https://http.cat      ").alignment(TextAlignment::End),
+            sized_box(prose("Copyright ©️ https://http.cat").alignment(TextAlignment::End))
+                .padding(Padding::trailing(20.)),
```
DJMcNab, line 203: We should still document this as a hack to avoid overlap with the scrollbar, but it is certainly a better solution.

xilem/src/view/sized_box.rs
```diff
@@ -209,3 +221,84 @@ struct BorderStyle {
     width: f64,
     color: Color,
 }
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Padding {
```
DJMcNab, line 226: All of this should be documented; we have `expect(missing_docs)` as a temporary crutch, but we want to get things fully documented.

xilem/src/view/sized_box.rs
```diff
@@ -209,3 +221,84 @@ struct BorderStyle {
...
+    top: f64,
```
DJMcNab, line 227: Can these fields be public?

xilem/src/view/sized_box.rs
```diff
@@ -209,3 +221,84 @@ struct BorderStyle {
...
+}
+
+impl From<f64> for Padding {
+    fn from(value: f64) -> Self {
+        Self::all(value)
+    }
+}
+
+// Follows CSS padding order for 4 values (top, trailing, bottom, leading)
+impl From<(f64, f64, f64, f64)> for Padding {
```
DJMcNab, line 281: We need to document why we use a different type to `Insets`, somewhere. Probably directly on `Padding`.

(That reason is so that these impls can exist)

xilem/src/view/sized_box.rs
```diff
@@ -209,3 +221,84 @@ struct BorderStyle {
...
+
+    pub const fn trailing(padding: f64) -> Self {
+        Self::new(0., padding, 0., 0.)
+    }
+
+    pub const fn bottom(padding: f64) -> Self {
+        Self::new(0., 0., padding, 0.)
+    }
+
+    pub const fn leading(padding: f64) -> Self {
```
DJMcNab, line 269: So in an LTR context this is left, and a RTL context this is right?

We should document that clearly. Especially since we don't actually handle bidi yet...

xilem/src/view/sized_box.rs
DJMcNab, line 226: Could this struct be moved into Masonry? I don't know a compelling argument for this to be Xilem specific.

DJMcNab (commented):
This looks really great, thank you! There are a few things I'd like to see changed before this lands, but this is idiomatic for how Xilem/Masonry are currently designed. I think  the behaviour with dynamic paddings is broken, which I'd like to see addressed (or a convincing check that it isn't).

The suggested name of `Frame` is certainly compelling, but should be a follow-up.

xilem/examples/http_cats.rs
```diff
@@ -47,13 +47,15 @@ enum ImageState {
...
-        let left_column = portal(flex((
+        let left_column = sized_box(portal(flex((
             prose("Status"),
             self.statuses
                 .iter_mut()
                 .map(Status::list_view)
                 .collect::<Vec<_>>(),
-        )));
+        ))))
+        .padding(Padding::leading(20.));
```
DJMcNab, line 57:
```suggestion
        .padding(Padding::leading(5.));
```

One thing to consider is that on Android, the amount of horizontal space is quite limited, which makes large paddings harder to justify (at least whilst we don't have a good dynamic layout). I also think that on desktop, a smaller padding here is nicer.

That being said, some padding here is definitely an improvement over none.

xilem/examples/http_cats.rs
DJMcNab, line 203: I've tried a few different numbers for this right padding, and none are quite right. The main problem is that the images from the API have the awkward black boxes around them with duplicated information. The main https://http.cat website crops them down in a lossy way, which also isn't ideal.

masonry/src/widget/sized_box.rs
viktorstrate: This is a mistake, it should request layout change.

masonry/src/widget/sized_box.rs
viktorstrate: I've added documentation

masonry/src/widget/sized_box.rs
viktorstrate: Fixed

masonry/src/widget/sized_box.rs
viktorstrate: It's fixed now

xilem/examples/http_cats.rs
viktorstrate: Sure. I added the comment back in.

xilem/src/view/sized_box.rs
viktorstrate: I've added documentation now, I'd like if you had a look at it to check that it follows the conventions of the project and that I didn't miss anything.

xilem/src/view/sized_box.rs
viktorstrate: Yes, I've made them public now.

xilem/src/view/sized_box.rs
viktorstrate: I moved the Padding struct to Masonry and completely removed the dependency on Insets.

xilem/src/view/sized_box.rs
viktorstrate: I have added a comment about it in the documentation of the struct attributes. Please let me know if that look good

xilem/src/view/sized_box.rs
viktorstrate: Yes, I've moved it.

xilem/examples/http_cats.rs
viktorstrate: I've accepted the change in the latest commit.

xilem/examples/http_cats.rs
viktorstrate: I ended up setting it to 15, so it's symmetrical with the left side. (10px for the scroll bar + 5px padding)

masonry/src/widget/sized_box.rs
```diff
@@ -512,5 +554,131 @@ mod tests {
...
+        let mut harness = TestHarness::create(widget);
+
+        assert_debug_snapshot!(harness.root_widget());
+        assert_render_snapshot!(harness, "label_box_with_outer_padding");
+    }
+
     // TODO - add screenshot tests for different brush types
 }
+
+// --- MARK: PADDING STRUCT ---
```
DJMcNab, line 592: This definitely shouldn't be after the tests. The usual style in Masonry would have the definition of `Padding` just before the definition of the `SizedBox` type, then these impls before the `impl SizedBox` section.

masonry/src/widget/sized_box.rs
```diff
@@ -512,5 +554,131 @@ mod tests {
...
+    /// The amount of padding for the bottom edge.
+    pub bottom: f64,
+    /// The amount of padding for the leading edge.
+    ///
+    /// For LTR contexts this is the left edge, for RTL it is the right edge.
+    pub leading: f64,
+}
+
+impl Padding {
+    /// Creates a new [Padding] object by specifying the amount of padding for each edge.
```
DJMcNab, line 612:
```suggestion
    /// Creates a new `Padding` by specifying the amount of padding for each edge.
```

See e.g. https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.new - if you're on this page, you already know you're talking about `Padding`

The term "object" also isn't needed or particularly meaningful in Rust.

masonry/src/widget/sized_box.rs
```diff
@@ -512,5 +554,131 @@ mod tests {
...
+
     // TODO - add screenshot tests for different brush types
 }
+
+// --- MARK: PADDING STRUCT ---
+
+/// Padding specifies the spacing between the edges of the box and the child view.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Padding {
+    /// The amount of padding for the top edge.
```
DJMcNab, line 597:
```suggestion
    /// The amount of padding in logical pixels for the top edge.
```

masonry/src/widget/sized_box.rs
```diff
@@ -180,6 +184,12 @@ impl SizedBox {
         self
     }
 
+    /// Builder style method for specifying the padding between the box and the child widget.
```
DJMcNab, line 187:
```suggestion
    /// Builder style method for specifying the padding added by the box.
    ///
    /// The padding will be inside any [borders](Self::border).
```

masonry/src/widget/sized_box.rs
```diff
@@ -512,5 +554,131 @@ mod tests {
...
+    }
+
+    /// Creates a new [Padding] object with padding only at the leading edge and zero padding for all other edges.
+    pub const fn leading(padding: f64) -> Self {
+        Self::new(0., 0., 0., padding)
+    }
+}
+
+impl From<f64> for Padding {
+    /// Converts the value to a [Padding] object with that amount of padding on all edges.
```
DJMcNab, line 664: I've looked at the generated docs page, and these are really hard to discover.

These docs are in the right place, but we should also mention these impls at the top level of `Padding`

xilem/src/view/sized_box.rs
```diff
@@ -3,7 +3,7 @@
 
 use std::marker::PhantomData;
 
-use masonry::widget;
+use masonry::widget::{self, Padding};
```
DJMcNab, line 6:
```suggestion
use masonry::widget;
pub use masonry::widget::Padding;
```
(possibly with a different ordering of the items)

xilem/src/view/sized_box.rs
```diff
@@ -103,11 +105,17 @@ impl<V, State, Action> SizedBox<V, State, Action> {
         self
     }
 
-    /// Builder style method for rounding off corners of this container by setting a corner radius
+    /// Builder style method for rounding off corners of this container by setting a corner radius.
```
DJMcNab, line 108: :fire: thanks!

DJMcNab (commented):
Looks good. Some tweaks to the docs and ordering in the file are still needed, but this is nearly ready.

Philipp-M:
> Rename sized_box to frame?

I'm in favor, but I think this should be a follow-up PR.

> Add SizedBoxExt for better ergonomics?

I guess we can do that as well. (Could also be a follow-up)

masonry/src/widget/sized_box.rs
viktorstrate: I moved it in the new commit.

masonry/src/widget/sized_box.rs
viktorstrate: Makes sense :)

masonry/src/widget/sized_box.rs
viktorstrate: Accepted in latest commit

masonry/src/widget/sized_box.rs
viktorstrate: Accepted in latest commit

xilem/src/view/sized_box.rs
viktorstrate: Accepted

masonry/src/widget/sized_box.rs
viktorstrate: I'm unsure how to best do this?

xilem/examples/http_cats.rs
```diff
@@ -10,6 +10,7 @@
 
 use std::sync::Arc;
 
+use masonry::widget::Padding;
```
DJMcNab, line 13: This should use the re-export from Xilem.

masonry/src/widget/sized_box.rs
DJMcNab: As in, adding some text to the doc comment on `Padding` describing that these implementations exist for convenience?

DJMcNab (approved):
The unresolved comments still apply (https://github.com/linebender/xilem/pull/736#discussion_r1836175033 and https://github.com/linebender/xilem/pull/736#discussion_r1836869677), plus this new one.

Otherwise this is really good, thanks!

masonry/src/widget/sized_box.rs
viktorstrate: Should I remove them? I just copied them from the other test cases in the file.

xilem/examples/http_cats.rs
viktorstrate: Fixed

masonry/src/widget/sized_box.rs
viktorstrate: Okay, I added this comment now, how does that look?

https://github.com/linebender/xilem/pull/736/files#diff-5cd89e6ede9989e07db2161ca9bb872fc743fe9ecfe65f1e43da72fa6379079cR30-R31

masonry/src/widget/sized_box.rs
DJMcNab: We can leave them from consistency. @PoignardAzur what are these snapshots actually testing?

masonry/src/widget/sized_box.rs
DJMcNab: I'd probably choose slightly different wording, but that form looks good. Thanks!

DJMcNab:
And yes, the two suggested follow-ups (probably renaming to Frame first) would be welcome.

PR #737, opened 2024-11-10
Add `zstack` view to compose views on top of each other

viktorstrate:
Adds a new `zstack` Xilem view (along with an accompanied Masonry widget).
The view is inspired by the [ZStack view in SwiftUI](https://developer.apple.com/documentation/swiftui/zstack), which composes the child view on top of each other.

This pull-request is still work-in-progress.

- [x] It doesn't allow for changing the layout when BoxConstraints of the children are different.
(top-left, top-right, bottom-left, bottom-right, center).
- [x] I want to update `http_cat` to put copyright on top of image like seen below.
Screenshot 2024-11-10 at 17 46 45


Philipp-M:
See also #591, which goes a similar direction.

I think to make a `zstack` really useful we would need transforms (`kurbo::Affine`) on widgets in general and relevant logic for pointer events etc. See also [here](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Add.20.60Affine.60.20transform.20to.20.60Widget.60.20trait.3F/near/472076600). I'm thinking about getting to that topic soon, as I think this is the main blocker for allowing fancy interactive vector graphics with vello and xilem.

I like the general direction, i.e. unifying a potential xilem_svg API using a zstack (instead of separate context such as in the web).

DJMcNab:
I don't think that arbitrary transforms should block this landing, though. I don't really have the bandwidth to look at this too closely at the moment, unfortunately.

viktorstrate:
The PR is now ready for review!

I have added the copyright on top of the image in the `http_cats` example. I'm concerned about the visibility on light images. Is there a way to add a border to the text? I would like to get suggestions on what to do in this case?

Screenshot 2024-11-16 at 22 56 38

DJMcNab:
Maybe adding a `sized_box` with a semi-transparent black background?

xilem/examples/http_cats.rs
```diff
@@ -197,11 +198,13 @@ impl Status {
...
                 .alignment(TextAlignment::Middle),
             FlexSpacer::Fixed(10.),
-            image,
-            // TODO: Overlay on top of the image?
-            // HACK: Trailing padding workaround scrollbar covering content
-            sized_box(prose("Copyright ©️ https://http.cat").alignment(TextAlignment::End))
-                .padding(Padding::trailing(15.)),
+            zstack((
+                image,
+                // HACK: Trailing spaces workaround scrollbar covering content
```
DJMcNab, line 203: Rebase artifact

xilem/src/view/zstack.rs
```diff
@@ -0,0 +1,219 @@
...
+        id_path: &[crate::core::ViewId],
+        message: DynMessage,
+        app_state: &mut State,
+    ) -> crate::core::MessageResult<Action, DynMessage> {
+        self.sequence
+            .seq_message(view_state, id_path, message, app_state)
+    }
+}
+
+// MARK: ZStackElement
```
DJMcNab, line 97: This seems to be a different format to the other marker comments?

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,324 @@
...
+use crate::{
+    vello::Scene, widget::WidgetMut, AccessCtx, BoxConstraints, LayoutCtx, PaintCtx, Point,
+    QueryCtx, RegisterCtx, Size, Widget, WidgetId, WidgetPod,
+};
+use accesskit::{NodeBuilder, Role};
+use smallvec::SmallVec;
+use tracing::trace_span;
+
+struct Child {
+    widget: WidgetPod<Box<dyn Widget>>,
```
DJMcNab, line 13: Is there a reason the children don't have an individual alignment?

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,324 @@
...
+    Trailing,
+    BottomLeading,
+    Bottom,
+    BottomTrailing,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VerticalAlignment {
+    Top,
+    Center,
```
DJMcNab, line 38:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum VerticalAlignment {
    Top,
    #[default]
    Center,
```
See: https://doc.rust-lang.org/std/default/trait.Default.html#enums

Our MSRV is higher than 1.62.

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+}
+
+// --- MARK: IMPL ZSTACK ---
+impl ZStack {
+    pub fn new() -> Self {
+        ZStack {
+            children: Vec::new(),
+            alignment: Alignment::default(),
+        }
```
DJMcNab, line 147:
```suggestion
        Self::default()
```

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    pub fn with_alignment(mut self, alignment: impl Into<Alignment>) -> Self {
+        self.alignment = alignment.into();
+        self
+    }
+
+    pub fn with_child(self, child: impl Widget) -> Self {
```
DJMcNab, line 155: This needs to document whether the child is at the front or the back.

masonry/src/widget/zstack.rs
DJMcNab, line 1: All of these types should have documentation. I recognise that this isn't the prevailing code style, but that doesn't mean we should continue that.

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,324 @@
...
+    pub fn set_alignment(this: &mut WidgetMut<'_, Self>, alignment: impl Into<Alignment>) {
+        this.widget.alignment = alignment.into();
+        this.ctx.request_layout();
+    }
+}
+
+// --- MARK: IMPL WIDGET---
+impl Widget for ZStack {
+    fn layout(&mut self, ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
+        let total_size = bc.max();
```
DJMcNab, line 212: This is suspect for a few reasons:
1) An infinite bound makes this fundamentally broken.
2) Even if all the children are small, this view takes up all available space.

I think the correct approach here might be a 2-pass solution, first to get the "overall" size, then to align widgets within that size. Allowing that is the reason that `run_layout` and `place_child` are different methods.

xilem/src/view/zstack.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+pub struct ZStack<Seq> {
+    sequence: Seq,
+    alignment: Alignment,
+}
+
+impl<Seq> ZStack<Seq> {
+    #[must_use]
```
DJMcNab, line 29: The usual style is to place this on the struct definition.

DJMcNab (commented):
This looks like a good foundation, and it should be a useful view. Some review thoughts from a first pass.

One thing which does concern me is event handling; I suspect that the behaviour will be hard to reason about. Ideally, we'd warn if more than one layer is interactive, but we don't currently have a mechanism to do that.

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,324 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum VerticalAlignment {
+    Top,
+    Center,
+    Bottom,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum HorizontalAlignment {
```
Philipp-M, line 43: We may want to check whether we can merge or reuse a few of these enums for different container widgets. (no action required here though, I think that should be done in follow-up PRs)

masonry/src/widget/zstack.rs
Philipp-M: I even omit these kind of `new` implementations nowadays, as they're kind of redundant.

masonry/src/widget/zstack.rs
Philipp-M: Ideally in the name already, something like `with_appended_child`... (Though TBF `flex` and `grid` has the same names for these methods, so that's likely a case for a follow-up PR).

Philipp-M:
> One thing which does concern me is event handling;

Can you expand on that a little bit?

masonry/src/widget/zstack.rs
DJMcNab: I find them useful occasionally - Rust-analyzer makes the `new` implementation much more obvious than a `::default`

viktorstrate:
> One thing which does concern me is event handling; I suspect that the behaviour will be hard to reason about.

Would it somehow be possible to pass the event to the front most child of which the cursor is within the bounding box?

Philipp-M:
Hmm, I think we need a more significant rework of the pointer-intersection logic in masonry anyways, i.e. allow the widgets to customize the pointer intersection behavior (as right now only the bounds are checked). Possibly with a cheaper hierarchical bounds check (BVH) before.

I have similar issues in #753.

I'm thinking about tackling this, as this is relevant for kurbo shapes with pointer event listeners as well.

Anyway, I think this PR (and the linked) just shows issues with the mentioned pointer/widget-intersection logic, which I think is orthogonal and should probably be handled in another PR....

masonry/src/widget/zstack.rs
DJMcNab: Yeah, this would be the right way to expose that at the Xilem level.

xilem/src/view/zstack.rs
viktorstrate: Fixed

masonry/src/widget/zstack.rs
viktorstrate: Thanks for the tip, I've changed it

masonry/src/widget/zstack.rs
viktorstrate: Changed in later commit

masonry/src/widget/zstack.rs
viktorstrate: I've written a fair bit of documentation in 42473c8, is there anything obvious that I missed?

masonry/src/widget/zstack.rs
viktorstrate: I added this in [78a901e](https://github.com/linebender/xilem/pull/737/commits/78a901ec43ff76ca81752740eb2480b85c1a3591), can you please take a look at it. I'm unsure if I did it right?

xilem/src/view/zstack.rs
viktorstrate: Fixed

xilem/examples/http_cats.rs
viktorstrate: Fixed now

masonry/src/widget/zstack.rs
viktorstrate: Should we implement this or keep it simple like it is now?

masonry/src/widget/zstack.rs
viktorstrate: A comment has been added

masonry/src/widget/zstack.rs
Philipp-M: I think we can implement it, I don't think it will be much more complex? It's a little bit more code, but it would be consistent to the other container views, and may be more straight-forward on the user-facing-side.
I think it's a little bit awkward to compose 2 `zstack` to achieve similar behavior.
(And the wrapper types (`Child`, `ZStackElement`) would actually make sense).

Philipp-M:
> Sure. If you have a child widget of the zstack which accepts events in only a small region, and wants to pass through the rest (for example, a circular button), I don't think that will work because of how Masonry works, that events can only bubble. I haven't thought too deeply about it, and it doesn't need to be addressed in this PR.

FWIW I *think* this should be fixed with #753, as well as supporting custom pointer intersection shapes (like the circular button), when I understand you correctly.

xilem/src/view/zstack.rs
```diff
@@ -0,0 +1,362 @@
...
+    ) -> MessageResult<Action, DynMessage> {
+        self.sequence
+            .seq_message(view_state, id_path, message, app_state)
+    }
+}
+
+// --- MARK: ZStackExt ---
+
+pub trait ZStackExt<State, Action>: WidgetView<State, Action> {
+    fn alignment(self, alignment: impl Into<ChildAlignment>) -> ZStackItem<Self, State, Action>
```
DJMcNab, line 125: This should have some documentation. In particular, it should specify that this alignment can only be used for a direct child of a `zstack`.

xilem/examples/http_cats.rs
```diff
@@ -197,12 +198,22 @@ impl Status {
...
+                    sized_box(
+                        prose("Copyright ©️ https://http.cat")
+                            .line_break_mode(LineBreaking::Clip)
+                            .alignment(TextAlignment::End),
+                    )
+                    .padding(4.)
+                    .rounded(4.)
+                    .background(Color::BLACK.multiply_alpha(0.5)),
+                )
+                // HACK: Trailing padding workaround scrollbar covering content
```
DJMcNab, line 213: I'm not sure this hack comment is entirely correct anymore.

DJMcNab (commented):
Handful of small things, but otherwise this looks good.

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,416 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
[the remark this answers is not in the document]
viktorstrate: Fixed

masonry/src/widget/zstack.rs
```diff
@@ -0,0 +1,416 @@
...
+            let child_size = ctx.run_layout(&mut child.widget, &loosened_bc);
+
+            max_size.width = child_size.width.max(max_size.width);
+            max_size.height = child_size.height.max(max_size.height);
+        }
+
+        // Second pass: place the children given the calculated max_size bounds.
+        let child_bc = BoxConstraints::new(Size::ZERO, max_size);
+        for child in &mut self.children {
+            let child_size = ctx.run_layout(&mut child.widget, &child_bc);
```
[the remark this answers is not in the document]
viktorstrate: Makes sense, it's fixed

xilem/src/view/zstack.rs
viktorstrate: Fixed

xilem/examples/http_cats.rs
viktorstrate: I've deleted it

masonry/src/widget/zstack.rs
viktorstrate: Added in bbb1d5f6bcbc1c5e1fc40cf7ee54b781caad2550

masonry/src/widget/zstack.rs
viktorstrate: I replaced on of the `run_layout` calls with a `child_size` call instead.

DJMcNab (approved):
Thanks!

xilem/src/view/zstack.rs
```diff
@@ -0,0 +1,378 @@
...
+            f(downcast)
+        };
+
+        (this, ret)
+    }
+}
+
+// MARK: Sequence
+
+/// A trait implementing `ViewSequence` for `ZStackElements`.
```
Philipp-M, line 300: Nit: I think using the plural for `ZStackElement` may suggest that there's a type `ZStackElements`, so maybe rather `ZStackElement`s or 
```
[`ZStackElement`]s
```
to directly link to the type.

(same for the other occurences of this here)

Philipp-M (approved):
Looks good, thanks for all the work!

xilem/src/view/zstack.rs
viktorstrate: Makes sense, I fixed it in the latest commit

viktorstrate:
I think the stars finally aligned ✨ and the CI is happy. It should be ready for merge now. Happy new year!

Philipp-M (approved):
Thanks again!

PR #738, opened 2024-11-11
Add `--no-fail-fast` to CI testing.

xStrom:
Most of the time is spent on compiling, so might as well see all the tests that fail.

Also the `cargo test --doc` step is moved after the screenshot failure upload to make it clearer that the doc tests don't produce any screenshots.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,10 @@ env:
 # Using cargo-hack also allows us to more easily test the feature matrix of our packages.
 # We use --each-feature & --optional-deps which will run a separate check for every feature.
 #
+# We use cargo-nextest, which has a faster concurrency model for running tests.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -269,6 +268,9 @@ jobs:
           name: masonry-snapshot-tests-${{ matrix.os }}
           path: masonry/src/widget/screenshots
 
+      - name: cargo test --doc
```
DJMcNab, line 271: Hmm. We likely do want these to produce screenshots - see #501.

DJMcNab (approved):
Not sure about the move of the doc tests, but otherwise looks good

.github/workflows/ci.yml
xStrom: Well, it won't be a big task to move it if  it makes sense with the new screenshot implementation. As for right now, it looks like we don't even have an implementation, so the details may change.

PR #739, opened 2024-11-11
Fix `missing_debug_implementations` in Xilem Core

DJMcNab:
This also updates to use `core::error::Error` now that it is stable.


Philipp-M (approved):
Thanks.

PR #740, opened 2024-11-12
Fix `elided_lifetimes_in_path` in Xilem Core

DJMcNab:


xStrom (approved):
Standard lint compliance.

DJMcNab:
There has been some discussion on Zulip. https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Standard.20Lint.20set/near/481912214

Essentially, the only conclusion we can really reach is that the tooling here is really bad. If tooling improved to allow enforcing a different set of choices (e.g. with a per-type attribute), I'd be all for swapping this back again. However, currently the only way to get consistency is to go with what Rust will enforce.

PR #741, opened 2024-11-12
Remove casts to f32 for text sizes

DJMcNab:
The main change is that text sizes in Masonry::theme are now `f32`. The rest of the changes are autofixed.


xStrom (approved):
Seems very reasonable.

PR #744, opened 2024-11-12
Fix `single_use_lifetimes` in Masonry

PoignardAzur:


masonry/src/paint_scene_helpers.rs
```diff
@@ -15,6 +15,10 @@ pub struct UnitPoint {
     v: f64,
 }
 
+#[expect(
+    single_use_lifetimes,
+    reason = "Anonymous lifetimes in `impl Trait` are unstable"
```
DJMcNab, line 20: Can we link to the rustc issue for this lint false positive here?

PoignardAzur: As far as I can see, there's no tracking issue beside https://dev-doc.rust-lang.org/unstable-book/language-features/anonymous-lifetime-in-impl-trait.html

DJMcNab: Hmm. https://github.com/rust-lang/rust/issues/129255 was the first result for `single_use_lifetimes` in rust-lang/rust.

PoignardAzur: Oh, right, I looked for the wrong thing.

DJMcNab (approved):

PR #746, opened 2024-11-13
Not for landing: Stress test of many variable font glyphs

DJMcNab:


DJMcNab:
I'm going to close this for now; we have done the experiments we needed to here.

PR #749, opened 2024-11-15
xilem_web: Add `ViewCtx::as_owned` to allow composing the view context in custom view contexts

Philipp-M:
Also adds `MessageThunk::enqueue_message` for convenience (basically `MessageThunk::push_message` delayed with `queueMicrotask`).

See https://github.com/slowtec/xilem-leaflet/pull/4 for a use-case of this.


flosse (approved):
Thank you, this helps a lot when developing with external JS libraries :+1:

PR #751, opened 2024-11-16
Add more robust `debug_assertions` verification to CI.

xStrom:
When a workspace uses `debug_assertions` for conditional compilation (*like Xilem does*), especially of the `#[cfg(debug_assertions)]` kind, it is not sufficient to just test with it either set to `true` or `false`. Both need testing as we don't know in advance which path contains faulty code.

An initial response to this issue was done in #431 which added some `--release --all-targets` steps to the CI.

This PR here improves that verification in some key ways:
* In addition to the stable toolchain, also verify MSRV.
* Only turn off the `debug_assertions` configuration option as that is the differentiator for our testing. The full `--release` option spends time on optimizations. It may be worth doing occasional `--release` tests to detect bugs in code optimization, but those checks aren't worth it as part of the `cargo clippy` / `cargo check` bundle we run most frequently.
* No longer use `--all-targets` which has known issues as explained in the rationale section of the CI script. Instead we duplicate the regular steps with `cargo-hack`.
* External dependencies are always built with `debug_assertions` set to `true` which allows for cache reuse. This is a trade-off. I think the cache reuse performance gains are worth it for regular CI runs. However some external dependency might e.g. sneak in a higher MSRV requirement into a `#[cfg(not(debug_assertions))]` block. Unlikely but possible. This is again where a less frequent `--release` check would help. Maybe scheduled, maybe just before publishing a new version. In any case, future work.

Additionally there is a new env variable `USING_DEBUG_ASSERTIONS` and a presence verification script `debug_assertions.sh`. These in combination make the CI script generic across workspaces regardless of `debug_assertions` usage. The env variable dictates whether we will spend the time doing the extra compiling.  The bash script will detect if the variable goes out of sync with actual usage in code.

### DIFF
--- a/.github/debug_assertions.sh
+++ b/.github/debug_assertions.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# Check all the standard Rust source files
+output=$(rg "debug_assertions" -g "*.rs" .)
+
+if [ -z "$output" ]; then
+	if [ "$USING_DEBUG_ASSERTIONS" = "true" ]; then
+		echo "Could not find any debug_assertions usage in Rust code."
+		echo "The CI script must be modified to not expect usage."
+		echo "Set USING_DEBUG_ASSERTIONS to false in .github/workflows/ci.yml."
+		exit 1
+	else
+		echo "Expected no debug_assertions usage in Rust code and found none."
+		exit 0
+	fi
+else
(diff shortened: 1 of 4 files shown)


DJMcNab (approved):
I thought I'd already approved this, but clearly not. Thanks

PR #753, opened 2024-11-18
Support transforms for each widget

Philipp-M:
Adds basic support for transforms (`kurbo::Affine`) of widgets in Masonry and Xilem, similarly as CSS transforms.
A new example `transforms` shows this in a playful way.
The pointer-intersection logic needed a core change to handle now possibly overlapping widgets.

Checkout [this](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Add.20.60Affine.60.20transform.20to.20.60Widget.60.20trait.3F) zulip thread for more info/discussion.


masonry/src/tree_arena.rs
```diff
@@ -268,6 +268,12 @@ impl<'a, Item> ArenaRefChildren<'a, Item> {
             .map(|child| child.arena_ref(self.id, self.parents_map.parents_map))
     }
 
+    pub fn iter(self) -> impl DoubleEndedIterator<Item = ArenaRef<'a, Item>> + ExactSizeIterator {
+        self.children
+            .iter()
+            .map(move |child| child.arena_ref(self.id, self.parents_map.parents_map))
+    }
```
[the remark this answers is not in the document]
Philipp-M: Yeah that's probably a good trade-off and at least avoids quadratic slowdown (which practically results in being completely "stuck" with big enough `N`).

But I think you mean a `HashMap`, i.e. something like `children: HashMap<u64, TreeNode<Item>>`?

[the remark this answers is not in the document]
PoignardAzur: Right, a hashmap.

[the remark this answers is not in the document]
Philipp-M: Ok, should this go into this PR? Or a separate PR, as I think it's somewhat orthogonal?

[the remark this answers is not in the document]
PoignardAzur: We'll see after #752 merges. Probably a separate PR. Either way, I'd prefer if this PR didn't add an `iter()` method.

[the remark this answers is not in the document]
Philipp-M: Sure, as I noted [here](https://github.com/linebender/xilem/pull/752#discussion_r1861927706) (and in zulip) that [this commit](https://github.com/linebender/xilem/pull/753/commits/14d1e6e366e5d834aaca155750b4a52fee9569f0) is temporary (a reason why I tried to isolate these changes, so that I can easily remove it again.

Alright then I'm waiting for the merger of #752 and implement it on top.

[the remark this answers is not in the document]
Philipp-M: Ok, good that you mention, otherwise I would've implemented that today or tomorrow.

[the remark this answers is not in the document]
PoignardAzur: See #772

Philipp-M:
I found a way to avoid the boilerplate on the masonry side, by using `WidgetState` and adding payload (just the transform for now) to the `WidgetPodInner`.
If we're having more of those custom global widget attributes, a structure of arrays for these attributes (separate arena?) might make sense (as well as a builder pattern for creating `WidgetPod`s).

But in this state I'm happy enough to merge this (after documentation and tests). As mentioned on zulip, avoiding even more boilerplate on the xilem level, is probably not reasonable, as either the API suffers (by using composition types, which doesn't allow further setting attributes of the underlying views after using a transform wrapper), or requires something like in xilem_web by only using composition types, which would be a big refactor (and a lot more complexity and code)...

Anyway to get a feeling for the added boilerplate on the xilem side, I've implemented transforms now for all views (and masonry also supports it for all widgets). I think it's not too bad and a good trade-off between complexity and boilerplate.

So apart of missing tests and docs, I think this is finished (but I might have missed things, so manual testing for now is appreciated).

masonry/src/passes/compose.rs
```diff
@@ -24,14 +24,21 @@ fn compose_widget(
...
-    if !parent_moved && !state.item.translation_changed && !state.item.needs_compose {
+    if !transformed && !state.item.needs_compose {
         return;
     }
 
+    let local_translation = state.item.translation + state.item.origin.to_vec2();
+
+    state.item.window_transform =
+        parent_window_transform * state.item.transform.then_translate(local_translation);
+    state.item.window_origin = state.item.window_transform.translation().to_point();
```
PoignardAzur, line 37: This code first applies the widget's transform, then the translation. Is that the right order? At first I was thinking it should be opposite, but I guess if the transform includes a rotation, you don't want the position vector to be rotated too.

We might want to add a comment to clarify this.

masonry/src/widget/button.rs
```diff
@@ -41,7 +41,7 @@ impl Button {
     ///
     /// let button = Button::new("Increment");
     /// ```
-    pub fn new(text: impl Into<ArcStr>) -> Button {
+    pub fn new(text: impl Into<ArcStr>) -> Self {
```
PoignardAzur, line 44: Not a dealbreaker, but I think this PR has a few too many drive-by fixes given its already large size.

masonry/src/widget/widget.rs
```diff
@@ -303,35 +324,36 @@ pub trait Widget: AsAny {
...
+            && widget.ctx.size().to_rect().contains(local_pos)
         {
-            return Some(child);
+            Some(*widget)
+        } else {
+            None
         }
+    } else {
+        None
     }
```
PoignardAzur, line 356: This is a core change in the PR, and it's one with a lot of semantic importance, so I'm giving it extra scrutiny. Some notes:

- This code has a lot of rightward drift that could be avoided with early returns.
- Again the difference between "declared layout rect" and "bounding rect of self and children" shows up. I think your code handles it correctly, but it does so implicitly. I think that distinction should be documented in a few places and given special attention in this function.
- I think `is_stashed` should be considered before the children. A stashed widget is never going to have un-stashed children. `accepts_pointer_interaction` is probably fine here, if we want to behave like `pointer-events: none` in the DOM, but that should be documented.

masonry/src/widget/widget_pod.rs
```diff
@@ -41,17 +46,29 @@ impl<W: Widget> WidgetPod<W> {
 
     /// Create a new widget pod with fixed id.
     pub fn new_with_id(inner: W, id: WidgetId) -> WidgetPod<W> {
+        Self::new_with_id_and_transform(inner, id, Affine::IDENTITY)
+    }
+
+    /// Create a new widget pod with a custom transform.
+    pub fn new_with_transform(inner: W, transform: Affine) -> WidgetPod<W> {
+        Self::new_with_id_and_transform(inner, WidgetId::next(), transform)
+    }
```
PoignardAzur, line 55: Okay, so I get why the PR does it this way, but I'm somewhat uncomfortable with adding a new constructor and all the special-casing that comes with it for a single attribute.

I don't know how else to do it, ideally it should be part of the styling refactor, but I don't want to block on that.

PoignardAzur (commented):
Sorry for the very late review.

This is not something anyone likes to hear, but I'm pretty confident this needs to be split into multiple PRs.

This PR essentially does three things:

- Replace the translation property with a general transform.
- Correct the pointer search algorithm to account for OOB sub-children.
- Add a way to set transform directly from creation.

I think all three of these changes would be worth their own PR. The second in particular feels like it needs some concepts (declared layout rect vs computed layout rect) to be reified and documented, or at least considered.

masonry/src/testing/harness.rs
```diff
@@ -418,7 +418,7 @@ impl TestHarness {
     ///
     /// Combines [`mouse_move`](Self::mouse_move), [`mouse_button_press`](Self::mouse_button_press), and [`mouse_button_release`](Self::mouse_button_release).
     pub fn mouse_click_on(&mut self, id: WidgetId) {
-        let widget_rect = self.get_widget(id).ctx().window_layout_rect();
+        let widget_rect = self.get_widget(id).ctx().bounding_rect();
```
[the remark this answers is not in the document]
Philipp-M: Whoops good catch, I think I just didn't pay too much attention to that part.

That's obviously wrong, as my intention was that the bounding_rect is in the window coordinate space. I've corrected this. Where are the other few places though?

*Maybe* `paint_rect` which TBH I think we should just remove for now, until we have a concrete use-case for it (e.g. damage regions).
It feels like we drag something along without having use for it currently, and potential to implement/adjust it wrongly without knowing if it's even how we want to do damage regions.
I guess this should otherwise be a (window coordinate) bounding box including the insets, ideally as an OOB in window coordinate space, but then again, it's not clear because it's not really used.

masonry/src/passes/compose.rs
Philipp-M: It's the right order, as you can see in the http_cats example (scroll) - at least when we want similar behavior as in CSS.
I've renamed `translation` in the last commit to `scroll_translation` so that it's more obvious what happens here, and I don't think `translation` is used in any other context, but I can of course revert that when there's arguments for it. Also added a comment.

masonry/src/widget/widget.rs
Philipp-M: > This code has a lot of rightward drift that could be avoided with early returns.

More than you have listed in the other points? I only see that `is_stashed` can be moved up (I've fixed this in the latest commit).

> Again the difference between "declared layout rect" and "bounding rect of self and children" shows up. I think your code handles it correctly, but it does so implicitly. I think that distinction should be documented in a few places and given special attention in this function.

Does it? the bounding rect is only used to filter out early as optimization, nothing else really. Otherwise I think it's documented what `bounding_rect` is, do you have more concrete suggestions?

> accepts_pointer_interaction is probably fine here, if we want to behave like pointer-events: none in the DOM, but that should be documented.

I agree that things should be more documented (and I've documented this in `Widget::find_widget_at_pos` now), but it's basically the same behavior as before with `get_child_at_pos` (which wasn't documented).

masonry/src/widget/widget_pod.rs
Philipp-M: Yes I don't think this should be a long term solution, what we could also do instead is something like this:

```rust
    /// Create a new widget pod with fixed id.
    pub fn new_with_id(inner: impl Into<CreatedWidget<W>>, id: WidgetId) -> WidgetPod<W> {
        Self::new_with_id_and_transform(inner, id, Affine::IDENTITY)
    }
```

and add the relevant `From` impls, like e.g. this:

```rust
impl<W: Widget> From<W> for CreatedWidget<W>{..}
impl<W: Widget> From<(W, Affine)> for CreatedWidget<W>{..}
```

for general attributes.
As mentioned somewhere else we could also introduce a builder type for creating widgets, maybe by just using `CreatedWidget` for this?

masonry/src/passes/compose.rs
```diff
@@ -68,9 +76,10 @@ fn compose_widget(
                 global_state,
                 widget,
                 state.reborrow_mut(),
-                moved,
-                translation,
+                transformed,
+                parent_transform,
             );
+            parent_state.bounding_rect = parent_state.bounding_rect.union(state.item.bounding_rect);
             parent_state.merge_up(state.item);
```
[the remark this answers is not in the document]
Philipp-M: You're right for at least the use case of intersecting pointer events and possibly doing some kind of damage regions (but I think this is basically the use case of bounding rects/boxes here). I've used an intersection of the (transformed) clip path and the child bounding rect instead here. There's for sure further optimization potential, but I think that should be probably done in follow-up PRs when it's more clear what shape clip_path will be.

Philipp-M:
> This is not something anyone likes to hear, but I'm pretty confident this needs to be split into multiple PRs.

I'm actually not so confident about that... If the PR would be much more complex, and easier to split into changes that could stand for themselves, I would've done so, but as it stands a lot of the code is intertwined, a few parts may or may not be isolated, but then I'm not really sure whether it's worth it to split these up. E.g. transforms without adjusted pointer-intersection is not correct when isolated.

I still need to ideally write some tests (any ideas what may be good?), probably document further, and revert at least the http_cats example (but right now it's good to see that scrolling is handled correctly).

PoignardAzur:
> I would've done so, but as it stands a lot of the code is intertwined, a few parts may or may not be isolated, but then I'm not really sure whether it's worth it to split these up.

I think you under-estimated how feasible it is to split that code up. Changes touches on the same areas of code, yes, but they're not semantically intertwined.

> E.g. transforms without adjusted pointer-intersection is not correct when isolated.

Pointer intersection is already incorrect with translation-only transforms. There's nothing tying adjusted pointer intersection and arbitrary transforms together.

masonry/src/widget/button.rs
Philipp-M: Was fixed on main in the meantime, I don't think there's any other drive-by fixes though?

xilem/examples/mason.rs
Philipp-M, line 1: I'm not sure whether I should revert the changes here, as I don't think they are particularly clean (rather a quick'n dirty hack to see/test whether transforms react to changes etc.).
On the other hand there's at least some eye-candy here (and a basic manual animated test for transforms).

xilem/src/view/mod.rs
```diff
@@ -50,3 +50,42 @@ pub use portal::*;
...
+
+    #[must_use]
+    fn translate(mut self, v: impl Into<crate::Vec2>) -> Self {
+        let transform = self.transform_mut();
+        *transform = transform.then_translate(v.into());
+        self
+    }
+
+    #[must_use]
+    fn transform(mut self, v: impl Into<crate::Affine>) -> Self {
```
Philipp-M, line 87: The naming here is a little bit unfortunate, as there's already `transform_mut`, I'm not sure of a better name for setting a general transform.

Philipp-M (commented):
I think this is now in a state that is mergeable/ready for review.

There could be more tests for sure, but I as noted elsewhere, I don't have much time, and I think it probably makes sense to not delay this too long (to avoid merge-conflicts etc.).

masonry/src/passes/accessibility.rs
```diff
@@ -91,7 +91,7 @@ fn build_accessibility_tree(
 // --- MARK: BUILD NODE ---
 fn build_access_node(widget: &mut dyn Widget, ctx: &mut AccessCtx) -> Node {
     let mut node = Node::new(widget.accessibility_role());
-    node.set_bounds(to_accesskit_rect(ctx.widget_state.window_layout_rect()));
```
PoignardAzur, line 94: I don't think this is correct. This should be the "layout" rect, not the bounding rect.

PoignardAzur (changes requested):
The changes to the Mason example are extremely distracting and need to be removed.

This and the `set_bounds` issue aside, there are a lot of nitpicks I could make, but you're right that it's probably better to expedite this PR. I'll address my concerns in follow-up PRs.

masonry/src/passes/accessibility.rs
Philipp-M: Hmm, reading the docs of accesskit, I think there's more to it, as the node can be transformed. I've pushed (untested) code how I think it needs to be handled according to the docs, I currently don't have a setup to quickly evaluate this. Maybe you can check again? I'm not sure though whether the scroll translation should be applied though...

Philipp-M:
> The changes to the Mason example are extremely distracting and need to be removed.

I guess that answers https://github.com/linebender/xilem/pull/753#discussion_r1912111418

> I'll address my concerns in follow-up PRs.

Yeah, I think that saves us both time (as IME writing and acting on review comments takes more time than just implementing the desired changes).
You can of course also push onto this branch.

PoignardAzur (approved):

Philipp-M:
Should I merge, or does anyone have some last-minute comments/suggestions?

PoignardAzur:
I think it's fine to merge.

PR #754, opened 2024-11-19
Unify Prose and Textbox with an underlying `TextRegion`

DJMcNab:
Remaining steps

- [x] Update Textbox to use this
- [x] Add internal padding

Follow-up work:
- [ ] Restore IME support
- [ ] Upstream new result of PlainEditor (once other steps are complete)


masonry/src/widget/sized_box.rs
```diff
@@ -82,6 +82,20 @@ impl Padding {
     /// A padding of zero for all edges.
     pub const ZERO: Padding = Padding::all(0.);
 
+    /// An padding which can be detected using, allowing parent views to determine if the padding is under user control.
```
jaredoconnell, line 85: It looks like a word is missing after "using".

jaredoconnell (approved):
It's always nice to reduce redundancy. I left a comment regarding a comment. I tested the relevant examples and didn't notice any new regressions.

This PR fixes tabbing between the two textboxes in the two_textbox exapmle, but if you have a selection it will remove it when you tab out of it. So that will need to be fixed in this or a future PR.

masonry/src/widget/sized_box.rs
waywardmonkeys: Also, "A padding" rather than "An ..."

waywardmonkeys:
When you renamed `TextRegion` to `TextArea`, a lot of things are still referring to the region terminology which makes it confusing.

Also, `TextArea` is kind of confusing since it makes think of the HTML tag `<textarea>` ... but don't have any real thoughts about if that's an issue or not.

DJMcNab:
> When you renamed `TextRegion` to `TextArea`, a lot of things are still referring to the region terminology which makes it confusing.

As mentioned on the Zulip, I'm still doing a pass over docs.

> Also, `TextArea` is kind of confusing since it makes think of the HTML tag `<textarea>` ... but don't have any real thoughts about if that's an issue or not.

I don't have a strong feeling about any particular name; this needing a name is largely an accident. Maybe we should prefer "TextCore" or something?

waywardmonkeys:
> > When you renamed `TextRegion` to `TextArea`, a lot of things are still referring to the region terminology which makes it confusing.
> 
> As mentioned on the Zulip, I'm still doing a pass over docs.

I meant in identifiers, not in docs:

```
                let mut text_region = Textbox::text_mut(&mut textbox);
                TextArea::reset_text(&mut text_region, "");
```

masonry/src/widget/prose.rs
```diff
@@ -1,630 +1,156 @@
...
-        if let StyleProperty::Brush(idx @ BrushIndex(1..)) = &property {
-            debug_panic!(
-                "Can't set a non-zero brush index ({idx:?}) on a `Label`, as it only supports global styling."
-            );
-        }
-        self.styles.insert(property)
+    /// Read the underlying text area. Useful for getting its ID.
+    // This is a bit of a hack, to work around `from_text_area_pod` not being
+    // able to set padding.
+    pub fn region_pod(&self) -> &WidgetPod<TextArea<false>> {
```
tomcur, line 84: Probably should be `pub fn text_area_pod`

tomcur (approved):
This looks good. I have only spot-checked the implementation, but the direction makes sense. The examples seem to work as expected. I've only tested LTR strings, because IME and copying into the application don't work yet, but I expect that would only uncover Parley issues. I haven't tested accessibility.

This makes an issue in Xilem's "mason" example a bit more obvious, where the Prose widget is not large enough. In this PR the bottom line of text is clipped.

DJMcNab:
Yeah, I believe that is https://github.com/linebender/parley/issues/165. I've added a workaround into `Prose` directly. Note that this issue is still visible in http_cats; I've added a workaround into http_cats itself for that one.

PR #756, opened 2024-11-22
Default to a line height of 1.2

DJMcNab:
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal


masonry/src/widget/screenshots/masonry__widget__textbox__tests__textbox_outline.png
DJMcNab, line 1: This change is surprising to me.

jaredoconnell: This moved in the opposite direction compared to the rest of the items in this PR. I wonder if this has anything to do with the alignment handling within textbox. Maybe padding only being counted once instead of both top and bottom?

dfrg (approved):
Parley should definitely default to something more reasonable but since we’re effectively copying web behavior there, it makes sense to also copy the default behavior of user agents here.

nicoburns (approved):

masonry/src/widget/screenshots/masonry__widget__textbox__tests__textbox_outline.png
DJMcNab: Looking at it again, it looks like the top line in all of the examples dropped, so I think this is fine.

PR #757, opened 2024-11-23
xilem_web: Add MemoizedStream view

flosse:
The `MemoizedStream` view is is basically a `MemoizedAwait` view that accepts a stream and invokes the callback for each stream item received.
Since the implementation is almost the same, the two views are just wrappers for the private `MemoizedInner` struct.
If the approach is okay, I can also add an example.

### DIFF
--- a/xilem_web/src/concurrent/memoized_await.rs
+++ b/xilem_web/src/concurrent/memoized_await.rs
@@ -1,16 +1,30 @@
 // Copyright 2024 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use std::{fmt, future::Future, marker::PhantomData};
+
+use futures::{Stream, StreamExt};
+use wasm_bindgen::{closure::Closure, JsCast, UnwrapThrowExt};
+use wasm_bindgen_futures::spawn_local;
+
 use crate::{
     core::{MessageResult, Mut, NoElement, View, ViewId, ViewMarker, ViewPathTracker},
     DynMessage, OptionalAction, ViewCtx,
 };
-use std::{future::Future, marker::PhantomData};
-use wasm_bindgen::{closure::Closure, JsCast, UnwrapThrowExt};
-use wasm_bindgen_futures::spawn_local;
 
-/// Await a future returned by `init_future` invoked with the argument `data`, `callback` is called with the output of the future. `init_future` will be invoked again, when `data` changes. Use [`memoized_await`] for construction of this [`View`]
-pub struct MemoizedAwait<State, Action, OA, InitFuture, Data, Callback, F, FOut> {
+/// Await a future returned by `init_future` invoked with the argument `data`, `callback` is called with the output of the future.
+/// `init_future` will be invoked again, when `data` changes. Use [`memoized_await`] for construction of this [`View`]
+pub struct MemoizedAwait<State, Action, OA, InitFuture, Data, Callback, F, FOut>(
+    MemoizedInner<State, Action, OA, InitFuture, Data, Callback, F, FOut>,
+);
+
+/// Await a stream returned by `init_stream` invoked with the argument `data`, `callback` is called with the items of the stream.
+/// `init_stream` will be invoked again, when `data` changes. Use [`memoized_stream`] for construction of this [`View`]
+pub struct MemoizedStream<State, Action, OA, InitStream, Data, Callback, F, StreamItem>(
+    MemoizedInner<State, Action, OA, InitStream, Data, Callback, F, StreamItem>,
+);
+
+struct MemoizedInner<State, Action, OA, InitFuture, Data, Callback, F, FOut> {
     init_future: InitFuture,
     data: Data,
     callback: Callback,
@@ -20,9 +34,26 @@ pub struct MemoizedAwait<State, Action, OA, InitFuture, Data, Callback, F, FOut>
 }
 
 impl<State, Action, OA, InitFuture, Data, Callback, F, FOut>
+    MemoizedInner<State, Action, OA, InitFuture, Data, Callback, F, FOut>
+where
+    FOut: fmt::Debug + 'static,
+    InitFuture: Fn(&Data) -> F,
+{
+    fn debounce_ms(mut self, milliseconds: usize) -> Self {
+        self.debounce_ms = milliseconds;
+        self
+    }
+
+    fn reset_debounce_on_update(mut self, reset: bool) -> Self {
+        self.reset_debounce_on_update = reset;
+        self
+    }
+}
+
+impl<State, Action, OA, InitFuture, Data, Callback, F, FOut>
     MemoizedAwait<State, Action, OA, InitFuture, Data, Callback, F, FOut>
 where
-    FOut: std::fmt::Debug + 'static,
+    FOut: fmt::Debug + 'static,
     F: Future<Output = FOut> + 'static,
     InitFuture: Fn(&Data) -> F,
 {
@@ -31,7 +62,7 @@ where
     ///
     /// The default for this is `0`
     pub fn debounce_ms(mut self, milliseconds: usize) -> Self {
-        self.debounce_ms = milliseconds;
+        self.0 = self.0.debounce_ms(milliseconds);
         self
     }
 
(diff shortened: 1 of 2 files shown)


xilem_web/Cargo.toml
```diff
@@ -181,3 +181,7 @@ features = [
 default = []
 # This interns some often used strings, such as element tags ("div" etc.), which slightly improves performance when creating elements at the cost of a bigger wasm binary
 intern_strings = ["wasm-bindgen/enable-interning"]
+
+[dev-dependencies]
+async-stream = "0.3.6"
+gloo-timers = { version = "0.3.0", features = ["futures"] }
```
flosse, line 187: I need to add these dependencies so that the doc test does not fail, but then clippy complains that these dependencies are not used. :thinking:

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -1,16 +1,30 @@
...
+    MemoizedInner<State, Action, OA, InitFuture, Data, Callback, F, FOut>,
+);
+
+/// Await a stream returned by `init_stream` invoked with the argument `data`, `callback` is called with the items of the stream.
+/// `init_stream` will be invoked again, when `data` changes. Use [`memoized_stream`] for construction of this [`View`]
+pub struct MemoizedStream<State, Action, OA, InitStream, Data, Callback, F, StreamItem>(
+    MemoizedInner<State, Action, OA, InitStream, Data, Callback, F, StreamItem>,
+);
+
+struct MemoizedInner<State, Action, OA, InitFuture, Data, Callback, F, FOut> {
```
Philipp-M, line 27: Maybe:

```suggestion
struct MemoizedFuture<State, Action, OA, InitFuture, Data, Callback, F, FOut> {
```

(but no strong feelings of the name, there are probably also better ones)

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -19,10 +33,27 @@ pub struct MemoizedAwait<State, Action, OA, InitFuture, Data, Callback, F, FOut>
...
+        self.debounce_ms = milliseconds;
+        self
+    }
+
+    fn reset_debounce_on_update(mut self, reset: bool) -> Self {
+        self.reset_debounce_on_update = reset;
+        self
+    }
+}
+
```
Philipp-M, line 52: I think we can omit this, I think it's just boilerplate without bringing value, but no strong feelings here either.

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -84,20 +169,78 @@ where
...
         reset_debounce_on_update: true,
         phantom: PhantomData,
-    }
+    })
+}
+
+/// Await a stream returned by `init_stream` invoked with the argument `data`, `callback` is called with the items of the stream.
+/// `init_stream` will be invoked again, when `data` changes.
+///
+/// The update behavior can be controlled, by [`debounce_ms`](`MemoizedStream::debounce_ms`) and [`reset_debounce_on_update`](`MemoizedStream::reset_debounce_on_update`)
```
Philipp-M, line 191:
```suggestion
/// The behavior of the `init_stream` invocation by changes to `data` can be customized by [`debounce_ms`](`MemoizedStream::debounce_ms`) and [`reset_debounce_on_update`](`MemoizedStream::reset_debounce_on_update`).
```

(maybe also something like this for `MemoizedAwait` as I think it's not entirely clear what I meant with "update behavior").

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -84,20 +169,78 @@ where
...
+///         html::div(format!("{state:?}")),
+///         memoized_stream(
+///             10,
+///             |n| {
+///                 let range = 0..*n;
+///                 stream! {
+///                     for i in range {
+///                         TimeoutFuture::new(500).await;
+///                          yield i;
+///                      }
```
Philipp-M, line 211:
```suggestion
///                         yield i;
///                     }
```

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -84,20 +169,78 @@ where
...
+///                 }
+///             },
+///             |state: &mut Vec<usize>, item: usize| {
+///                 state.push(item);
+///             },
+///         ),
+///     )
+/// }
+/// ```
```
Philipp-M, line 220: :+1:  on doc-comment.

I wonder whether it makes sense to have something like I've done [here](https://github.com/Philipp-M/trui/blob/dd24a2921eb0587dfffbbafea4a3e62b92f39c43/examples/async_event_handler_stream.rs#L53-L60). To know in what the state of the stream is.

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -148,7 +291,7 @@ impl MemoizedAwaitState {
 }
 
 #[derive(Debug)]
-enum MemoizedAwaitMessage<Output: std::fmt::Debug> {
+enum MemoizedAwaitMessage<Output: fmt::Debug> {
```
Philipp-M, line 294: Maybe a different name here as well, as it's not just for the `MemoizedAwait` anymore?

Philipp-M (approved):
I added a few comments, no real blockers though, thanks!

(Ideally we would also have an example showing a use-case for this, but I don't want to block on that...)

xilem_web/src/concurrent/memoized_await.rs
flosse: I use `StreamMessage::Started` here instead of `StreamMessage::Started(T)` 
because there can be a big difference between the start of the stream and the reception of the first element.
Does that make sense?

xilem_web/Cargo.toml
flosse: I just marked it with `ignore`

xilem_web/src/concurrent/memoized_await.rs
flosse: I've just realized that it's easier and cleaner if you leave the responsibility to the user. For example you might want to add some payload to the `Started` variant. And even existing streams could be wrapped.

flosse:
> (Ideally we would also have an example showing a use-case for this, but I don't want to block on that...)

Of course, that will come.
Unfortunately, this week looks very full for me, but apart from myself, probably nobody is waiting at the moment :wink:

xilem_web/web_examples/streams/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+    current_search: Option<AbortHandle>,
+}
+
+const DEBOUNCE_MILLIS: usize = 500;
+
+fn app_logic(state: &mut AppState) -> impl Element<AppState> {
+    log::debug!("Run app logic");
+
+    let search_stream = memoized_stream(
+        (state.search_term.clone(), state.db.clone()),
```
flosse, line 30: This is a bit messy, because the first argument (`state.search_term`) is what should be monitored,
the second (`state.db`) is only there to make it available for stream generation (in `create_search_stream`).

flosse: @Philipp-M 
I have now changed the implementation so that the app state is injected again before the future is initialized.
This has two advantages:
1. Only the data that is to be monitored for changes needs to implement `PartialEq` and  `Clone`
2. the point in time at which the future is created is now easier to grasp.

The API now looks like this:

```rust
memoized_await(
    50, // data to watch
    |state, watched_data| {
        state.async_call_started = true; // before it was debounced
        my_aync_call(state.non_watched_data, watched_data)
    },
    |state, result| { / * ... */ }
)

```

Philipp-M (commented):
Thanks for the example!

I hope it's not too much to ask for and I think we can also use this and iterate in later PRs.
Though I would've thought of something that's more of a real-world use-case (and possibly use html instead of console log, as this more obviously shows what the example is about).
Maybe with inspiration from something like [this](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API#examples), with e.g. something like this: https://docs.rs/wasm-streams/latest/wasm_streams/readable/struct.ReadableStream.html#method.into_stream

xilem_web/src/concurrent/memoized_await.rs
```diff
@@ -84,23 +146,80 @@ where
     State: 'static,
     Action: 'static,
     Data: PartialEq + 'static,
-    FOut: std::fmt::Debug + 'static,
+    FOut: fmt::Debug + 'static,
     F: Future<Output = FOut> + 'static,
-    InitFuture: Fn(&Data) -> F + 'static,
+    InitFuture: Fn(&mut State, &Data) -> F + 'static,
```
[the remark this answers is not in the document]
flosse: it's not a final solution, but i changed it now at least to a immutable reference:
```diff
-InitFuture: Fn(&mut State, &Data) -> F + 'static,
+InitFuture: Fn(&State, &Data) -> F + 'static,
```

PR #758, opened 2024-11-23
Update to Linebender lint set v2.

xStrom:
Some examples needed minor adjustments and `xilem_web` needed an exception.

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -1,14 +1,16 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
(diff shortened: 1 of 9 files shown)


xilem/src/lib.rs
```diff
@@ -1,14 +1,16 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(clippy::trivially_copy_pass_by_ref)]
-// LINEBENDER LINT SET - v1
+//! An experimental Rust native UI framework.
```
DJMcNab, line 4: Oops :sweat_smile: (I should have written something here before)

xStrom: I just copy-pasted the GitHub description field. Feel free to update both. 🚀

PR #762, opened 2024-11-28
Re-add IME support

tomcur:
This adds IME support back into Masonry. This sticks close to https://github.com/linebender/parley/pull/111, except that during IME compose, this version doesn't allow changing the selection anchor, making the code simpler. For reference, there's also https://github.com/linebender/parley/pull/136.

This tweaks the focus update pass: when a widget with IME is unfocused, Masonry sends the platform's IME disable event to the newly focused widget. As a workaround, we synthesize an IME disable event in the focus pass and send it to the widget that is about to be unfocused. A complication is that the handling of that event can request focus to a different widget, and in particular, can request itself to be focused again. This handles that case, too.

Remaining work is setting the IME candidate region to be near the current selection and to make a decision on cursor/selection hiding when the platform sends a `None` cursor.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -405,13 +405,13 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
     let prev_focused = root.global_state.focused_widget;
     let was_ime_active = root.global_state.is_ime_active;
 
-    let synthesize_ime_disabled =
-        was_ime_active && prev_focused != root.global_state.next_focused_widget;
-    if synthesize_ime_disabled {
-        // IME was active, but the next focused widget is going to receive the
-        // Ime::Disabled event. Synthesize an Ime::Disabled event here and send it to
-        // the widget about to be unfocused.
+    let disable_ime = was_ime_active && prev_focused != root.global_state.next_focused_widget;
+    if disable_ime {
+        // IME was active, but the next focused widget is going to receive the Ime::Disabled event
+        // sent by the platform. Synthesize an Ime::Disabled event here and send it to the widget
+        // about to be unfocused.
         run_on_text_event_pass(root, &TextEvent::Ime(winit::event::Ime::Disabled));
+        root.global_state.emit_signal(RenderRootSignal::EndIme);
     }
 
     // Note: handling of the Ime::Disabled event sent above may have changed the next focused
@@ -475,14 +475,7 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
     // Refocus if the focused widget changed. Or, if the focused widget was going to change, the
     // handling of the synthesized Ime::Disabled event above may have changed the focus back to the
     // original widget. In that case, also refocus. This is messy.
-    if prev_focused != next_focused || synthesize_ime_disabled {
-        let is_ime_active = if let Some(id) = next_focused {
-            root.widget_arena.get_state(id).item.accepts_text_input
-        } else {
-            false
-        };
-        root.global_state.is_ime_active = is_ime_active;
-
+    if prev_focused != next_focused || disable_ime {
         // We send FocusChange event to widget that lost and the widget that gained focus.
         // We also request accessibility, because build_access_node() depends on the focus state.
         if let Some(prev_focused) = prev_focused {
@@ -501,19 +494,20 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
                 ctx.widget_state.request_accessibility = true;
                 ctx.widget_state.needs_accessibility = true;
             });
-        }
 
-        if prev_focused.is_some() && was_ime_active {
-            root.global_state.emit_signal(RenderRootSignal::EndIme);
-        }
-        if next_focused.is_some() && is_ime_active {
-            root.global_state.emit_signal(RenderRootSignal::StartIme);
-        }
+            let widget_state = root.widget_arena.get_state(next_focused).item;
 
-        if let Some(id) = next_focused {
-            let ime_area = root.widget_arena.get_state(id).item.get_ime_area();
+            root.global_state.is_ime_active = widget_state.accepts_text_input;
+            if widget_state.accepts_text_input {
+                root.global_state.emit_signal(RenderRootSignal::StartIme);
+            }
+
             root.global_state
-                .emit_signal(RenderRootSignal::new_ime_moved_signal(ime_area));
+                .emit_signal(RenderRootSignal::new_ime_moved_signal(
+                    widget_state.get_ime_area(),
+                ));
+        } else {
+            root.global_state.is_ime_active = false;
         }
     }
 
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -411,7 +411,18 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
         // sent by the platform. Synthesize an Ime::Disabled event here and send it to the widget
         // about to be unfocused.
         run_on_text_event_pass(root, &TextEvent::Ime(winit::event::Ime::Disabled));
+        // Disable the IME, which was enabled specifically for this widget.
+        // Note that if the newly focused widget also requires IME, we will request it 
+        // again - this resets the platform's state, ensuring that partial IME 
+        // inputs do not "travel" between widgets
         root.global_state.emit_signal(RenderRootSignal::EndIme);
+        if prev_focused == root.global_state.next_focused_widget {
+             tracing::warn!(id = prev_focused.map(|id| id.trace()), "request_focus called whilst handling Ime::Disabled");
+             // In this unlikely case, the rest of this handler will short-circuit, and IME will never be re-enabled 
+             // for this widget. The resultant `ImeEnabled` event will be routed to this widget as it is the focused
+             // widget. We don't handle this as above to avoid loops
+            root.global_state.emit_signal(RenderRootSignal::StartIme);
+        }
     }
 
     // Note: handling of the Ime::Disabled event sent above may have changed the next focused
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -6,7 +6,7 @@ use std::collections::HashSet;
 use cursor_icon::CursorIcon;
 use tracing::{info_span, trace};
 
-use crate::passes::event::run_on_pointer_event_pass;
+use crate::passes::event::{run_on_pointer_event_pass, run_on_text_event_pass};
 use crate::passes::{enter_span, enter_span_if, merge_state_up, recurse_on_children};
 use crate::render_root::{RenderRoot, RenderRootSignal, RenderRootState};
 use crate::tree_arena::ArenaMut;
@@ -14,8 +14,6 @@ use crate::{
     PointerEvent, QueryCtx, RegisterCtx, TextEvent, Update, UpdateCtx, Widget, WidgetId,
     WidgetState,
 };
-
-use super::event::run_on_text_event_pass;
 
 // --- MARK: HELPERS ---
 fn get_id_path(root: &RenderRoot, widget_id: Option<WidgetId>) -> Vec<WidgetId> {
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -394,6 +394,13 @@ pub(crate) fn run_update_focus_chain_pass(root: &mut RenderRoot) {
 // --- MARK: UPDATE FOCUS ---
 pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
     let _span = info_span!("update_focus").entered();
+    // If the next-focused widget is disabled, stashed or removed, we set
+    // the focused id to None
+    if let Some(id) = root.global_state.next_focused_widget {
+        if !root.is_still_interactive(id) {
+            root.global_state.next_focused_widget = None;
+        }
+    }
 
     let prev_focused = root.global_state.focused_widget;
     let was_ime_active = root.global_state.is_ime_active;
@@ -417,6 +424,13 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
         // re-enabled for this widget. Re-enable IME here; the resultant `Ime::Enabled` event sent
         // by the platform will be routed to this widget as it remains the focused widget. We don't
         // handle this as above to avoid loops.
+        //
+        // First do the disabled, stashed or removed check again.
+        if let Some(id) = root.global_state.next_focused_widget {
+            if !root.is_still_interactive(id) {
+                root.global_state.next_focused_widget = None;
+            }
+        }
         if prev_focused == root.global_state.next_focused_widget {
             tracing::warn!(
                 id = prev_focused.map(|id| id.trace()),
@@ -426,13 +440,6 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
         }
     }
 
-    // If the focused widget is disabled, stashed or removed, we set
-    // the focused id to None
-    if let Some(id) = root.global_state.next_focused_widget {
-        if !root.is_still_interactive(id) {
-            root.global_state.next_focused_widget = None;
-        }
-    }
     let next_focused = root.global_state.next_focused_widget;
 
     // "Focused path" means the focused widget, and all its parents.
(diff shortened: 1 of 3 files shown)


masonry/src/text/editor.rs
```diff
@@ -517,14 +581,10 @@ where
...
         // TODO: Do we need to be non-dirty?
         // FIXME: `Selection` should make this easier
-        if index >= self.buffer.len() {
-            Cursor::from_index(
-                &self.layout,
-                self.buffer.len().saturating_sub(1),
-                Affinity::Upstream,
-            )
+        if index > self.buffer.len() {
+            Cursor::from_byte_index(&self.layout, self.buffer.len(), Affinity::Upstream)
```
tomcur, line 585: This may be a current Parley bug.

masonry/src/widget/text_area.rs
```diff
@@ -734,8 +788,17 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+                // event is routed to the newly focused widget. Do IME clean up here.
+                if !focused && self.editor.is_composing() {
+                    let (fctx, lctx) = ctx.text_contexts();
+                    self.editor.transact(fctx, lctx, |txn| txn.clear_compose());
+                    self.rendered_generation = self.editor.generation();
+                    ctx.request_layout();
+                } else {
+                    ctx.request_render();
+                }
             }
```
tomcur, line 802: This is a bit unfortunate.

DJMcNab: I think sending `Ime::Disabled` to the newly unfocused widget in the focus update pass would be a reasonable solution here. It does mean that the new widget gets `Disabled, Enabled`. But that's probably fine?

masonry/src/text/editor.rs
```diff
@@ -674,6 +729,17 @@ where
...
+            // TODO: underline currently doesn't show up, maybe the brush is invisible?
+            builder.push(
+                parley::style::StyleProperty::UnderlineBrush(Some(T::default())),
+                preedit_range.clone(),
+            );
+            builder.push(
+                parley::style::StyleProperty::Underline(true),
+                preedit_range.clone(),
+            );
+        }
```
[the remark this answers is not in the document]
DJMcNab: I think that would probably be idiomatic. It does mean that it becomes meaningful to have multiple brushes, even if most of the time you'd want it to be the same.

A rainbow gradient for the composing text would be pretty funny...

masonry/src/text/editor.rs
tomcur: Fixed here and upstream https://github.com/linebender/parley/pull/187.

masonry/src/text/editor.rs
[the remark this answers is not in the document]
DJMcNab: Thinking about this again, I'm inclined to leave this as out-of-scope for this PR. This also gets onto the question of underline and strikethrough brushes, which is extra added complication.

masonry/src/widget/text_area.rs
```diff
@@ -709,7 +713,50 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+                            {
+                                // The IME has started composing. Delete the current selection and
+                                // send a TextChange event with the selection removed, but without
+                                // the composing preedit text.
+                                self.editor.transact(fctx, lctx, |txn| {
+                                    txn.delete_selection();
+                                });
+                                submit_text = Some(self.text().to_string());
+                            }
+
```
tomcur, line 738: Note `PlainEditorTxn::set_compose` also removes the selection from the buffer when the IME newly starts composing (i.e., the first compose after IME is enabled or after a `clear_compose`). That seemed like a useful abstraction for Parley. The Masonry code here as a special-case also removes the selection, because Masonry might not want to send preedit text in TextChange actions, but does want to notify about the removed selection.

masonry/src/widget/text_area.rs
```diff
@@ -709,7 +713,50 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+                            if !self.editor.is_composing() && self.editor.selected_text().is_some()
+                            {
+                                // The IME has started composing. Delete the current selection and
+                                // send a TextChange event with the selection removed, but without
+                                // the composing preedit text.
+                                self.editor.transact(fctx, lctx, |txn| {
+                                    txn.delete_selection();
+                                });
+                                submit_text = Some(self.text().to_string());
+                            }
```
tomcur, line 737: Note `PlainEditorTxn::set_compose` also removes the selection from the buffer when the IME newly starts composing (i.e., the first compose after or after a `clear_compose`). That seemed like a useful abstraction for Parley. The Masonry code here as a special-case also removes the selection, because Masonry might not want to send preedit text in `TextChange` actions, but does want to notify about the removed selection.

masonry/src/widget/text_area.rs
```diff
@@ -709,7 +713,50 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+                            self.editor
+                                .transact(fctx, lctx, |txn| txn.set_compose(text, *cursor));
+                        }
+                    }
+                    winit::event::Ime::Commit(text) => {
+                        self.editor
+                            .transact(fctx, lctx, |txn| txn.insert_or_replace_selection(text));
+                        submit_text = Some(self.text().to_string());
+                    }
+                    _ => {}
```
DJMcNab, line 748:
```suggestion
                    winit::event::Ime::Enabled => {}
```
We know we don't need to handle `Enabled`, but we might need to handle some other event kinds.

masonry/src/widget/text_area.rs
```diff
@@ -720,12 +767,15 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
-        // TODO: Implement IME, then flip back to EDITABLE.
-        false
+        EDITABLE
     }
 
     fn on_access_event(&mut self, ctx: &mut EventCtx, event: &AccessEvent) {
         if event.action == accesskit::Action::SetTextSelection {
+            if self.editor.is_composing() {
+                return;
+            }
```
DJMcNab, line 777: We should check with @mwcampbell if this is right. The scenario is:
1) The user is inputting some text which uses a pre-edit through their IME
2) Through their accessibility tool, they have tried to select some text, without their IME finishing the input

What should we do in that case? This code just ignores the second request, which is the easiest method to reason about. The alternatives are quite messy.

masonry/src/widget/text_area.rs
DJMcNab: I still think this would be the right approach. It does make our pass code slightly more complicated, but makes the IME handling much more consistent.

masonry/src/widget/text_area.rs
```diff
@@ -738,8 +788,17 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
 
     fn update(&mut self, ctx: &mut UpdateCtx, event: &Update) {
         match event {
-            Update::FocusChanged(_) => {
-                ctx.request_render();
+            Update::FocusChanged(focused) => {
+                // HACK: currently, when moving focus away from a text area, the Ime::Disabled
+                // event is routed to the newly focused widget. Do IME clean up here.
+                if !focused && self.editor.is_composing() {
```
DJMcNab, line 794: As a code style thing, I'd probably write this as an:
```rust
FocusChanged(false) if self.editor.is_composing() => {
```
arm. But I think this code should go anyway, so that's moot.

DJMcNab (approved):
This works really well for me. It does show that underlines for an emoji is in the wrong place, but that's not the fault of this PR.

Some minor points to be considered. I don't think there's anything blocking, but we should also check-in with Matt.

masonry/src/text/editor.rs
```diff
@@ -214,9 +230,75 @@ where
...
+
+        if let Some(cursor) = cursor {
+            // Select the location indicated by the IME.
+            self.editor.set_selection(Selection::new(
+                self.editor.cursor_at(start + cursor.0),
+                self.editor.cursor_at(start + cursor.1),
+            ));
+        } else {
+            // IME indicates nothing is to be selected: collapse the selection to a
+            // caret just in front of the preedit.
```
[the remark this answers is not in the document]
tomcur: It was more a less an arbitrary choice I lifted from https://github.com/linebender/parley/pull/111, but on my platform, Firefox, Chrome and LibreOffice Writer all place the caret in front of the preedit text (and they appear to never make a spanning selection).

Firefox:

20241129_14h10m32s_grim

masonry/src/widget/text_area.rs
```diff
@@ -508,7 +512,7 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
     fn on_text_event(&mut self, ctx: &mut EventCtx, event: &TextEvent) {
         match event {
             TextEvent::KeyboardKey(key_event, modifiers_state) => {
-                if !key_event.state.is_pressed() {
+                if !key_event.state.is_pressed() || self.editor.is_composing() {
```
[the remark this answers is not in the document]
tomcur: On my platform (Wayland through wlroots), at least clipboard actions get sent while composing, perhaps more.

masonry/src/widget/text_area.rs
tomcur: Gone. :wind_face:

masonry/src/widget/text_area.rs
tomcur: Yes, I think this is better. The focus update pass now sends a synthesized `Ime::Disabled` event (if IME was active) just prior to switching focus to the next focused widget.

masonry/src/text/editor.rs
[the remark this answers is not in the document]
DJMcNab: Hmm. Is that hitting this condition where the IME doesn't provide a preedit cursor?

masonry/src/text/editor.rs
```diff
@@ -214,9 +230,75 @@ where
...
+    ///
+    /// This removes the IME preedit text.
+    pub fn clear_compose(&mut self) {
+        if let Some(preedit_range) = self.editor.compose.clone() {
+            self.editor.buffer.replace_range(preedit_range.clone(), "");
+            self.editor.compose = None;
+            self.update_layout();
+
+            self.editor
+                .set_selection(self.editor.cursor_at(preedit_range.start).into());
```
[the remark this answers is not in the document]
tomcur: `PlainEditor::layout_dirty` gets us part of the way there, but as you mentioned, we can't set the next selection until the layout is updated (it is forced to fit within the current layout). It might work if we have some `PlainEditor::defer_set_selection` or something similar, that simply waits with setting the selection until inside the `update_layout` method.

masonry/src/text/editor.rs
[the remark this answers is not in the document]
tomcur: (I haven't figured out steps yet with this IME that sends a `cursor: None`.)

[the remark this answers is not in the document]
DJMcNab: I've done a little bit more research into this, and it seems like Winit gets into a borked state. Any choice seems reasonable here. Alternatively, maybe we should just not have a visible selection?

masonry/src/passes/update.rs
```diff
@@ -471,6 +474,12 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
         // We also request accessibility, because build_access_node() depends on the focus state.
         if let Some(prev_focused) = prev_focused {
             if root.widget_arena.has(prev_focused) {
+                if was_ime_active {
+                    // IME was active, but the next focused widget is going to receive the
+                    // Ime::Disabled event. Synthesize an Ime::Disabled event here and send it to
+                    // the widget about to be unfocused.
+                    run_on_text_event_pass(root, &TextEvent::Ime(winit::event::Ime::Disabled));
+                }
```
DJMcNab, line 480: One edge case here is that this causes inconsistencies if `set_focus` ends up being called in this event.
At the very least, we need to move the setting of `self.global_state.focused_widget` on line 504 to use `next_focused` instead.

tomcur: Ah, of course. I tried to be careful, e.g. with the position where the event is triggered, but missed the `EventCtx::set_focus` method.

masonry/src/passes/update.rs
```diff
@@ -402,6 +403,19 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
...
+    let was_ime_active = root.global_state.is_ime_active;
+
+    let synthesize_ime_disabled =
+        was_ime_active && prev_focused != root.global_state.next_focused_widget;
+    if synthesize_ime_disabled {
+        // IME was active, but the next focused widget is going to receive the
+        // Ime::Disabled event. Synthesize an Ime::Disabled event here and send it to
+        // the widget about to be unfocused.
+        run_on_text_event_pass(root, &TextEvent::Ime(winit::event::Ime::Disabled));
+    }
```
DJMcNab, line 415: Should we emit the EndIme signal here?

masonry/src/text/editor.rs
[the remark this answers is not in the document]
tomcur: It's all slightly messy, with different implementations making different choices. Hiding the cursor completely is what Winit suggests, so perhaps hiding the caret is the best option.

Probably we should then also call `accesskit::node::clear_text_selection`.

masonry/src/passes/update.rs
tomcur: If this is hit, the `EndIme` signal is emitted in the refocus steps below (line 507). I'm happy with moving it up. There may be some more opportunities to clean this code up a bit; there appear at least to be some invariants that the current code misses, for example, `was_ime_active` appears to imply `prev_focused.is_some()`, and `is_ime_active` implies `next_focused.is_some()`.

tomcur: I've simplified IME in the focus pass a bit, behavior should be the same.

masonry/src/passes/update.rs
```diff
@@ -402,6 +403,19 @@ pub(crate) fn run_update_focus_pass(root: &mut RenderRoot) {
...
     let prev_focused = root.global_state.focused_widget;
+    let was_ime_active = root.global_state.is_ime_active;
+
+    let disable_ime = was_ime_active && prev_focused != root.global_state.next_focused_widget;
+    if disable_ime {
+        // IME was active, but the next focused widget is going to receive the Ime::Disabled event
+        // sent by the platform. Synthesize an Ime::Disabled event here and send it to the widget
+        // about to be unfocused.
+        run_on_text_event_pass(root, &TextEvent::Ime(winit::event::Ime::Disabled));
+        root.global_state.emit_signal(RenderRootSignal::EndIme);
```
DJMcNab, line 414: A much more targetted fix, with fewer non-local impacts might look like:
```suggestion
        // Disable the IME, which was enabled specifically for this widget.
        // Note that if the newly focused widget also requires IME, we will request it 
        // again - this resets the platform's state, ensuring that partial IME 
        // inputs do not "travel" between widgets
        root.global_state.emit_signal(RenderRootSignal::EndIme);
        if prev_focused == root.global_state.next_focused_widget {
             tracing::warn!(id = prev_focused.map(|id| id.trace()), "request_focus called whilst handling Ime::Disabled");
             // In this unlikely case, the rest of this handler will short-circuit, and IME will never be re-enabled 
             // for this widget. The resultant `ImeEnabled` event will be routed to this widget as it is the focused
             // widget. We don't handle this as above to avoid loops
            root.global_state.emit_signal(RenderRootSignal::StartIme);
        }
```

DJMcNab (approved):
A potential simplification of the new pass code.

Either applying this or not, I think we can still land this. The only accessibilty integration problem might occur if there is a pre-edit, and not having IME is a much bigger accessibility fail. I'm unlikely to be able to review again before Monday (at least), but I'm happy for it to land as-is (presuming there aren't any massive new changes, of course)

tomcur:
I'm going to defer the decision on cursor hiding to a different PR, as it may hit on AccessKit accessibility, too.

masonry/src/passes/update.rs
tomcur: I like it. We've hill-climbed the focus pass IME stuff to something no longer _as_ messy.

tomcur: I'm going to look into one more aspect of the focus update pass about the behavior surrounding disabling.

PR #764, opened 2024-11-28
Set a transform on the root AccessKit node

mwcampbell:
This is better than having to apply the scale factor to the bounding rectangles of all nodes, especially when we take externally generated AccessKit nodes, such as those generated by Parley, into account.

### DIFF
--- a/masonry/src/passes/accessibility.rs
+++ b/masonry/src/passes/accessibility.rs
@@ -19,7 +19,7 @@ fn build_accessibility_tree(
     mut widget: ArenaMut<'_, Box<dyn Widget>>,
     mut state: ArenaMut<'_, WidgetState>,
     rebuild_all: bool,
-    scale_factor: f64,
+    scale_factor: Option<f64>,
 ) {
     let _span = enter_span_if(
         global_state.trace.access,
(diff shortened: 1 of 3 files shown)


masonry/src/passes/accessibility.rs
```diff
@@ -89,10 +89,7 @@ fn build_accessibility_tree(
 // --- MARK: BUILD NODE ---
 fn build_access_node(widget: &mut dyn Widget, ctx: &mut AccessCtx) -> Node {
     let mut node = Node::new(widget.accessibility_role());
-    node.set_bounds(to_accesskit_rect(
-        ctx.widget_state.window_layout_rect(),
-        ctx.scale_factor,
```
DJMcNab, line 94: Can we remove `AccessCtx::scale_factor`?

dfrg (approved):
I’m not able to properly test this but the code looks good to me and the change makes sense.

masonry/src/passes/accessibility.rs
DJMcNab: Hmm, I see it's used in the `root_widget` accessibility. I wonder if we should somehow make this a property? of the root widget specifically?

It seems like it's only useful for that one place (analogy: the paint method doesn't know about the scale factor)

DJMcNab (approved):
This is better, but it seems like `AccessCtx::scale_factor` is now of unclear value, except for the root widget.

The equivalent code for paint is done at the render time.

masonry/src/passes/accessibility.rs
mwcampbell: I didn't like the idea of passing the scale factor down through all the `AccessCtx` instances either. I think I've come up with something cleaner now.

DJMcNab (approved):
I agree that this new change is nicer, and can land.

PR #767, opened 2024-12-03
Stop IME just before shutting down

DJMcNab:
This works around a janky behaviour where IME content is transferred to new instances of the same app.

This might not be needed on other systems, but I don't think that this change can cause any harm either.

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -472,7 +472,9 @@ impl MasonryState<'_> {
                 accesskit_adapter.update_if_active(|| tree_update);
             }
             WinitWindowEvent::CloseRequested => {
-                //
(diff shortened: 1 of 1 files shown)


masonry/src/event_loop_runner.rs
```diff
@@ -471,7 +471,11 @@ impl MasonryState<'_> {
                 };
                 accesskit_adapter.update_if_active(|| tree_update);
             }
-            WinitWindowEvent::CloseRequested => event_loop.exit(),
+            WinitWindowEvent::CloseRequested => {
+                //
```
tomcur, line 475: Unfinished comment

Philipp-M (approved):

tomcur (approved):

PR #766, opened 2024-12-03
Update to upstream unbatched PlainEditor

DJMcNab:
This is https://github.com/linebender/parley/pull/192

PR #768, opened 2024-12-03
Clear compose in TextArea::reset_text

tomcur:
The user may be composing when the `TextArea` text is programmatically changed. This change clears the compose state in order to adhere to the `PlainEditor` requirements. It is disruptive, but we already warn about disruption (and about IME in particular) in the documentation of `TextArea::reset_text`.

This doesn't reset the platform's IME state, we could do, but that's potentially even more disruptive, and looks like it's a bit harder to wire up. We could also re-insert the preedit text. That'd be nicest to do if we can query `PlainEditor` for the current preedit text, which requires an API change there.

### DIFF
--- a/masonry/src/widget/text_area.rs
+++ b/masonry/src/widget/text_area.rs
@@ -356,6 +356,15 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
     /// This is likely to be disruptive if the user is focused on this widget,
     /// as it does not retain selections, and may cause undesirable interactions with IME.
     pub fn reset_text(this: &mut WidgetMut<'_, Self>, new_text: &str) {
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
I do already have this in #766, but I think landing it now also makes sense.

tomcur:
Ah, seems like we crossed each other. But yeah, let's land this now, as it's quite easy to cause crashes in the current state.

tomcur:
Note for posterity: there's a separate issue in Xilem where views like `Textbox` use `PlainEditor::text()` to check whether to use `reset_text`. Because `PlainEditor::text()` currently returns the text including the preedit, whereas Masonry hides the preedit from Xilem, the resetting is a bit overzealous.

PR #769, opened 2024-12-05
use cargo rdme for tree arena

KGrewal1:
Use `cargo rdme` for crate readme and check in CI as mentioned in https://github.com/linebender/xilem/pull/752#discussion_r1867737085

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,9 +96,13 @@ jobs:
         with:
           tool: cargo-rdme
 
-      - name: cargo rdme
+      - name: cargo rdme masonry
         run: |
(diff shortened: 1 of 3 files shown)


DJMcNab (approved):
Looks good! Thanks

You don't need to ping me in the PR description, especially for a change which can be reviewed by anyone.

DJMcNab (commented):
Looking at it again, we should adopt some of the differences from https://github.com/linebender/color/pull/38.

Those are:
1) Using `--heading-base-level=0` (evaluating this for Masonry would be a reasonable follow-up, not for this PR)
2) Adding an explanation block comment at the top of the README

There are no intra-doc-links yet, so the hack used for those isn't needed.

PR #770, opened 2024-12-05
heading level and doc for masonry readme

KGrewal1:
Update the Masonry readme with documentation on how it is generated and change the heading level to be consistent with Color

### DIFF
--- a/masonry/README.md
+++ b/masonry/README.md
@@ -16,6 +16,11 @@
 
 [tracing_tracy]: https://crates.io/crates/tracing-tracy
 
+<!-- We use cargo-rdme to update the README with the contents of lib.rs.
+To edit the following section, update it in lib.rs, then run:
+cargo rdme --workspace-project=color --heading-base-level=0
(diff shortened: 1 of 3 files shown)


masonry/README.md
```diff
@@ -16,6 +16,11 @@
 
 [tracing_tracy]: https://crates.io/crates/tracing-tracy
```
DJMcNab, line 17: This line is the equivalent of the `libm` line in: https://github.com/linebender/color/blob/9775885b74e37f2de2a1713f80c607ba84e474c1/color/README.md#L21-L23

Therefore, it should be below the new comment and the links block should have the same comment as in color.

DJMcNab (approved):
Looks good! Again, let me know when you're ready for this to merge.

KGrewal1:
Fine with merging now if this looks good to you!

KGrewal1:
Made change to add the comment, so good to merge on my side!

PR #771, opened 2024-12-05
Add special case when window is resized to zero

PoignardAzur:
There have been reported crashes with Xilem when minimizing an app with Windows 11 (see also https://github.com/linebender/vello/issues/371).

This fix is based on the solution proposed in:

https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/Windows.2011.20minimizing

I've tested it on Linux just enough to confirm it doesn't crash. Haven't tested in on Windows yet, so I'm not sure it actually fixes the crash on minimize.

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -479,6 +479,12 @@ impl MasonryState<'_> {
                 event_loop.exit();
             }
             WinitWindowEvent::Resized(size) => {
+                if size.width == 0 || size.height == 0 {
+                    self.handle_suspended(event_loop);
+                    return;
+                } else if let WindowState::Suspended { .. } = &mut self.window {
+                    self.handle_resumed(event_loop);
+                }
                 self.render_root
                     .handle_window_event(WindowEvent::Resize(size));
             }


xStrom:
This fixes the crash, but after restoring the window from the minimized state no events are processed any more. So clicking on buttons, including the X to close window, doesn't do anything.

```
09:50:39.763Z  WARN Got window event whilst suspended or before window created event=CursorMoved { device_id: DeviceId(DeviceId(0)), position: PhysicalPosition { x: 674.0, y: 1.0 } }
09:50:39.764Z  WARN Got window event whilst suspended or before window created event=CursorMoved { device_id: DeviceId(DeviceId(0)), position: PhysicalPosition { x: 674.0, y: 0.0 } }
09:50:39.770Z  WARN Got window event whilst suspended or before window created event=CursorLeft { device_id: DeviceId(DeviceId(0)) }
09:50:40.083Z  WARN Got window event whilst suspended or before window created event=CloseRequested
09:50:40.661Z  WARN Got window event whilst suspended or before window created event=CloseRequested
```

PoignardAzur:
Since I don't have a Windows 11 machine, I'm not the person best placed to fix this. Closing this PR to avoid giving the impression I'm working on it.

PR #772, opened 2024-12-06
Use Hashmap to track TreeArena node's children

PoignardAzur:
This should improve performance in cases with where a widget has a very large number of direct children.

### DIFF
--- a/tree_arena/src/tree_arena_safe.rs
+++ b/tree_arena/src/tree_arena_safe.rs
@@ -19,7 +19,7 @@ use std::collections::HashMap;
 struct TreeNode<T> {
     id: NodeId,
     item: T,
-    children: Vec<TreeNode<T>>,
+    children: HashMap<NodeId, TreeNode<T>>,
 }
 
 // TODO - ArenaRefChildren and ArenaMutChildren might be easier to document if they were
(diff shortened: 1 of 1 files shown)


tree_arena/src/tree_arena_safe.rs
```diff
@@ -420,23 +414,19 @@ impl<'arena, T> ArenaMutChildren<'arena, T> {
     #[must_use]
     pub fn remove_child(&mut self, child_id: impl Into<NodeId>) -> Option<T> {
         let child_id = child_id.into();
-        let i = self
-            .children
-            .iter()
-            .position(|child| child.id == child_id)?;
+        let child = self.children.remove(&child_id)?;
 
-        fn remove_children<I>(
```
Philipp-M, line 428: I don't think that rename is really necessary, but no strong opinion there...

Philipp-M (approved):
Thanks, that definitely improves the issues I have faced.

What we could also consider is something like `FxHashMap` or `hashbrown` to further improve access of children (either in this PR or follow-ups).

I also wonder, whether something like the js-framework-benchmark would make sense to implement for either Xilem and/or Masonry (as native bench of course), so that we can quickly/roughly evaluate performance improvements/regressions, I think it covers common tree-mutations and access quite well.

PR #777, opened 2024-12-09
Render one frame before showing the window to avoid flashing

### LINKED ISSUE #590: White flash on app startup on Windows 11 dark mode

Long0x0:
Fixes #590.

Demo:
capture

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -264,7 +264,6 @@ impl MasonryState<'_> {
                 let window = event_loop.create_window(attributes).unwrap();
 
                 let adapter = Adapter::with_event_loop_proxy(&window, self.proxy.clone());
-                window.set_visible(visible);
                 let window = Arc::new(window);
                 // https://github.com/rust-windowing/winit/issues/2308
                 #[cfg(target_os = "ios")]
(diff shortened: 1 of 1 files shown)


jaredoconnell (commented):
I can confirm that this works great on Windows 10. I'll let the other reviewers review the code itself.

masonry/src/event_loop_runner.rs
```diff
@@ -280,12 +279,18 @@ impl MasonryState<'_> {
...
-                    window,
+                    window: window.clone(),
                     surface,
                     accesskit_adapter: adapter,
                 };
                 self.render_root
                     .handle_window_event(WindowEvent::Rescale(scale_factor));
+                // Render one frame before showing the window to avoid flashing
+                if visible {
+                    let (scene, _tree_update) = self.render_root.redraw();
```
DJMcNab, line 290: Discarding this tree update feels likely to be very bad.

Long0x0: Yeah, ignoring the tree update could lead to issues. I’ve updated the code to include it, following the approach used in handling the `RedrawRequested` event. Let me know if it looks good now.

DJMcNab:
I'm not set up to test accessibility on my machine yet. But this looks right to me now. I'm treating this similarly to #767, where it's a targeted fix which shouldn't lead to regressions.

DJMcNab (approved):

PR #778, opened 2024-12-09
Consistently use `FontWeight` as the name

waywardmonkeys:
Right now, there are uses of `Weight` from Fontique, `TextWeight` from Masonry, and `FontWeight` from Parley.

We should just use a single name.


waywardmonkeys:
This is related to https://github.com/linebender/parley/pull/211

DJMcNab (approved):
This has been a consequence of just doing whatever auto-import works. I've know it's been inconsistent, but sometimes `FontWeight` just wouldn't become available, so I went with the `Weight` which would auto-import as expected.

Thanks!

PR #779, opened 2024-12-09
Disable debug info compilation in CI to boost performance.

xStrom:
Turning off debug info in CI gives us tiny speedups, with potential for more [in some scenarios](https://davidlattimore.github.io/posts/2024/02/04/speeding-up-the-rust-edit-build-run-cycle.html#avoid-linking-debug-info). Most notably there is 30% space savings on artifact size in Linux, really helping with cache pressure and compression/decompression time.

We don't really need the benefits of debug info in the CI environment. We're not attaching debuggers there and in case of panics we can still get a backtrace. Thus it's just wasted effort to generate the debug info and then to link it in and cache it.

## Benchmarks

I did some minor benchmarking on three different OS setups with Rust 1.83. Every step was repeated four times (1 warmup, then 3 for measuring), but stuff like CPU boost frequency etc was not controlled for. Even so, the numbers are useful for ballparking.

I set `CARGO_INCREMENTAL` to `0` to match the CI and measured `cargo clippy -p xilem --locked --profile ci --all-features`.

| Windows 11 23H2 | Old         | New         | Diff |
|-----------------|-------------|-------------|------|
| Cold duration   | 17.40s      | 17.21s      | -1%  |
| Warm duration   | 0.74s       | 0.73s       | -1%  |
| Cache size      | 490 801 867 | 485 287 057 | -1%  |

| macOS 10.15.4 | Old         | New         | Diff |
|---------------|-------------|-------------|------|
| Cold duration | 30.58s      | 30.34s      | -1%  |
| Warm duration | 1.27s       | 1.21s       | -5%  |
| Cache size    | 207 200 857 | 200 717 130 | -3%  |

| Ubuntu 23.10  | Old         | New         | Diff |
|---------------|-------------|-------------|------|
| Cold duration | 24.50s      | 23.50s      | -4%  |
| Warm duration | 0.84s       | 0.87s       | +4%  |
| Cache size    | 597 456 507 | 411 918 698 | -31% 🚀 |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,6 +129,6 @@ time = "0.3.36"
 [profile.ci]
 inherits = "dev"
 debug = 0 # Don't compile debug info to reduce compilation artifact size for cache benefits.
-strip = "debuginfo" # Implied with debug = 0 since Rust 1.77, still useful for older.
+strip = "debuginfo" # Implied by debug = 0 since Rust 1.77, but still needed for an older MSRV.
 [profile.ci.package."*"]
 debug-assertions = true # Keep always on for dependencies for cache reuse.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,5 +128,7 @@ time = "0.3.36"
 
 [profile.ci]
 inherits = "dev"
+debug = 0 # Don't compile debug info to reduce compilation artifact size for cache benefits.
+strip = "debuginfo" # Implied with debug = 0 since Rust 1.77, still useful for older.
 [profile.ci.package."*"]
 debug-assertions = true # Keep always on for dependencies for cache reuse.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,7 +279,7 @@ jobs:
           save-if: ${{ github.event_name != 'merge_group' }}
 
       - name: Run cargo nextest
-        run: cargo nextest run --workspace --locked --all-features --no-fail-fast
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -128,5 +128,7 @@ time = "0.3.36"
 
 [profile.ci]
 inherits = "dev"
+debug = 0 # Don't compile debug info to reduce compilation artifact size for cache benefits.
+strip = "debuginfo" # Implied with debug = 0 since Rust 1.77, still useful for older.
```
DJMcNab, line 132: This sentence doesn't scan to me - `older` doesn't really work as a subject. Maybe something like:
```suggestion
strip = "debuginfo" # Implied by debug = 0 since Rust 1.77 - but our MSRV is older.
```

DJMcNab (approved):
Smaller cache sizes are great! It's especially important for the snapshots LFS workaround

Cargo.toml
DJMcNab: Nice!

xStrom: It's meant to be copy-pastable and in Xilem's case the MSRV is actually newer than 1.77. I adjusted the comment but made the MSRV reference a bit more conditional.

waywardmonkeys: MSRV will be 1.82 as soon as this gets updated Peniko or Vello.

DJMcNab: Maybe we should just not have the comment at all? The redundancy doesn't have any real cost

PR #780, opened 2024-12-09
xilem_web: Add up- and download example

flosse:


DJMcNab:
I think that calling this upload is quite misleading, as it strongly suggests that there is a server involved. I'd normally say open/save are probably the equivalents for local files, noting that the save operation as you have to implement it on the web is much closer semantically to download.

flosse:
> I think that calling this upload is quite misleading, as it strongly suggests that there is a server involved.

Well, I guess it depends on the perspective, but I understand your point of course.

> I'd normally say open/save are probably the equivalents for local files, noting that the save operation as you have to implement it on the web is much closer semantically to download.

I am totally open to the naming :smile:

I have now renamed it to open/save.

flosse:
> I think the fact that this example needs to use `Rc<RefCell<...>>` points at some core architectural problems with Xilem,

... or at least with `xilem_web`.

> so if nothing else it's worth including it to highlight these problems.

I don't have enough understanding of the background to say anything intelligent here.
I'm trying to collect use cases that are not (yet) trivial to solve from a user perspective.

> Otherwise, code seems perfectly fine.

Thanks for the review!

Philipp-M:
> > I think the fact that this example needs to use `Rc<RefCell<...>>` points at some core architectural problems with Xilem,
> 
> ... or at least with `xilem_web`.

FWIW I don't think it's necessary anymore, since we have access to the app state in the lifecycle methods of the `View` trait.
`AfterBuild` etc. views need to be updated to also have access to the app state (PR welcome, I can review that), which should make interior mutability unnecessary on a quick look.

PR #782, opened 2024-12-11
Replace CI `RUST_MIN_VER_PKGS` with `--workspace --ignore-private`.

xStrom:
The CI `RUST_MIN_VER_PKGS` configuration variable was meant to track packages that are intended for publishing in order to limit MSRV checks to only those packages. This same goal can be achieved universally without the need for a configuration variable via the `--ignore-private` `cargo-hack` option.

The immediate effect on this repo is that `xilem_core` is no longer part of the MSRV check because it has `publish = false`. That is fine, because the MSRV check will run just before publishing, triggered by the PR that will remove `publish = false`, ensuring that the MSRV is accurate before publishing.

### DIFF
--- a/tree_arena/Cargo.toml
+++ b/tree_arena/Cargo.toml
@@ -1,27 +1,22 @@
 [package]
 name = "tree_arena"
 version = "0.1.0"
-description = "An arena allocated tree designed for Linebender"
+description = "An arena allocated tree."
 keywords = ["arena", "tree"]
-categories = ["gui"]
+categories = ["algorithms", "data-structures"]
 edition.workspace = true
-rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
 homepage.workspace = true
+rust-version.workspace = true
 
+publish = false # Remove just before publishing
 
 [package.metadata.docs.rs]
 all-features = true # all features are enabled - this means that the generated docs are for the safe version
 # There are no platform specific docs.
 default-target = "x86_64-unknown-linux-gnu"
 targets = []
-# rustdoc-scrape-examples tracking issue https://github.com/rust-lang/rust/issues/88791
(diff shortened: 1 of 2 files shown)


tree_arena/Cargo.toml
```diff
@@ -1,27 +1,22 @@
...
 
+publish = false # Remove just before publishing
 
 [package.metadata.docs.rs]
 all-features = true # all features are enabled - this means that the generated docs are for the safe version
 # There are no platform specific docs.
 default-target = "x86_64-unknown-linux-gnu"
 targets = []
-# rustdoc-scrape-examples tracking issue https://github.com/rust-lang/rust/issues/88791
-# cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"] # no examples yet
```
DJMcNab, line 20: Hmm. I'd be surprised if scrape-examples failed with no examples. Maybe we should just have this?

DJMcNab (approved):
Both xilem_core and tree-arena will be checked when their corresponding crates are checked anyway, just not exhaustively with all features.

tree_arena/Cargo.toml
xStrom: It wouldn't fail but this removal is for consistency and reduced noise, as every other package that doesn't enable it doesn't have it either.

I don't expect major examples to be introduced any time soon for this package. If they do appear at some point we can reintroduce it.

Also, if an example does get introduced, merely uncommenting that line (or having it already be uncommented) isn't sufficient. Something like the following is needed as well:

```toml
[[example]]
name = "mason"
# This actually enables scraping for all examples, not just this one.
# However it is possible to set doc-scrape-examples to false for other specific examples.
doc-scrape-examples = true
```

PR #784, opened 2024-12-11
Add TOML formatting with Taplo

PoignardAzur:
Add Taplo config.
Format TOM config files.
Add Taplo step to CI.


.github/workflows/ci.yml
```diff
@@ -81,6 +81,14 @@ jobs:
       - name: Run cargo fmt
         run: cargo fmt --all --check
 
+      - name: Install Taplo
+        uses: uncenter/setup-taplo@v1
```
DJMcNab, line 85: I'm not a huge fan of using another third party actions provider, but I see that wgpu also uses this.

It doesn't seem like the maintainer of this action actually uses it, which is slightly concerning.
If we just depended on a specific hash rather than just `@v1`, the security concern goes away

.github/workflows/ci.yml
```diff
@@ -81,6 +81,14 @@ jobs:
...
+        with:
+          version: "0.9.3"
+
+      - name: run `taplo fmt`
```
DJMcNab, line 89:
```suggestion
      - name: Run taplo fmt
```
would be consistent with the other projects.

.taplo.toml
```diff
@@ -0,0 +1,11 @@
+# See https://taplo.tamasfe.dev/configuration/file.html
+# and https://taplo.tamasfe.dev/configuration/formatter-options.html
+
+[formatting]
+# This is a matter of taste, but expanded inline arrays make tables harder to read.
+array_auto_collapse = true
```
DJMcNab, line 6: This seems to already be the default

.taplo.toml
```diff
@@ -0,0 +1,11 @@
...
+inline_table_expand = false
+
+# Aligning comments with the largest line creates
+# diff noise when neighboring lines are changed.
+align_comments = false
```
DJMcNab, line 11: Agreed

xStrom, line 11:
```suggestion
align_comments = false

column_width = 100
indent_string = "    "
```

xStrom (commented):
Let's bump the column width to 100 and the indentation size to 4 spaces. Both of these values are exactly what `rustfmt` does with Rust code.

DJMcNab (commented):
I'd like to get a second opinion from @xStrom on the actions design, but otherwise this looks right to me.

.github/workflows/ci.yml
DJMcNab: For reference, this wouldn't have been my immediate reaction if we used the [`taiki-e/install-action`](https://github.com/taiki-e/install-action) here.

.taplo.toml
xStrom: Yeah not much point in defining `array_auto_collapse` here as it matches the default.

.taplo.toml
```diff
@@ -0,0 +1,11 @@
+# See https://taplo.tamasfe.dev/configuration/file.html
+# and https://taplo.tamasfe.dev/configuration/formatter-options.html
+
+[formatting]
+# This is a matter of taste, but expanded inline arrays make tables harder to read.
+array_auto_collapse = true
+inline_table_expand = false
```
xStrom, line 7: Not sure about setting `inline_table_expand` to `false`. The expansion **only** happens when the line exceeds the length defined in `column_width` (which we should have at 100).

With it `false`, the following happens:

```toml
winapi = { version = "0.3.9", features = [
    "winbase",
    "libloaderapi",
    "errhandlingapi",
    "winuser",
    "shellscalingapi",
    "shobjidl",
    "combaseapi",
    "dxgi1_3",
    "dwmapi",
    "wincon",
    "fileapi",
    "processenv",
    "winbase",
    "winerror",
    "handleapi",
    "shellapi",
    "winnls",
] }

# .. gets converted into:

winapi = { version = "0.3.9", features = ["winbase", "libloaderapi", "errhandlingapi", "winuser", "shellscalingapi", "shobjidl", "combaseapi", "dxgi1_3", "dwmapi", "wincon", "fileapi", "processenv", "winbase", "winerror", "handleapi", "shellapi", "winnls"] }
```

Also, just so I understand your claim, are you saying that the following expansion that we currently have makes the `[workspace.lints]` table hard to read?

```toml
rust.unexpected_cfgs = { level = "warn", check-cfg = [
    'cfg(FALSE)',
    'cfg(tarpaulin_include)',
] }
```

PoignardAzur: I've found while testing it that the formatter was really eager to break up inline tables even when it didn't seem needed to stay below the column limit. I'll give it another look.

.github/workflows/ci.yml
PoignardAzur: I just looked into adding taplo to the install-action list, but the format of taplo's release artifacts seems to change from release to release, so I'm not comfortable taking on that work.

Using a specific hash for uncenter's action seems easy enough.

PoignardAzur:
I've addressed review comments.

Taplo with a column width of 100 seems to mostly leave inline tables be, so I've removed the ` inline_table_expand = false` part.

xStrom:
The config options look good now. Still two things though:
* The CI step should not have backticks, they won't be converted into CSS.
* One TOML file needs adjustment to pass the CI. 😆

PoignardAzur:
Huh, I really thought I'd fixed both.

xStrom (approved):
Looks good, thanks!

alerque:
> the indentation size to 4 spaces

Why go with something different than what cargo itself generates? That's just going to be an ongoing conflict with other tools. TOML is not Rust, I would suggest 2 space indent to match other tools that edit the manifest and lock files in the Rust toolchain.

xStrom:
Well we had far more 4 spaces already for one.

I am curious though which cargo command generates indents? I know `Cargo.lock` has them, but there it's only a single space.

xStrom:
The Rust Style Guide has a page called [Cargo.toml conventions](https://doc.rust-lang.org/stable/style-guide/cargo.html#cargotoml-conventions). It states:

> Use the same line width and indentation as Rust code.

So there you go, the Rust docs seem to agree with our chosen options of 100 and 4 spaces.

alerque:
> So there you go

Indeed. I guess I'll take my beef upstream to `cargo` tooling that doesn't do that. Something I just bumped into a couple weeks ago was reformatting my manifest and I wasn't too happy about it. Maybe it was `cargo edit` or `cargo set-version`. I'll check it out.

In any event if that's what the Rust style guide suggests, then unless/until that gets changed upstream I'd say carry on! I'll likely adjust my own `taplo` usage to match if that's really the case.

PR #785, opened 2024-12-11
Improve IME area reporting and set IME area in `TextArea`

tomcur:
This adds the required context methods for (re)setting the IME area. These are added to the mutable context types, which I believe is the right place.

`TextArea` now uses `PlainEditor::ime_cursor_area` to report its IME area.

Note Masonry currently reports new IME cursor areas only when the widget itself moves or the focus changes to a new widget, but Parley's `PlainEditor` determines the IME cursor area based on the selection and preedit state of the editor, meaning the area can change based on events as well. Other widgets may choose to have different behavior still.

(It appears Masonry currently doesn't account for the IME area potentially changing based on relayout, which may be a bug.)

This replaces the old IME area reporting with a check at the end of the rewrite pass to see if IME is active, and if it is, whether the IME area has moved. If it has, the new area is reported to the platform.

### DIFF
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -74,6 +74,7 @@ pub(crate) struct RenderRootState {
     pub(crate) text_layout_context: LayoutContext<BrushIndex>,
     pub(crate) mutate_callbacks: Vec<MutateCallback>,
     pub(crate) is_ime_active: bool,
+    pub(crate) last_sent_ime_area: Rect,
     pub(crate) scenes: HashMap<WidgetId, Scene>,
     /// Whether data set in the pointer pass has been invalidated.
     pub(crate) needs_pointer_pass: bool,
@@ -170,6 +171,7 @@ impl RenderRoot {
                 text_layout_context: LayoutContext::new(),
                 mutate_callbacks: Vec::new(),
                 is_ime_active: false,
+                last_sent_ime_area: Rect::new(f64::NAN, f64::NAN, f64::NAN, f64::NAN),
                 scenes: HashMap::new(),
                 needs_pointer_pass: false,
                 trace: PassTracing::from_env(),
@@ -263,6 +265,13 @@ impl RenderRoot {
         let _span = info_span!("text_event");
         let handled = run_on_text_event_pass(self, &event);
         run_update_focus_pass(self);
+
+        if matches!(event, TextEvent::Ime(winit::event::Ime::Enabled)) {
+            // Reset the last sent IME area, as the platform reset the IME state and may have
+            // forgotten it.
+            self.global_state.last_sent_ime_area =
+                Rect::new(f64::NAN, f64::NAN, f64::NAN, f64::NAN);
+        }
         self.run_rewrite_passes();
 
         handled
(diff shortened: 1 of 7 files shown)


masonry/src/render_root.rs
```diff
@@ -74,6 +74,7 @@ pub(crate) struct RenderRootState {
     pub(crate) text_layout_context: LayoutContext<BrushIndex>,
     pub(crate) mutate_callbacks: Vec<MutateCallback>,
     pub(crate) is_ime_active: bool,
+    pub(crate) last_sent_ime_area: Rect,
```
DJMcNab, line 77: Maybe there's a better place for this comment, but I think the magic value should be documented.
```suggestion
    /// We ensure that the first valid IME area is sent by using an invalid initial value
    pub(crate) last_sent_ime_area: Rect,
```

Maybe the right solution is to have a:
```rust
/// We ensure that the initial IME area is sent by storing an invalid initial value.
const INITIAL_IME_RECT: Rect = Rect::new(f64::NaN, f64::NaN, f64::NaN, f64::NaN);
```

The comment wording is first-pass and can be refined.

masonry/src/widget/text_area.rs
```diff
@@ -299,6 +299,25 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
     }
 }
 
+// --- MARK: HELPERS ---
+impl<const EDITABLE: bool> TextArea<EDITABLE> {
+    /// Get the IME area from the editor, accounting for padding.
+    ///
+    /// This should only be called when the editor layout is available.
+    fn ime_area(&self) -> Rect {
```
DJMcNab, line 307: This being repeated so often to me suggests that it should be a method on `Widget`, so that there is only one source of truth needed. That would be a bit of a bigger change to wire up, and I don't have all the specifics.

That is, this is a non-blocking conern

masonry/src/render_root.rs
```diff
@@ -489,6 +498,19 @@ impl RenderRoot {
...
+                .global_state
+                .focused_widget
+                .expect("IME is active without a focused widget");
+            let ime_area = self.widget_arena.get_state(widget).item.get_ime_area();
+            if self.global_state.last_sent_ime_area != ime_area {
+                self.global_state.last_sent_ime_area = ime_area;
+                self.global_state
+                    .emit_signal(RenderRootSignal::new_ime_moved_signal(ime_area));
+            }
+        }
```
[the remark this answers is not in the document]
DJMcNab: This seems valid to me.

masonry/src/render_root.rs
```diff
@@ -489,6 +498,19 @@ impl RenderRoot {
             self.global_state
                 .emit_signal(RenderRootSignal::RequestRedraw);
         }
+
+        if self.global_state.is_ime_active {
+            let widget = self
+                .global_state
+                .focused_widget
+                .expect("IME is active without a focused widget");
+            let ime_area = self.widget_arena.get_state(widget).item.get_ime_area();
```
DJMcNab, line 507: It's conceivable that you have enough context to call a `Widget` method with a `QueryCtx` here.

DJMcNab (approved):
I think this is an improvement. A few peripheral suggestions, and a potential refactor which might be more idiomatic

masonry/src/render_root.rs
tomcur: Applied with a second pass to the wording.

masonry/src/render_root.rs
DJMcNab: Hmm, that's true, and it's probably too messy. Without having a new invalidation method, it doesn't really work.

PR #787, opened 2024-12-14
Tweak documentation for readability

PoignardAzur:
Apply many small tweaks I found while going over the documentation.
Rename some variables.
Add asserts to `BoxConstraints::constrain_aspect_ratio`.
Flesh out the WidgetState documentation.
Introduce the concept of "zombie flags".

### DIFF
--- a/masonry/src/doc/05_pass_system.md
+++ b/masonry/src/doc/05_pass_system.md
@@ -12,15 +12,15 @@
 
 </div>
 
-Masonry has a set of **passes**, which are computations run over a subset of the widget tree during every frame.
+Masonry's internal architecture is based on a set of **passes**, which are computations run over a subset of the widget tree during every frame.
 
 Passes can be split into roughly three categories:
 
-- **Event passes:** triggered by user interaction.
-- **Rewrite passes:** run after every event pass, may run multiple times until all invalidation flags are cleared.
-- **Render passes:** run just before rendering a new frame.
+- **Event passes** are triggered by user interaction.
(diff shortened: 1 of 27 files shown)


PoignardAzur:
I'd say this is ready for review.

It's a bit of a kitchen sink PR, but almost all the changes are to documentation or comments only, and almost none of them change executable code.

Some changes remove long-dead code, bitrotted comments, outdated documentation, etc. Masonry's documentation has a tendency to refer to old ideas as if we were just about to implement them, so I think this spring cleaning will make it more readable.

jaredoconnell (approved):
This is an improvement.

masonry/src/doc/05_pass_system.md
```diff
@@ -106,6 +106,13 @@ It is called when new widgets are added to the tree, or existing widgets are rem
 
 It will call the `register_children()` widget method on container widgets whose children changed, then the `update()` method with the `WidgetAdded` event on new widgets.
 
+<!-- TODO - document UPDATE DISABLED --- -->
```
DJMcNab, line 109: I wouldn't use all caps here for the pass names

masonry/src/event.rs
```diff
@@ -313,6 +312,7 @@ pub enum Update {
 }
 
 impl PointerEvent {
+    /// Create a [`PointerEvent::PointerLeave`] event with dummy values.
     pub fn new_pointer_leave() -> Self {
```
DJMcNab, line 316: Maybe this method should be at the bottom, as something most users won't need (which users would need this?)

DJMcNab (approved):

PR #792, opened 2024-12-15
Progress bar animation

richard-uk1:


masonry/src/widget/progress_bar.rs
```diff
@@ -40,7 +48,40 @@ impl ProgressBar {
...
+            label,
+            animate: false,
+            anim_pixel_position: 0.,
+            anim_prev_interval: None,
+            next_anim_token: None,
+        }
+    }
+
+    /// Note - if calling this in a lifecycle method, you must also request an anim frame.
+    pub fn animate(mut self, animate: bool) -> Self {
```
PoignardAzur, line 62: I think this comment is above the wrong function.

masonry/src/widget/progress_bar.rs
```diff
@@ -174,6 +230,41 @@ impl Widget for ProgressBar {
...
+        self.reset_animation();
+        ctx.request_anim_frame();
+        ctx.request_paint_only();
+    }
+
+    fn on_anim_frame(&mut self, ctx: &mut UpdateCtx, interval: u64) {
+        tracing::debug!("progress_bar: on_anim_frame");
+        // pixel per 1ms
+        const NS_PER_PIXEL: f64 = 1_000_000.;
+        const DURATION_BETWEEN_ANIMATIONS: Duration = Duration::from_millis(2_500);
```
PoignardAzur, line 244: It's not super clear to me why you want the progress bar to have an animation every 2.5 seconds?

PoignardAzur (changes requested):
Sorry it took me forever to review. You'll probably need to either rewrite this from the ground up or deal with some thorny merge conflicts, given the refactors we've done since.

Overall, I like this PR. The design is pretty elegant and avoids relying on Winit too much. One thing I would like before we can consider merging this is better support deterministic testing in TestHarness.

There's a bunch of issues that need to be fixed that I assume are due to this being a draft: lack of documentation, `examples/animation.rs` is copy-pasted from the hello example, I think Instant doesn't work with webassembly, etc.

masonry/src/widget/progress_bar.rs
richard-uk1: It's arbitrary - I'm copying the way progress bars work on windows (with periodic movement to indicate progress). My motivation for having timers is triggering events at specific intervals in a simple way without needing to feed in messages from external tasks etc. Happy to go with whatever setup we wanted for animation on the progress bar. In any case it should be optional, with a `with_animation` on at least the xilem view. If I didn't do this already I should before merging.

masonry/src/testing/harness.rs
```diff
@@ -110,6 +111,7 @@ pub struct TestHarness {
     has_ime_session: bool,
     ime_rect: (LogicalPosition<f64>, LogicalSize<f64>),
     title: String,
+    timers: TimerQueue,
```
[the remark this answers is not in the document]
richard-uk1: I want to solve testing but didn't have an idea how to do it. Happy to go with your approach.

PR #794, opened 2024-12-16
Fix `clippy::question_mark` lint

waywardmonkeys:


jaredoconnell (approved):
The `?` operator is a cool feature in Rust.

PR #795, opened 2024-12-17
Sized Box: Use a brush for the border

waywardmonkeys:
This allows for a richer appearance by using any valid brush for the border.


waywardmonkeys:
I'm not excited about the clone of the brush here (but it would've happened anyway further down).  You can' pass `&border.brush` in Xilem because there's no `From<&Brush> for Brush` in Peniko (and maybe there should be?).

I'm not sure that any API being `Into<Color>` makes all that much sense. There isn't a rich selection of types that convert to `peniko::Color`.

DJMcNab (approved):
The clone is definitely unfortunate for gradients, but should be cheap for solid and images at least.

PR #797, opened 2024-12-18
xilem: Button with customisable label(text)

ArtyomSinyugin:


xilem/src/view/button.rs
```diff
@@ -60,8 +69,8 @@ where
         _ctx: &mut ViewCtx,
         mut element: Mut<Self::Element>,
     ) {
-        if prev.label != self.label {
-            widget::Button::set_text(&mut element, self.label.clone());
+        if prev.label.label != self.label.label {
+            widget::Button::set_text(&mut element, self.label.label.clone());
```
DJMcNab, line 73: I *think* this might be able to use `self.label.rebuild(prev.label, /* state */, ctx, widget::Button::label_mut(&mut element))`

Absolutely this needs to handle all of the properties properly.

xilem/src/view/button.rs
```diff
@@ -50,7 +52,14 @@ where
     type ViewState = ();
 
     fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
-        ctx.with_leaf_action_widget(|ctx| ctx.new_pod(widget::Button::new(self.label.clone())))
+        ctx.with_leaf_action_widget(|ctx| ctx.new_pod(
+            widget::Button::from_label(widget::Label::new(self.label.label.clone())
```
DJMcNab, line 56: I think this is the best we can do with the current design of `xilem::Pod`, but it's not a good final state.

xilem/src/view/button.rs
```diff
@@ -1,43 +1,45 @@
...
 pub use masonry::PointerButton;
 
 use crate::core::{DynMessage, Mut, View, ViewMarker};
 use crate::{MessageResult, Pod, ViewCtx, ViewId};
+use crate::view::Label;
 
 /// A button which calls `callback` when the primary mouse button (normally left) is pressed.
 pub fn button<State, Action>(
-    label: impl Into<ArcStr>,
+    label: Label,
```
DJMcNab, line 14: I'm not a *massive* fan of all the callsites being `button(label("text"), ...)`. I.e. that most users will just want to use the unitary `Label` input.

I'm not sure what the right way to fix that would be. *Maybe* it would be to `impl Into<Label> for &str/Arc<str>` and use `impl Into<Label>`  here?

DJMcNab (commented):
This is a refactor I've been meaning to do myself, so thanks for taking it on. Unfortunately, as it stands it isn't directly usable, but there's a viable path towards making it be so.

ArtyomSinyugin:
@DJMcNab Thank you for comments. 
I do my best to figure out the best solution for this task. 

I'll be fine if you decide to do it yourself in a better way in case you need to do it a little bit faster. 
For me it is just an interesting way to improve my skill in Rust and to learn UI in Rust.

xilem/src/view/button.rs
ArtyomSinyugin: It would be great to write self.label.rebuild, but rebuild method is not public. 

Now I am looking for solution similar as for sized_box, but it is not easy for me to figure out do I really need to rewrite code with something like V: WidgetView and etc. or there is more simple way.

xilem/src/view/button.rs
ArtyomSinyugin: I find a way to use `impl Into<Label>` after implementing `impl<T> From<T> for xilem::view::Label where T: Into<ArcStr>`

xilem/src/view/button.rs
ArtyomSinyugin: Okey. So, I do not need to change this code yet..

xilem/src/view/label.rs
```diff
@@ -21,8 +21,9 @@ pub fn label(label: impl Into<ArcStr>) -> Label {
 }
 
 #[must_use = "View values do nothing unless provided to Xilem."]
+#[derive(PartialEq)]
 pub struct Label {
-    label: ArcStr,
+    pub(in crate::view) label: ArcStr,
 
     // Public for variable_label as a semi-interims state.
```
DJMcNab, line 28: We should add button to this comment and move it up (probably also worth fixing my typo of `interims` when I meant `interim`)

xilem/src/view/label.rs
```diff
@@ -21,8 +21,9 @@ pub fn label(label: impl Into<ArcStr>) -> Label {
 }
 
 #[must_use = "View values do nothing unless provided to Xilem."]
+#[derive(PartialEq)]
```
DJMcNab, line 24: We shouldn't have this.

xilem/src/view/button.rs
```diff
@@ -50,18 +51,34 @@ where
...
     fn rebuild(
         &self,
         prev: &Self,
-        _: &mut Self::ViewState,
-        _ctx: &mut ViewCtx,
+        state: &mut Self::ViewState,
+        ctx: &mut ViewCtx,
         mut element: Mut<Self::Element>,
     ) {
         if prev.label != self.label {
```
DJMcNab, line 73: This check is not needed - the `rebuild` internally will re-do the same work

xilem/src/view/button.rs
```diff
@@ -50,18 +51,34 @@ where
     type ViewState = ();
 
     fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
-        ctx.with_leaf_action_widget(|ctx| ctx.new_pod(widget::Button::new(self.label.clone())))
+        ctx.with_leaf_action_widget(|ctx| {
+            ctx.new_pod(widget::Button::from_label(
+                widget::Label::new(self.label.label.clone())
```
DJMcNab, line 56:
```suggestion
                // TODO: Use `Label::build` here - currently impossible because `Pod` uses `WidgetPod` internally
                widget::Label::new(self.label.label.clone())
```

DJMcNab (commented):
This is looking really good! The last thing I think this needs is actual usage in an example. For example, changing the text colour of some buttons (e.g. the clear button in `calc`).

xilem/src/view/label.rs
ArtyomSinyugin: Oh. Now I understand, I'll fix it.

ArtyomSinyugin:
> This is looking really good! The last thing I think this needs is actual usage in an example. For example, changing the text colour of some buttons (e.g. the clear button in `calc`).

Thank you. I fixed all tasks and updated calc example: CE button is MEDIUM_VIOLET_RED when clicked now and color returns to WHITE after any number is pressed (do not blame me for this, I am really do not know how to do UI). 

Now I need only rustup.

DJMcNab (commented):
I'd suggest that changing the example by applying something like:
```diff
diff --git a/xilem/examples/calc.rs b/xilem/examples/calc.rs
index 33a2358ad..5d5dfad3d 100644
--- a/xilem/examples/calc.rs
+++ b/xilem/examples/calc.rs
@@ -50,12 +50,15 @@ struct Calculator {
     numbers: [String; 2],
     result: Option<String>,
     operation: Option<MathOperator>,
-    clear_button: Color,
 }
 
 impl Calculator {
     fn get_current_number(&self) -> String {
-        self.numbers[self.current_num_index].clone()
+        self.current_number().to_string()
+    }
+
+    fn current_number(&self) -> &str {
+        &self.numbers[self.current_num_index]
     }
 
     fn set_current_number(&mut self, new_num: String) {
@@ -73,7 +76,6 @@ impl Calculator {
 
     fn clear_entry(&mut self) {
         self.clear_current_entry_on_input = false;
-        self.clear_button = Color::MEDIUM_VIOLET_RED;
         if self.result.is_some() {
             self.clear_all();
             return;
@@ -82,9 +84,6 @@ impl Calculator {
     }
 
     fn on_entered_digit(&mut self, digit: &str) {
-        if self.clear_button != Color::WHITE {
-            self.clear_button = Color::WHITE;
-        }
         if self.result.is_some() {
             self.clear_all();
         } else if self.clear_current_entry_on_input {
@@ -204,7 +203,6 @@ fn num_row(nums: [&'static str; 3], row: i32) -> impl GridSequence<Calculator> {
 const DISPLAY_FONT_SIZE: f32 = 30.;
 const GRID_GAP: f64 = 2.;
 fn app_logic(data: &mut Calculator) -> impl WidgetView<Calculator> {
-    let label = label("CE").brush(data.clear_button);
     grid(
         (
             // Display
@@ -222,7 +220,15 @@ fn app_logic(data: &mut Calculator) -> impl WidgetView<Calculator> {
             ))
             .grid_item(GridParams::new(0, 0, 4, 1)),
             // Top row
-            expanded_button(label, Calculator::clear_entry).grid_pos(0, 1),
+            expanded_button(
+                label("CE").brush(if data.get_current_number().is_empty() {
+                    Color::MEDIUM_VIOLET_RED
+                } else {
+                    Color::WHITE
+                }),
+                Calculator::clear_entry,
+            )
+            .grid_pos(0, 1),
             expanded_button("C", Calculator::clear_all).grid_pos(1, 1),
             expanded_button("DEL", Calculator::on_delete).grid_pos(2, 1),
             operator_button(MathOperator::Divide).grid_pos(3, 1),
@@ -290,7 +296,6 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
         numbers: ["".into(), "".into()],
         result: None,
         operation: None,
-        clear_button: Color::WHITE,
     };
 
     let app = Xilem::new(data, app_logic);

```
would be more idiomatic. I can't apply this myself, because you have made this PR from your fork's default branch (`main`).

ArtyomSinyugin:
> would be more idiomatic. I can't apply this myself, because you have made this PR from your fork's default branch (`main`).

I've copied your version of calc.rs to my fork.

PR #798, opened 2024-12-18
Update to git Parley and Vello.

waywardmonkeys:
This brings the new releases of Color and Peniko and start making sure things are good for the upcoming releases of Parley and Vello.


xilem_web/src/svg/common_attrs.rs
```diff
@@ -67,10 +67,11 @@ fn join(iter: &mut impl Iterator<Item: std::fmt::Display>, sep: &str) -> String
 fn brush_to_string(brush: &Brush) -> String {
     match brush {
         Brush::Solid(color) => {
-            if color.a == 0 {
+            let [r, g, b, a] = color.to_rgba8().to_u8_array();
+            if a == 0 {
                 "none".into()
             } else {
-                format!("#{:02x}{:02x}{:02x}", color.r, color.g, color.b)
+                format!("#{:02x}{:02x}{:02x}", r, g, b)
```
waywardmonkeys, line 74: @Philipp-M Was it intentional here to drop the alpha channel?

If not, then we can use the new `LowerHex` formatter for `Color`....

Philipp-M: I think this is originally from Raph, but IIRC this is indeed intentional, see also [here](https://stackoverflow.com/questions/6042550/svg-fill-color-transparency-alpha), the alpha channel gets separately applied with the `fill-opacity` and `stroke-opacity` attribute

waywardmonkeys: I've changed this code in the new version now and added a test for the behavior to make sure it is doing the right thing.

Cargo.toml
```diff
@@ -104,13 +104,13 @@ clippy.duplicated_attributes = "allow"
 masonry = { version = "0.2.0", path = "masonry" }
 xilem_core = { version = "0.1.0", path = "xilem_core" }
 tree_arena = { version = "0.1.0", path = "tree_arena" }
-vello = "0.3"
-wgpu = "22.1.0"
+vello = { git = "https://github.com/linebender/vello", rev = "a71236c7c8da10a6eaad4602267663339620835a" }
+wgpu = "23.0.1"
 kurbo = "0.11.1"
-parley = { git = "https://github.com/linebender/parley", rev = "1a8740d8d86ebf751201e45e89bb71019340137d", features = [
+parley = { git = "https://github.com/linebender/parley", rev = "352d156d3cda07fe9986c1e594aed176238eddf3", features = [
```
DJMcNab, line 110:
```suggestion
parley = { git = "https://github.com/linebender/parley", rev = "16b62518d467487ee15cb230fdb28530d89a8cf6", features = [
```
is the same commit, but on main.

DJMcNab (approved):

Cargo.toml
waywardmonkeys: Done locally. Also has to update `Cargo.lock`.

waywardmonkeys:
Something looks wrong with the image layout tests... not sure what.

waywardmonkeys:
I think the new ones are correct and due to the sampling improvements in Vello.

DJMcNab (approved):

waywardmonkeys:
This *will* need a follow up to improve how exports are done and `color::palette` is accessed so that there's a nice story for user code (as well as within the libraries).

waywardmonkeys:
> This _will_ need a follow up to improve how exports are done and `color::palette` is accessed so that there's a nice story for user code (as well as within the libraries).

I did that (at least good enough for now) in #800.

PR #799, opened 2024-12-19
Fix and un-allow `clippy::use_self` lints

waywardmonkeys:


DJMcNab (approved):
I have confirmed that these (fc4df58eb53f9792a17bfab8725b36a6285ed6c7) are the results of `cargo clippy --fix` with `use_self` re-enabled. I have not manually reviewed the changes.

Thanks!

PR #800, opened 2024-12-20
Use `palette` and `Color` from top level Xilem and Masonry

waywardmonkeys:


waywardmonkeys:
This intentionally doesn't move anything around from the theme. That contains non-color data and wouldn't necessarily be appropriate here.

masonry/src/lib.rs
```diff
@@ -178,7 +178,7 @@ pub use cursor_icon::{CursorIcon, ParseError as CursorIconParseError};
 pub use kurbo::{Affine, Insets, Point, Rect, Size, Vec2};
 pub use parley::layout::Alignment as TextAlignment;
 pub use parley::style::FontWeight;
-pub use vello::peniko::{Color, Gradient};
+pub use vello::peniko::{color::palette, Color, Gradient};
```
DJMcNab, line 181: For reviewers, this is the new change to public API (it wasn't obvious to me on a first pass).

DJMcNab (approved):
I think this is an improvement. Our re-exports are generally a mess, but `masonry::vello::peniko::color::pallette` is clearly obscene.

masonry/src/lib.rs
waywardmonkeys: And a similar in `xilem/src/lib.rs` re-exporting from here (as it already did for `Color`).

PR #803, opened 2024-12-29
Add missing `request_render` call when setting text selection from AccessKit

mwcampbell:
Without this, the updated text selection is neither painted nor reflected in an updated accessibility tree until the user presses a key.


DJMcNab:
What needs to happen to land this? This file being edited has been moved into the `widgets` (rather than `widget`) module, but otherwise the diff should still be correct and an improvement.

PR #807, opened 2025-01-10
Remove redundant code in calc_masonry example

### LINKED ISSUE #796: FIXME comment about mutate pass not being run
In the Masonry calculator example, there is this bit of code:

```
            Update::HoveredChanged(true) => {
                ctx.mutate_later(&mut self.inner, move |mut inner| {
                    SizedBox::set_border(&mut inner, Color::WHITE, 3.0);
                });
                // FIXME - This is a monkey-patch for a problem where the mutate pass isn't run after this.
                // Should be fixed once the pass spec RFC is implemented.
                ctx.request_anim_frame();
            }
```

Is that FIXME still valid and if so, is there an issue tracking the missing functionality?

PoignardAzur:
Fixes #796


DJMcNab (approved):
Tested manually, the hovered borders work

PR #809, opened 2025-01-10
Improve documentation of TestHarness

PoignardAzur:
Add scale_factor parameter.
Add links to doc module.


PoignardAzur:
I'm not gonna lie, I'm super satisfied by the UX of those links. It's nice that rustdoc enforces that the links are correct (if not the jump targets).

PR #810, opened 2025-01-10
Reorder context impl blocks

PoignardAzur:
This makes it so that each context types has its most relevant, unique methods appear earlier in the documentation.


waywardmonkeys (approved):
Assuming you didn't change things other than the order (and the broken intradoc link fix).

PoignardAzur:
I wish there was a tool that could compare two versions of your codebase and say "Yup, it's all the same except things are in different places / with different names".

PR #812, opened 2025-01-10
Remove fill_major_axis flex attribute

PoignardAzur:
See https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/About.20Flex.20fill_major_axis for why. It's not a standard Flex attribute.

### DIFF
--- a/masonry/src/widget/flex.rs
+++ b/masonry/src/widget/flex.rs
@@ -23,7 +23,6 @@ pub struct Flex {
     direction: Axis,
     cross_alignment: CrossAxisAlignment,
     main_alignment: MainAxisAlignment,
-    fill_major_axis: bool,
     children: Vec<Child>,
     old_bc: BoxConstraints,
     gap: Option<f64>,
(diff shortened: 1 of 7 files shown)


jaredoconnell:
I fixed the stopwatch example and removed a stale reference to a function. However, this is not ready for merging. The design of the variable clock example needs to be fixed. It appears that things are taking up all available space individually when only the sum of things should take up all available space.
We'll need to look into how that example was designed and figure out the better alternative implementation.

PoignardAzur:
> I fixed the stopwatch example

Shoot, I forgot to check examples.

> It appears that things are taking up all available space individually when only the sum of things should take up all available space.

Yup, I had a lot of visual problems like this when I tried removing `BoxConstraints::min`, hence my trying a more piecemal approach.

It's weird that we get the problem here, though. Each timezone is constrained in a SizedBox with `.height(72.)`, there shouldn't be a way for it to expand to available space.

PR #815, opened 2025-01-11
Document most items in Masonry

PoignardAzur:
Move `#[allow(missing_docs)]` to two modules. Missing docs now warn everywhere else by default.

Remove `#[expect(rustdoc::broken_intra_doc_links)]` and fix all broken links.


masonry/src/contexts.rs
```diff
@@ -764,59 +767,48 @@ impl_context_method!(
...
-        /// widget can be considered hovered when some other widget has captured the
-        /// pointer (for example, when clicking one widget and dragging to the
-        /// next). The documentation should clearly state the resolution.
+        /// [captured]: crate::doc::doc_06_masonry_concepts#pointer-capture
         pub fn is_hovered(&self) -> bool {
             self.widget_state.is_hovered
         }
 
-        /// Whether the pointer is captured by this widget.
+        /// Whether the mouse is [captured] by this widget.
```
jaredoconnell, line 790: Is there a reason the other changes switch from mouse to pointer, and this line changes from pointer to mouse?

masonry/src/contexts.rs
```diff
@@ -764,59 +767,48 @@ impl_context_method!(
...
-        /// [`register_for_focus`]: UpdateCtx::register_for_focus
-        /// [`Update::FocusChanged`]: crate::Update::FocusChanged
+        /// [text focus]: crate::doc::doc_06_masonry_concepts#text-focus
         /// [`has_focus`]: Self::has_focus
         pub fn is_focused(&self) -> bool {
             self.global_state.focused_widget == Some(self.widget_id())
         }
 
-        /// The (tree) focus status of a widget.
+        /// Whether this widget or any of its descendants is focused.
```
jaredoconnell, line 810: The verb should agree with the plurality of the nearest verb, which is descendants, which is plural. So this should be "are focused"

masonry/src/event.rs
```diff
@@ -181,65 +187,105 @@ impl From<PointerButton> for PointerButtons {
...
 pub enum PointerEvent {
+    /// A pointer was pressed.
     PointerDown(PointerButton, PointerState),
+    /// A pointer was released.
     PointerUp(PointerButton, PointerState),
+    /// A pointer was moved.
     PointerMove(PointerState),
+    /// A pointer entered the window.
     PointerEnter(PointerState),
+    /// A pointer left the window.
```
jaredoconnell, line 204: As documented earlier in this PR, these events can also be triggered due to some edge cases. It may be worth briefly noting that here.

masonry/src/event.rs
```diff
@@ -181,65 +187,105 @@ impl From<PointerButton> for PointerButtons {
...
+    /// A mouse wheel event.
     MouseWheel(LogicalPosition<f64>, PointerState),
+    /// During a file drag-and-drop operation, a file was kept over the window.
     HoverFile(PathBuf, PointerState),
+    /// During a file drag-and-drop operation, a file was dropped on the window.
     DropFile(PathBuf, PointerState),
+    /// A file drag-and-drop operation was cancelled.
     HoverFileCancel(PointerState),
+    /// A pinch gesture was detected.
     Pinch(f64, PointerState),
```
jaredoconnell, line 215: It would be helpful for users to document some of the data stored in the enum's parameters, like the meaning behind the MouseWheel event's logical position, and pinch's floating point value.

jaredoconnell (commented):
This is a great improvement to the documentation. I have a few minor comments.

PoignardAzur:
(For transparency, I'll mention I made heavier user of Copilot than usual for this PR. I don't think it's wrong, but I should probably have signaled this right away, and planned to do my own code review a day after, otherwise it starts to feel like I'm doing a high noise-to-signal contribution and expecting reviewers to filter the mistakes.)

masonry/src/contexts.rs
PoignardAzur: Nope, "any of its descendants" is singular.

masonry/src/event.rs
PoignardAzur: I added some notes, though the "units" used will need to be better documented.

masonry/src/contexts.rs
jaredoconnell: Nope, "any of" is plural. "one of" would be the singular version.

masonry/src/event.rs
PoignardAzur: Done for PointerLeave. PointerEnter doesn't have edge cases.

masonry/src/contexts.rs
PoignardAzur: Huh... I just checked, and apparently it is for enumerable nouns. TIL.

jaredoconnell (approved):
This is a great improvement. I don't see any issues.

masonry/src/event_loop_runner.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+// We use allow because expect(missing_docs) is noisy with rust-analyzer.
```
DJMcNab, line 4: Can you clarify what this means? I'm not seeing any noise if I change this to `expect`

PoignardAzur: I don't see a warning any more either. I guess they fixed whatever bug there was.

PR #816, opened 2025-01-11
Rename `WidgetArena::widget_states` to `WidgetArena::states`

PoignardAzur:


jaredoconnell (approved):
I guess there is enough context for the more brief name.

PR #818, opened 2025-01-12
xilem: Docs updating

ArtyomSinyugin:
**Draft**
This PR is the proposal for xilem docs starting page. But I think I will be able to make some more commits to improve documentation.

### DIFF
--- a/xilem/README.md
+++ b/xilem/README.md
@@ -14,22 +14,130 @@
 
 </div>
 
-Xilem is a UI toolkit.
-It combines ideas from Flutter, SwiftUI, and Elm.
-Like all of these, it uses lightweight view objects, diffing them to provide minimal updates to a retained UI.
-Like SwiftUI, it is strongly typed.
-For more details on Xilem's reactive architecture see [Xilem: an architecture for UI in Rust].
+<!-- We use cargo-rdme to update the README with the contents of lib.rs.
+To edit the following section, update it in lib.rs, then run:
+cargo rdme --workspace-project=color --heading-base-level=0
+Full documentation at https://github.com/orium/cargo-rdme -->
 
-Xilem's reactive layer is built on top of a wide array of foundational Rust UI projects, e.g.:
-* Widgets are provided by [Masonry], which is a fork of the now discontinued [Druid] UI toolkit.
+<!-- Intra-doc links used in lib.rs should be evaluated here.
+See https://linebender.org/blog/doc-include/ for related discussion. -->
+
+[`flex`]: https://docs.rs/xilem/latest/xilem/view/fn.flex.html
+[`grid`]: https://docs.rs/xilem/latest/xilem/view/fn.grid.html
+[`sized box`]: https://docs.rs/xilem/latest/xilem/view/fn.sized_box.html
+[`button`]: https://docs.rs/xilem/latest/xilem/view/fn.button.html
+[`checkbox`]: https://docs.rs/xilem/latest/xilem/view/fn.checkbox.html
+[`image`]: https://docs.rs/xilem/latest/xilem/view/fn.image.html
+[`label`]: https://docs.rs/xilem/latest/xilem/view/fn.label.html
+[`portal`]: https://docs.rs/xilem/latest/xilem/view/fn.portal.html
+[`progress bar`]: https://docs.rs/xilem/latest/xilem/view/fn.progress_bar.html
+[`prose`]: https://docs.rs/xilem/latest/xilem/view/fn.prose.html
+[`spinner`]: https://docs.rs/xilem/latest/xilem/view/fn.spinner.html
+[`task`]: https://docs.rs/xilem/latest/xilem/view/fn.task.html
+[`textbox`]: https://docs.rs/xilem/latest/xilem/view/fn.textbox.html
+[`variable label`]: https://docs.rs/xilem/latest/xilem/view/fn.variable_label.html
+[`zstack`]: https://docs.rs/xilem/latest/xilem/view/fn.zstack.html
+[weight]: https://docs.rs/parley/latest/parley/style/struct.FontWeight.html
+
+<!-- cargo-rdme start -->
+
+`Xilem` is a UI toolkit. It combines ideas from `Flutter`, `SwiftUI`, and `Elm`.
+Like all of these, it uses lightweight view objects, diffing them to provide
+minimal updates to a retained UI. Like `SwiftUI`, it is strongly typed. For more
+details on `Xilem`'s reactive architecture see `Xilem`: an [architecture for UI in Rust].
+
+`Xilem`'s reactive layer is built on top of a wide array of foundational Rust UI projects, e.g.:
+
+* Widgets are provided by [Masonry], which is a fork of the now discontinued `Druid` UI toolkit.
 * Rendering is provided by [Vello], a high performance GPU compute-centric 2D renderer.
-* GPU compute infrastructure is provided by [wgpu].
-* Text support is provided by [Parley], [Fontique], [Swash], and [Skrifa]. 
+* GPU compute infrastructure is provided by wgpu.
+* Text support is provided by [Parley], [Fontique], [swash], and [skrifa].
 * Accessibility is provided by [AccessKit].
 * Window handling is provided by [winit].
 
-Xilem can currently be considered to be in an alpha state.
-Lots of things need improvements.
+`Xilem` can currently be considered to be in an alpha state. Lots of things need improvements.
+
+### Example
+The simplest app looks like this:
+```rust
+use winit::error::EventLoopError;
+use xilem::view::{button, flex, label};
+use xilem::{EventLoop, WidgetView, Xilem};
+
+#[derive(Default, Debug)]
+struct AppState {
+    num: i32,
+}
+
+fn app_logic(data: &mut AppState) -> impl WidgetView<AppState> {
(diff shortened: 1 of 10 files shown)


xilem/src/lib.rs
```diff
@@ -1,8 +1,100 @@
...
+//! <div align="center" style="margin-bottom: 1em">
+//!
+//! <h1 style="margin: 0; line-height: 1em">Xilem</h1>
+//! <h6 style="margin-top: 0">An experimental Rust native UI framework</h6>
+//!
+//! [Linebender Website] · [Github] · [Community in the Xi Zulip] · [Examples]
+//!
+//! </div>
+//!
+//! Xilem is a UI toolkit. It combines ideas from `Flutter`, `SwiftUI`, and `Elm`. Like all of these, it uses lightweight view objects, diffing them to provide minimal updates to a retained UI. Like `SwiftUI`, it is strongly typed. For more details on Xilem's reactive architecture see Xilem: an [architecture for UI in Rust].
```
DJMcNab, line 14: This line is too long.

xilem/src/lib.rs
```diff
@@ -1,8 +1,100 @@
...
+//!     num: i32,
+//! }
+//!
+//! fn app_logic(data: &mut AppState) -> impl WidgetView<AppState> {
+//!     flex((label(format!("{}", data.num)), button("increment", |data: &mut AppState| data.num+=1)))
+//! }
+//!
+//! fn main() -> Result<(), EventLoopError> {
+//!     let app = Xilem::new(AppState::default(), app_logic);
+//!     app.run_windowed(EventLoop::with_user_event(), "Counter".into())?;
```
DJMcNab, line 45: A trick we've used elsewhere is:
```suggestion
//!     # return;
//!     app.run_windowed(EventLoop::with_user_event(), "Counter".into())?;
```
which ensures that the doctest doesn't try to actually run the app.

DJMcNab (commented):
Thanks for providing a foundation here. The CI needs to be green to land this.

xilem/src/lib.rs
waywardmonkeys: Just do: "rust,no_run" after the triple ticks to open the code block if you don't want it to run and just want it to compile.

DJMcNab: Oh yeah, that's better!

xilem/src/lib.rs
PoignardAzur, line 14: If you want to converge the documentation's main page with the README, you should probably integrate it with `cargo rdme` the same way we did with Masonry, and add it to CI.

DJMcNab: I avoided making that suggestion here as I intend to do it as a follow-up. If it were done in this PR, that would be excellent.

ArtyomSinyugin:
> Thanks for providing a foundation here. The CI needs to be green to land this.

Thank you for comments. Before I left some red alerts for purpose: I want to do some more work inside this PR. So, now I am going to look at Masonry docs and then will make some inserts and updates.

xilem/src/lib.rs
ArtyomSinyugin: I will try to do this.

ArtyomSinyugin: It’s done. But I removed all original text from readme. Maybe I should retrieve a part of it and make it intact…

DJMcNab (changes requested):
As mentioned, the new README content needs to be added to the "formatting" CI job.

I'm willing to make the corresponding changes to the README myself, outside of this PR, if you prefer.

xilem/README.md
```diff
@@ -1,85 +1,153 @@
...
+[swash]: https://crates.io/crates/swash
+[skrifa]: https://crates.io/crates/skrifa
 [AccessKit]: https://crates.io/crates/accesskit
-[winit]: https://crates.io/crates/winit
-[Xilem: an architecture for UI in Rust]: https://raphlinus.github.io/rust/gui/2022/05/07/ui-architecture.html
-[Rust code of conduct]: https://www.rust-lang.org/policies/code-of-conduct
+[`flex`]: https://docs.rs/xilem/latest/xilem/view/flex/
+[`grid`]: https://docs.rs/xilem/latest/xilem/view/grid/
+[`lens`]: core::lens
+[`map state`]: core::map_state
```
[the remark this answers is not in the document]
DJMcNab: Your first point is the very subtle thing, but these specific links to the web ONLY apply to the README.md file, not to the lib.rs. That is achieved by placing them *above* the `cargo-rdme start` section, so that they are the preferred link definition.

Your second point is true, but there isn't really a better link we could use, unfortunately.

For the third point, we choose to use `/latest/`, and just live with the fact that these links will become outdated if people are browsing ancient versions of our README.

> So, I think it is useful to write the list of view elements in lib.rs, but there should be a way to restrict cargo rdme in including all information from lib.rs.

Fwiw, I don't really agree that the full list of view elements is very useful, either way.
The point of a README file is to give an overview - the individual page of the `views` module is where you can find all of the available views. We absolutely should point to the documentation of the Xilem Core views, though

xilem/README.md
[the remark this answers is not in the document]
ArtyomSinyugin: I tried to do this task. Hope it's all right.

xilem/README.md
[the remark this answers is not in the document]
ArtyomSinyugin: All links exept xilem_core now like `https://docs.rs/xilem/0.1.0/xilem/view/fn.flex.html`

If it is needed, I will remove `view elements` from main page of the docs. But In my opinion with given example and building elements it is the most straightforward way to say `just build your app`.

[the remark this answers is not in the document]
ArtyomSinyugin: Finally I understand how it works. Thank you. I've pushed changes. 
But if there are any mistakes I don't mind if you fix it outside of this PR.

DJMcNab:
I've converted to draft based on your **Draft** in the top-level body of the PR. When you're ready for this to be reviewed/land, please remove it from draft (or one of us can finish it off if you ask)

ArtyomSinyugin:
> I've converted to draft based on your **Draft** in the top-level body of the PR. When you're ready for this to be reviewed/land, please remove it from draft (or one of us can finish it off if you need)

I think I've finished. 
I would like to describe in docs some more elements, but for `image`, `portal`, `progress_bar`, prose`, `spinner`, `task`, etc. it take much more time for me to figure out how these elements work. 

Maybe I'll do the rest later, maybe in the next PR (if nobody does it in the near future).

PoignardAzur (approved):
LGTM.

jaredoconnell (approved):

xilem/src/lib.rs
```diff
@@ -1,8 +1,104 @@
...
+//!
+//! * [`flex`][crate::view::flex]: layout defines how items will be arranged in a row or column
+//! * [`grid`][crate::view::grid]: layout divides a window into regions and defines the relationship
+//!   between inner elements in terms of size and position
+//! * [`sized_box`][crate::view::sized_box]: forces its child to have a specific width and/or height
+//! * [`button`][crate::view::button]: basic button element
+//! * [`image`][crate::view::image]: displays a bitmap image
+//! * [`portal`][crate::view::portal]: a scrollable region
+//! * [`progress_bar`][crate::view::progress_bar]: progress bar element
+//! * [`prose`][crate::view::prose]: displays immutable text which can be selected within
```
tomcur, line 78:
```suggestion
//! * [`prose`][crate::view::prose]: displays immutable, selectable text
```

tomcur (approved):
This looks good. I've provided some minor README suggestions inline.

xilem/src/lib.rs
```diff
@@ -1,8 +1,104 @@
...
+//! A key feature of Xilem's architecture is that `Counter`, the application's state, is an arbitrary `'static` Rust type.
+//! In this example, `app_logic` is the root component, which creates the view value it returns.
+//! This, in turn, leads to corresponding Masonry widgets being created, in this case a button and a label.
+//! When the button is pressed, the number will be incremented, and then `app_logic` will be re-ran.
+//! The returned view will be compared with its previous value, whichi will minimally update the contents of these widgets.
+//! In this case, because the button is still the same, it will not be updated, and the label will show its new value.
+//!
+//! More examples can be found [in the repository][xilem_examples]
+//!
+//! **Note: The linked examples are for the `main` branch of Xilem. If you are using a released version, please view the examples in the tag for that release**
```
tomcur, line 60:
```suggestion
//! **Note: The linked examples are for the `main` branch of Xilem. If you are using a released version, please view the examples in the tag for that release.**
```

xilem/src/lib.rs
```diff
@@ -1,8 +1,104 @@
...
+//!     Ok(())
+//! }
+//! ```
+//!
+//! A key feature of Xilem's architecture is that `Counter`, the application's state, is an arbitrary `'static` Rust type.
+//! In this example, `app_logic` is the root component, which creates the view value it returns.
+//! This, in turn, leads to corresponding Masonry widgets being created, in this case a button and a label.
+//! When the button is pressed, the number will be incremented, and then `app_logic` will be re-ran.
+//! The returned view will be compared with its previous value, whichi will minimally update the contents of these widgets.
+//! In this case, because the button is still the same, it will not be updated, and the label will show its new value.
```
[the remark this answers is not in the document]
DJMcNab: This is already in:

> and the label will show its new value.

You're saying that we want a step "before" that? Seems reasonable

tomcur (approved):

PR #819, opened 2025-01-13
shadow and slider

Cupnfish:
https://github.com/user-attachments/assets/ca6e9c75-3a44-44cb-a04e-ba8f9ecd0bbe

This is a simple implementation with many areas that can still be improved. It serves as a reference for those who want to implement corresponding functionalities in the future. Currently, I haven't thoroughly researched the testing part, and the tests are mostly not passing. However, it can run a simple example


DJMcNab:
A few notes:
- What is @snooowfire's role in this PR?
- It might be worth splitting this PR into two PRs the slider (first?) and the shadow.
- Shadow might be able to use [`draw_blurred_rounded_rect`](https://docs.rs/vello/latest/vello/struct.Scene.html#method.draw_blurred_rounded_rect) (optionally, because it's somewhat expensive).

PoignardAzur:
> * Shadow might be able to use [`draw_blurred_rounded_rect`](https://docs.rs/vello/latest/vello/struct.Scene.html#method.draw_blurred_rounded_rect) (optionally, because it's somewhat expensive).

It is?

DJMcNab:
Sure - the per-pixel maths is significantly more expensive than just blitting a solid colour. Unless you mean that we should have a fast-path in Vello (i.e. splitting blurred_rounded_rect into the solid part and the non-solid part), which I guess would be not too unreasonable

PoignardAzur:
> Sure - the per-pixel maths is significantly more expensive than just blitting a solid colour.

Oh, I thought you meant it had some sort of hidden cost.

This specific use-case is the one we added blurred rectangles for, so I wouldn't worry about the per-pixel cost.

masonry/src/widget/sized_box.rs
```diff
@@ -520,6 +587,24 @@ impl Widget for SizedBox {
...
+                .to_rounded_rect(corner_radius);
+
+            scene.draw_blurred_rounded_rect_in(
+                &shadow_rect,
+                Affine::translate(shadow.offset),
+                shadow_rect.rect(),
+                shadow.color,
+                shadow.corner_radius.unwrap_or(corner_radius.top_left),
+                shadow.blur_radius,
+            );
```
jaredoconnell, line 606: Why are you using the clipped version of the blurred rect? The `_in` version. The unclipped version is how I would expect it to look.
image

PR #820, opened 2025-01-13
Implement very basic widget inspector

PoignardAzur:


masonry/src/passes/update.rs
```diff
@@ -584,6 +584,17 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot) {
 
     let pointer_pos = root.last_mouse_pos.map(|pos| (pos.x, pos.y).into());
 
+    if root.global_state.inspector_state.is_picking_widget {
+        if let Some(pos) = pointer_pos {
+            root.global_state.inspector_state.hovered_widget = root
+                .get_root_widget()
+                .find_widget_at_pos(pos)
```
DJMcNab, line 591: This doesn't seem to find the label inside a `Button`, for some reason

masonry/src/tracing_backend.rs
```diff
@@ -142,14 +142,11 @@ pub(crate) fn try_init_test_tracing() -> Result<(), SetGlobalDefaultError> {
...
+    // Default level is INFO unless a level is passed.
+    // We tried using DEBUG in --dev, but it was too noisy.
     // DEBUG should print a few logs per low-density event.
     // INFO should only print logs for noteworthy things.
-    let default_level = if cfg!(debug_assertions) {
-        LevelFilter::DEBUG
-    } else {
-        LevelFilter::INFO
-    };
+    let default_level = LevelFilter::INFO;
```
DJMcNab, line 149: This is a slightly weird change to have in this PR. I don't fundamentally disagree with it, but it seems quite orthogonal

masonry/src/passes/event.rs
```diff
@@ -103,6 +105,30 @@ pub(crate) fn run_on_pointer_event_pass(root: &mut RenderRoot, event: &PointerEv
...
+        let target_widget_id = get_pointer_target(root, event.position());
+        if let Some(target_widget_id) = target_widget_id {
+            root.global_state
+                .emit_signal(RenderRootSignal::WidgetSelectedInInspector(
+                    target_widget_id,
+                ));
+        }
+        root.global_state.inspector_state.is_picking_widget = false;
+        root.global_state.inspector_state.hovered_widget = None;
+        return Handled::Yes;
```
DJMcNab, line 129: The widget under the cursor isn't hovered once this happens until you move the mouse. That's *fine*, but it would be nicer if it fixed itself.

Fwiw, the previous hover state being retained is probably the right behaviour, as it lets you inspect the hovered widget.

masonry/src/event_loop_runner.rs
```diff
@@ -732,6 +732,17 @@ impl MasonryState<'_> {
                 render_root::RenderRootSignal::ShowWindowMenu(position) => {
                     window.show_window_menu(position);
                 }
+                render_root::RenderRootSignal::WidgetSelectedInInspector(widget_id) => {
+                    let (widget, state) = self.render_root.widget_arena.get_pair(widget_id);
+                    let widget_name = widget.item.short_type_name();
+                    let display_name = if let Some(debug_text) = widget.item.get_debug_text() {
+                        format!("{widget_name}<{debug_text}>")
```
DJMcNab, line 739: This is unfortunately sparse for a lot of widgets. I suppose that makes a good follow-up?

masonry/src/event_loop_runner.rs
```diff
@@ -732,6 +732,17 @@ impl MasonryState<'_> {
...
+                    } else {
+                        widget_name.into()
+                    };
+                    println!("Widget selected in inspector: {widget_id} - {display_name}");
+                    println!("{:#?}", state.item);
```
DJMcNab, line 744: Whereas this is extremely large. I don't necessarily think there's anything we can do about it though...

masonry/src/event_loop_runner.rs
```diff
@@ -732,6 +732,17 @@ impl MasonryState<'_> {
...
+                }
```
DJMcNab, line 745: There would be reasonable arguments for printing the broad shape of the children tree as well. Not necessarily the full widget state, but the names and debug text. Maybe? limited to a certain depth(/length?)

DJMcNab (commented):
This should help make some things a bit easier to reason about.

Can we please also document this in Masonry's README/lib.rs?

masonry/src/passes/event.rs
```diff
@@ -181,6 +207,15 @@ pub(crate) fn run_on_text_event_pass(root: &mut RenderRoot, event: &TextEvent) -
...
+
+        if key.physical_key == PhysicalKey::Code(KeyCode::F11)
+            && key.state == ElementState::Pressed
+            && handled == Handled::No
+        {
+            root.global_state.inspector_state.is_picking_widget =
+                !root.global_state.inspector_state.is_picking_widget;
+            handled = Handled::Yes;
+        }
     }
```
DJMcNab, line 219: Would it be worth having a similar keybinding for toggling debug paint? That's probably a follow-up/different PR.

masonry/src/passes/update.rs
PoignardAzur: The search uses pointer routing, which labels inside buttons opt out of. We should ideally fix that, but it requires more design work.

masonry/src/tracing_backend.rs
PoignardAzur: The main reason is that the log noise was making the printed text harder to see (though less so now that I've added the full debug print).

PoignardAzur: I rolled it back for now.

jaredoconnell (approved):
Seems like a good basic inspector. It's nice that we can now toggle debug paint without messing with the environment variable.

jaredoconnell:
I do agree with Daniel that some documentation in the README would be good. I needed to scan the code to find out how to use this.

DJMcNab (approved):

masonry/README.md
```diff
@@ -108,6 +108,12 @@ The following feature flags are available:
 - `tracy`: Enables creating output for the [Tracy](https://github.com/wolfpld/tracy) profiler using [`tracing-tracy`][tracing_tracy].
   This can be used by installing Tracy and connecting to a Masonry with this feature enabled.
 
+### Debugging features
+
+Masonry apps currently ship with two debugging features built in:
+- A rudimentary widget inspector - toggled by F11 key.
```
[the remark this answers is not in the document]
jaredoconnell: Technically it can be toggled off with the key.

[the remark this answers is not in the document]
PoignardAzur: I think "toggled" is right.

jaredoconnell:
Do you think the Ubuntu failure has anything to do with Github Actions' switch from Ubuntu 22 to 24?

waywardmonkeys:
@jaredoconnell Yes, and it is fixed on `main`

PoignardAzur:
Alright, I'll rebase.

PR #821, opened 2025-01-14
Make `DriverCtx` publicly constructable again

DJMcNab:
Follow-up to #808

Also adds a test to limit the risk of future regression.


waywardmonkeys (approved):

waywardmonkeys:
Thanks for this!

PR #823, opened 2025-01-14
Move cursor to end of text area when text is changed

PoignardAzur:
This matches the behavior of other frameworks in similar situations.


DJMcNab (approved):
I agree here. If you're calling reset text, the user isn't actively editing.

jaredoconnell (commented):
I agree that when the text changes, the cursor should go to the end.
What I'm not sure about is moving the cursor to the end when losing focus. I checked and on the web the position is retained when tabbing off of a textbox and tabbing back to it.

PoignardAzur:
Yeah, and now that I think about it, the case I cared about (lose focus, text changes, get focus back) already gets the behavior I want (cursor is at the end) with the first commit.

I'll remove the second commit.

jaredoconnell (approved):

PR #826, opened 2025-01-15
Document the IME cursor area workaround

DJMcNab:


masonry/src/render_root.rs
```diff
@@ -127,7 +127,7 @@ pub(crate) struct RenderRootState {
     /// Whether an IME session is active.
     pub(crate) is_ime_active: bool,
 
-    /// The area in which text is being edited.
+    /// The cursor area last sent to the platform.
```
DJMcNab, line 130: (Ideally I'd choose to expand this here, with something like "\n\nUsed to avoid an infinite loop on some platforms" but that has been vetoed)

Philipp-M (approved):

PR #827, opened 2025-01-16
Edit TreeArena tests

PoignardAzur:
Document tests more.
Fix a lifetime error in unsafe API.

Remove the `mem::swap` tests. While the action they were testing was possible, sharing data between two separate arenas isn't meant to be part of the API and is unlikely to happen in the wild.


tree_arena/tests/basic_tests.rs
```diff
@@ -31,107 +44,77 @@ fn arena_tree_test() {
...
-    // node 1 from tree a now believes it is node 4 from tree b
-    assert_eq!(node_1_a.id(), 4_u64, "Node 1 id should be 4");
-    // however it still contains the item from tree a
-    assert_eq!(*node_1_a.item, 'a', "Node 1 item should be 'a'");
-    // and we can access the nodes in tree b, that node 4 was able to
-    assert_eq!(
-        *node_1_a.children.get_child(2_u64).unwrap().item,
-        'g',
-        "Node 2 item in tree b should be g"
-    );
```
DJMcNab, line 136: It's not clear to me that this test is sufficiently useless to justify removing. This is both:
1) an operation which can be performed safely
2) A weird thing to do.

and so I think we should test that it doesn't cause UB.

.github/workflows/ci.yml
```diff
@@ -292,6 +292,9 @@ jobs:
           # See also https://github.com/linebender/vello/pull/610
           SKIP_RENDER_TESTS: ${{ matrix.skip_gpu }}
 
+      - name: Run cargo nextest on unsafe arena
+        run: cargo nextest run -p tree_arena --locked --no-fail-fast --no-default-features
```
DJMcNab, line 296: Ideally we'd use miri here, I think?

waywardmonkeys: We run Miri in the `color` CI so you can take a peek there.

tree_arena/tests/basic_tests.rs
KGrewal1: Yes this was my motivation for having the test, or something similar, as `mem::swap` should be safe, but wanted to make sure that doing something with it wouldn't allow any of the preconditions to be broken (specifically that you couldn't access a parent from a child)

PoignardAzur: I guess I'll re-add it with a comment, then. I'll try to make a shorter version.

DJMcNab (commented):
I'd like to get @KGrewal1's review here as well, if you're willing to do so.

KGrewal1:
LGTM!



 In terms of logic of the PR, this looks good to me: most changes are renames so diff larger than acc changes. On the unsafe side are using `reborrow` for `has` on the `ArenaMutList` (which is correct and reduces duplication of the logic), and fixing the signature of `find` so it takes `&self` by reference (and not `self`), bringing it in line with the safe version and with the appropriate lifetime.

On the safe tree changes, only

DJMcNab (approved):

PR #831, opened 2025-01-16
Apply is/has naming convention for flags and methods

PoignardAzur:
Rename has_pointer_capture to is_pointer_capture_target.
Rename has_focus to has_focus_target.
Rename is_focused to is_focus_target.
Add has_hovered flag.
Add ChildHoverChanged event.

### DIFF
--- a/masonry/src/passes/update.rs
+++ b/masonry/src/passes/update.rs
@@ -340,7 +340,7 @@ fn update_focus_chain_for_widget(
         return;
     }
 
-    // Replace has_focus to check if the value changed in the meantime
+    // Replace has_focused to check if the value changed in the meantime
     state.item.has_focused = global_state.focused_widget == Some(id);
     let had_focus = state.item.has_focused;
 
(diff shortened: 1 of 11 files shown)


waywardmonkeys (commented):
I understand why you're doing it, but `has_focused` reads pretty poorly for an English speaker.

PoignardAzur:
True. Let's go for "has_focus_target" for now, then.

waywardmonkeys:
For the commit message to be correct, you'll have to edit the main summary at the top of this PR page, right?

PoignardAzur:
Done.

masonry/src/passes/update.rs
```diff
@@ -677,6 +686,21 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot) {
         }
     }
 
+    if prev_hovered_widget != next_hovered_widget {
+        if let Some(prev_hovered_widget) = prev_hovered_widget {
+            run_single_update_pass(root, prev_hovered_widget, |widget, ctx| {
```
DJMcNab, line 691: This panics if `prev_hovered_widget` is no longer in the tree...

PoignardAzur: Fixed in #834.

PR #832, opened 2025-01-16
Add screenshots to documentation

PoignardAzur:
Create include_screenshot macro


PoignardAzur:
For now I've added a screenshot to the Align widget only. A full PR would add screenshots to all of them.

Follow-up work needed:

- Checking whether the docrs trick actually works.
- Checking whether this method will work with `cargo rdme`.
- Making better screenshots than the ones we currently use.

PoignardAzur:
The screenshots could stand to be improved (a lot of them are way too big), but for now this is already a marked improvement.

masonry/examples/readme.md
```diff
@@ -1,7 +1,42 @@
...
+Simples possible Masonry app.
+
+## `two_textboxes`
+
+Example used to test text input and text focus.
+
+
+## `calc_masonry`
+
+![alt text](screenshots/calc_masonry__tests__base_screenshot.png)
```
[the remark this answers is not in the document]
PoignardAzur: I lean towards not including any alt text. The most obvious alt text I can think to add would either be super verbose or just a restatement of the paragraph next to it.

masonry/src/testing/screenshots.rs
```diff
@@ -6,6 +6,38 @@
...
+            "masonry-v", env!("CARGO_PKG_VERSION"), "/masonry/src/", $path,
+            ")",
+        )
+    };
+}
+
+// TODO:
+// - Check whether the docs.rs trick actually works.
+// - Check whether this method will work with cargo rdme.
+// - Make better screenshots than the ones we currently use.
```
DJMcNab, line 26: This needs an issue

masonry/src/testing/screenshots.rs
```diff
@@ -6,6 +6,38 @@
...
+            "(", "https://media.githubusercontent.com/media/linebender/xilem/",
+            "masonry-v", env!("CARGO_PKG_VERSION"), "/masonry/src/", $path,
+            ")",
+        )
+    };
+}
+
+// TODO:
+// - Check whether the docs.rs trick actually works.
+// - Check whether this method will work with cargo rdme.
```
DJMcNab, line 25: So this is a slightly messy one. The short answer is that it won't, and the long answer is that you can make it work, but it's messy, and the even longer answer is that even after making it work, I don't know if it will work when published to crates.io.

Either way, I'm not sure that this comment as-is is useful in this PR. Maybe there should be a comment at `/src/lib.rs`. Something like: `// TODO: Add screenshot. This can't use include_screenshot as that doesn't work with cargo-rdme`

masonry/src/testing/screenshots.rs
```diff
@@ -6,6 +6,38 @@
...
+            "![", $($caption,)? "]",
+            "(", "https://media.githubusercontent.com/media/linebender/xilem/",
+            "masonry-v", env!("CARGO_PKG_VERSION"), "/masonry/src/", $path,
+            ")",
+        )
+    };
+}
+
+// TODO:
+// - Check whether the docs.rs trick actually works.
```
DJMcNab, line 24: This isn't really actionable as a comment here. I'd probably just remove it

PoignardAzur:
Created #851 and 852 to track follow-up work.

DJMcNab (approved):
Thanks! This trick might be worth making a blog post on, although the LFS trickery makes it more complicated.

PR #833, opened 2025-01-17
Emoji picker example improvements

jaredoconnell:
This PR includes improvements to page selector.

Specifically it now shows the full percentage range instead of just the start of it, and it hides the buttons when you reach the end of the range.

For some reason it panics when I get to the end.

### DIFF
--- a/xilem/examples/emoji_picker.rs
+++ b/xilem/examples/emoji_picker.rs
@@ -57,9 +57,12 @@ fn picker(data: &mut EmojiPagination) -> impl WidgetView<EmojiPagination> {
             };
             let view = flex((
                 // TODO: Expose that this button corresponds to the label below for accessibility?
-                sized_box(button(label(emoji.display).text_size(200.0 / data.size as f32), move |data: &mut EmojiPagination| {
-                    data.last_selected = Some(idx);
-                }))
+                sized_box(button(
+                    label(emoji.display).text_size(200.0 / data.size as f32),
+                    move |data: &mut EmojiPagination| {
+                        data.last_selected = Some(idx);
+                    },
+                ))
                 .expand_width(),
                 sized_box(
                     prose(emoji.name)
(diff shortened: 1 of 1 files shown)


DJMcNab:
The panic is as discussed in xilem#831, and is unrelated. It's the equivalent of https://github.com/linebender/xilem/pull/761 but for hover

PoignardAzur:
Panic should be fixed by #834.

xilem/examples/emoji_picker.rs
```diff
@@ -91,19 +91,25 @@ fn paginate(
...
-    let percentage = (current_start * 100) / max_count;
+    let current_end = (current_start + count_per_page).min(max_count);
+    let percentage_start = (current_start * 100) / max_count;
+    let percentage_end = (current_end * 100) / max_count;
 
     flex((
         // TODO: Expose that this is a previous page button to accessibility
-        button("<-", move |data| {
-            *data = current_start.saturating_sub(count_per_page);
+        (current_start != 0).then(|| {
```
DJMcNab, line 100: Ideally, we'd disable the button instead of hiding it, but that's not currently exposed to Xilem (it will likely end up looking similar to transform)

xilem/examples/emoji_picker.rs
DJMcNab, line 1: Whilst we're here, could we increase the font size on the buttons (which is possible since #797)

DJMcNab (approved):
Thanks! This is nice

xilem/examples/emoji_picker.rs
jaredoconnell: Done. It looked bad so I got the idea of using emoji. This isn't great for universal compatibility, but of all examples this is the most appropriate. This may even help with default accessibility.
image

DJMcNab: (I also meant the buttons associated with each option)

jaredoconnell: That makes sense. Pushed. It's definitely even better now.

DJMcNab (approved):
This breaks down a little bit on horizontal Android, but overall that's not worth worrying about. Thanks!

jaredoconnell:
> This breaks down a little bit on horizontal Android, but overall that's not worth worrying about. Thanks!

Good to know. A long-term goal is definitely to improve responsiveness. I'll be available to review a follow up PR if you choose to do that.

PR #834, opened 2025-01-18
Fix `run_single_update_pass` to avoid crashes

PoignardAzur:


waywardmonkeys (commented):
Is this something where there could be a test to ensure it doesn’t regress?

PoignardAzur:
There should probably be one, yes.

jaredoconnell (approved):
I can confirm that this fixes the problem I found on my branch. This is also a code improvement. Less duplication.
Adding a test to catch regressions would be a good idea.

PoignardAzur:
Leaving tests for later, sorry.

PR #840, opened 2025-01-20
Update to Vello 0.4, Peniko 0.3.1, Color 0.2.3

waywardmonkeys:
This also lets us update to using `Color::from_rgb8()` when there is no need for alpha to be specified.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,13 +104,13 @@ clippy.duplicated_attributes = "allow"
 masonry = { version = "0.2.0", path = "masonry" }
 xilem_core = { version = "0.1.0", path = "xilem_core" }
 tree_arena = { version = "0.1.0", path = "tree_arena" }
-vello = { git = "https://github.com/linebender/vello", rev = "d5ebe12737861c521d7ce6ce02b800278367d07d" }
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -104,13 +104,13 @@ clippy.duplicated_attributes = "allow"
...
 tree_arena = { version = "0.1.0", path = "tree_arena" }
-vello = { git = "https://github.com/linebender/vello", rev = "d5ebe12737861c521d7ce6ce02b800278367d07d" }
+vello = "0.4.0"
 wgpu = "23.0.1"
 kurbo = "0.11.1"
 parley = { git = "https://github.com/linebender/parley", rev = "16b62518d467487ee15cb230fdb28530d89a8cf6", features = [
     "accesskit",
 ] }
-peniko = "0.3.0"
+peniko = "0.3.1"
```
DJMcNab, line 113: Hmm, do we need this dependency? I guess we use it in Xilem web?

DJMcNab (approved):

Cargo.toml
waywardmonkeys: Correct.

PR #841, opened 2025-01-21
Add features to help test Textbox

PoignardAzur:


PoignardAzur:
This is marked as a draft because the screenshot is currently incorrect (the selected area isn't highlighted).

PoignardAzur:
Aaand fixed.

jaredoconnell (approved):

masonry/src/event.rs
```diff
@@ -458,6 +459,30 @@ impl PointerEvent {
...
+    /// **selected** is the part of the preedit text that should be selected.
+    ///
+    /// ## Panics
+    ///
+    /// If **selected** isn't a substring of **text**.
+    pub fn preedit_with_cursor(text: String, selected: String) -> Self {
+        let Some(offset) = text.find(&selected) else {
+            panic!("Error building Preedit event: '{selected}' not found in '{text}'");
+        };
+        let span = (offset, selected.len());
```
DJMcNab, line 482: In most cases for a pre-edit, we expect the second parameter to be of the form `(x, x)`. We should also be trying to test that case

masonry/src/widget/text_area.rs
```diff
@@ -477,6 +477,32 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
...
+    /// Set the selection to the given byte range.
+    ///
+    /// No-op if either index is not a char boundary.
+    pub fn select_byte_range(this: &mut WidgetMut<'_, Self>, start: usize, end: usize) {
+        let (fctx, lctx) = this.ctx.text_contexts();
+        this.widget
+            .editor
+            .driver(fctx, lctx)
+            .select_byte_range(start, end);
+        this.ctx.request_render();
```
DJMcNab, line 490: Why is this needed? You request a layout directly below

masonry/src/widget/textbox.rs
```diff
@@ -105,6 +105,8 @@ impl Textbox {
         this.widget.clip = clip;
         this.ctx.request_layout();
     }
+
+    //pub fn set
```
DJMcNab, line 109: What is this for?

masonry/src/event.rs
```diff
@@ -3,6 +3,7 @@
 
 //! Events.
 
+use std::panic;
```
DJMcNab, line 6: `panic!` is part of the prelude, and we don't seem to be using anything from the `panic` module here. Indeed, this still compiles removing this import.

I'm surprised this doesn't give a warning of any kind.

masonry/src/widget/text_area.rs
PoignardAzur: Oh wow, I left a lot of cruft from debugging.

PR #842, opened 2025-01-21
Remove some cruft

PoignardAzur:
Remove outdated TODO comments.
Rewrite some other TODO comments.
Remove references to tarpaulin.

(cargo-tarpaulin is a tool used for code coverage. Masonry hasn't used it for years, and anyway we'll probably use `#[coverage]` annotations once they stabilize.)

### DIFF
--- a/masonry/src/render_root.rs
+++ b/masonry/src/render_root.rs
@@ -435,21 +435,12 @@ impl RenderRoot {
     /// Returns an update to the accessibility tree and a Vello scene representing
     /// the widget tree's current state.
     pub fn redraw(&mut self) -> (Scene, TreeUpdate) {
-        if self.root_state().needs_layout {
-            // TODO - Rewrite more clearly after run_rewrite_passes is rewritten
-            self.run_rewrite_passes();
-        }
-        if self.root_state().needs_layout {
-            warn!("Widget requested layout during layout pass");
-            self.global_state
-                .emit_signal(RenderRootSignal::RequestRedraw);
-        }
+        self.run_rewrite_passes();
 
         // TODO - Handle invalidation regions
-        (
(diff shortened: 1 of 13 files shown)


masonry/src/render_root.rs
```diff
@@ -435,21 +435,12 @@ impl RenderRoot {
...
+        self.run_rewrite_passes();
 
         // TODO - Handle invalidation regions
-        (
-            run_paint_pass(self),
-            run_accessibility_pass(self, self.scale_factor),
-        )
+        let scene = run_paint_pass(self);
+        let tree_update = run_accessibility_pass(self, self.scale_factor);
+        (scene, tree_update)
```
jaredoconnell, line 443: The called function doesn't use needs_layout. Is that change of logic okay?

DJMcNab: `run_rewrite_passes` does call `run_layout_pass`

masonry/src/widget/image.rs
```diff
@@ -214,7 +214,7 @@ mod tests {
             harness.render()
         };
 
-        // TODO - write comparison function
+        // TODO - Use kompari instead
```
DJMcNab, line 217:
```suggestion
        // TODO - Use Kompari instead
```

DJMcNab (approved):

masonry/src/render_root.rs
PoignardAzur: > The called function doesn't use needs_layout

All the passes do their own short-circuiting.

Here's the one for layout: https://github.com/linebender/xilem/blob/299429b641e8cdc931939769dc5d72723376574e/masonry/src/passes/layout.rs#L193-L196

PR #843, opened 2025-01-21
Rename modules to follow new convention

PoignardAzur:
Rename module files from `foobar/mod.rs` to `foobar/_foobar.rs`.

The full name is for readability when multiple module files are open in a text editor. The leading underscore is so these names appear first in file hierarchies.


PoignardAzur:
There should probably be an open debate as to whether we adopt this convention.

The upsides are as listed in the commit message.
The downsides are that this is surprising for someone reading the code eg on Github.

DJMcNab (changes requested):
I strongly object to this - we should not break from idiomatic Rust without very good reason.
I'm happy to be overruled if there's strong consensus in the other direction, especially if this is to be localised to Masonry.

I understand that you use an editor which is derivative of the VScode project, and so would suggest you use the [Customise Tab Labels](https://code.visualstudio.com/docs/getstarted/userinterface#_customize-tab-labels) feature for this purpose.

PoignardAzur:
> I know that vscode has some neat tricks around file organisation contributed by extensions

It doesn't, actually: https://github.com/microsoft/vscode/issues/27286

In any case, I'm not going to push for this if there's strong objection.

canklot:
> I know that vscode has some neat tricks around file organisation contributed by extensions

You can sort your files in the VS code using SortMyFiles extensions 
https://marketplace.visualstudio.com/items?itemName=CanklotSoftware.SortMyFiles

PR #847, opened 2025-01-22
Apply rustfmt to entire project

PoignardAzur:
This PR applies Rust fmt with these two options uncommented:

```toml
imports_granularity = "Module"
group_imports = "StdExternalCrate"
```

I'm creating it as a support for explaining the problems I have with the second option.

### DIFF
--- a/masonry/src/widget/mod.rs
+++ b/masonry/src/widget/mod.rs
@@ -37,7 +37,6 @@ mod variable_label;
 mod widget_arena;
 mod zstack;
 
-pub use self::image::Image;
 pub use align::Align;
 pub use button::Button;
 pub use checkbox::Checkbox;
@@ -55,15 +54,15 @@ pub use split::Split;
 pub use text_area::TextArea;
 pub use textbox::Textbox;
 pub use variable_label::VariableLabel;
+pub(crate) use widget_arena::WidgetArena;
 pub use widget_mut::WidgetMut;
+pub(crate) use widget_pod::CreateWidget;
 pub use widget_pod::WidgetPod;
 pub use widget_ref::WidgetRef;
+pub(crate) use widget_state::WidgetState;
 pub use zstack::{Alignment, ChildAlignment, HorizontalAlignment, VerticalAlignment, ZStack};
 
-pub(crate) use widget_arena::WidgetArena;
-pub(crate) use widget_pod::CreateWidget;
-pub(crate) use widget_state::WidgetState;
-
+pub use self::image::Image;
 use crate::{Affine, Size};
 
 // These are based on https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -141,23 +141,22 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use masonry::dpi::LogicalSize;
+pub use masonry::event_loop_runner::{EventLoop, EventLoopBuilder};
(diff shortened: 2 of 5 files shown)


masonry/src/widget/mod.rs
```diff
@@ -55,15 +54,15 @@ pub use split::Split;
...
 pub use widget_ref::WidgetRef;
-pub use zstack::{Alignment, ChildAlignment, HorizontalAlignment, VerticalAlignment, ZStack};
-
-pub(crate) use widget_arena::WidgetArena;
-pub(crate) use widget_pod::CreateWidget;
 pub(crate) use widget_state::WidgetState;
+pub use zstack::{Alignment, ChildAlignment, HorizontalAlignment, VerticalAlignment, ZStack};
 
+pub use self::image::Image;
 use crate::{Affine, Size};
```
PoignardAzur, line 66: The re-export of `Image` is split from the other re-exports and bundled with non-exported imports.

xilem/src/lib.rs
```diff
@@ -141,23 +141,22 @@ use std::collections::HashMap;
...
+pub use masonry::widget::LineBreaking;
 use masonry::widget::{RootWidget, WidgetMut};
+pub use masonry::{dpi, palette, Affine, Color, FontWeight, TextAlignment, Vec2};
 use masonry::{event_loop_runner, FromDynWidget, Widget, WidgetId, WidgetPod};
+/// Tokio is the async runner used with Xilem.
+pub use tokio;
 use view::{transformed, Transformed};
 use winit::error::EventLoopError;
 use winit::window::{Window, WindowAttributes};
+pub use xilem_core as core;
```
PoignardAzur, line 154: `pub` re-exports are mixed with imports.

PoignardAzur (commented):
After writing this, I think my main objection is that rustfmt's grouping ignores the visibility level of use directives. If they added a setting that let you split imports and re-exports then 90% of my objections go away.

masonry/src/widget/mod.rs
PoignardAzur: We could probably address that by factoring `ObjectFit` out (thus removing the `use crate::` part) and doing `pub use self::` for all the other re-exports.

xStrom:
For the re-export special casing I found [rustfmt#4070](https://github.com/rust-lang/rustfmt/issues/4070) but there isn't really much going on there.

For the `pub(crate)` situation there's [a bit more activity](https://github.com/rust-lang/rustfmt/issues/5083#issuecomment-1951011779) and so maybe eventually it'll get addressed.

Improvements to this autoformatting would definitely be nice. Now the question is, are these problems big enough to not do any autoformatting? Do you consider the current state of these features unacceptable?

PoignardAzur:
Note that we already do run `cargo fmt` in CI in basically all our projects. The question is a bit narrower than "do we auto-format", it's "do we use `group_imports = "StdExternalCrate"` in our rustfmt.tom files?".

(`imports_granularity = "Module"` is fine)

And in any case, the answer is "No" because it's an unstable setting, for pretty much the reasons I listed and others.

So the question is "do we use an always-commented version of that setting that people will occasionally apply manually?", which is more or less the status quo with the Xilem repo.

I guess we could just always include a comment with a link to this PR in our standard rustfmt config.

xStrom:
I created [website#87](https://github.com/linebender/linebender.github.io/issues/87) to track the shortcomings of `group_imports = "StdExternalCrate"`.

PoignardAzur:
Then I'm closing this in favor of #848.

PR #848, opened 2025-01-23
Reorganize Masonry modules

PoignardAzur:


PoignardAzur:
First commit just moves modules around and adds temporary re-exports in `lib.rs` to avoid changing too many files.

jaredoconnell (approved):
If we want to approve this design, I don't see any issues. Daniel suggested splitting a few things up, so it's worth considering that, too.

PR #849, opened 2025-01-23
Update ARCHITECTURE.md file

PoignardAzur:
Note: some of the doc depends on #832 and #848 which haven't been merged yet.


PoignardAzur:
> I see that this follow extantly doesn't follow our one sentence per line guidelines. I'd like any new writings to follow that, but I'm not going to block on it here.

I remember deliberately choosing to do this, because I expect this file to mostly be read for text editors and the "one line per sentence" approach hurts readability there. I don't know if I'd stand by that reasoning.

PR #853, opened 2025-01-27
feat: add slider widget with horizontal and vertical support

Cupnfish:
- Implement Slider widget with customizable colors, step size, and radii
- Add horizontal and vertical axis support with proper coordinate handling
- Include hover effects and accessibility features
- Add comprehensive tests for interaction and rendering
- Integrate slider into widget gallery example with direction switching
- Add documentation and example usage
- Fix edge cases for min/max values and step adjustments

There is still an issue where my tests are not outputting images for some reason. Can anyone help me with this?


Cupnfish:
#819

xilem/src/view/slider.rs
```diff
@@ -0,0 +1,265 @@
...
+                            MasonrySlider::$set_method($widget, v);
+                        }
+                    }
+                )*
+            };
+        }
+    };
+}
+
+// 定义所有需要处理的属性
```
DJMcNab, line 165: I don't know that we have a definitive policy here, but certainly I can only accept comments in English, and I'm pretty sure that all potential reviewers will have a similar position.

xilem/src/view/slider.rs
```diff
@@ -0,0 +1,265 @@
...
+    pub fn on_editing_changed(
+        mut self,
+        on_editing_changed: impl Fn(&mut State, bool) -> Action + Send + Sync + 'static,
+    ) -> Self {
+        self.on_editing_changed = Some(Box::new(on_editing_changed));
+        self
+    }
+}
+
+macro_rules! define_slider_properties {
```
DJMcNab, line 137: This is... quite unexpected. I can see arguments in favour of this, but I don't think this is at the right level of abstraction.

xilem/src/view/slider.rs
```diff
@@ -0,0 +1,265 @@
...
+        }
+
+        macro_rules! update_slider {
+            ($widget:expr, $current:expr, $prev:expr) => {
+                $(
+                    if $current.$field != $prev.$field {
+                        if let Some(v) = $current.$field {
+                            MasonrySlider::$set_method($widget, v);
+                        }
+                    }
```
DJMcNab, line 158: I'm not convinced that this behaviour is right.  I think it would fall down if `prev` is `Some(...)` and `current` is `None`.

xilem/src/view/slider.rs
```diff
@@ -0,0 +1,265 @@
...
+    #[must_use]
+    /// Set the slider's hover glow spread radius.
+    pub fn with_hover_glow_spread_radius(mut self, spread_radius: f64) -> Self {
+        self.hover_glow_spread_radius = Some(spread_radius);
+        self
+    }
+
+    #[must_use]
+    /// Set the callback for editing state changes.
+    pub fn on_editing_changed(
```
DJMcNab, line 128: I was hoping to see some use case for this.
Also "editing state changes" is extremely vague terminology.

xilem/src/view/slider.rs
```diff
@@ -0,0 +1,265 @@
...
+        mut self,
+        on_editing_changed: impl Fn(&mut State, bool) -> Action + Send + Sync + 'static,
+    ) -> Self {
+        self.on_editing_changed = Some(Box::new(on_editing_changed));
+        self
+    }
+}
```
DJMcNab, line 135: Hmm, should there not also be an `on_value_changed` property?

masonry/src/widgets/slider.rs
DJMcNab, line 1: I've not reviewed this code

xilem/src/view/slider.rs
Philipp-M: Agree, it's slightly more boilerplatey to "manually expand" this macro, but certainly better and a lot more understandable than macros creating macros ([speaking of experience](https://github.com/linebender/xilem/blob/3ef6efa8ce6fc3bc6e10460ced781129f54448dd/crates/xilem_web/src/interfaces.rs#L215)). Though I think it's showing that we need to find better ways to avoid this kind of boilerplate.

masonry/src/widgets/slider.rs
```diff
@@ -0,0 +1,854 @@
...
+            is_hovered: false,
+            hover_glow_color: Color::from_rgba8(255, 255, 255, 50),
+            hover_glow_blur_radius: 5.0,
+            hover_glow_spread_radius: 2.0,
+            track_rect: None,
+        }
+    }
+
+    /// Sets the slider's color.
+    pub fn with_color(mut self, color: impl Into<Color>) -> Self {
```
waywardmonkeys, line 87: This should probably be with_thumb_color and rename the struct member as well.

also, should this and the other colors be a brush instead? That would permit gradients…

masonry/src/widgets/slider.rs
```diff
@@ -0,0 +1,854 @@
...
+            step: 1.0,
+            color: theme::PRIMARY_LIGHT,
+            track_color: theme::PRIMARY_DARK,
+            editing: false,
+            is_dragging: false,
+            grab_anchor: None,
+            thumb_radii: RoundedRectRadii::from_single_radius(5.0),
+            track_radii: RoundedRectRadii::from_single_radius(2.0),
+            is_hovered: false,
+            hover_glow_color: Color::from_rgba8(255, 255, 255, 50),
```
waywardmonkeys, line 79: Should there be something in the theme for this instead of a hardcoded color here?

waywardmonkeys (commented):
All that I have time for right now…

xilem/src/view/slider.rs
Cupnfish: https://developer.apple.com/documentation/swiftui/slider/init(value:in:step:oneditingchanged:)
The parameter name `on_editing_changed` is taken from SwiftUI's slider implementation. Do you think it's better to keep this or use `on_value_changed`?

Cupnfish: Do you mean `on_change`? It can get the current changed value.

DJMcNab: Oh I see. That makes more sense. Sorry

DJMcNab (commented):
I'm happy to approve the Xilem side. I would still like some documentation for why you'd set `on_editing_changed`, but I won't block on it.

@PoignardAzur can you please review the Masonry/widget code?

masonry/src/widgets/slider.rs
```diff
@@ -0,0 +1,845 @@
...
+    fn get_cursor(&self, ctx: &QueryCtx, _pos: Point) -> CursorIcon {
+        if ctx.is_disabled() {
+            CursorIcon::Wait
+        } else if self.is_dragging {
+            CursorIcon::Grabbing
+        } else if self.is_hovered {
+            CursorIcon::Grab
+        } else {
+            CursorIcon::Default
+        }
```
PoignardAzur, line 451: IIRC disabled widgets don't affect the cursor.

Also, if your `get_cursor` methods can return different values based on status, you should call `request_cursor_icon_change` when the status changes. (Otherwise, you may have cases where the icon doesn't change until the mouse moves.)

masonry/src/widgets/slider.rs
```diff
@@ -0,0 +1,845 @@
...
+    }
+
+    #[test]
+    fn vertical_slider() {
+        let [slider_id] = widget_ids();
+        let slider = Slider::new(Axis::Vertical, 0.0, 100.0, 75.0).with_id(slider_id);
+
+        let mut harness = TestHarness::create_with_size(slider, Size::new(40.0, 200.0));
+        assert_render_snapshot!(harness, "vertical_slider");
+    }
```
PoignardAzur, line 492: It's great that you added lots of screenshot tests, but the screenshots haven't been added to source control.

PoignardAzur (commented):
The code calculating the slider values and the tests checking that code are pretty intense, and I didn't inspect them in detail. I think it *may* be possible to write a simpler version of that code, but I won't block on that.

The Widget code itself seems mostly fine and idiomatic. I'm fine with merging it once the problems I've pointed out are fixed.

PR #854, opened 2025-01-27
Use nightly feature `arbitrary_self_types`

PoignardAzur:


PoignardAzur:
Here's an example of what the user code ends up looking like:

https://github.com/linebender/xilem/pull/854/files#diff-d438593c075826291708861b124a9be8c7070379eba402835abf79ac492b4ad6L291-R296

Issue #857, opened 2025-02-02
`portal` scroll bar stays activated when mouse is released outside of window


richard-uk1:
Steps to repro:
 1. create an app with a portal area (I used vertical scroll)
 2. run the app
 3. mouse down on the scroll bar, then drag the mouse outside the window and release
 4. move mouse over scrollbar again

expected: scroll bar doesn't move with mouse
actual: scroll bar moves with mouse

PR #858, opened 2025-02-04
Fix scrollbars when pointer capture is released

DJMcNab:
Fixes https://github.com/linebender/xilem/issues/857

cc @PoignardAzur - I don't know that the model for pointer capture is entirely coherent from a widget's perspective at the moment. In particular, since any widget can call `release_pointer` at any time, and there's no way for the widget with pointer capture to know.

I'm also not certain that other widgets don't have this same issue. I do think something in our model might need to be refined slightly here.

### DIFF
--- a/masonry/src/widgets/scroll_bar.rs
+++ b/masonry/src/widgets/scroll_bar.rs
@@ -154,7 +154,7 @@ impl Widget for ScrollBar {
                 }
                 ctx.request_render();
             }
-            PointerEvent::PointerUp(_, _) => {
+            PointerEvent::PointerUp(_, _) | PointerEvent::PointerLeave(_) => {
                 self.grab_anchor = None;
                 ctx.request_render();
             }
@@ -233,7 +233,7 @@ mod tests {
 
     use super::*;
     use crate::assert_render_snapshot;
-    use crate::core::PointerButton;
+    use crate::core::{PointerButton, PointerState};
     use crate::testing::{widget_ids, TestHarness, TestWidgetExt};
 
     #[test]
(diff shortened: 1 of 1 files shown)


jaredoconnell:
This PR does what you intended for it to do. However, the behavior isn't how users may expect it to work. For example, if I drag a scrollbar in a web browser, I can keep holding the click while dragging outside of the window, and the scrollbar will continue to follow my cursor on the appropriate axis.
In this situation, it just loses attachment when you leave the bounds of the window. In the prior code if you went back in the window it would continue being attached in a broken way, but would not update outside of the window.

Would the current design allow for dragging while the cursor is outside of the window, or would further changes be required to make that work?

DJMcNab:
>  In this situation, it just loses attachment when you leave the bounds of the window. In the prior code if you went back in the window it would continue being attached in a broken way, but would not update outside of the window.

This isn't the behaviour on my machine. As far as I know, this is how we describe this in the API we have now.
It seems like there might need to be a fix on winit's side for your platform?

PoignardAzur:
> In particular, since any widget can call release_pointer at any time, and there's no way for the widget with pointer capture to know.

That seems like an oversight on our part. A widget should only be able to release its own capture status.

PoignardAzur:
Testing both the main branch and this PR, I agree that our current model of pointer capture doesn't work, and especially the fact that we remove capture on PointerLeave.

Honestly I'm not sure I had a good model of how winit worked when I coded that. I think I was thinking of `WindowEvent::CursorLeft` as "the cursor disappeared" when the proper event to represent that is probably `DeviceEvent::Removed`.

This PR works as a band-aid to be less inconsistent in this case, but we'd probably want to keep pointer capture even when the pointer leaves the window rect.

masonry/src/widgets/scroll_bar.rs
```diff
@@ -154,7 +154,7 @@ impl Widget for ScrollBar {
                 }
                 ctx.request_render();
             }
-            PointerEvent::PointerUp(_, _) => {
+            PointerEvent::PointerUp(_, _) | PointerEvent::PointerLeave(_) => {
                 self.grab_anchor = None;
                 ctx.request_render();
             }
```
PoignardAzur, line 160: At this point I wonder if we should add an `Update::PointerCaptureChanged` event.

PoignardAzur:
Let's leave it open. We'll merge it if we haven't had a better fix before the end of the month.

PoignardAzur:
Well it took a bit more than a month, but I've submitted a proper fix. I think we can close this.

PR #859, opened 2025-02-06
Fix accessibility documentation

PoignardAzur:


PoignardAzur:
@mwcampbell I'd appreciate if you could review this week.

Otherwise I'll assume the documented wording is okay, since it's mostly stuff we discussed during a video call.

DJMcNab (approved):
I've not carefully reviewed this, but it's not great to leave it sitting unmerged.

PoignardAzur:
CI got held up on formatting. I wonder if we should add a pre-push hook that runs cargo fmt, because (1) it's faster than anything else we do in CI (2) it's the check that's most likely to fail for dumb reasons.

Issue #584, opened 2024-09-07
Xilem requires sibling dependency on Winit and Masonry


TechComet:
required import winit and masonry features without import him to project

default example
https://github.com/linebender/xilem/blob/main/xilem/examples/flex.rs

required to (without required winit and masonry import in project)
```
use xilem::{
    widget::{CrossAxisAlignment, MainAxisAlignment},
    view::{button, flex, label, sized_box, Axis, FlexExt as _, FlexSpacer},
    EventLoop, WidgetView, Xilem,
    ArcStr,
    error::EventLoopError,
};

/// A component to make a bigger than usual button
fn big_button(
    label: impl Into<ArcStr>,
    callback: impl Fn(&mut i32) + Send + Sync + 'static,
) -> impl WidgetView<i32> {
    sized_box(button(label, callback)).width(40.).height(40.)
}

fn app_logic(data: &mut i32) -> impl WidgetView<i32> {
    flex((
        FlexSpacer::Fixed(30.0),
        big_button("-", |data| {
            *data -= 1;
        }),
        FlexSpacer::Flex(1.0),
        label(format!("count: {}", data)).text_size(32.).flex(5.0),
        FlexSpacer::Flex(1.0),
        big_button("+", |data| {
            *data += 1;
        }),
        FlexSpacer::Fixed(30.0),
    ))
    .direction(Axis::Horizontal)
    .cross_axis_alignment(CrossAxisAlignment::Center)
    .main_axis_alignment(MainAxisAlignment::Center)
}

fn main() -> Result<(), EventLoopError> {
    let app = Xilem::new(0, app_logic);
    app.run_windowed(EventLoop::with_user_event(), "Centered Flex".into())?;
    Ok(())
}
```

DJMcNab:
We do not have any examples written in Arabic.
Please explain what issue you're seeing, and what you're expecting to see. In its current form, this issue contains scant information.

I *think* you are trying to express that it's not possible to use Xilem without importing Masonry and winit into your project? This is a known issue, but it's not been a priority to address, as we're still in pre-alpha.
Pull requests to add targetted re-exports to make this possible are welcome, however.

Unfortunately, due to how Rust examples work, we cannot force them to use re-exports. This can be seen in this [stackoverflow discussion](https://stackoverflow.com/questions/72760584/prevent-access-to-dependencies-in-examples-in-rust-project). This might make a good clippy lint?

TechComet:
It's my problem in type
see issues my title

[...]

PR #860, opened 2025-02-08
expose winit

richard-uk1:
I don't think there is a reason not to, and you need winit for some types so currently you must define the crate yourself, which risks version errors etc.

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -163,6 +163,8 @@ pub use xilem_core as core;
 /// Tokio is the async runner used with Xilem.
 pub use tokio;
 
+pub use winit;
+
 mod any_view;
 mod driver;
 mod one_of;


DJMcNab (commented):
So, this is trying to solve one half of #584?

There are a couple of things to think about here:
1) This is also an intermediate state, see [this comment](https://github.com/linebender/xilem/blob/8d59974f2f43573ddc0cc30b1a1bda6b5e7112f0/xilem/src/lib.rs#L215).
2) There are reasonable use cases for using Xilem without winit

I think I'm happiest just leaving it in the state that you need to depend on winit for the moment, especially as you already need to depend on Masonry.

PoignardAzur (approved):
LGTM.

PoignardAzur:
To expand on the above, I think we'll eventually decouple Masonry from winit (having vocabulary crates for winit events would be a good first step) but until that happens, re-exporting it is fine, and is in line with how we treat other dependencies.

richard-uk1:
Yes if you decouple then you'd want to reverse this out.

PR #861, opened 2025-02-10
xilem: make a Pod method pub to allow creating custom buttons in an app

eugenesvk:
I wanted to create a custom button view, but when needing to pass to the masonry type could not convert the pod to a widgetpod.
So I'd like to make this method pub to allow that

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -314,17 +314,29 @@ impl<W: Widget + FromDynWidget> Pod<W> {
 }
 
 impl<W: Widget + FromDynWidget + ?Sized> Pod<W> {
-    fn erased(self) -> Pod<dyn Widget> {
+    /// Type-erase the contained widget.
+    ///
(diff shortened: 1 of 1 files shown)


DJMcNab (commented):
Thanks! I'm looking forward to see what you create.
I do however think that all three of the methods in that impl block should be public.

If we're making something public, it's best practise to add documentation, even if it's just a small comment. E.g. for the method here:
- Create the Masonry widget for this view. In most cases, you will add the return value as a child to another Masonry widget, using a method on that widget type.

eugenesvk:
Sure, can make all pub, didn't add docs because I don't fully understand the various elements of the XilemView-on-top-of-MasonryWidget architecture, so please check the description

DJMcNab:
Thank you! I've reworded some of the docs - my previous comment was written quickly, and reading it back revealed that it was misleading.

DJMcNab (approved):

PR #862, opened 2025-02-10
masonry: make util pub to allow customizing buttons in an app

eugenesvk:
Similar to https://github.com/linebender/xilem/pull/861 this would help creating custom elements while relying on the already used/tested library utilities

### DIFF
--- a/masonry/src/lib.rs
+++ b/masonry/src/lib.rs
@@ -148,7 +148,7 @@
 // TODO - Add logo
 
 #[macro_use]
-mod util;
+pub mod util;
 
 #[cfg(doc)]
 pub mod doc;


DJMcNab:
What is the specific functionality you need access to from this module.

DJMcNab:
I'll defer this to @PoignardAzur. To unblock you, I'd suggest just copying these items into your own crate, or even just hardcoding the gradient directions.

I'd be interested in hearing your report on how implementing your own button type goes. My current expectation is that the way we support styling (in the sense of implementing design systems) would be to use custom implementations of widgets, so I'm hoping that goes well for you.

Could you either create a thread [on Zulip](https://xi.zulipchat.com/#narrow/channel/317477-masonry) when you have something to share?

PoignardAzur (approved):
I think this is fine for now, though we'll probably remove that module eventually.

Issue #851, opened 2025-01-23
Integrate screenshots in documentation


PoignardAzur:
#832 added a way to add screenshots to documentation

What we need after that PR is merged is to:

- Check whether the docs.rs trick actually works.
- Document the tagging pattern needed to make it work.
- Find a way to add screenshots that will still work with cargo rdme.

PoignardAzur:
We've had a discussion somewhere (can't find where), where Daniel pointed out that the current setup might burn through our LFS quotas because people will download the image every time they consult our doc.

@DJMcNab is this still a problem?

DJMcNab:
We discussed it in office hours two weeks ago (https://docs.google.com/document/d/1cL4CywyDJ-AdTyQTYTySYJWCsaq0nT50uV9rliFM6zU/edit?tab=t.0#heading=h.evjimzdnrevx).

It might still be an issue - the docs from GitHub are a bit unclear. The decision we made was to launch it as currently planned, then:
- If it becomes an issue, and chews through our LFS bandwidth, we pay for a single pack of LFS backwidth/data ($5), which should give us enough breathing room, as it provides 50x the current bandwidth
- Release a 0.3.1 which moves the screenshots to a GitHub pages/repository which doesn't use LFS (not necessarily the long-term solution, just to stop the "bleeding")
- (Maybe) ask the docs.rs team to remove the docs for 0.3.0, if it's still getting enough back-traffic

Certainly if someone has the will to move to a properly hosted solution on linebender.org (the exact technical details there are TBD, because we want to keep the screenshots for all previous tags available), then that would be welcome.

PoignardAzur:
Could we publish the screenshots in a Github release and link to those? I don't think Github release downloads are metered?

PoignardAzur:
> This is now a blocker for 0.3.0 (with a possible solution being to just remove the screenshots from the docs entirely).

Given that we're in the middle of a release train, and that fixing the doc can easily be done in a minor release, I lean towards doing that unless someone finds a long-term solution *really* fast.

PoignardAzur:
See also this zulip thread: [#linebender > cfg(docsrs)](https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/cfg.28docsrs.29/with/516854696)

ofluffydev:
Could we not using some cdn instead of having tons of images take of lfs bandwidth?

PoignardAzur:
[...]

Issue #392, opened 2024-06-12
Write Xilem documentation


PoignardAzur:
Xilem's lack of documentation is one of the problems new users complain about the most often. We need to avoid staying in that state for too long.

We should at least document:

- All pub items.
- Basic usage patterns.
- How to create your first app.
- The general project architecture.

TechComet:
+1

fardarter:
If someone wants to teach me:

- basic usage patterns
- how to create your first app,

I'll write them up. 

I don't even know rust yet (aside from very trivial syntax), so you'd get the view from someone very attentive to assumptions.

DJMcNab:
Thank you for your offer. Some starting points.

## Basic usage patterns

- Xilem works based on a mapping of your application state to `View`s in `app_logic`
- These create widgets, which your users interact with
- When these interactions happen, callbacks in your views are called
- These update the app state, and so your `app_logic` is called again
- The new view is then applied to the widgets.

## How to create your first app

You'd add a dependency to Xilem (probably on the main branch), then set up your app:

```rust
fn app_logic(data: &mut i32) -> impl WidgetView<i32> {
     flex((label(format!("{data}")), button("increment", |data: &mut i32| data+=1)))
}

fn main() -> Result<(), EventLoopError> {
    let app = Xilem::new(0, app_logic);
    app.run_windowed(EventLoop::with_user_event(), "Centered Flex".into())?;
    Ok(())
}
```
A slightly more advanced pattern would use a custom struct in place of the i32, there.

TechComet:
how open new window ?

DJMcNab:
Xilem and Masonry currently only support a single window. To open that window, you should see our examples.

Feel free to open a Zulip discussion around menus so we can mentor your implementation of them

ArtyomSinyugin:
I opened [this](https://github.com/linebender/xilem/pull/818) PR to make some docs improvements.

PR #864, opened 2025-02-11
Add a "gallery"-type example for Xilem's Label View

eugenesvk:
There are existing issues re. the general state of docs with Xilem
https://github.com/linebender/xilem/issues/392
https://github.com/linebender/xilem/issues/851

so this PR is a suggestion on covering one area I missed when reading the docs: being able to have
- a **condensed overview** of
- **all** available options and, more importantly, 
- their **visual style**
- in an **interactive** manner (ideally at runtime, which this PR doesn't address)

So this is an example of a simple gallery-type app that covers Xilem's **Label View** (but not Masonry Label Widget) where you can simply vertically scroll to see all the available options and even some notes on interaction issues: e.g., the fact that a label with a Start alignment starts in the middle because its parent vertical flex has a Center alignment by default for its horizontal axis. Now, you could find this explanation deep down in Masonry tests, but that would require a lot of digging around (and Masonry screenshots wouldn't show it). 

After identifying an issue this example also shows a suggested fix by using `.flex`, though if I do the override, I lose the portal's vertical scrolling, which might be a 🐞bug or just a missing understanding on my part of how these portal/flex/.flex elements interact, in which case it would be nice to change to a better example and to have a shorter alignment section that didn't have to note these issues.

Xilem Label View

Now, there are a few missing pieces, but think this might be still be a good initial fix:

As far as I understand mostly blocked by missing functionality support in the framework:
- add rust code generating each element in a tooltip
- add the same code in a context menu as a "copy" command
  - is this something the debugging F11/F12 tools could do?
- add URL support for doc links

But also just not yet implemented:
- add non-desktop platforms
- add to a CI build so you can simply download a binary and use it as a reference without having to build everything yourself
  - have both debug/release builds so that you can use F11/F12 inspectors?
    - is a release build useful for basic examples?

(these are also added in the Todos comment section in the example itself)

### DIFF
--- a/xilem/Cargo.toml
+++ b/xilem/Cargo.toml
@@ -91,6 +91,10 @@ path = "examples/emoji_picker.rs"
 # cdylib is required for cargo-apk
 crate-type = ["cdylib"]
 
+[[example]]
+name = "label_xilem_view"
+path = "examples/label_xilem_view.rs"
+
 [lints]
 workspace = true
 
--- a/xilem/examples/label_xilem_view.rs
+++ b/xilem/examples/label_xilem_view.rs
@@ -1,7 +1,4 @@
 //! An illustration of the various options for Xilem's Label View (based on Masonry Label Widget)
-#![cfg_attr(not(debug_assertions),allow(non_snake_case,non_upper_case_globals,non_camel_case_types))]
-#![cfg_attr(    debug_assertions ,allow(non_snake_case,non_upper_case_globals,non_camel_case_types,unused_imports,unused_mut,unused_variables,dead_code,unused_assignments,unused_macros))]
-
 // TODOs:
   // add rust code generating each element in a tooltip
   // add the same code in a context menu as a "copy" command
@@ -13,7 +10,7 @@
 
 use winit::window::Window;
 use masonry::dpi::LogicalSize;
-use xilem::view::{MainAxisAlignment,CrossAxisAlignment,FlexExt,FlexParams,};
+use xilem::view::{CrossAxisAlignment,FlexExt,FlexParams,};
 #[derive(Default)]
 struct AppState {}
 
@@ -27,7 +24,7 @@ use xilem::{EventLoop,WidgetView,Xilem,FontWeight,TextAlignment,LineBreaking,pal
 use masonry::core::ArcStr;
 use masonry::parley::fontique;
 use masonry::peniko::Color;
-const label_color:Color = css::ROYAL_BLUE;
+const LABEL_COLOR:Color = css::ROYAL_BLUE;
 
 fn title_prose(text:impl Into<ArcStr>) -> Prose {
   prose(text).text_size(18.0).alignment(TextAlignment::Justified).brush(css::GREEN)
@@ -36,15 +33,11 @@ fn txt_prose(text:impl Into<ArcStr>) -> Prose {
   prose(text).text_size(14.0).alignment(TextAlignment::Start).brush(Color::from_rgb8(0x11,0x11,0x11))
 }
 fn lc(text:impl Into<ArcStr>) -> Label { //colored label
-  label(text).brush(label_color)
+  label(text).brush(LABEL_COLOR)
 }
-fn app_logic(D:&mut AppState) -> impl WidgetView<AppState> {
-  let cb = |D:&mut AppState|{}; // empty callback for empty buttons
-  // let lc = Color::from_rgb8(0x77,0x77,0x77); //119 css::LIGHT_GRAY=211 label color
-  // let lc = Color::from_rgb8(0x0a,0x0a,0x0a);
-  let lC = css::ROYAL_BLUE;
-  let tC = css::GREEN; //title
-  let mC = Color::from_rgb8(0x11,0x11,0x11); //main text
+fn app_logic(_d:&mut AppState) -> impl WidgetView<AppState> {
+  let m_c = Color::from_rgb8(0x11,0x11,0x11); //main text
+  let l_c = LABEL_COLOR;
   let mut i = 1;
 
   portal(
@@ -55,10 +48,10 @@ fn app_logic(D:&mut AppState) -> impl WidgetView<AppState> {
     • F11 to toggle a rudimentary widget inspector
     • F12 to toggle paint widget layout rectangles")
   } else {txt_prose("This is not a debug build, so F11 widget inspector and F12 widget rectangles tools are not available)\ngithub.com/linebender/xilem/tree/main/masonry#debugging-features")},
-  label(format!("Label: Serif Bold 14 {label_color:?}")).text_size(14.0).weight(FontWeight::BOLD) // float bold=700, FontWeight::parse("normal") for css
+  label(format!("Label: Serif Bold 14 {LABEL_COLOR:?}")).text_size(14.0).weight(FontWeight::BOLD) // float bold=700, FontWeight::parse("normal") for css
     .font(fontique::GenericFamily::Serif)
     .alignment(TextAlignment::Start)
-    .brush(lC) //impl Into<peniko:Brush> brush sets text color, but gradients and images are also supported Solid(color::AlphaColor<Srgb>) Gradient(Gradient) Image(Image),
+    .brush(l_c) //impl Into<peniko:Brush> brush sets text color, but gradients and images are also supported Solid(color::AlphaColor<Srgb>) Gradient(Gradient) Image(Image),
     .line_break_mode(LineBreaking::Overflow) //WordWrap Clip Overflow
     ,
   title_prose(format!("§{i} .alignment")),{i+=1;},
@@ -76,7 +69,7 @@ fn app_logic(D:&mut AppState) -> impl WidgetView<AppState> {
   //   ).width(200f64).height(70f64).padding(Padding::from(0.))
   //    .background(css::LIGHT_GRAY) // .border(css::RED,0.).rounded(RoundedRectRadii::from_single_radius(0.))
   // ,),
-  (l("  •grid in a 200×70 sized_box to make labels same-width (one per row in 4×1 table)").alignment(TextAlignment::Justified).brush(mC),
+  (l("  •grid in a 200×70 sized_box to make labels same-width (one per row in 4×1 table)").alignment(TextAlignment::Justified).brush(m_c),
   sized_box(
     grid((
       lc("1/4 alignment Start"    	).alignment(TextAlignment::Start    	).grid_pos(0,0),
@@ -87,14 +80,14 @@ fn app_logic(D:&mut AppState) -> impl WidgetView<AppState> {
     ).width(200f64).height(70f64).padding(Padding::from(0.))
      .background(css::LIGHT_GRAY) //.border(css::RED,0.).rounded(RoundedRectRadii::from_single_radius(0.))
   ,),
-  (l("  •unboxed (constrained by root parent's flex in a portal)\n  (Start=Middle: parent Flex container ≝CrossAxisAlignment::Center,\n  so the alignment for a label starts at the center)").alignment(TextAlignment::Justified).brush(mC),
-  l("  can be fixed with a custom per-element override .flex(FlexParams::new(1.0,CrossAxisAlignment::Start))").alignment(TextAlignment::Justified).brush(mC),
+  (l("  •unboxed (constrained by root parent's flex in a portal)\n  (Start=Middle: parent Flex container ≝CrossAxisAlignment::Center,\n  so the alignment for a label starts at the center)").alignment(TextAlignment::Justified).brush(m_c),
+  l("  can be fixed with a custom per-element override .flex(FlexParams::new(1.0,CrossAxisAlignment::Start))").alignment(TextAlignment::Justified).brush(m_c),
   lc("1/4 alignment Start"    	).alignment(TextAlignment::Start    	),
   lc("2/4 alignment Middle"   	).alignment(TextAlignment::Middle   	),
   lc("3/4 alignment End"      	).alignment(TextAlignment::End      	),
   lc("4/4 alignment Justified"	).alignment(TextAlignment::Justified	),
   ),
-  (l("  •flex in a 500×140 sized_box (🐞? unboxed .flex override removes Portal scrolling)").alignment(TextAlignment::Justified).brush(mC),
+  (l("  •flex in a 500×140 sized_box (🐞? unboxed .flex override removes Portal scrolling)").alignment(TextAlignment::Justified).brush(m_c),
   sized_box(flex((
     lc("1/4 alignment Start"                             	).alignment(TextAlignment::Start    	),
     lc("1/4 alignment Start + CrossAxisAlignment::Start "	).alignment(TextAlignment::Start    	).flex(FlexParams::new(1.0,CrossAxisAlignment::Start)),
@@ -107,7 +100,7 @@ fn app_logic(D:&mut AppState) -> impl WidgetView<AppState> {
   ),
   (title_prose(format!("§{i} .line_break_mode")),{i+=1;},
   txt_prose("  3 options: ≝WordWrap Clip Overflow\n  https://docs.rs/masonry/latest/masonry/widget/enum.LineBreaking.html"),
-  l("  •grid in a 340×120 box to make labels same-width (one per row in 3×1 table)").alignment(TextAlignment::Justified).brush(mC),
+  l("  •grid in a 340×120 box to make labels same-width (one per row in 3×1 table)").alignment(TextAlignment::Justified).brush(m_c),
   sized_box(
     grid((
       lc("1/3 WordWrap: break at word boundaries = abcd-efgh-ijkl-mnop-qrst-uvwx-yz" 	).line_break_mode(LineBreaking::WordWrap	).grid_pos(0,0),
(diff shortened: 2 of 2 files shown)


DJMcNab:
My initial feedback here is that Label is that we shouldn't be encouraging developers to reach for `label` in most cases.
The label is designed *solely* to be used as a child of other non-interactive widgets (such as the extant button and checkbox). This is because it doesn't allow the user to select within the text, and therefore doesn't allow copy-and-paste from the text, which makes it annoying. Note that this isn't a position that I've carefully reasoned about (e.g., what about on Android where the idiomatic way to select such (short) text is to use the OCR built-in to the app selector).

I also want to check (as a self-calibration exercise only), was this code LLM assisted?

eugenesvk:
The galleries aren't meant to encourage anything, it's purely documenting what's available. Similar to the regular docs, but in a better format so that you don't to click on an argument, then an enum, then read its variants and still have a vague idea as to how it would look in practice.

Feel free to add a disclaimer at the top guiding to better options and pointing out the limitations!

> such as the extant button

That's actually precisely the use case that lead me to create this example as a reference for myself when trying to learn about to how add labels to a button! It's just that the button in this case is more complex than typical and has several labels so that, for example, you could have these great self-documenting buttons that show their shortcuts

button9


(no LLM involved. Have you tried it yourself, does any LLM give good info on Xilem/Masonry? Would it be able to uncover and explain the label|flex alignment interaction?)

DJMcNab:
That makes sense to me. I think there's something in the naming of label which needs to change, probably. I would agree that we should probably have a paragraph at the start of this, explaining why Label exists and how to use it (which should be copied into Label's docs).

That does look pretty cool. Is that saying that the confirm action happens if you press enter or alt-k?

I'm not sure about the flex issue. I know that our layout story is still... in flux at the moment.

I don't know when I'll find time to review this.

> (no LLM involved. Have you tried it yourself, does any LLM give good info on Xilem/Masonry?)

No, I've not tried it myself. I'd hope that they don't, to be honest, because we're still at a stage of refining our APIs.
The reason I asked that is because of the weird formatting of the Rust code, which struck me as sufficiently unusual that I had alarm bells going off.

The formatting definitely needs to be changed to follow the Rust style guide. The reason that this isn't happening is the eternally vexing rustfmt issue, https://github.com/rust-lang/rustfmt/issues/3863. We have a few workarounds in the past, e.g. https://github.com/linebender/vello/pull/499

eugenesvk:
> Is that saying that the confirm action happens if you press enter or alt-k?

Yes, though just <kbd>K</kbd>, this is just an experimental replacement for the "&Accelerators", which are bad because of their inconsistent position, requiring users to scan, and poor highlight (especially consider a tiny underscore for a thin letter i̲ - try to find it vs the first O̲
i_under
)

> weird formatting of the Rust code

Which LLMs have you seen that are able to do this? That would be at least something of use from them...

I find the default "fmt"'s style an unaligned mess, and don't use it myself, so this is a result of whatever the cargo tool managed to do...
(Due to the very high verbosity of Xilem's code and 4 spaces this has the downside of being too wide, but then a simple dynamic word-wrap editor shortuct can easily bring all the zigzags back!)

Like what's the point of this zigzaggy mess where you can't easily compare the grid position coordinates
```rs
flex((
  lc("1/4").alignment(TextAlignment::Start).grid_pos(0, 0),
  lc("2/4").alignment(TextAlignment::Middle).gird_pos(0, 1),
  lc("3/4").alignment(TextAlignment::End).grid_pos(0, 2),
  lc("4/4").alignment(TextAlignment::Justified).grid_pos(0, 3),
))
```
when tables have been invented a long time ago, so you could have a proper tabular alignment, allowing you to skip identical syntax like `()` and also identical words, focusing only on the difference, which would be `0,1,2,3`
```rs
flex((
  lc("1/4").alignment(TextAlignment::Start    ).grid_pos(0,0),
  lc("2/4").alignment(TextAlignment::Middle   ).gird_pos(0,1),
  lc("3/4").alignment(TextAlignment::End      ).grid_pos(0,2),
  lc("4/4").alignment(TextAlignment::Justified).grid_pos(0,3),
))
``

Though at any rate, the CI should just commit the formatting changes on PRs, that's what robots do well, no need to involve people manually running commands that fit this project's style

File: xilem/examples/gallery_label_xilem_view.rs [hunk 1]
```diff
@@ -0,0 +1,178 @@
...
+    // sized_box(
+    //   flex((
+    //     lc("1/4 alignment Start"     ).alignment(TextAlignment::Start        ),
+    //     lc("2/4 alignment Middle"    ).alignment(TextAlignment::Middle       ),
+    //     lc("3/4 alignment End"       ).alignment(TextAlignment::End          ),
+    //     lc("4/4 alignment Justified" ).alignment(TextAlignment::Justified    ),
+    //   ))
+    //   ).width(200f64).height(70f64).padding(Padding::from(0.))
+    //    .background(css::LIGHT_GRAY) // .border(css::RED,0.).rounded(RoundedRectRadii::from_single_radius(0.))
+    // ,),
```
DJMcNab, line 84: This will need to be resolved before we can merge; it's not clear what it's doing

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,178 @@
...
+        • F11 to toggle a rudimentary widget inspector
+        • F12 to toggle paint widget layout rectangles")
+    } else {txt_prose("This is not a debug build, so F11 widget inspector and F12 widget rectangles tools are not available)\ngithub.com/linebender/xilem/tree/main/masonry#debugging-features")},
+    label(format!("Label: Serif Bold 14 {LABEL_COLOR:?}")).text_size(14.0).weight(FontWeight::BOLD) // float bold=700, FontWeight::parse("normal") for css
+        .font(fontique::GenericFamily::Serif)
+        .alignment(TextAlignment::Start)
+        .brush(l_c) //impl Into<peniko:Brush> brush sets text color, but gradients and images are also supported Solid(color::AlphaColor<Srgb>) Gradient(Gradient) Image(Image),
+        .line_break_mode(LineBreaking::Overflow) //WordWrap Clip Overflow
+        ,
+    title_prose(format!("§{i} .alignment")),{i+=1;},
```
DJMcNab, line 70: I think we should break this out into multiple expressions, if we're incrementing the counter inside the expression.

DJMcNab, line 70: Nice use of the section symbol! It's a nice little showcase that we just find the fonts for stuff like that correctly (even though it is a fairly easy case)

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,178 @@
...
+#![allow(
+    unused_assignments,
+    reason = "allows NOT having to track section increments {i+=1;}, removing the last one."
+)]
+
+use masonry::dpi::LogicalSize;
+use winit::window::Window;
+use xilem::view::{CrossAxisAlignment, FlexExt, FlexParams};
+#[derive(Default)]
+struct AppState {}
```
DJMcNab, line 23: I'd probably move this from out of the set of imports

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,178 @@
...
+    let mut i = 1;
+
+    portal(
+    flex((
+    (txt_prose("Xilem view::Label formats vGit@25-02 #25b12ad (in a ↕-scrollable area)").text_size(18.0),
+    if cfg!(debug_assertions) {txt_prose(
+     "This is a debug build, so you can use github.com/linebender/xilem/tree/main/masonry#debugging-features:
+        • F11 to toggle a rudimentary widget inspector
+        • F12 to toggle paint widget layout rectangles")
+    } else {txt_prose("This is not a debug build, so F11 widget inspector and F12 widget rectangles tools are not available)\ngithub.com/linebender/xilem/tree/main/masonry#debugging-features")},
```
DJMcNab, line 63: This is really nice!

DJMcNab (changes requested):
It will be nice to get something like this in. It's great to see an exploration of the full range of options here.

This is still blocked on the formatting. I think the 2024 style edition might have resolved this not being formatted, but I've not checked it.

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,178 @@
...
+fn txt_prose(text: impl Into<ArcStr>) -> Prose {
+    prose(text)
+        .text_size(14.0)
+        .alignment(TextAlignment::Start)
+        .brush(Color::from_rgb8(0x11, 0x11, 0x11))
+}
+fn lc(text: impl Into<ArcStr>) -> Label {
+    //colored label
+    label(text).brush(LABEL_COLOR)
+}
```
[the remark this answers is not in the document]
eugenesvk: the minimal name here is so that all the props fit in a single screen width, which with the waste of 4 spaces per indent level and 3 indents in some cases is a challenge

So the only important part in 
`lc("4/4 alignment Justified"    ).alignment(TextAlignment::Justified    ).grid_pos(0,3),` is alignment/grid info, so whether it's `l` or `lc` doesn't really matter, but it fits in 100 chars

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,178 @@
...
+fn app_logic(_d: &mut AppState) -> impl WidgetView<AppState> {
+    let m_c = Color::from_rgb8(0x11, 0x11, 0x11); //main text
+    let l_c = LABEL_COLOR;
+    let mut i = 1;
+
+    portal(
+    flex((
+    (txt_prose("Xilem view::Label formats vGit@25-02 #25b12ad (in a ↕-scrollable area)").text_size(18.0),
```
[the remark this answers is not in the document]
eugenesvk: Since the framework can change a lot and docs can lag, this references how old/relevant the examples are without having to search for the source cargo file

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,178 @@
...
+use xilem::view::{CrossAxisAlignment, FlexExt, FlexParams};
+#[derive(Default)]
+struct AppState {}
+
+use masonry::core::ArcStr;
+use masonry::parley::fontique;
+use masonry::peniko::Color;
+use winit::error::EventLoopError;
+use xilem::view::{
+    flex, grid, label, label as l, portal, prose, sized_box, Axis, GridExt, Label, Padding, Prose,
```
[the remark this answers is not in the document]
eugenesvk: the logic is the same as for `lc`s, but in this case it doesn't help as much, so removed the abbreviation

xilem/examples/gallery_label_xilem_view.rs
eugenesvk: Does `↕-scrollable area` work for you (in the title)? Or ≝ default signifier in `4 options: ≝Start Middle End Justified`? I see squares there

xilem/examples/gallery_label_xilem_view.rs
eugenesvk: repurposed this into a comparison of how labels spread within a flex vs grid

xilem/examples/gallery_label_xilem_view.rs
eugenesvk: don't understand this, meanwhile just changed the order so that `unused_assignments` allowance isn't needed anymore

xilem/examples/gallery_label_xilem_view.rs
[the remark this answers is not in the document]
eugenesvk: That follows from the desire to differentiate Xilem's Label View from Masonry Label Widget

What title would you like?

eugenesvk:
I don't get it, but locally on Windows  `cargo fmt --all` reverts the changes suggested by the failing CI's fmt

xilem/examples/gallery_label_xilem_view.rs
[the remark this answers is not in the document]
DJMcNab: Probably just `label_gallery`. This is already in the `xilem/examples` folder, so saying Xilem isn't needed. And it's not making a custom view, which is where you'd want a "view specific" example.

xilem/examples/gallery_label_xilem_view.rs
DJMcNab: Yes. Both of those resolve for me. The ↕ resolves to an emoji

DJMcNab: Well like `i+=1` is not a meaningful view; it only happens to work here because it has type `()`, which is the empty tuples, and nested tuples are allowed. We shouldn't have assignment expressions inside view writing code

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,348 @@
...
+    txt_prose("  6 options: ≝Start Middle End Justified Left Right\n  https://docs.rs/parley/latest/parley/layout/enum.Alignment.html")
+    ),
+    (label("  • flex in a 200×220 sized_box to show the impact of constraints (buttons to change alignment)").alignment(TextAlignment::Justified).brush(m_c),
+    flex((
+      button("Δ1/6", |da:&mut AppState|{da.realign("l1i1");}),
+      button("Δ2/6", |da:&mut AppState|{da.realign("l1i2");}),
+      button("Δ3/6", |da:&mut AppState|{da.realign("l1i3");}),
+      button("Δ4/6", |da:&mut AppState|{da.realign("l1i4");}),
+      button("Δ5/6", |da:&mut AppState|{da.realign("l1i5");}),
+      button("Δ6/6", |da:&mut AppState|{da.realign("l1i6");}),
```
DJMcNab, line 209: It's not all that clear what these buttons are useful for?

Even so, with having them, I think they could probably also be changed into an enumarted loop over a sub-array of the state.

xilem/examples/gallery_label_xilem_view.rs
```diff
@@ -0,0 +1,348 @@
...
+        CrossAxisAlignment::Start => CrossAxisAlignment::Center,
+        CrossAxisAlignment::Center => CrossAxisAlignment::End,
+        CrossAxisAlignment::End => CrossAxisAlignment::Baseline,
+        CrossAxisAlignment::Baseline => CrossAxisAlignment::Fill,
+        CrossAxisAlignment::Fill => CrossAxisAlignment::Start,
+    }
+}
+
+impl AppState {
+    fn realign(&mut self, pos: impl AsRef<str> + std::fmt::Debug) {
```
DJMcNab, line 106: I don't think that this is a great signature; if you really think that repeating `it.x = cycle(it.x)` is too much repetition, then maybe we can try `cycle(&mut it.x)`.
Stringly typed field access is not generally idiomatic in Rust code.

xilem/examples/gallery_label_xilem_view.rs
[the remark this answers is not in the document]
eugenesvk: It will be git-up-to-date, but can mismatch the released crate version / docs you read on docs.rs, so can help avoid some confusion

Also, the CI doesn't check everything, for example, if new styling options are added, they won't break anything

DJMcNab (changes requested):
This is still blocked on applying the Rust style guide.

I know that rustfmt breaks with big expressions; that is another reason to suggest breaking it up into multiple expressions (perhaps a component per section)?

xilem/examples/gallery_label_xilem_view.rs
eugenesvk: For playing with various styling in real time. While currently for some styles all the variants are already included, so buttons are of limited use when you want to see e.g. the whole "table" of labels aligned the same way, that's not true for all styles - Cross-alignment doesn't list all the combos. Neither would font families (though those don't have dynamic buttons/pickers at the moment) and colors/sizes

xilem/examples/gallery_label_xilem_view.rs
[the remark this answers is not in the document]
eugenesvk: But the user will not see "xilem/examples" when using the final executable as a reference

xilem/examples/gallery_label_xilem_view.rs
eugenesvk: I don't understand the proposed change to fix it without having to manually track index increments. 
The narrow issue of empty tuples could be fixed via `{i+=1;title_prose(format!("§ {i} .alignment"))}` which results in a single valid view result

xilem/examples/gallery_label_xilem_view.rs
eugenesvk: You're right, it's not too much repetition, and the strings are bad, updated to remove them

xilem/examples/gallery_label_xilem_view.rs
DJMcNab: Ah - when I said expressions above, I meant "statements". Sorry for the confusion

xilem/Cargo.toml
```diff
@@ -91,6 +91,10 @@ path = "examples/emoji_picker.rs"
 # cdylib is required for cargo-apk
 crate-type = ["cdylib"]
 
+[[example]]
+name = "gallery_label_xilem_view"
+path = "examples/gallery_label_xilem_view.rs"
+
```
xStrom, line 97: This example doesn't need to be defined in `Cargo.toml`. The others that are defined there are hacks for Android examples, which this one doesn't need.

```suggestion
```

PR #866, opened 2025-02-12
Add a button view+widget with up to 9 labels

eugenesvk:
Allowing the creation of helpful UI buttons like this, documenting which keys activate it
button9

It supports label padding as well, so you could in principle position labels wherever you like

I've noticed that even empty labels have >0 size (4) and, I guess, makes the button slightly bigger than necessary. Is this an effect of some global layout constant?
- Is it possible to disable it for all labels but the middle one?

My idea is for this to be identical in its default instance to a regular button.

I've also thought about skipping empty children (though currently it's suprisingy cumbersome to test whether a label is empty, so thought of adding some `is` method to the label to check whether its string is empty), but then wasn't clear on the proper mechanism for instantiating such children once their text is changed, so thought it might not be an issue if empty labels always exist. 
- Or is it an issue that empty labels are always created/tracked?

Updated docs/tests and added a `gallery` type example similar to https://github.com/linebender/xilem/pull/864

Button9Help

### DIFF
--- a/masonry/src/widgets/button9.rs
+++ b/masonry/src/widgets/button9.rs
@@ -8,10 +8,10 @@ use smallvec::{smallvec, SmallVec};
 use tracing::{trace, trace_span, Span};
 use vello::Scene;
 
-use crate::core::{self,
-  AccessCtx, AccessEvent, Action, ArcStr, BoxConstraints, EventCtx, LayoutCtx, PaintCtx,
-  PointerButton, PointerEvent, QueryCtx, TextEvent, Update, UpdateCtx, Widget, WidgetId,
-  WidgetMut, WidgetPod,
+use crate::core::{
+    self, AccessCtx, AccessEvent, Action, ArcStr, BoxConstraints, EventCtx, LayoutCtx, PaintCtx,
+    PointerButton, PointerEvent, QueryCtx, TextEvent, Update, UpdateCtx, Widget, WidgetId,
+    WidgetMut, WidgetPod,
 };
 use crate::kurbo::{Insets, Size, Vec2};
 use crate::theme;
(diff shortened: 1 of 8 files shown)


PoignardAzur:
This feels like the kind of case where we might want a "widget nursery" kind of crate, like druid had. It wouldn't need to be a separate repository, though, I don't think that separation was really useful.

eugenesvk:
What do you think of just making the default widgets overpowered and add all the extra functionality like this to them?

PoignardAzur:
I'm a bit iffy on that. The complexity quickly adds up.

DJMcNab:
I think that this is best placed to live in your own project which needs it, at the moment.
I do think that you've hit upon a real issue, which is that having a single string in a button is very limiting. However, the implementation strategy used in this PR doesn't seem ideal to me; it's both limiting and too powerful for the most common use cases.

My suggested design for this kind of button would be to allow it to contain an arbitrary widget, so long as it and all of its descendants disallow pointer interaction. The main challenge there is how you handle e.g. hover styling, because reasonably implementations will want to e.g. increase font weight or change other variable font parameters whilst the button is hovered, which requires more design. But that wouldn't block an MVP.

I'd like to close this PR, as I don't see any reasonable path towards it being merged.

eugenesvk:
Turning a button into a container of arbitrary widgets sounds way more complicated

> The main challenge there is how you handle e.g. hover styling

Indeed. But another one is presumably the user would have to do the work of all the inner widget layout within this button container instead of adding tweaks to an existing widget

> because reasonably implementations will want to e.g. increase font weight or change other variable font parameters whilst the button is hovered

What would you use that for?

DJMcNab:
Consider the tab buttons at the side (on desktop) of https://m3.material.io/ - when you hover over them, the font is emboldened, which makes it clear that the text is associated with the relevant chip.

eugenesvk:
That effect is way too subtle to make it clear (only `e` is somewhat different), especially compared to the much bigger and visible background color change as is common in buttons

DJMcNab:
We've discussed this in [#office hours > 2025-04-24](https://xi.zulipchat.com/#narrow/channel/359642-office-hours/topic/2025-04-24/with/514151655), and agreed to close this. As I say, if you need this in a project, the interim solution is to import this widget into your own code - support writing your own widgets in your own framework.

We definitely do want to provide support for something like this, but as I mentioned previously, this is not a principled strategy to get there.

PR #870, opened 2025-02-17
Update to current `parley` `main`

waywardmonkeys:
This updates to the version using the updated objc2 and binding crates for Fontique.

It also has a minor API change.

### DIFF
--- a/masonry/src/widgets/label.rs
+++ b/masonry/src/widgets/label.rs
@@ -386,7 +386,7 @@ impl Widget for Label {
         };
         if self.alignment_changed {
             self.text_layout
-                .align(Some(alignment_width), self.alignment);
+                .align(Some(alignment_width), self.alignment, false);
         }
         let text_size = Size::new(alignment_width.into(), self.text_layout.height().into());
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,9 +58,9 @@ dependencies = [
  "accesskit",
  "accesskit_consumer",
  "hashbrown",
- "objc2",
+ "objc2 0.5.2",
  "objc2-app-kit",
- "objc2-foundation",
+ "objc2-foundation 0.2.2",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


masonry/src/widgets/label.rs
```diff
@@ -386,7 +386,7 @@ impl Widget for Label {
         };
         if self.alignment_changed {
             self.text_layout
-                .align(Some(alignment_width), self.alignment);
+                .align(Some(alignment_width), self.alignment, false);
```
DJMcNab, line 389: Obviously I'm not super happy about an ugly boolean parameter appearing here. For reference, it seems to be from https://github.com/linebender/parley/pull/241, meaning whether we align when overflowing.
The lack of tests in that PR to make it more comprehensible isn't ideal...

In this case:
1) I agree that `false` maintains the same behaviour here. I think that true would maybe be a better default, but that should be a follow-up
2) I think that https://github.com/linebender/parley/issues/247 should block parley 0.3 (cc @tomcur); the fact that every use in https://github.com/linebender/parley/pull/241 needed to be documented with the parameter name as a comment strongly supports this...

waywardmonkeys:
@DJMcNab I've removed this from the auto-merge because I had to update a single screenshot test. @tomcur You might want to take a peek just to be sure all is well.

masonry/src/widgets/label.rs
tomcur: > I think that https://github.com/linebender/parley/issues/247 should block parley 0.3

Yes, that issue is blocking 0.3. Nico opened a PR fixing it yesterday: https://github.com/linebender/parley/pull/278.

> I agree that false maintains the same behaviour here. I think that true would maybe be a better default, but that should be a follow-up

I believe `false` (start-alignment for lines that overflow) is what web does, and currently that's the planned default for the `AlignmentOptions` introduced in Nico's PR.

nicoburns:
> @DJMcNab I've removed this from the auto-merge because I had to update a single screenshot test. @tomcur You might want to take a peek just to be sure all is well.

It looks to me like the old version is including trailing whitespace in the width of the line for the purposes of alignment, and the new version is ignoring. If so, then I believe the new version is correct. Eyeballing it, the text looks more centered in the new screenshot.

PR #871, opened 2025-02-17
Update `float-cmp` and `pollster`

waywardmonkeys:
These are the direct dependencies that have non-semver compatible updates (apart from wgpu).


DJMcNab (approved):
I'm not sure why this is needed, but it doesn't hurt anything

waywardmonkeys:
So that when I pull in an update via GitHub dependency, I wont get some duplicated deps.

Issue #869, opened 2025-02-15
Blurry 1px-wide borders on Windows


eugenesvk:
Was trying to replicate a Windows button style, and set the border to 1px width instead of 2 (and changed the color), but the button has a "fuzzy" border, see this scaled up image

- top is default windows button from a Wordpad, it has a crisp border exactly 1-px width both vertically and horizontally

- bottom is a Masonry button widget, which is not universally 1px (and also the color is washed out, so it's blended with the button color?)
<img width="148" alt="Image" src="https://github.com/user-attachments/assets/a2dfcd86-7f56-456c-828a-53267857c2c1" />

Is there a way to preserve the exact color and border size?

After some experiments: there seems to be 2 issues here
  - Border is incorrectly drawn "halfway", on top of the inner rect, not adjacent to it, so only half of the border width gets show, the other half is obscured by the background. And `stroke width` ≠ `border width`
  - Border is drawn before background, so they get "blended" at the edge?

For example, here green border is drawn with a corrected positioning - it's drawn **adjacent** to the inner rect, not on top of it (it's drawn on top of the
 - `rect - ½stroke_width`, but background is filling the 
 - `rect - 1stroke_width`)

, so it's not blended (red one is)
<img width="143" alt="Image" src="https://github.com/user-attachments/assets/76339872-ec11-471a-bf52-1a09aba6c016" />

But if red border is drawn on top of the background, then even its incorrect positioning doesn't matter - you still get crisp border

<img width="129" alt="Image" src="https://github.com/user-attachments/assets/7b65665f-0e0c-43ca-8678-64b2bbf349a3" />

(these are buttons with stroke width `1/dpi_scaling_factor` so you get 1px)

eugenesvk:
Interestingly enough, the color wash out is inconsistent:
- left: `1`px border stroke new
- right: `1`px border stroke with dashes of `0` offset and `1` 0` pattern (so, I guess, would be identical to a solid line)

On the right top border is proper non-washed-out color, but then its other borders are even worse than on the left as they have a "checkered" dashed pattern (and they are also not 1px, just like the left one)
<img width="270" alt="Image" src="https://github.com/user-attachments/assets/60647102-aebc-41e0-a6b1-89e5f909f094" />

burrbull:
There were render improvements in last egui release. Maybe this is helpful.
https://github.com/emilk/egui/pull/5669

DJMcNab:
There are a couple of relevant things here:
[...]

PR #872, opened 2025-02-18
Add DPI scale factor to the render root state

eugenesvk:
so that it's available to widgets that can fix blurry lines due to bad overlapping positioning of its internal elements

partially addresses https://github.com/linebender/xilem/issues/869

### DIFF
--- a/masonry/src/app/render_root.rs
+++ b/masonry/src/app/render_root.rs
@@ -147,6 +147,7 @@ pub(crate) struct RenderRootState {
     /// Pass tracing configuration, used to skip tracing to limit overhead.
     pub(crate) trace: PassTracing,
     pub(crate) inspector_state: InspectorState,
+    pub(crate) scale_factor: f64,
 }
 
 pub(crate) struct MutateCallback {
@@ -290,6 +291,7 @@ impl RenderRoot {
                     is_picking_widget: false,
                     hovered_widget: None,
                 },
+                scale_factor,
             },
             widget_arena: WidgetArena {
                 widgets: TreeArena::new(),
@@ -339,6 +341,7 @@ impl RenderRoot {
         match event {
             WindowEvent::Rescale(scale_factor) => {
                 self.scale_factor = scale_factor;
+                self.global_state.scale_factor = scale_factor;
                 self.request_render_all();
                 Handled::Yes
             }
--- a/masonry/src/core/contexts.rs
+++ b/masonry/src/core/contexts.rs
@@ -777,6 +777,11 @@ impl_context_method!(
         pub fn to_window(&self, widget_point: Point) -> Point {
             self.widget_state.window_transform * widget_point
         }
+
+        /// Get DPI scaling factor.
+        pub fn get_scale_factor(&self) -> f64 {
+            self.global_state.scale_factor
+        }
     }
 );
 


masonry/src/core/contexts.rs
```diff
@@ -777,6 +777,11 @@ impl_context_method!(
         pub fn to_window(&self, widget_point: Point) -> Point {
             self.widget_state.window_transform * widget_point
         }
+
+        /// Get DPI scaling factor.
+        pub fn get_scale_factor(&self) -> f64 {
```
DJMcNab, line 782: Why is this made available for `ComposeCtx`? If the scale factor changes, as it currently stands, `compose` won't be re-run, so using this in `ComposeCtx` would lead to stale results.

eugenesvk: Excluded ComposeCtx

DJMcNab:
I'll defer to @PoignardAzur, but I am happy to approve in principle, as a stop-gap until we get a more coherent design.

(This would also be useful for e.g. `Image` type widgets, to know which resolution of an image to load from the remote)

Obviously it is true that this value is pretty meaningless for widgets with a transform applied, but that's pretty unavoidable; Transforms are designed to break these kinds of things anyway.

masonry/src/core/contexts.rs
DJMcNab: Thanks! I think the set provided there is mostly reasonable.

masonry/src/app/render_root.rs
```diff
@@ -339,6 +341,7 @@ impl RenderRoot {
         match event {
             WindowEvent::Rescale(scale_factor) => {
                 self.scale_factor = scale_factor;
+                self.global_state.scale_factor = scale_factor;
```
PoignardAzur, line 344: There seems to be some redundancy there. You should probably remove `RenderRoot::scale_factor` and have `RenderRootState::scale_factor` be the only canonical value.

masonry/src/core/contexts.rs
```diff
@@ -780,6 +780,23 @@ impl_context_method!(
...
+    PaintCtx<'_>,
+    AccessCtx<'_>,
+    {
+        /// Get DPI scaling factor.
+        pub fn get_scale_factor(&self) -> f64 {
+            self.global_state.scale_factor
+        }
+    }
+);
+
```
PoignardAzur, line 799: Given that changing the scale factor only causes a re-render, I would recommend only giving access to it in PaintCtx and AccessCtx. Basically assume it's an implementation detail of the rendering and it's not accessible to app logic.

PoignardAzur (approved):

masonry/src/core/contexts.rs
eugenesvk: limited to 2

masonry/src/app/render_root.rs
eugenesvk: removed from Root

masonry/src/core/contexts.rs
```diff
@@ -780,6 +780,15 @@ impl_context_method!(
     }
 );
 
+// Methods on selected context types.
+// Access selected global state information.
+impl_context_method!(PaintCtx<'_>, AccessCtx<'_>, {
+    /// Get DPI scaling factor.
```
DJMcNab, line 786:
```suggestion
    /// Get DPI scaling factor.
    ///
    /// This is not required for most widgets, and should be used only for precise 
    /// rendering, such as rendering single pixel lines or selecting image variants.
    /// This is currently only provided in the render stages, as these are the only passes which
    /// are re-run when the scale factor changes.
    ///
    /// Note that accessibility nodes and paint results will automatically be scaled by Masonry.
    /// This also doesn't account for the widget's current transform, which cannot currently be
    /// accessed by widgets directly.
```

DJMcNab (commented):
Thank you for your patience. I don't need you to do anything else on this; I'll apply the suggestions and merge tomorrow.

PR #873, opened 2025-02-18
Add arbitrary properties to widgets

PoignardAzur:
Add `Properties[Mut]` argument to widget methods.
Add a third TreeArena to sort per-widget arbitrary property values.

For the type-to-value map, I considered the following crates:
- https://docs.rs/typemap/latest/typemap/
- https://crates.io/crates/typemap_rev
- https://crates.io/crates/typemap-ors
- https://github.com/chris-morgan/anymap
- https://github.com/reivilibre/anymap3

Of these, anymap3 is the only one actively maintained (last commit less than 12 months ago). The code source itself is extremely light and simple; we may or may not decide to roll out our own implementation down the line.

Add a BackgroundBrush property used by SizedBox as a proof of concept.

Note that SizedBox still has its `background` field, but we should expect future widgets to use almost *exclusively* properties; properties usually shouldn't be redundant with local fields.

To get there, we'll first need to better integrate properties in Xilem.


PoignardAzur:
Remaining work:

- Add way to create widgets with pre-filled property values.
- Add way to change property values from WidgetMut.
- Have at least one widget use property values.
- Write unit tests.
- Document feature.

PoignardAzur:
I'm not super happy about the documentation I wrote, but it will do for now. The most important is that we don't start adding undocumented features and maintain "everything should be documented" discipline.

PoignardAzur:
> It is not clear to me why the properties are not part of WidgetState. As far as I can tell (see the first sentence), its ownership and usage is identical to how WidgetState is used. If these were combined, this PR would also be 10x easier to review.

It's mostly for lifetime and DX reasons. Essentially, writing

```rust
x = props.get::<SomeValue>();
```

is shorter and in my strongly-held opinion *better* than

```rust
x = ctx.props.get::<SomeValue>();
x = ctx.props().get::<SomeValue>();
x = ctx.properties().get::<SomeValue>();
```

Either way, `x` is very likely to be a reference (unless we mandate `SomeValue: Copy`, which has its own implications). Holding that reference will create lifetime problems if we also call mutating methods on `ctx`, which most implementations will.

Hence, props and ctx probably need to be passed to widget methods separately.

---

I think all your other points boil down to "The WidgetProperty trait is badly designed", and now that you point it out, I agree.

Being able to associated arbitrary data with arbitrary widgets is a fundamental part of the design. But I agree that properties shouldn't have "opinions" on what bits of context need to update when they're changed, so this part needs to go back to the drawing board.

Let's move the discussion [to zulip](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Widget.20properties).

PoignardAzur:
> As I've already mentioned, it makes this PR unreviewable.

That's a one-time cost, though.

> I'll cover lifetimes in a bit, but I would claim that the DX cost of having an extra argument on every single method is... massive.

I disagree that adding one more argument to all Widget methods is a massive DX cost. It was the status quo with Druid (except no, because Druid's methods had data *and* env arguments) and people didn't complain. I did notice that removing those extra arguments made Masonry feel more readable, but not "block a feature over it" more readable.

(Especially since we've removed manual recursion for most passes, so end users don't need to worry about *calling* these methods.)

> However, this solution is extremely bad, if it is at all avoidable. An alternative I'd like to explore would be to add something like:
> 
> ```rust
> impl LayoutCtx<'a> {
>     fn take_properties<'ref>(&'ref mut self)->&'a mut Properties;
> }
> ```

I can try tinkering with lifetimes if you think we really to find another syntax, but I think there's basically no way to do that within Rust's borrowing rules.

If calling a method lets you mutate something that you can reach through your `LayoutCtx`, then the method **must** mutably borrow `LayoutCtx` as long as you can mutate that something (aside from loopholes like interior mutability). There no possible way to declare `take_properties` that will let you write

```rust
let value = ctx.take_properties().get::<Foo>();
ctx.request_whatever();
x = value.foo;
```

Also, while right now it might feel like we're adding an extra argument and leaving it unused, I expect that in the future widgets will use properties *a lot*. Hence why I think it's important to keep the "reading a property" code as short as possible. I expect that a lot of widgets may have dozens of lines in a row of reading properties and calling methods from them.

PoignardAzur:
(Also, I agree the PR is pretty hard to review, but if you mean "unreviewable" literally, then I have to disagree. Most of the changes are only plumbing, and don't affect program logic.)

DJMcNab:
For what it's worth, I'm really excited about this feature. The use of this API for flex and other layout properties will be a game-changer for ergonomics and composability, and it has other very appealing properties.

---

> Also, while right now it might feel like we're adding an extra argument and leaving it unused, I expect that in the future widgets will use properties a lot. Hence why I think it's important to keep the "reading a property" code as short as possible. I expect that a lot of widgets may have dozens of lines in a row of reading properties and calling methods from them.

I agree completely here. If at some future point in time, this API of every method always being given the properties is the correct one, I'd be happy to move to it! However, the form I'm proposing allows us to add properties as/when we see fit, allowing us to evaluate that later change on its own merits.

> There no possible way to declare `take_properties` that will let you write
> 
> ```rust
> let value = ctx.take_properties().get::<Foo>();
> ctx.request_whatever();
> x = value.foo;
> ```

I have constructed a [counterexample](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dd07d31768e1cff85dca39a72a1c155b), which contains that exact snipped (except with indentation), which compiles and runs.

Most users should of course use:
```rust
let props = ctx.take_properties();
let value = props.get::<Foo>();
```

The exact storage of the arena of states would need to change, to contain something like:
```rust
struct WidgetData {
    properties: Properties,
    state: WidgetState
}
```

(there's an aside here I'm not including to avoid derailing things too much)

> but not "block a feature over it" more readable.

I'm not at all saying that we should block the feature over this; I'm not sure how you reached that conclusion. I'm saying that I would strongly prefer that we use a backwards-compatible addition of this feature whilst it's still not properly used at all internally.

> (Also, I agree the PR is pretty hard to review, but if you mean "unreviewable" literally, then I have to disagree. Most of the changes are only plumbing, and don't affect program logic.)

I gave this an earnest attempt at review, and had to give up and say "to summarise my understanding", as in "I don't know that what I'm saying is correct, because I think I've probably missed some important parts". But you're right, unreviewable is not the right term, as that's a universal statement - I should have said "unreviewable by me". Apologies

PoignardAzur:
I've given it some thought, and I want to keep the architecture proposed by the PR, extra parameter and all, unless you or someone is very strongly opposed to it. I think implementing the PR some other way (eg with the proposed WidgetData type) would still be a major refactor with a massive diff, only it would also not be the architecture I want and we'd end up having the huge refactor later on anyway.

That said, I'm looking into ways to make the PR more digestible and reviewable. Worst case scenario, I can give you a list of files with non-plumbing changes, or we can schedule time to have a live code review.

Also, there's the problems we mentioned with WidgetProperty that still need to be fixed.

DJMcNab:
I'm not going to block on the argument change. I still very strongly disagree with it, given the current state of the feature and the widget set; but I can see how if it does get used as you claim it will be, it would make sense.

Architecturally I still don't think adding another arena for this value is the right choice. I admit that it requires a first massive refactor to make the arena be of `struct WidgetData { state: WidgetState }`. And actually doing that transformation in this PR would not be good - it would need to be a predecessor. But semantically, the properties are not meaningfully different from being "state".

I'm unlikely to be able to prioritise reviewing this.

masonry/src/core/properties.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use anymap3::{AnyMap, Entry};
+
+// TODO - Add PropertyValue wrapper struct that implements receiver trait.
+// Return PropertyValue<T> instead of Option<T> from methods.
```
DJMcNab, line 7: Why?

masonry/src/core/widget_pod.rs
```diff
@@ -60,9 +61,23 @@ impl<W: Widget + ?Sized> WidgetPod<W> {
             inner: WidgetPodInner::Create(CreateWidget {
                 widget: inner,
                 transform,
+                properties: Properties::new(),
             }),
         }
     }
+
+    // TODO - Remove transform, have it as a special-case property instead.
```
DJMcNab, line 69: Hmm, interesting!

DJMcNab (approved):
I still disagree both with the architecture of a new arena, and a new parameter added to each widget method.

However, there is not will to implement this in a different way, so we've decided to move forward with this.

masonry/src/core/properties.rs
PoignardAzur: The idea is that widget methods will use a lot of properties. Instead of having one `if let Some(foo) = ...` block per property access, which is three lines minimum with standard rustfmt, they should instead call methods directly on the optional property, for which we need something implementing the receiver trait.

PR #874, opened 2025-02-20
Improve bounding rect code

PoignardAzur:
Document concepts of "layout rect" and "bounding rect".
Deprecate `local_layout_rect()` method.
Add `global_layout_rect()` method.
Remove `transform_changed()` method.
Improve bounding rect merging code.
Tweak `TestHarness::mouse_move_to`.
Tweak WidgetState doc.

### DIFF
--- a/masonry/src/core/contexts.rs
+++ b/masonry/src/core/contexts.rs
@@ -632,12 +632,8 @@ impl LayoutCtx<'_> {
         self.get_child_state(child).baseline_offset
     }
 
-    /// Get the given child's layout rect.
-    ///
-    /// ## Panics
-    ///
-    /// This method will panic if [`LayoutCtx::run_layout`] and [`LayoutCtx::place_child`]
-    /// have not been called yet for the child.
+    // TODO - Remove (used in Flex)
+    #[doc(hidden)]
     #[track_caller]
     pub fn child_layout_rect(&self, child: &WidgetPod<impl Widget + ?Sized>) -> Rect {
         self.assert_layout_done(child, "child_layout_rect");
@@ -667,7 +663,7 @@ impl LayoutCtx<'_> {
     #[track_caller]
     pub fn child_size(&self, child: &WidgetPod<impl Widget + ?Sized>) -> Size {
         self.assert_layout_done(child, "child_size");
-        self.get_child_state(child).layout_rect().size()
+        self.get_child_state(child).size
     }
 
     /// Skips running the layout pass and calling [`LayoutCtx::place_child`] on the child.
@@ -766,16 +762,6 @@ impl_context_method!(
         /// [`layout`]: Widget::layout
         pub fn size(&self) -> Size {
             self.widget_state.size
-        }
-
-        /// The layout rect of the widget in window coordinates.
-        ///
-        /// This is the layout [size](Self::size) and [window_origin](Self::window_origin) combined.
-        ///
-        /// See [layout rect documentation](crate::doc::doc_06_masonry_concepts#layout-rect)
-        /// for details.
-        pub fn global_layout_rect(&self) -> Rect {
-            Rect::from_origin_size(self.widget_state.window_origin(), self.widget_state.size)
         }
 
         // TODO - Remove
--- a/masonry/src/core/contexts.rs
+++ b/masonry/src/core/contexts.rs
@@ -768,12 +768,19 @@ impl_context_method!(
             self.widget_state.size
         }
 
-        // TODO - Remove? A widget doesn't really have a concept of its own "origin",
-        // it's more useful for the parent widget.
-        /// The layout rect of the widget.
+        /// The layout rect of the widget in window coordinates.
         ///
-        /// This is the layout [size](Self::size) and origin (in the parent's coordinate space) combined.
-        pub fn layout_rect(&self) -> Rect {
+        /// This is the layout [size](Self::size) and [window_origin](Self::window_origin) combined.
+        ///
+        /// See [layout rect documentation](crate::doc::doc_06_masonry_concepts#layout-rect)
+        /// for details.
+        pub fn global_layout_rect(&self) -> Rect {
+            Rect::from_origin_size(self.widget_state.window_origin(), self.widget_state.size)
+        }
+
+        // TODO - Remove
+        #[allow(dead_code, reason = "Only used in tests")]
+        pub(crate) fn local_layout_rect(&self) -> Rect {
             self.widget_state.layout_rect()
         }
 
@@ -788,7 +795,10 @@ impl_context_method!(
             self.widget_state.window_origin()
         }
 
-        /// The axis aligned bounding rect of this widget in window coordinates.
+        /// The bounding rect of the widget in window coordinates.
+        ///
+        /// See [bounding rect documentation](crate::doc::doc_06_masonry_concepts#bounding-rect)
+        /// for details.
         pub fn bounding_rect(&self) -> Rect {
             self.widget_state.bounding_rect()
         }
--- a/masonry/src/core/contexts.rs
+++ b/masonry/src/core/contexts.rs
@@ -1040,13 +1040,6 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, UpdateCtx<'_>, {
         self.request_layout();
     }
 
-    /// Indicate that the transform of this widget has changed.
-    pub fn transform_changed(&mut self) {
-        trace!("transform_changed");
-        self.widget_state.transform_changed = true;
-        self.request_compose();
-    }
-
     /// Indicate that a child is about to be removed from the tree.
     ///
     /// Container widgets should avoid dropping `WidgetPod`s. Instead, they should
@@ -1083,7 +1076,8 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, UpdateCtx<'_>, {
     /// It behaves similarly as CSS transforms
     pub fn set_transform(&mut self, transform: Affine) {
         self.widget_state.transform = transform;
-        self.transform_changed();
+        self.widget_state.transform_changed = true;
+        self.request_compose();
     }
 });
 
--- a/masonry/src/core/widget_state.rs
+++ b/masonry/src/core/widget_state.rs
@@ -300,6 +300,24 @@ impl WidgetState {
         self.window_transform.translation().to_point()
     }
 
+    /// Return the result of applying the widget's clip shape (if any) to the given rect.
+    ///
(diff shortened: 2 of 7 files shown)


masonry/src/core/widget_state.rs
```diff
@@ -300,6 +302,24 @@ impl WidgetState {
         self.window_transform.translation().to_point()
     }
 
+    /// Return the result of applying the widget's clip shape (if any) to the given rect.
+    ///
+    /// Both the argument and the result are in global coordinates.
+    ///
+    /// Returns None if the given rect is out of bounds of the clip.
```
waywardmonkeys, line 309: Backticks on `None`.

masonry/src/core/widget_state.rs
```diff
@@ -300,6 +302,24 @@ impl WidgetState {
         self.window_transform.translation().to_point()
     }
 
+    /// Return the result of applying the widget's clip shape (if any) to the given rect.
```
waywardmonkeys, line 305: You say "clip shape" here, but the code calls it a `clip_path` ... I think best to have a single term

masonry/src/core/widget_state.rs
```diff
@@ -300,6 +302,24 @@ impl WidgetState {
...
+    ///
+    /// Both the argument and the result are in global coordinates.
```
waywardmonkeys, line 307: Global to what? Window space, right?

masonry/src/passes/compose.rs
```diff
@@ -75,18 +75,10 @@ fn compose_widget(
             );
             let parent_bounding_rect = parent_state.bounding_rect;
 
-            // This could be further optimized by more tightly clipping the child bounding rect according to the clip path.
```
waywardmonkeys, line 78: This note about optimization is not present after the code was moved. Is it still a valid future thing to do?

waywardmonkeys: Or is that the new call to `overlaps`? Either way, the logic in this code has changed, so the PR title that this is just adding documentation starts to become misleading.

waywardmonkeys: Ah, you renamed it after I started looking at it.

PoignardAzur: > This note about optimization is not present after the code was moved. Is it still a valid future thing to do?

I'd have to ask @Philipp-M. I'm not sure the note about optimization made sense in the original PR. I'd rather remove it unless we can point at a specific optimization we expect to do.

PoignardAzur:
@Philipp-M Any chance you can review this weekend? Otherwise, I expect to merge on Monday.

masonry/src/core/contexts.rs
```diff
@@ -730,12 +730,19 @@ impl_context_method!(
...
         ///
-        /// This is the layout [size](Self::size) and origin (in the parent's coordinate space) combined.
-        pub fn layout_rect(&self) -> Rect {
+        /// This is the layout [size](Self::size) and [window_origin](Self::window_origin) combined.
+        ///
+        /// See [layout rect documentation](crate::doc::doc_06_masonry_concepts#layout-rect)
+        /// for details.
+        pub fn global_layout_rect(&self) -> Rect {
+            Rect::from_origin_size(self.widget_state.window_origin(), self.widget_state.size)
+        }
```
Philipp-M, line 741: I'm not sure whether this is really useful at all, as this doesn't really represent the (global) widgets position, bounds or size (due to potential transformations).

I wonder whether we should get rid of "layout rects"?
I think they cause more confusion than being helpful.

masonry/src/doc/06_masonry_concepts.md
```diff
@@ -100,6 +100,26 @@ A widget is considered "interactive" if it can still get text and/or pointer eve
 Stashed and disabled widget are non-interactive.
 
 
+## Layout rect
+
+A widget's layout rect includes its self-declared size, and the position attributed by its parent (local position) / parents (global position).
+
+Pointer events cannot target a widget if they are outside its layout rect, though they can target its children (see **Bounding rect** section).
+
+A layout rect doesn't necessarily have a formal definition; it's generally where the widget will be drawn, though a widget can be drawn outside its layout rect; it's usually a widget's own space and containers will try to avoid having its children's layout rects overlap.
```
Philipp-M, line 109: I think this section should somehow explain, that a widget can be transformed by its transform.

Pointer events also only reach the widget, when they intersect the *transformed* widget, thus I don't think it's currently correct. Try for example transformations on the left portal in the xilem http_cats example, to see what I mean.

masonry/src/doc/06_masonry_concepts.md
```diff
@@ -100,6 +100,26 @@ A widget is considered "interactive" if it can still get text and/or pointer eve
...
+
+
+## Bounding rect
+
+A widget's bounding rect is a rectangle inside of which pointer events might affect either the widget or its descendants.
+
+In general, the bounding rect is a union or a widget's layout rect and the bounding rects of all its descendants.
+
+The bounding rects of the widget tree form a kind of "bounding volume hierarchy": when looking to find which widget a pointer is on, Masonry will automatically exclude any widget if the pointer is outside its bounding rect.
```
Philipp-M, line 118: Might be worth mentioning that bounding rects currently are in window space and axis aligned.

masonry/src/passes/compose.rs
Philipp-M: Well since the bounds might need to be bigger when transforming the AABB (due to not being able to tighly fit a transformed AABB), it's probably larger afterwards.
An optimization might be an OOB to more tightly represent the widget, or more optimally an arbitrary shape.
I'm not sure how useful that comment is, but at least it hints that this is not yet optimal.

masonry/src/testing/harness.rs
```diff
@@ -472,8 +472,7 @@ impl TestHarness {
     #[track_caller]
     pub fn mouse_move_to(&mut self, id: WidgetId) {
         let widget = self.get_widget(id);
-        let local_widget_center = (widget.ctx().size() / 2.0).to_vec2().to_point();
-        let widget_center = widget.ctx().widget_state.window_transform * local_widget_center;
+        let widget_center = widget.ctx().global_layout_rect().center();
```
Philipp-M, line 475: I don't think this change is correct, in that a transformed widget might not be hit.

Philipp-M (commented):
I'm a little bit uncomfortable with this change in this form TBH.

I do think we may even want to get rid of layout rects completely, and just keep the "OOB" (ignoring children, i.e. transformed local layout rect?) which is defined by the `origin` and `size` * `window_transform` to keep it more simple.

masonry/src/passes/compose.rs
PoignardAzur: I'm fine not hinting at a future optimization, because I highly doubt there will be one. Tightly clipping an arbitrary path in an oriented bounding box is a hard problem, and it's unclear that it would have any performance benefit at all.

99.9% of cases are going to be AABBs anyway, for which the current code is close to optimal.

PoignardAzur:
On second thought, I agree with @Philipp-M that the "layout rect" concept doesn't really make sense anymore with arbitrary transforms.

I've edited the documentation to reflect this.

masonry/src/core/contexts.rs
```diff
@@ -594,12 +594,8 @@ impl LayoutCtx<'_> {
...
-    /// Get the given child's layout rect.
-    ///
-    /// ## Panics
-    ///
-    /// This method will panic if [`LayoutCtx::run_layout`] and [`LayoutCtx::place_child`]
-    /// have not been called yet for the child.
+    // TODO - Remove (used in Flex)
+    #[doc(hidden)]
     #[track_caller]
     pub fn child_layout_rect(&self, child: &WidgetPod<impl Widget + ?Sized>) -> Rect {
```
Philipp-M, line 600: If it's only used in Flex maybe:
```suggestion
    pub(crate) fn child_layout_rect(&self, child: &WidgetPod<impl Widget + ?Sized>) -> Rect {
```

Might also be worth to fix the parts where the code is used in `Flex`?
So that we can get rid of it entirely here?

masonry/src/doc/06_masonry_concepts.md
```diff
@@ -100,6 +100,26 @@ A widget is considered "interactive" if it can still get text and/or pointer eve
...
+
+A widget's bounding rect is a window-space axis-aligned rectangle inside of which pointer events might affect either the widget or its descendants.
+
+In general, the bounding rect is a union or a widget's layout rect and the bounding rects of all its descendants.
+
+The bounding rects of the widget tree form a kind of "bounding volume hierarchy": when looking to find which widget a pointer is on, Masonry will automatically exclude any widget if the pointer is outside its bounding rect.
+
+<!-- TODO - Include illustration. -->
+
+<!-- TODO - Add section about clip paths and pointer detection. -->
```
Philipp-M, line 113: Do you plan to handle the TODOs in this PR?

masonry/src/doc/06_masonry_concepts.md
```diff
@@ -100,6 +100,26 @@ A widget is considered "interactive" if it can still get text and/or pointer eve
...
+
+
+## Layout rect
+
+Previous versions of Masonry had a concept of a widget's "layout rect", composed of its self-declared size and the position attributed by its parent.
+
+However, given that widgets can have arbitrary transforms, the concept of an axis-aligned layout rect doesn't really make sense anymore.
```
Philipp-M, line 120: Not sure if it's even worth mentioning anymore, when all the methods are all "private" and planned to be deleted?

masonry/src/passes/compose.rs
Philipp-M: Yeah I'm not too attached to that comment, feel free to remove it.

Philipp-M (approved):
Thanks, I'm much more comfortable with this.
I think the PR could already be handling the added TODOs (as IME merged TODOs are less likely to be handled, and I don't think they are a lot of work to handle), but I'm not going to block on this.

masonry/src/core/contexts.rs
PoignardAzur: I think that's a whole PR on its own.

masonry/src/doc/06_masonry_concepts.md
PoignardAzur: We need to refactor how clip paths work at some point, it would be part of that.

masonry/src/doc/06_masonry_concepts.md
PoignardAzur: I'd like to keep it at least until all references are removed from the code, including private code.

PR #875, opened 2025-02-24
add canvas widget

richard-uk1:
I'm playing with making a vector path editor in `xilem`. This requires some extra stuff over what is currently available so I'm going to maintain a fork. I will try to merge upstream anything I think is more widely useful.

This PR allows the user to draw directly onto the scene fragment for a widget. Currently, the user must recreate the draw function every time there is a change in state. I wonder if there should be separated out state and draw parts, so the function needs to be updated less often.

### DIFF
--- a/masonry/src/widgets/canvas.rs
+++ b/masonry/src/widgets/canvas.rs
@@ -1,0 +1,137 @@
+// Copyright 2025 the Xilem Authors and the Druid Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#![warn(missing_docs)]
+
+//! A canvas widget.
+
+use accesskit::{Node, Role};
+use smallvec::SmallVec;
+use tracing::{trace_span, Span};
+use vello::kurbo::Size;
+use vello::Scene;
+
+use crate::core::{
+    AccessCtx, AccessEvent, BoxConstraints, EventCtx, LayoutCtx, PaintCtx, PointerEvent, QueryCtx,
+    RegisterCtx, TextEvent, Update, UpdateCtx, Widget, WidgetId, WidgetMut,
+};
+
+/// A widget allowing custom drawing.
+pub struct Canvas {
+    draw: Box<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>,
+    // TODO: pointer events
+}
+
+// --- MARK: BUILDERS ---
+impl Canvas {
+    /// Create a new canvas with the given draw function
+    pub fn new(draw: impl Fn(&mut Scene, Size) + Send + Sync + 'static) -> Self {
+        Self {
+            draw: Box::new(draw),
+        }
+    }
+}
+
+// --- MARK: WIDGETMUT ---
+impl Canvas {
+    /// Update the draw function
+    pub fn update_draw(
+        this: &mut WidgetMut<'_, Self>,
+        draw: impl Fn(&mut Scene, Size) + Send + Sync + 'static,
+    ) {
+        this.widget.draw = Box::new(draw);
+        this.ctx.request_render();
+    }
+}
+
+// --- MARK: IMPL WIDGET ---
+impl Widget for Canvas {
+    fn on_pointer_event(&mut self, _ctx: &mut EventCtx, _event: &PointerEvent) {
+        // TODO: ensure coordinates are correct and pass to callback
+    }
+
+    fn accepts_pointer_interaction(&self) -> bool {
+        true
+    }
+
+    fn on_text_event(&mut self, _ctx: &mut EventCtx, _event: &TextEvent) {}
+
+    fn on_access_event(&mut self, _ctx: &mut EventCtx, _event: &AccessEvent) {}
+
+    fn register_children(&mut self, _ctx: &mut RegisterCtx) {}
+
+    fn update(&mut self, _ctx: &mut UpdateCtx, _event: &Update) {}
+
+    fn layout(&mut self, _ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
+        // use as much space as possible - caller can size it as necessary
+        bc.max()
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, scene: &mut Scene) {
+        (self.draw)(scene, ctx.size())
+    }
+
+    fn accessibility_role(&self) -> Role {
+        Role::Canvas
+    }
+
+    fn accessibility(&mut self, _ctx: &mut AccessCtx, _node: &mut Node) {
+        // TODO: should probably give the caller the opportunity to handle accessibility
+    }
+
+    fn children_ids(&self) -> SmallVec<[WidgetId; 16]> {
+        SmallVec::new()
+    }
+
+    fn make_trace_span(&self, ctx: &QueryCtx<'_>) -> Span {
+        trace_span!("Canvas", id = ctx.widget_id().trace())
+    }
+
+    fn get_debug_text(&self) -> Option<String> {
+        None
+    }
+}
+
+// --- MARK: TESTS ---
+#[cfg(test)]
+mod tests {
+    use insta::assert_debug_snapshot;
+    use vello::kurbo::{Affine, BezPath, Stroke};
+    use vello::peniko::{Color, Fill};
+
+    use super::*;
+    use crate::assert_render_snapshot;
+    use crate::testing::TestHarness;
+
+    #[test]
+    fn simple_canvas() {
+        let canvas = Canvas::new(|scene, size| {
+            let scale = Affine::scale_non_uniform(size.width, size.height);
+            let mut path = BezPath::new();
+            path.move_to((0.1, 0.1));
+            path.line_to((0.9, 0.9));
+            path.line_to((0.9, 0.1));
+            path.close_path();
+            path = scale * path;
+            scene.fill(
+                Fill::NonZero,
+                Affine::IDENTITY,
+                Color::from_rgb8(100, 240, 150),
+                None,
+                &path,
+            );
+            scene.stroke(
+                &Stroke::new(4.),
+                Affine::IDENTITY,
+                Color::from_rgb8(200, 140, 50),
+                None,
+                &path,
+            );
+        });
+
+        let mut harness = TestHarness::create(canvas);
+
+        assert_debug_snapshot!(harness.root_widget());
+        assert_render_snapshot!(harness, "hello");
+    }
+}
--- a/masonry/src/testing/harness.rs
+++ b/masonry/src/testing/harness.rs
@@ -715,7 +715,10 @@ impl TestHarness {
             // Remove '<test_name>.new.png' file if it exists
             let _ = std::fs::remove_file(&new_path);
             new_image.save(&new_path).unwrap();
-            panic!("Snapshot test '{test_name}' failed: No reference file");
+            panic!(
+                "Snapshot test '{test_name}' failed: No reference file\n\
+            New screenshot created with `.new.png` extension. If correct, change to `.png`"
+            );
         }
     }
 }
--- a/masonry/src/widgets/mod.rs
+++ b/masonry/src/widgets/mod.rs
@@ -11,6 +11,7 @@ mod tests;
 
 mod align;
 mod button;
+mod canvas;
 mod checkbox;
 mod flex;
 mod grid;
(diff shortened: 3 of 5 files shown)


masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+        this.widget.draw = Box::new(draw);
+        this.ctx.request_render();
+    }
+}
+
+// --- MARK: IMPL WIDGET ---
+impl Widget for Canvas {
+    fn on_pointer_event(&mut self, _ctx: &mut EventCtx, _event: &PointerEvent) {
+        // TODO: ensure coordinates are correct and pass to callback
```
DJMcNab, line 50: This comment is misleading. Something has gone wrong if the canvas widget is managing the pointer event. At that point, the user should be writing their own widget, I believe.

A canvas is basically equivalent to an Image, but containing a vector scene (ish).

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn paint(&mut self, ctx: &mut PaintCtx, scene: &mut Scene) {
+        (self.draw)(scene, ctx.size())
+    }
+
+    fn accessibility_role(&self) -> Role {
+        Role::Canvas
+    }
+
+    fn accessibility(&mut self, _ctx: &mut AccessCtx, _node: &mut Node) {
+        // TODO: should probably give the caller the opportunity to handle accessibility
```
DJMcNab, line 79: This would probably just be an alt-text like mechanism, although maybe we should also be strongly encouraging that a caption exist?

We haven't handled this for Image, which this is very similar to.

DJMcNab (commented):
Apart from the pointer event TODO comment, this seems reasonable as an addition to Masonry.

richard-uk1:
One question I have: can a scene fragment be retained and appended to the scene on paint?

DJMcNab:
Masonry is already doing that internally; your paint callback will only be called once, except when it gets replaced.

richard-uk1:
OK ready for review.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
+// Copyright 2025 the Xilem Authors and the Druid Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#![warn(missing_docs)]
+
```
PoignardAzur, line 5: I don't think the annotation is needed. Either way, we should avoid adding undocumented code.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+/// A widget allowing custom drawing.
+pub struct Canvas {
+    draw: Arc<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>,
+    alt_text: Option<String>,
+}
+
+// --- MARK: BUILDERS ---
+impl Canvas {
+    /// Create a new canvas with the given draw function
```
PoignardAzur, line 29:
```suggestion
    /// Create a new canvas with the given draw function.
```

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+        Self {
+            draw,
+            alt_text: None,
+        }
+    }
+
+    /// Set the text that will be used to communicate the meaning of the canvas to
+    /// those using screen readers.
+    ///
+    /// Users are strongly encouraged to set alt text for the canvas.
```
PoignardAzur, line 45: I have mixed feelings about that sentence. It's a common piece of advice, but one I'd argue is often misinterpreted as "Apply descriptive alt text everywhere".

When to add non-empty alt text or not could be the subject of an entire article; and there's often situations where adding a description makes a page *less* friendly to screen readers.

Since we don't have such an article written yet, I'd default to something like:

```suggestion
    /// Users are strongly encouraged to set alt text for the canvas.
    /// If possible, the alt-text should succinctly describe what the canvas represents.
    ///
    /// If the canvas is decorative or too hard to describe through text, users should set alt text to `""`.
```

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+
+// --- MARK: WIDGETMUT ---
+impl Canvas {
+    /// Update the draw function
+    pub fn update_draw(
+        this: WidgetMut<'_, Self>,
+        draw: impl Fn(&mut Scene, Size) + Send + Sync + 'static,
+    ) {
+        Self::update_from_arc(this, Arc::new(draw));
+    }
```
PoignardAzur, line 60: I think a name like `set_painter` would be more in line with similar methods in other widgets.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// Users are strongly encouraged to set alt text for the canvas.
+    pub fn with_alt_text(mut self, alt_text: impl Into<String>) -> Self {
+        self.alt_text = Some(alt_text.into());
+        self
+    }
+}
+
+// --- MARK: WIDGETMUT ---
+impl Canvas {
```
PoignardAzur, line 53: There should also be a method to set/remove the alt text.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+// --- MARK: IMPL WIDGET ---
+impl Widget for Canvas {
+    fn on_pointer_event(&mut self, _ctx: &mut EventCtx, _event: &PointerEvent) {}
+
+    fn accepts_pointer_interaction(&self) -> bool {
+        false
+    }
```
PoignardAzur, line 78: I don't think you should set this flag to false. It makes Canvas transparent to pointers, meaning for example that if there's a widget behind the canvas, that widget will get pointer events, affect the pointer icon, etc.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn update(&mut self, _ctx: &mut UpdateCtx, _event: &Update) {}
+
+    fn layout(&mut self, _ctx: &mut LayoutCtx, bc: &BoxConstraints) -> Size {
+        // use as much space as possible - caller can size it as necessary
+        bc.max()
+    }
+
+    fn paint(&mut self, ctx: &mut PaintCtx, scene: &mut Scene) {
+        (self.draw)(scene, ctx.size());
+    }
```
PoignardAzur, line 95: So, to be clear, this is a static canvas, right? The draw function isn't expected to paint different things from frame to frame?

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,165 @@
...
+        Role::Canvas
+    }
+
+    fn accessibility(&mut self, _ctx: &mut AccessCtx, node: &mut Node) {
+        // TODO: is this correct?
+        if let Some(text) = &self.alt_text {
+            node.set_description(text.clone());
+        } else {
+            node.clear_description();
+        }
```
PoignardAzur, line 107: You can skip the `clear_description`. Node values are persisted between `accessibility` calls, you're rebuilding the node from scratch every time.

masonry/src/widgets/canvas.rs
richard-uk1: Happy to go with what you suggest - I'm not a domain expert.

masonry/src/widgets/canvas.rs
DJMcNab: I'm not certain that I agree; I can see arguments either way (since the canvas isn't filled by default, for example). But since if it did intercept pointer events, they would still bubble up to the parents, then either way is fine.

masonry/src/widgets/canvas.rs
DJMcNab: I'm not sure what the first clause of the second sentence is trying to say. But I agree, you don't need clear_description, because by default there is no description.

masonry/src/widgets/canvas.rs
richard-uk1: I think so - it could be used for things like icons. But what should be different if it were more dynamic?

masonry/src/widgets/canvas.rs
richard-uk1: This must be from a copy/paste - not sure how it snuck in.

masonry/src/widgets/canvas.rs
richard-uk1: I suggest setting it to `true` for now, then seeing how that works in practice.

richard-uk1:
Should be ready for review again (assuming CI is green). I will squash before merging, as I don't think there is anything to be gained by keeping the commits.

Philipp-M:
> I will squash before merging, as I don't think there is anything to be gained by keeping the commits.

Not necessary though, as this is the default anyways when merging (with the title/text in the PR top comment as commit description), so keeping the commits should give a better timeline of this PR shouldn't it?

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,92 @@
...
+/// use masonry::TextAlignment;
+/// use masonry::parley::fontique;
+///
+/// label("Text example.")
+///     .brush(palette::css::RED)
+///     .alignment(TextAlignment::Middle)
+///     .text_size(24.0)
+///     .weight(FontWeight::BOLD)
+///     .with_font(fontique::GenericFamily::Serif)
+/// ```
```
DJMcNab, line 28: I don't think these docs are right...

richard-uk1: Oops sorry I missed this!

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,197 @@
...
+        this.ctx.request_render();
+    }
+
+    pub fn set_alt_text(mut this: WidgetMut<'_, Self>, alt_text: String) {
+        this.widget.alt_text = Some(alt_text);
+        this.ctx.request_accessibility_update();
+    }
+
+    pub fn remove_alt_text(mut this: WidgetMut<'_, Self>) {
+        this.widget.alt_text = None;
```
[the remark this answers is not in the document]
PoignardAzur: You'd have to ask @mwcampbell for an authoritative answer, but from what I remember:

- Setting no alt-text means you "forgot" to add it, and some screen reader tools might add a default one for you.
- Setting an empty alt text means you consider that the item shouldn't be described by screen readers.

At least I think that's how ARIA treats it.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    /// Set the text that will be used to communicate the meaning of the canvas to
+    /// those using screen readers.
+    ///
+    /// Users are encouraged to set alt text for the canvas.
+    /// If possible, the alt-text should succinctly describe what the canvas represents.
+    ///
+    /// If the canvas is decorative or too hard to describe through text, users should set alt text to `""`.
```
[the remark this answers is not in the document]
DJMcNab: It's a good point. My thinking here does need to be focused on the user experience, and having a snarky message isn't that. But I still think there is an important distinction between the two cases, and using the same handling for both is doing a disservice to the end user. That is, the cases are:
1) Not in the accessibility tree/only for visual display purposes. In that case setting to the empty string is the right behaviour. I agree completely with that.
2) The content of this widget is potentially important to all users, but we've decided not to make it accessible because doing so would be "too hard". https://github.com/linebender/xilem/pull/875#discussion_r1995378124 seems to answer this; that if it's "too hard" to add alt text, then we should just not add it. That would then use the screen reader's default handling of "developer didn't bother to make this accessible".

This sentence in the docs conflates those two cases, I think. Having both be the same case has one of two consequences:
1) It just completely hides that important information is missing from the user; i.e. creates an unknown unknown; OR
2) It forces the user to doubt all items which are indicated as display-only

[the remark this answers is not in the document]
mwcampbell: This comment is entirely correct.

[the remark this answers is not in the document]
PoignardAzur: Which one?

DJMcNab (changes requested):
(Marking as request changes for the sake of GitHub's UI, to indicate that this is waiting on the author)

PR #879, opened 2025-02-28
Add methods to MasonryState to allow external renderer

cfagot:
Adds methods to MasonryState that allows external renderer enough access to control rendering of Masonry/Xilem.

This pull request does not provide a good example of how this would be done, but I have a fairly simple example here: https://github.com/cfagot/space_survival. This example shows how xilem can be rendered inside an application which has other wgpu render managers.

### DIFF
--- a/masonry/src/app/event_loop_runner.rs
+++ b/masonry/src/app/event_loop_runner.rs
@@ -774,4 +774,61 @@ impl MasonryState<'_> {
             self.render_cx.set_present_mode(surface, present_mode);
         }
     }
+
+    pub fn get_render_device_and_queue(&self) -> Option<(&wgpu::Device, &wgpu::Queue)> {
+        if let WindowState::Rendering { surface, .. } = &self.window {
+            let dev_id = surface.dev_id;
+            let device = &self.render_cx.devices[dev_id].device;
+            let queue = &self.render_cx.devices[dev_id].queue;
+            Some((device, queue))
+        } else {
+            None
+        }
+    }
+
+    pub fn get_adapter(&self) -> Option<&wgpu::Adapter> {
+        if let WindowState::Rendering { surface, .. } = &self.window {
+            let dev_id = surface.dev_id;
+            Some (self.render_cx.devices[dev_id].adapter())
+        } else {
+            None
+        }
+    }
+
+    /// Get next frame to present, handling resize if necessary.
+    pub fn get_next_frame(&mut self) -> Result<wgpu::SurfaceTexture, wgpu::SurfaceError> {
+        let WindowState::Rendering {
+            window, surface, ..
+        } = &mut self.window
+        else {
+            tracing::warn!("Tried to render whilst suspended or before window created");
+            return Err(wgpu::SurfaceError::Lost);
+        };
+        // https://github.com/rust-windowing/winit/issues/2308
+        #[cfg(target_os = "ios")]
+        let size = window.outer_size();
+        #[cfg(not(target_os = "ios"))]
+        let size = window.inner_size();
+        let width = size.width;
+        let height = size.height;
+
+        if surface.config.width != width || surface.config.height != height {
+            self.render_cx.resize_surface(surface, width, height);
+        }
+
+        surface.surface.get_current_texture()
+    }
+
+    /// Handle accesskit tree updates. This is called by external renderers at the conclusion of
+    /// rendering the masonry scene.
+    pub fn handle_tree_update(&mut self, tree_update: accesskit::TreeUpdate) {
+        let WindowState::Rendering {
+            accesskit_adapter, ..
+        } = &mut self.window
+        else {
+            debug_panic!("Suspended inside event");
+            return;
+        };
+        accesskit_adapter.update_if_active(|| tree_update);
+    }
 }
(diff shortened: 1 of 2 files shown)


cfagot:
> I don't know wgpu enough to review this, but I'm not seeing anything outrageous.

Yeah, I should add that the api I added isn't necessarily carefully thought out -- it is just what I needed to drive rendering externally. But I've been sitting on it for months and haven't needed to change anything so it is likely sufficient for others too.

masonry/src/app/event_loop_runner.rs
```diff
@@ -774,4 +774,65 @@ impl MasonryState<'_> {
...
+    pub fn handle_tree_update(&mut self, tree_update: accesskit::TreeUpdate) {
+        let WindowState::Rendering {
+            accesskit_adapter, ..
+        } = &mut self.window
+        else {
+            debug_panic!("Suspended inside event");
+            return;
+        };
+        accesskit_adapter.update_if_active(|| tree_update);
+    }
```
DJMcNab, line 837: I'm somewhat struggling to understand what these semantically mean...
You should never be using these methods, as you're using `EventLoopRunner`; you should instead have your own event loop, and manage the window and accesskit adapter yourself.

masonry/src/app/event_loop_runner.rs
```diff
@@ -774,4 +774,65 @@ impl MasonryState<'_> {
...
+
+    /// Access to the wgpu adapter.
+    pub fn get_adapter(&self) -> Option<&wgpu::Adapter> {
+        if let WindowState::Rendering { surface, .. } = &self.window {
+            let dev_id = surface.dev_id;
+            Some(self.render_cx.devices[dev_id].adapter())
+        } else {
+            None
+        }
+    }
```
DJMcNab, line 800: These seem reasonable to me, in the "intentionally not very carefully reasoned about" way.

masonry/src/app/mod.rs
```diff
@@ -11,6 +11,7 @@ mod tracing_backend;
 pub use app_driver::{AppDriver, DriverCtx};
 pub use event_loop_runner::{
     run, run_with, EventLoop, EventLoopBuilder, EventLoopProxy, MasonryState, MasonryUserEvent,
+    WindowState,
```
DJMcNab, line 14: - Marking this `pub` in #309 was probably a mistake
- Exporting it as an "we've not carefully thought about it" item is *fine*.
- Since you're exporting it, you should give it some docs to that effect, namely "internal implementation detail of the event loop runner, exposed to unblock use cases we haven't had the bandwidth to reason about"
- This is why we should enable `unreachable_pub` and `unnameable_types`

masonry/src/app/event_loop_runner.rs
cfagot: If you want to do any rendering outside of Masonry/Xilem you can't call RedrawRequested because it will simply draw masonry. Instead you need to reproduce much of that call integrated with the external renderer.

In my demo app, here is where I render:
https://github.com/cfagot/space_survival/blob/main/src/render_mgr.rs#L65

And here is where the Xilem rendering happens (called from above):
https://github.com/cfagot/space_survival/blob/main/src/xilem_render.rs#L52

The xilem renderer gets called as one of potentially many render managers (in this case one of two).

The api I added was what I needed to to integrate with Masonry.

> you should instead manage the window and accesskit adapter yourself

MasonryState owns everything. Masonry isn't set up for running on a borrowed window, adapter, or access kit.

masonry/src/app/mod.rs
cfagot: Ok, I'll add a doc to WindowState.

masonry/src/app/event_loop_runner.rs
DJMcNab: Yes, it will draw Masonry - because `MasonryState` is what you use if you want a window which is managed by Masonry. If you're wanting to manage a Masonry window yourself, the way to do that is to create your own `RenderRoot`, and wire up the events you care about wiring up yourself.

The `redraw` method on `RenderRoot` is public, and it's what you would use for this use case.

DJMcNab: But you're explicitly not making a Masonry app? If you're not using any of Masonry's handling for events.
`MasonryState` can be used if you want to make Masonry *one* of your windows in an existing app, but it is very much not designed to be a general winit+wgpu context.
`MasonryState` assumes that it can render into your window. At the moment, this only occurs during `RedrawRequested`, but I'm not willing to make that guarantee any further, that if we need to fix a bug by redrawing at a different time we won't do so (concretely, this also includes frame pacing, which is planned, although it's still a long ways off). And that means that anything which uses this method would necessarily be broken at that point.

I'd recommend just copying similar logic to `MasonryState` into your app, for this use case. If there's something it uses internally which needs to be public for an embedding use case, that's an issue which we can absolutely address by making the things you need public.

This whole conversation has made me think that exporting `MasonryState` might have been a mistake, if it serves as an attractive nuisance.

cfagot: Ok, that makes sense. For now this is still the best way for me to manage the context, but since the framework is in flux and can change a lot I understand why you might not wan to integrate this. Keeping the fork up to date is only a minor inconvenience to me since it's only a few lines at the end of a file.

I'll go ahead and close the PR.

PR #883, opened 2025-03-08
Update to parley 0.3

mwcampbell:
This update requires us to handle one breaking API change in parley.


mwcampbell:
Can anyone tell why cargo clippy keeps failing on macOS? I don't see an actual error message, just an error exit status.

waywardmonkeys:
You have to fix the tests that call the same API

burrbull:
> I don't see an actual error message

There is expander.

masonry/examples/custom_widget.rs
```diff
@@ -138,7 +138,13 @@ impl Widget for CustomWidget {
 
         let mut text_layout = text_layout_builder.build(&self.0);
         text_layout.break_all_lines(None);
-        text_layout.align(None, Alignment::Start, false);
+        text_layout.align(
+            None,
+            Alignment::Start,
+            AlignmentOptions {
```
DJMcNab, line 144: I think we should use `AlignmentOptions::default()` here. The reason that the other parameters aren't in the options is because that `Default` impl exists (https://github.com/linebender/parley/pull/283#issuecomment-2665296623), and we are using exactly the default value.

DJMcNab (approved):
One small nit, but otherwise LGTM.

Thanks!

masonry/examples/custom_widget.rs
mwcampbell: Should I use `AlignmentOptions::default()` in the label widget too, or just in this example? And do you mean that I should use `AlignmentOptions::default()` *instead* of explicitly specifying the one option, or use `..AlignmentOptions::default()` to allow for future extensibility?

DJMcNab: Apologies for the abiguity. The first for both. That is, use `AlignmentOptions::default()` as the full parameter in both cases.

PR #885, opened 2025-03-14
Free properties as well

DJMcNab:
I found in #882 that properties (as introduced in #873) are never freed.


PoignardAzur (approved):
LGTM.

PR #886, opened 2025-03-14
Change how some widgets are rendered

PoignardAzur:
Make the border and background disjoint.
Correctly interpret the border width.


PoignardAzur:
This renders widgets with a border more consistently to prepare for the properties refactor.

Note that the changes are subtle enough that I had to manually remove the existing snapshots to re-render them.

(The progress bar screenshots in particular might be pixel-identical, but I still updated them for consistency.)

DJMcNab:
As a temporary process, would you mind running these updated screenshots through [Oxipng](https://crates.io/crates/oxipng) (There's a CLI)? When we move to Kompari, this will be done by default, but we don't have it built-in yet.

I'd like to see how well it does for Masonry screenshots

DJMcNab (commented):
I've not really carefully reviewed, but I'm happy for this to land once the Oxipng thing is done (or determined to not work for some reason)

PoignardAzur:
> As a temporary process, would you mind running these updated screenshots through [Oxipng](https://crates.io/crates/oxipng) (There's a CLI)?

Size went from 200K to 116K.

DJMcNab (approved):

PR #887, opened 2025-03-15
Add Xilem view for the Split widget

viktorstrate:
### Highlights

- Make the Masonry Split widget generic over its children (fixes a todo)
- Adds a new Xilem view for the widget
- Updated the `http_cats` example to use it for the main divider.

---

I'm unsure how exactly the `message` function works on a Xilem view as it was hard to find documentation on that, so I've tried my best to implement it, by looking at other views. Please double check that I did it right.

I also changed some defaults on the split widget. First, I changed `draggable` to default to true, since I believe this is the intended behaviour for the view, similar to how a button is not disabled by default.

I changed the constructors of the widget to not include the `split_axis` but rather let it be optional with vertical (row) being the default axis. I feel this made it more ergonomic to integrate with Xilem.

### DIFF
--- a/xilem/src/view/split.rs
+++ b/xilem/src/view/split.rs
@@ -1,3 +1,6 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
 use std::marker::PhantomData;
 
 use masonry::widgets::{self, Axis};
@@ -5,6 +8,34 @@ use xilem_core::{DynMessage, View, ViewId, ViewMarker, ViewPathTracker};
 
 use crate::{Pod, ViewCtx, WidgetView};
 
+/// A container containing two other widgets, splitting the area either horizontally or vertically.
+///
+/// The split view will by default be draggable,
+/// which means that the relative size of each child view can be changed by dragging the mouse.
+///
+/// # Examples
+/// To create a split view, provide it with two child views.
+///
+/// ```ignore
+/// use xilem::view::{split, label};
+///
+/// split(
+///     label("Left view"),
+///     label("Right view")
+/// )
+/// ```
+///
+/// The split axis and split point can be changed.
+/// For the full list of modifiers see the [`Split`] struct.
+///
+/// ```ignore
+/// use xilem::view::{split, label};
+///
+/// split(label("Left view"), label("Right view"))
+///     .split_axis(Axis::Horizontal)
+///     .split_point(0.25)
+/// ```
+///
 pub fn split<State, Action, ChildA, ChildB>(
     child1: ChildA,
     child2: ChildB,
@@ -14,13 +45,13 @@ where
     ChildB: WidgetView<State, Action>,
 {
     Split {
-        split_axis: Axis::Vertical,
+        split_axis: Axis::Horizontal,
         split_point: 0.5,
         min_size: (0.0, 0.0),
         bar_size: 6.0,
         min_bar_area: 6.0,
         solid_bar: false,
-        draggable: false,
+        draggable: true,
         child1,
         child2,
         phantom: PhantomData,
@@ -45,7 +76,12 @@ pub struct Split<ChildA, ChildB, State, Action = ()> {
 }
 
 impl<ChildA, ChildB, State, Action> Split<ChildA, ChildB, State, Action> {
-    /// Set the split axis
+    /// Set the split axis.
+    ///
+    /// Horizontal split axis means that the children are left and right.
+    /// Vertical split axis means that the children are up and down.
+    ///
+    /// The default split point is horizontal.
     pub fn split_axis(mut self, axis: Axis) -> Self {
         self.split_axis = axis;
         self
@@ -141,11 +177,16 @@ where
         let (child1, child1_state) = ctx.with_id(CHILD1_VIEW_ID, |ctx| self.child1.build(ctx));
         let (child2, child2_state) = ctx.with_id(CHILD2_VIEW_ID, |ctx| self.child2.build(ctx));
 
-        let widget_pod = ctx.new_pod(widgets::Split::new_pod(
-            self.split_axis,
-            child1.into_widget_pod(),
-            child2.into_widget_pod(),
-        ));
+        let widget_pod = ctx.new_pod(
+            widgets::Split::new_pod(child1.into_widget_pod(), child2.into_widget_pod())
+                .split_axis(self.split_axis)
+                .split_point(self.split_point)
+                .min_size(self.min_size.0, self.min_size.1)
+                .bar_size(self.bar_size)
+                .min_bar_area(self.min_bar_area)
+                .draggable(self.draggable)
+                .solid_bar(self.solid_bar),
+        );
 
         (widget_pod, (child1_state, child2_state))
     }
--- a/xilem/src/view/split.rs
+++ b/xilem/src/view/split.rs
@@ -4,7 +4,7 @@
 use std::marker::PhantomData;
 
 use masonry::widgets::{self, Axis};
-use xilem_core::{DynMessage, View, ViewId, ViewMarker, ViewPathTracker};
+use xilem_core::{DynMessage, MessageResult, View, ViewId, ViewMarker, ViewPathTracker};
 
 use crate::{Pod, ViewCtx, WidgetView};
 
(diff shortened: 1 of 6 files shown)


xilem/src/view/split.rs
```diff
@@ -0,0 +1,276 @@
...
+            widgets::Split::set_draggable(&mut element, self.draggable);
+        }
+
+        if prev.solid_bar != self.solid_bar {
+            widgets::Split::set_bar_solid(&mut element, self.solid_bar);
+        }
+
+        let child1_element = widgets::Split::child1_mut(&mut element);
+        self.child1
+            .rebuild(&prev.child1, &mut view_state.0, ctx, child1_element);
```
Philipp-M, line 231:
```suggestion
        ctx.with_id(CHILD1_VIEW_ID, |ctx| {
            self.child1
                .rebuild(&prev.child1, &mut view_state.0, ctx, child1_element)
        });
```

Same for the other child.

As explained below as well, tree-structure might change in this function (within children), so we need to track this via the id path, so that messages can be routed correctly in `View::message`.

xilem/src/view/split.rs
```diff
@@ -0,0 +1,276 @@
...
+        view_state: &mut Self::ViewState,
+        id_path: &[xilem_core::ViewId],
+        message: DynMessage,
+        app_state: &mut State,
+    ) -> xilem_core::MessageResult<Action, DynMessage> {
+        let (view_id, rest) = id_path
+            .split_first()
+            .expect("message should contain id_path to child in Split view");
+
+        if view_id == &CHILD1_VIEW_ID {
```
Philipp-M, line 262: > I'm unsure how exactly the message function works on a Xilem view as it was hard to find documentation on that, so I've tried my best to implement it, by looking at other views. Please double check that I did it right.

I'll try to explain:

Xilem uses generational ids, similar as bevy (actually I think it was inspired by bevy).

So when it's possible that a child view can change, it gets assigned a new id (generation) when that happens.
I think the `ViewSequence` impl of `Option<impl ViewSequence>` is probably a good somewhat simple example to get the idea.

It uses these ids to route messages to the right views, as the views are ephemeral representations, and we need to check IRC in `View/Sequence::rebuild` whether the structure has changed, and if this is the case assign new ids, so that potential async messages don't end up in the wrong views.

When these children views don't change their structure they can have fixed ids (see e.g. `[impl ViewSequence; N]` for an example, in the same file as `Option` in `xilem_core` - `sequence.rs`).

The *actual* "id" of a view is resulting of all the elements of an `id_path` which should be a unique path to a view.
It's not really an id, as some views don't take messages, and thus don't need an explicit element/id within this id_path, it's more routing information for views that take events. 

So in your case, as this view doesn't change the structure of the children views, it's fine to assign fixed ids for each of these child views (something distinctive for debugging is good).

*But* they need to be different, otherwise all messages will be routed to the first `child1`.

While it's not entirely up-to-date anymore (generational ids were added later, to avoid having to assign ids for every view, when it's not necessary): [this](https://raphlinus.github.io/rust/gui/2022/05/07/ui-architecture.html#identity-and-id-paths) explains the origin of this mechanism.

viktorstrate:
@Philipp-M Thank you for the detailed explaination. I have updated the rebuild function to add the view ids.

xilem/src/view/split.rs
```diff
@@ -0,0 +1,280 @@
...
+    ///
+    /// If this is `false` (the default), the bar will be drawn as two parallel lines.
+    pub fn solid_bar(mut self, solid: bool) -> Self {
+        self.solid_bar = solid;
+        self
+    }
+}
+
+const CHILD1_VIEW_ID: ViewId = ViewId::new(0);
+const CHILD2_VIEW_ID: ViewId = ViewId::new(0);
```
Philipp-M, line 161: > *But* they need to be different, otherwise all messages will be routed to the first child1.

I think that went probably under the radar.
So to be more explicit:
`CHILD1_VIEW_ID` needs to be something different than `CHILD2_VIEW_ID`, so e.g.

```suggestion
const CHILD1_VIEW_ID: ViewId = ViewId::new(5436781);
const CHILD2_VIEW_ID: ViewId = ViewId::new(5436782);
```

(btw. just my preference, and more of a nit, you could use `match` in `View::message`)

viktorstrate: Right, that was an oversight, I've pushed an update. Is there any particular reason to using pseudo-random numbers rather than simply 0 and 1?

DJMcNab: The reason you'd want to use a random key would be to be able to identify after-the-fact which `View`s were used in the tree just by inspecting the view id path. However, we've not done that for any other Views.

viktorstrate: Alright, should I change it to be 0 and 1 then, or is it fine as is?

xilem/src/view/split.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub fn split_axis(mut self, axis: Axis) -> Self {
+        self.split_axis = axis;
+        self
+    }
+
+    /// Set the split point as a fraction of the split axis.
+    ///
+    /// The value must be between `0.0` and `1.0`, inclusive.
+    /// The default split point is `0.5`.
+    pub fn split_point(mut self, split_point: f64) -> Self {
```
DJMcNab, line 94: Maybe make this `#[track_caller]`

xilem/src/view/split.rs
```diff
@@ -0,0 +1,274 @@
...
+        assert!(
+            (0.0..=1.0).contains(&split_point),
+            "split_point must be in the range [0.0-1.0]!"
```
DJMcNab, line 97:
```suggestion
            "split_point must be in the range [0.0-1.0], got {split_point}."
```

Is that a usual range syntax? I'd normally say `0.0..=1.0` or `[0.0, 1.0]`

Edit: Oh I see, this is retained from the previous... Hmm

xilem/src/view/split.rs
```diff
@@ -0,0 +1,274 @@
...
+        );
+        self.split_point = split_point;
+        self
+    }
+
+    /// Set the minimum size for both sides of the split axis.
```
DJMcNab, line 103: These are both in logical pixels, right?

xilem/src/view/split.rs
```diff
@@ -0,0 +1,274 @@
...
+        match id_path.split_first() {
+            Some((&CHILD1_VIEW_ID, rest)) => {
+                self.child1
+                    .message(&mut view_state.0, rest, message, app_state)
+            }
+            Some((&CHILD2_VIEW_ID, rest)) => {
+                self.child2
+                    .message(&mut view_state.1, rest, message, app_state)
+            }
+            _ => unreachable!("invalid id_path for Split view message: {:?}", id_path),
```
DJMcNab, line 271: We normally warn/error in these cases, e.g. https://github.com/linebender/xilem/blob/aaa84034efbdf7621ab8197f74e32ac7fb5c4bbb/xilem/src/view/spinner.rs#L89-L91, rather than panic.

We don't think it's possible to reach this state, if all other views are written correctly. But falling over isn't something we should do lightly.

DJMcNab (approved):
A few minor details. Nothing too significant

xilem/src/view/split.rs
viktorstrate: I've added `#[track_caller]` to all methods in the view that have an assert.

xilem/src/view/split.rs
viktorstrate: I changed it for both the View and the Widget 👍

xilem/src/view/split.rs
viktorstrate: That makes sense, I've changed it.

Do you think we should do the same for the asserts when building the view?
Maybe change all asserts to `debug_assert!` and then print an error and bound the values within their valid range to handle it more gracefully in release mode?

xilem/src/view/split.rs
DJMcNab: Yeah, 0 and 1 is probably best.

DJMcNab:
Running this in `http_cats` runs into the perennial frame pacing issue, that mouse motion isn't coalesced, so each of the mouse events in a frame run a full layout pass, which is reasonably quick at 5ms, but is still slower than the events come in, which is very bad.

Luckily in release mode, it's fast enough. This doesn't block this PR.

DJMcNab:
It's not really a bug in `Split`. Basically, we currently forward every mouse event to the active widget, which they then dutifully handle.

We can fix it by getting frame pacing in, which is hard. There's some evidence that suggests an "easier" version of frame pacing, which runs a bus at the vsync rate would be sufficient here, but I've not looked into it in great detail. See https://github.com/aevyrie/bevy_framepace for an example (although we'd probably want to have a per-widget opt-out of coalesced events)

PR #890, opened 2025-03-17
Add note about `split-debuginfo="unpacked"` in README

PoignardAzur:


README.md
```diff
@@ -155,6 +155,17 @@ cargo update -p package_name --precise 0.1.1
 
 </details>
 
+## Footprint
+
+The Xilem repository includes a lot of projects and examples, most of them pulling a lot of dependencies.
+
+If you contribute to Xilem on Linux or MacOS, we recommend using `split-debuginfo` in your `config.toml` file to reduce the size of the `target/` folder:
```
DJMcNab, line 162:
```suggestion
If you contribute to Xilem on Linux or MacOS, we recommend using [`split-debuginfo`](https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo) in your
[`.cargo/config.toml`](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure) file to reduce the size of the `target/` folder:
```

It might also be worth mentioning that this is already the default on MacOS.

README.md
```diff
@@ -155,6 +155,17 @@ cargo update -p package_name --precise 0.1.1
...
+
+```toml
+[profile.dev]
+split-debuginfo="unpacked" # Leave debuginfo in separate files for smaller footprint.
```
DJMcNab, line 166:
```suggestion
# One debuginfo file per dependency, to reduce file size of tests/examples.
# Note that this setting value is not supported on Windows.
# See https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
split-debuginfo="unpacked"
```

README.md
```diff
@@ -155,6 +155,17 @@ cargo update -p package_name --precise 0.1.1
 
 </details>
 
+## Footprint
```
DJMcNab, line 158: I'm not familiar with `Footprint` as a common term for this (and so don't find it compelling as a section title, as it doesn't strongly indicate what will be in the section).

Maybe "Recommended Cargo Config"?

DJMcNab (approved):
I'd like to see a handful of tweaks here, but otherwise LGTM

README.md
PoignardAzur: The default on MacOS is "packed", which is not quite as good IMO.

DJMcNab: Oh. Fair enough - I only skimmed this!

PR #891, opened 2025-03-17
Update Cargo profiles

PoignardAzur:
Some minor improvements to config values when building dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,9 +106,13 @@ accesskit = "0.17.0"
 accesskit_winit = "0.23.0"
 time = "0.3.36"
 
+[profile.dev.package."*"]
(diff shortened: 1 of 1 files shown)


DJMcNab:
The change for CI needs numbers; that is, how much longer does a clean build take on each platform, and how much smaller is the cache.

I lean slightly against the dev profile change, for a few reasons:
1) it has the possibility to make debugging more confusing, because you'd be stepping through optimised code
2) it makes our clean build dev times worse (how much?), when they already leave a lot to be desired. That is, e.g. for users checking out Xilem's examples, it takes longer to get something running.
3) it hides from us the realities of our debug mode performance. End-users are unlikely to set this themselves.

It would be easier to evaluate this with a compilation time vs performance table for both options (the current state and the change). It might also make sense to include opt-level=1.

PR #892, opened 2025-03-17
Add Properties to Button

PoignardAzur:
Depends on #904.

Add BackgroundColor, BorderColor, BorderWidth, CornerRadius and Padding properties.
Modularize button painting code.
Paint border after background to get slightly better border appearance.
Update screenshots.
Add screenshot test using properties.


RagibHasin (approved):
I've read through the code, and it looks good to me.

The one thing I think is that the `CornerRadius` property should wrap a `kurbo::RoundedRectRadii` for complete control, should it not? Anyway, if that is desirable, it can be done in a followup.

PoignardAzur:
Yeah, I don't remember my exact reasoning for not using RoundedRectRadii, but you're right that I should probably use it. We'll leave it for a future PR.

PR #895, opened 2025-03-18
Pair image with copyright notice in http_cats example.

platlas:
While on slow internet connection I noticed that copyright notice is conflicting with the spinner: 
http_cat-screenshot

So I paired the image with the copyright notice to display it only when image is loaded. I hope this is not complicating the example code.


DJMcNab (approved):
Thanks! I moved where the new variable is defined slightly

PR #896, opened 2025-03-18
Refactor example tests to run more efficiently

PoignardAzur:


masonry/examples/calc_masonry.rs
```diff
@@ -423,22 +424,3 @@ fn main() {
...
-
-    #[test]
-    fn screenshot_test() {
-        let mut harness = TestHarness::create(build_calc());
-        assert_debug_snapshot!(harness.root_widget());
-        assert_render_snapshot!(harness, "initial_screenshot");
-
-        // TODO - Test clicking buttons
-    }
-}
```
DJMcNab, line 444: It's not obvious to me why you've moved these tests from their individual example files into all one module; a `#[cfg(test)] mod tests` should still work in this new pattern.

masonry/examples/tests.rs
DJMcNab, line 1: Can we make this an integration test rather than an example? For prior art, see https://github.com/bevyengine/bevy/pull/3597#discussion_r859047263 (although note that in this case, we want it to actually be a "proper" integration test, not just an example.

DJMcNab (commented):
Getting this to be quicker is excellent; I think there are a few ways to make this more idiomatic, which I believe should work.

masonry/examples/calc_masonry.rs
PoignardAzur: ... Huh. That's true.

masonry/examples/tests.rs
PoignardAzur: I'll see if I can make it work.

masonry/examples/calc_masonry.rs
PoignardAzur: I gave it a try. One problem is that the tests end up running *and failing* when you do `cargo test --examples`, because in that case cargo runs unit tests for all examples even if they're listed as not having tests, but because their module structure is different the snapshot files are different as well.

As far as I know `cargo test --examples` isn't run in CI, and `cargo test` works fine, but it's still a footgun I'm uncomfortable leaving around.

(You could address that with some kind of cfg trick, but at that point I don't think it's worth the effort.)

DJMcNab: Hmm, so this is https://github.com/rust-lang/cargo/issues/13668...

I've never liked the full module path in the screenshot name anyway (it makes it really hard to tell what the actual *name* of the screenshot is), but now isn't the time to fix that.

I really strongly dislike the "tests for all examples are in a single file". In fact, since you think that's required for this PR, I'm now against this change entirely. It misses the point of the tests in the examples, which are to demonstrate that you can (and should be) testing your UI code.

PoignardAzur: For me it was to check that the examples didn't get broken by a change, something that kept happening to me before.

If this is a deal-breaker, I can always go with a cfg-based solution.

PoignardAzur: Actually, turns out making a cfg-based solution is harder than I thought.

At this point I'm seriously considering changing the naming convention of screenshot files. Like you say, it's really hard to read right now, and removing the "module" part of it would make this PR a lot easier.

DJMcNab (changes requested):
Marking this as requested reviews to give my understanding of the current state; that moving all of the tests into a single file is not viable, but fixing that requires changing the naming scheme for the screenshots

masonry/examples/calc_masonry.rs
DJMcNab: A solution which sets `harness = false` for the examples, and bails out from  `main` immediately if `cfg(test)` is true should work.

PoignardAzur:
Superseded by #945.

PR #897, opened 2025-03-18
Change sizes of various screenshots

PoignardAzur:


DJMcNab:
Have you run these screenshots through Oxipng again? I think something like `oxipng -o 6 -s $(git diff --name-only main HEAD -- '*.png')` (untested) should work for that. I realise that doing it manually is a pain, but it is a significant win, and we'll get it done automatically when we move to Kompari.

PoignardAzur:
Done.

DJMcNab (approved):
I've not carefully reviewed, but this seems like a big improvemenet

PR #899, opened 2025-03-19
Document more items

PoignardAzur:
Make missing_docs annotations more granular.

### DIFF
--- a/masonry/src/theme.rs
+++ b/masonry/src/theme.rs
@@ -1,9 +1,9 @@
 // Copyright 2019 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//! Theme keys and initial values.
+//! Default values used by various widgets in their paint methods.
 
-#![allow(missing_docs)]
+#![allow(missing_docs, reason = "Names are self-explanatory.")]
 
 use crate::kurbo::Insets;
 use crate::peniko::Color;
--- a/masonry/src/widgets/button.rs
+++ b/masonry/src/widgets/button.rs
@@ -79,7 +79,7 @@ impl Button {
         Label::set_text(&mut Self::label_mut(this), new_text);
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Get a mutable reference to the label.
     pub fn label_mut<'t>(this: &'t mut WidgetMut<'_, Self>) -> WidgetMut<'t, Label> {
         this.ctx.get_mut(&mut this.widget.label)
     }
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -299,7 +299,7 @@ impl Flex {
         self.children.len()
     }
 
-    /// Returns true if this flex container has no children (widgets or spacers).
+    /// Returns `true` if this flex container has no children (widgets or spacers).
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
@@ -457,16 +457,20 @@ impl Flex {
         this.ctx.request_layout();
     }
 
-    /// Add a non-flex child widget.
+    /// Insert a non-flex child widget at the given index.
     ///
-    /// See also [`with_child`].
+    /// # Panics
     ///
-    /// [`with_child`]: Flex::with_child
+    /// Panics if the index is larger than the number of children.
     pub fn insert_child(this: &mut WidgetMut<'_, Self>, idx: usize, child: impl Widget) {
         Self::insert_child_pod(this, idx, WidgetPod::new(child).erased());
     }
 
-    /// Add a non-flex child widget.
+    /// Insert a non-flex child widget wrapped in a [`WidgetPod`] at the given index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is larger than the number of children.
     pub fn insert_child_pod(
         this: &mut WidgetMut<'_, Self>,
         idx: usize,
@@ -480,7 +484,11 @@ impl Flex {
         this.ctx.children_changed();
     }
 
-    /// Add a flex child widget.
+    /// Insert a flex child widget at the given index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is larger than the number of children.
     pub fn insert_flex_child(
         this: &mut WidgetMut<'_, Self>,
         idx: usize,
@@ -490,7 +498,11 @@ impl Flex {
         Self::insert_flex_child_pod(this, idx, WidgetPod::new(child).erased(), params);
     }
 
-    /// Add a flex child widget.
+    /// Insert a flex child widget wrapped in a [`WidgetPod`] at the given index.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is larger than the number of children.
     pub fn insert_flex_child_pod(
         this: &mut WidgetMut<'_, Self>,
         idx: usize,
@@ -503,22 +515,30 @@ impl Flex {
     }
 
     // TODO - remove
-    /// Add a spacer widget with a standard size.
+    /// Insert a spacer widget with a standard size at the given index.
     ///
     /// The actual value of this spacer depends on whether this container is
     /// a row or column, as well as theme settings.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is larger than the number of children.
     pub fn insert_default_spacer(this: &mut WidgetMut<'_, Self>, idx: usize) {
         let key = axis_default_spacer(this.widget.direction);
         Self::insert_spacer(this, idx, key);
         this.ctx.request_layout();
     }
 
-    /// Add an empty spacer widget with the given size.
+    /// Insert an empty spacer widget with the given size at the given index.
     ///
     /// If you are laying out standard controls in this container, you should
     /// generally prefer to use [`add_default_spacer`].
     ///
     /// [`add_default_spacer`]: Flex::add_default_spacer
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is larger than the number of children.
     pub fn insert_spacer(this: &mut WidgetMut<'_, Self>, idx: usize, mut len: f64) {
         if len < 0.0 {
             tracing::warn!("add_spacer called with negative length: {}", len);
@@ -531,6 +551,10 @@ impl Flex {
     }
 
     /// Add an empty spacer widget with a specific `flex` factor.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is larger than the number of children.
     pub fn insert_flex_spacer(this: &mut WidgetMut<'_, Self>, idx: usize, flex: f64) {
         let flex = if flex >= 0.0 {
             flex
@@ -549,7 +573,7 @@ impl Flex {
     ///
     /// # Panics
     ///
-    /// Panics if the index is out of bounds.
+    /// Panics if the index is larger than the number of children.
     pub fn remove_child(this: &mut WidgetMut<'_, Self>, idx: usize) {
         let child = this.widget.children.remove(idx);
         if let Child::Fixed { widget, .. } | Child::Flex { widget, .. } = child {
@@ -560,11 +584,11 @@ impl Flex {
 
     /// Returns a mutable reference to the child widget at `idx`.
     ///
-    /// Returns None if the child at `idx` is a spacer.
+    /// Returns `None` if the child at `idx` is a spacer.
     ///
     /// # Panics
     ///
-    /// Panics if the index is out of bounds.
+    /// Panics if the index is larger than the number of children.
     pub fn child_mut<'t>(
         this: &'t mut WidgetMut<'_, Self>,
         idx: usize,
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -294,12 +294,12 @@ impl Flex {
         self
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Returns the number of children (widgets and spacers) this flex container has.
     pub fn len(&self) -> usize {
         self.children.len()
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Returns true if this flex container has no children (widgets or spacers).
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
@@ -390,7 +390,11 @@ impl Flex {
         this.ctx.children_changed();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Add a non-flex child widget with a pre-assigned id.
+    ///
+    /// See also [`with_child_id`].
+    ///
+    /// [`with_child_id`]: Flex::with_child_id
     pub fn add_child_id(this: &mut WidgetMut<'_, Self>, child: impl Widget, id: WidgetId) {
         let child = Child::Fixed {
             widget: WidgetPod::new_with_id(child, id).erased(),
@@ -476,7 +480,7 @@ impl Flex {
         this.ctx.children_changed();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Add a flex child widget.
     pub fn insert_flex_child(
         this: &mut WidgetMut<'_, Self>,
         idx: usize,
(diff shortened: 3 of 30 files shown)


PoignardAzur:
This PR documents all the items that I *can* document. Still missing are:

- Items that I think would be too noisy to document (eg constants).
- xilem_web.
- Code in xilem that I don't know well enough to document.
- Code in Masonry that I expect to be refactored before we ever document them.

masonry/src/widgets/flex.rs
```diff
@@ -548,7 +558,13 @@ impl Flex {
         this.ctx.request_layout();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Returns a mutable reference to the child widget at `idx`.
+    ///
+    /// Returns None if the child at `idx` is a spacer.
```
waywardmonkeys, line 563: The `None` should have backticks around it.

masonry/src/widgets/flex.rs
```diff
@@ -294,12 +294,12 @@ impl Flex {
...
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Returns the number of children (widgets and spacers) this flex container has.
     pub fn len(&self) -> usize {
         self.children.len()
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Returns true if this flex container has no children (widgets or spacers).
```
waywardmonkeys, line 302: We used to be consistent in having `true` wrapped in backticks. We should get back to that consistency.

masonry/src/widgets/grid.rs
```diff
@@ -113,8 +108,13 @@ fn new_grid_child(params: GridParams, widget: WidgetPod<dyn Widget>) -> Child {
 
 // --- MARK: IMPL GRIDPARAMS ---
 impl GridParams {
-    #[expect(missing_docs, reason = "TODO")]
+    /// Get grid parameters with the given values.
```
waywardmonkeys, line 111: This creates rather than "gets" ...

masonry/src/widgets/grid.rs
```diff
@@ -181,7 +183,13 @@ impl Grid {
         this.ctx.request_layout();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Insert a child widget at the giving index.
```
waywardmonkeys, line 186: "at the given `idx`" (2 changes)

masonry/src/widgets/grid.rs
```diff
@@ -191,7 +199,13 @@ impl Grid {
         Self::insert_grid_child_pod(this, idx, WidgetPod::new(child).erased(), params);
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Insert a child widget already wrapped in a [`WidgetPod`] at the giving index.
```
waywardmonkeys, line 202: Same as above.

masonry/src/widgets/grid.rs
```diff
@@ -204,31 +218,35 @@ impl Grid {
...
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Set the spacing between grid items.
     pub fn set_spacing(this: &mut WidgetMut<'_, Self>, spacing: f64) {
         this.widget.grid_spacing = spacing;
         this.ctx.request_layout();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Set the number of columns of the grid.
```
waywardmonkeys, line 227: This makes it seem like the names of these functions are not so great given the wide difference between what the docs say and what a reasonable interpretation of the name of the function might have been.

masonry/src/widgets/grid.rs
```diff
@@ -204,31 +218,35 @@ impl Grid {
...
+    /// # Panics
+    ///
+    /// Panics if `idx` is out of bounds.
     pub fn child_mut<'t>(
         this: &'t mut WidgetMut<'_, Self>,
         idx: usize,
-    ) -> Option<WidgetMut<'t, dyn Widget>> {
-        let child = this.widget.children[idx].widget_mut()?;
-        Some(this.ctx.get_mut(child))
+    ) -> WidgetMut<'t, dyn Widget> {
```
waywardmonkeys, line 247: Confusing to be changing API in the middle of documentation changes.

masonry/src/widgets/grid.rs
```diff
@@ -246,7 +264,11 @@ impl Grid {
         this.ctx.request_layout();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Removes a child widget at the specified index.
```
waywardmonkeys, line 267: Sometimes we use "specified" and sometimes "given" ... should decide on one (everywhere).

masonry/src/widgets/grid.rs
PoignardAzur: I stand by it. The behavior doesn't change, we were always returning `Some`, the only consumer was unwrapping it, etc.

I could technically split it out into a second PR and add a TODO in this PR or something, but I don't see a benefit.

masonry/src/theme.rs
```diff
@@ -1,9 +1,9 @@
 // Copyright 2019 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//! Theme keys and initial values.
+//! Default values used by various widgets in their paint methods.
 
-#![allow(missing_docs)]
+#![allow(missing_docs, reason = "Names are self-explanatory.")]
```
DJMcNab, line 6: There are still some useful docs possible here; see https://docs.rs/color/0.2.3/color/palette/css/index.html.

Not to block on that for this change, but I'd maybe say:
```suggestion
// TODO: Docs as in https://docs.rs/color/0.2.3/color/palette/css/index.html.
#![allow(missing_docs, reason = "Names are reasonably self-explanatory.")]
```

masonry/src/widgets/button.rs
```diff
@@ -79,7 +79,7 @@ impl Button {
         Label::set_text(&mut Self::label_mut(this), new_text);
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Get a mutable reference to the label.
```
DJMcNab, line 82: This is *true*, but you also get the `MutateCtx` along for the ride. I'm not sure what the right way to say this succinctly is.

masonry/src/widgets/grid.rs
```diff
@@ -52,25 +56,23 @@ impl Grid {
...
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Builder-style method for setting the spacing between grid items.
     pub fn with_spacing(mut self, spacing: f64) -> Self {
         self.grid_spacing = spacing;
         self
     }
 
-    /// Builder-style variant of [`Grid::add_child`].
```
DJMcNab, line 61: I do think it's useful to link to the runtime equivalent; see how I've documented things on `TextArea` for example.
That doesn't need to be in this PR though; it's a follow-up cleanup pass. And I agree that the old wording here was actively misleading, so changing it is good.

masonry/src/widgets/grid.rs
```diff
@@ -113,8 +108,13 @@ fn new_grid_child(params: GridParams, widget: WidgetPod<dyn Widget>) -> Child {
 
 // --- MARK: IMPL GRIDPARAMS ---
 impl GridParams {
-    #[expect(missing_docs, reason = "TODO")]
+    /// Get grid parameters with the given values.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the width or height is less than or equal to zero.
```
DJMcNab, line 115: This also panics if `x` or `y` are negative, right? Is there some reason this shouldn't also be called out that I'm missing.

masonry/src/widgets/grid.rs
```diff
@@ -150,27 +150,29 @@ impl GridParams {
 impl Grid {
     /// Add a child widget.
     ///
-    /// See also [`with_child`].
-    ///
-    /// [`with_child`]: Grid::with_child
+    /// See also [`with_child`](Grid::with_child).
```
DJMcNab, line 153: Similar thing here, that "See also" is a bit too oblique here. However, that would be part of the same aforementioned consistency pass.

masonry/src/widgets/grid.rs
```diff
@@ -181,7 +183,13 @@ impl Grid {
         this.ctx.request_layout();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Insert a child widget at the giving index.
+    ///
+    /// This lets you control the order in which the children are drawn.
```
DJMcNab, line 188:
```suggestion
    /// This lets you control the order in which the children are drawn. Children are drawn in
    /// order of their index (i.e. children with higher indices are above any children with lower index).
```

masonry/src/widgets/grid.rs
```diff
@@ -191,7 +199,13 @@ impl Grid {
         Self::insert_grid_child_pod(this, idx, WidgetPod::new(child).erased(), params);
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Insert a child widget already wrapped in a [`WidgetPod`] at the giving index.
+    ///
+    /// This lets you control the order in which the children are drawn.
```
DJMcNab, line 204:
```suggestion
    /// This lets you control the order in which the children are drawn. Children are drawn in
    /// order of their index (i.e. children with higher indices are above any children with lower index).
```

masonry/src/widgets/grid.rs
DJMcNab: I think it not being called out proactively is concerning, as this is quite a small part of a big PR.
That is, I agree that changing it here is fine; doing the documentation revealed the absurdity of the `widget_mut` method, and following it through led to this change naturally.

Fwiw, one possibility is that the intention from @jaredoconnell was that the grid would support spacers of some kind (maybe grid lines?). But I can't really see how that would be meaningful with this API.

masonry/src/widgets/progress_bar.rs
```diff
@@ -33,11 +33,11 @@ pub struct ProgressBar {
...
     /// Create a new `ProgressBar`.
     ///
-    /// `progress` is a number between 0 and 1 inclusive. If it is `NaN`, then an
-    /// indefinite progress bar will be shown.
-    /// Otherwise, the input will be clamped to [0, 1].
-    pub fn new(mut progress: Option<f64>) -> Self {
-        clamp_progress(&mut progress);
+    /// The progress value will be clamped to [0, 1].
+    ///
+    /// A `None` value (or NaN) will show an indeterminate progress bar.
```
DJMcNab, line 38: I'm not a huge fan of NaN being meaningful API-wise; but that isn't for this PR to address.

masonry/src/widgets/root_widget.rs
```diff
@@ -14,15 +14,14 @@ use crate::core::{
 };
 use crate::kurbo::Size;
 
-// TODO: This is a hack to provide an accessibility node with a Window type.
-// This should eventually be removed.
-#[expect(missing_docs, reason = "TODO")]
+// TODO: This should eventually be removed once accesskit does that for us.
```
DJMcNab, line 17: Is there an accesskit issue this can link to?

masonry/src/widgets/sized_box.rs
```diff
@@ -327,7 +327,9 @@ impl SizedBox {
 
 // --- MARK: WIDGETMUT ---
 impl SizedBox {
-    #[expect(missing_docs, reason = "TODO")]
+    /// Give this container a child widget.
+    ///
+    /// If this container already has a child, it will be overwritten.
     pub fn set_child(this: &mut WidgetMut<'_, Self>, child: impl Widget) {
```
DJMcNab, line 333: Incidentally, this was an extremely surprising API to me. I don't know that it causes any complication, but it didn't match my mental model of how widgets like this should be handled.

masonry/src/widgets/sized_box.rs
```diff
@@ -337,7 +339,9 @@ impl SizedBox {
         this.ctx.request_layout();
     }
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Remove the child widget.
+    ///
+    /// (If this widget has no child, this method does nothing.)
```
DJMcNab, line 344: Should it `debug_panic` instead (again, not really for this PR)?

xilem/src/view/grid.rs
```diff
@@ -195,16 +192,19 @@ impl<W: Widget + FromDynWidget + ?Sized> SuperElement<Pod<W>, ViewCtx> for GridE
         // There is not much else, beyond purposefully failing, that can be done here,
         // because there isn't enough information to determine an appropriate spot
         // for the widget.
-        Self::Child(child.erased(), GridParams::new(1, 1, 1, 1))
+        Self {
+            child: child.erased(),
+            // TODO - Should be 0, 0?
```
DJMcNab, line 197: Ideally we'd follow the rules in [grid-auto-flow](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-flow). But that's again not really for this PR

xilem/src/view/portal.rs
```diff
@@ -8,7 +8,7 @@ use masonry::widgets;
 use crate::core::{DynMessage, Mut, ViewMarker};
 use crate::{MessageResult, Pod, View, ViewCtx, ViewId, WidgetView};
 
-/// A view which puts `child` into a scrollable region.
+/// Returns a view which puts `child` into a scrollable region.
```
DJMcNab, line 11: I don't think saying "returns a" is meaningfully better than "a" here. Conceptually, I think of this as having an elided `[The component for ]a view which ...`

That is, component functions never do anything other than returning a view, it's only noise to specify that the thing it does with the view is return it.

masonry/src/widgets/button.rs
PoignardAzur: Honestly, for some of these items the signature is self-explanatory; the documentation only serves to say "Yes, this does what the signature implies".

masonry/src/theme.rs
PoignardAzur: I'm not sure I'd even want these one-line docs. I'd be worried they could get out of sync with the actual values.

xilem/src/view/variable_label.rs
```diff
@@ -12,7 +12,7 @@ use super::{Label, label};
 use crate::core::{DynMessage, Mut, ViewMarker};
 use crate::{MessageResult, Pod, View, ViewCtx, ViewId};
 
-/// A view for displaying non-editable text, with a variable [weight](masonry::parley::style::FontWeight).
+/// Returns a view for displaying non-editable text, with a variable [weight](masonry::parley::style::FontWeight).
```
DJMcNab, line 15: This is still "returns a"

xilem/src/view/textbox.rs
```diff
@@ -13,6 +13,7 @@ use crate::{Color, MessageResult, Pod, TextAlignment, ViewCtx, ViewId};
 
 type Callback<State, Action> = Box<dyn Fn(&mut State, String) -> Action + Send + Sync + 'static>;
 
+/// Returns a view which displays editable text.
```
DJMcNab, line 16: As is this

xilem/src/view/prose.rs
```diff
@@ -10,6 +10,7 @@ use vello::peniko::Brush;
 use crate::core::{DynMessage, Mut, ViewMarker};
 use crate::{Color, MessageResult, Pod, TextAlignment, View, ViewCtx, ViewId};
 
+/// Returns a view which displays selectable text.
```
DJMcNab, line 13: And this

DJMcNab (approved):
Once the last "returns a" are fixed, I'm happy for this to land

masonry/src/widgets/grid.rs
PoignardAzur: I can add a TODO, but otherwise I'd rather not go down that hole.

PoignardAzur:
I think I've addressed all comments.

I've added a few TODOs in places where I didn't want to change the code.

masonry/src/widgets/sized_box.rs
PoignardAzur: Agreed, not addressing that this PR (I'm not sure it even warrants a TODO comment).

masonry/src/theme.rs
PoignardAzur: I'm leaving that as-is unless you consider it a serious blocker.

DJMcNab: I don't consider this a blocker.

DJMcNab (approved):
Thanks! It's great to get so many things documented here

PR #903, opened 2025-03-21
masonry: Remove `focus` from `PointerState`

waywardmonkeys:
This is unused.


masonry/src/core/event.rs
```diff
@@ -269,10 +269,6 @@ pub struct PointerState {
     /// The number of successive clicks registered. This is used to detect e.g. double-clicks.
     pub count: u8,
 
-    // TODO - Find out why this was added, maybe remove it.
```
DJMcNab, line 272: The quickest search for `pub focus: bool` in Druid brought up:
https://github.com/linebender/druid/blob/b831b5fe1597d5ec1fc3379cc1b39f3dd106e220/druid/src/mouse.rs#L39-L46

and https://github.com/linebender/druid/pull/842

That seems like reasonable functionality to have (indeed, I want it on other platforms, even thought it's not idiomatic). But I presume that winit doesn't sufficiently expose it to us?

DJMcNab (approved):
I'm happy to approve this, but it's probably worth making an issue to track the MacOS native behaviour.

waywardmonkeys:
Updated the first comment with more of the context after Daniel dug up the history.

DJMcNab:
Fwiw, I don't fully agree with the reasoning given in the PR description - we need to effectively make this decision per-widget. However, winit doesn't give us enough information to do so, so we can't really do better.

PR #904, opened 2025-03-21
Tighten screenshot testing

### LINKED ISSUE #893: Screenshot tests have false negatives
This is a bit worrying. While I've been mucking in Masonry widget code, I've noticed that some screenshots failed to update.

Apparently our diff infrastructure considers that this:

Image

is roughly equivalent to this:

Image

I haven't looked into why in detail yet. My best guess is that either the algorithm nv_flip uses considers dark grey and white to be very similar colors, or the difference is in a small enough number of pixels compared to the size of the image that the image is considered unchanged.

As an aside, I wonder how well kompari handles these images.

PoignardAzur:
Remove nv-flip dependency and use pixel-by-pixel comparison instead.
Reduce error tolerance in tests.
Update a few outdated screenshots.

Create `assert_failing_render_snapshot` macro to make sure our diffing infrastructure catches regressions.

Fixes #893.


PoignardAzur:
So far a tolerance value of 16 seems to be the sweet spot.

The test suite passes CI (so it works on both MacOS and Linux-with-lavapipe), but it also stops me if I try to introduce small errors like swapping the stroke join, changing a widget's border width, moving text by a tenth of a pixel, etc. Very encouraging.

PoignardAzur:
> but it also stops me if I try to introduce small errors like swapping the stroke join

I wonder if we should add deliberate errors to our test suite, with `#[should_panic]` tests. Shouldn't be *too* hard.

PoignardAzur:
Aaaaand done. I'm pretty happy with the new macro.

PoignardAzur:
This PR is now tested on at leat N machines:

- My workstation (Ubuntu)
- My laptop (Ubuntu)
- CI (Windows)
- CI (Ubuntu with lavapipe)
- @RagibHasin's machine (Windows)
- @jaredoconnell's machine (MacOS)

I'm now fairly confident it can be merged, pending the oxipng stuff.

waywardmonkeys (approved):
I think I'm okay with this for now, hoping that it is an intermediate step along the way to using Kompari.

PoignardAzur:
It is, yes. The diffing model especially is a lot closer to what kompari does.

waywardmonkeys:
This bit of documentation is now out of date:

```
The rendered screenshot is compared against an existing file checked in your project, and panics if the reference file is meaningfully different (using [the NVIDIA FLIP algorithm](https://developer.nvidia.com/blog/flip-a-difference-evaluator-for-alternating-images/)) or if there isn't one.
```

PR #905, opened 2025-03-21
Remove tempfile direct dependency

PoignardAzur:
`tempfile` on Unix pulls `rustix`, which has a 9.19s single-core build time on my machine, and was unused in our code.

`tempfile` is still a indirect dependency through `uds_windows`, which is fine because in that dependency path (windows-only) doesn't pull `rustix`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1902,7 +1902,6 @@ dependencies = [
  "serde",
  "serde_json",
  "smallvec",
- "tempfile",
  "time",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
It's sad that this didn't get reported by `unused_crate_dependencies`. To be clear, this isn't the fault of our CI setup; this is because of Rust's conflating of example and test dependencies as "dev" dependencies

The PR title/description bury the lead; the most important thing is that this dependency is entirely unused.

PoignardAzur:
> To be clear, this isn't the fault of our CI setup; this is because of

Because of?

DJMcNab:
Expanded

PR #907, opened 2025-03-21
[WIP] Decouple winit dependency

PoignardAzur:

PR #908, opened 2025-03-22
Isolate winit imports to `app` module

PoignardAzur:
This is the first step towards #907.

This doesn't split the crate yet, but adds a translation layer between winit and Masonry to facilitate that split.


nicoburns (approved):
I'm definitely keen to see these types split out into their own crate. But if that's planned for a followup then I'm happy to approve this as-is.

PR #909, opened 2025-03-23
Split most of the crate into masonry_core sub-crate.

PoignardAzur:
This change is done very carefully to minimize the number of changed files, and use almost only file renames. This is both for avoiding merge conflicts, keeping file history unbroken, and keeping the PR easy to review.

A future PR will rename masonry to masonry_core, masonry_winit to masonry change all the doc tests and snapshots that depend on those names, etc.

After *that*, future PRs will have actual code changes splitting winit away from masonry_core.

See also https://github.com/linebender/xilem/pull/907.


PoignardAzur:
Mhh. I'm not happy with the number of LFS files being considered added/removed instead of renamed.

PR #911, opened 2025-03-24
Mark `masonry/screenshots/*.png` as LFS files

PoignardAzur:
These files we previously stored as blobs as-is. This stores them using the same LFS method as other screenshots.


PoignardAzur:
I'm wondering if there's any strong reason to have the pattern for LFS files be `<some_path>/*.png` and not just `*.png`.

Is there any case where we expect to add PNG images to the repo and not use LFS?

DJMcNab:
Test input data should not be stored in LFS; only the "canonical" outputs/results. That's because we're intentionally treating the LFS as a potentially fallible step, due to GitHub's weirdness around it (primarily the bandwidth limitation), but also in case we needed to move off GitHub.

That is, we need to be able to gracefully restore to a non-LFS state, even if we somehow lost access to all LFS files (including historical ones).

DJMcNab (approved):
I'm not a massive fan of storing screenshots in unique directories per module/group of modules. However, this change is accurate for the current state of the repository.

PoignardAzur:
> Test input data should not be stored in LFS; only the "canonical" outputs/results.

Then how about something like `**/screenshots/*.png`? And document that pattern somewhere? (Maybe in ARCHITECTURE.md)

> I'm not a massive fan of storing screenshots in unique directories per module/group of modules.

I doubt we'll ever get to a state where we have a single screenshot folder for the entire workspace. I'd expect at least each crate to have its own folder.

(Right now it's only masonry, but at some point it's probably going to be masonry_core, masonry and xilem.)

PR #912, opened 2025-03-25
Fix pointer capture

### LINKED ISSUE #857: `portal` scroll bar stays activated when mouse is released outside of window

PoignardAzur:
Create new PointerLost event.
No longer remove pointer capture on PointerLeave.
Fix small bugs and implementation errors.

Fixes #857.
Supersedes #858.

### DIFF
--- a/masonry/src/app/event_loop_runner.rs
+++ b/masonry/src/app/event_loop_runner.rs
@@ -608,7 +608,7 @@ impl MasonryState<'_> {
                     }
                     winit::event::TouchPhase::Cancelled => {
                         self.render_root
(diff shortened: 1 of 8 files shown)


jaredoconnell (approved):
This is exactly what was needed.

PR #913, opened 2025-03-25
Add BackgroundGradient property

PoignardAzur:
Depends on #892.


RagibHasin:
Shouldn't  we have a `Background` enum property with variants for gradient, solid color, image, Vello scene? Otherwise if we add each variant as its own property and some widget gets more than one set at a time, it would be a problem.

PoignardAzur:
Yeah, this isn't my best work. I'll make the API cleaner.

masonry_core/src/properties/types/gradient.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Gradient that transitions between two or more colors along a line.
+    Linear {
+        /// The angle defining the gradient line's direction.
+        /// Zero degrees points upwards and positive angles represent clockwise rotation.
+        angle: f64,
+    },
+}
+
+#[derive(Clone, Debug)]
+pub struct Gradient {
```
DJMcNab, line 19: This should have docs (we should also mention Peniko's gradient type here, including it's lack of translation invariance as the reason for this being a separate type)

At the very least, the docs should mention that this is a widget property (maybe also what built-in widgets make use of it, but that's more questionable).

DJMcNab: Ha, indeed; exactly my point - this isn't a widget property, but I thought it was. That's what docs are for - this should link up to the `Background` property.

masonry_core/src/widgets/button.rs
```diff
@@ -219,16 +220,27 @@ impl Widget for Button {
         let border_width = props.get::<BorderWidth>().unwrap_or(&DEFAULT_BORDER_WIDTH);
         let border_radius = props.get::<CornerRadius>().unwrap_or(&DEFAULT_BORDER_RADII);
 
+        // TODO - Add DEFAULT_BACKGROUND_GRADIENT constant.
```
DJMcNab, line 223: Why? What does that constant achieve/why isn't it in this PR

masonry_core/src/widgets/button.rs
DJMcNab, line 1: Should the docs have changed (to mention that `Gradient` is amongst the properties which button supports)?

DJMcNab (changes requested):
I would strongly encourage removing https://github.com/linebender/xilem/blob/825cfd88ebba3c7c838921a9755fe233c19eec50/masonry_core/src/properties/mod.rs#L8-L11. The main issue with this code is that methods which should be documented aren't, because of the crutch of that line. That actual change shouldn't be in this PR

The explained reasoning, that properties are self-explanatory doesn't hold water for me; there's always something to say, even if it's that "this is a widget property which [does x]". And it's clear that having it on quickly regresses documentation writing, even for the things that lint allowal isn't intended to cover.

I do agree that needing to document fields sucks, but I think localised `#[expect(missing_docs, reason="obvious"]` is still better.

Other than the docs, this seems great. It could probably also do with a test.

masonry_core/src/properties/types/gradient.rs
PoignardAzur: This should hopefully be clearer once the Property trait is merged.

The general idea is you have a module with properties, and you have a sub-module with values used *in* properties that aren't properties themselves.

Yes, I should probably document that.

masonry_core/src/widgets/button.rs
PoignardAzur: Consistency with the other properties.

I agree it's not super important.

masonry_core/src/widgets/button.rs
PoignardAzur: I'm fine leaving that undocumented for now.

PoignardAzur:
I wish Rust had a `missing_field_docs` and `missing_variant_docs` sub-lints. It's almost never a good idea to leave types and methods undocumented, but it's very often unnecessary to document public fields.

masonry_core/src/properties/types/gradient.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::kurbo::{Point, Rect};
+use crate::peniko::color::{ColorSpaceTag, HueDirection};
+use crate::peniko::{ColorStops, ColorStopsSource, Extend};
+
+/// Properties for the supported [`Gradient`] types.
+///
+/// This mirrors [`peniko::GradientKind`](crate::peniko::GradientKind),
+/// but uses a layout-invariant representation: instead of saying
+/// "The gradient goes from point A to point B", we declare things like
+/// "The gradient has angle X", and A and B and computed dynamically from widget layout.
```
DJMcNab, line 13:
```suggestion
/// "The gradient has angle X", and A and B are computed dynamically from widget layout.
```
?

masonry_core/src/properties/types/gradient.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// [CSS Color Module Level 4 § 12.4]: https://drafts.csswg.org/css-color/#hue-interpolation
+    pub hue_direction: HueDirection,
+    /// Color stop collection.
+    pub stops: ColorStops,
+}
+
+impl Gradient {
+    /// Creates a [`Linear`](GradientShape::Linear) gradient.
+    ///
+    /// `angle` is in radians, with zero pointing upwards.
```
DJMcNab, line 59: It might be worth adding something like:
```suggestion
    /// `angle` is in radians, with zero pointing upwards, and higher values rotating the gradient clockwise.
    /// This matches how [CSS gradients are defined](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient#angle).
    ///
    /// That is, for an `angle` of 0, the first stop will be at the bottom, and
    /// for an `angle` of [`π/2`](core::f32::consts::FRAC_PI_2) (90°), the first
    /// stop will be aligned with the left edge.
```

Since increasing radians almost always go anticlockwise usually (and 0 indicates pointing to the left)

masonry_core/src/widgets/button.rs
DJMcNab: I'd strongly prefer that we did document it. But I guess this PR isn't a regression on that front :(

DJMcNab (approved):
I don't like that this isn't documented on the `Button` side, but that isn't a regression in this PR.

PR #914, opened 2025-03-26
Rename `masonry` crate to `masonry_core` and `masonry_winit` crate to `masonry`

PoignardAzur:
The main change is to change the names of the crates as registered in the `Cargo.toml` files. The folders these projects are in aren't renamed, but since most snapshot test files include their crate name, a lot of files end up being renamed as well.


PoignardAzur:
Once this PR is merged, the remaining work will be actually splitting masonry and masonry_core, so mostly removing dependencies from the `Cargo.toml` files.

DJMcNab (approved):
My gut reaction is that having the module `masonry_core::core` is quite unfortunate. However, I don't know that there's a reasonable alternative.

One thing which slightly worries me is that the re-exports *might* break source links on `docs.rs` - have you checked that the way docs are rendered with this change is sensible?

I think landing this quickly is good - we can reason about any other changes at a later point.

PoignardAzur:
> One thing which slightly worries me is that the re-exports might break source links on docs.rs - have you checked that the way docs are rendered with this change is sensible?

Rustdoc handles reexports mostly gracefully (by pretending the re-exported module are part of the parent crate).

Aside from CI not breaking, I've tried clicking around in the doc and didn't see anything obviously wrong.

PR #915, opened 2025-03-26
Split winit dependency from masonry_core

PoignardAzur:
Builds on top of #914.

This is the final step to splitting masonry_core from masonry.


PoignardAzur:
Note that this split isn't perfect:

- Ideally, I'd like to move most of the tracing dependencies out of `masonry_core`, but there's no easy way to do that right now.
- There are places where we'd want the `masonry_core` doc to link to `masonry` items. I don't really know how to do that either.

We should probably create follow-up issues once this is merged.

nicoburns (approved):

PR #917, opened 2025-04-03
support font weight option on prose (like label)

richard-uk1:
This is a small PR that adds font weight to prose. There are other properties, but this is the only one I needed now :)

### DIFF
--- a/xilem/src/view/prose.rs
+++ b/xilem/src/view/prose.rs
@@ -127,6 +127,9 @@ impl<State, Action> View<State, Action, ViewCtx> for Prose {
                 StyleProperty::FontSize(self.text_size),
             );
         }
+        if prev.weight != self.weight {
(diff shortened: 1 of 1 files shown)


xilem/src/view/prose.rs
```diff
@@ -86,6 +95,7 @@ impl<State, Action> View<State, Action, ViewCtx> for Prose {
             .with_brush(self.text_brush.clone())
             .with_alignment(self.alignment)
             .with_style(StyleProperty::FontSize(self.text_size))
+            .with_style(StyleProperty::FontWeight(self.weight))
```
DJMcNab, line 98: This field also needs to be handled in rebuild.

DJMcNab:
@richard-uk1 are you planning to continue working on this?

PoignardAzur (approved):

DJMcNab (approved):

PR #921, opened 2025-04-08
Virtual Scrolling for Xilem

DJMcNab:
The semantics of how this integration works are really ugly - it really feels like there's a better model hiding in here somewhere, but that's somewhat by-the-by.

However, it does work.

### DIFF
--- a/xilem/examples/virtual_cats.rs
+++ b/xilem/examples/virtual_cats.rs
@@ -10,6 +10,7 @@
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use masonry::core::ArcStr;
 use masonry::widgets::Alignment;
 use tokio::sync::mpsc::Sender;
 use vello::peniko::{Blob, Image};
@@ -32,7 +33,7 @@ struct VirtualCats {
 #[derive(Debug)]
 struct Status {
     code: u32,
-    message: &'static str,
+    message: ArcStr,
     image: ImageState,
 }
 
@@ -83,7 +84,7 @@ impl VirtualCats {
         } else {
             Either::B(sized_box(spinner()).width(80.).height(80.))
         };
-        flex((img, prose(item.message)))
+        flex((prose(item.message.clone()), img))
     }
 
     fn view(&mut self) -> impl WidgetView<Self> + use<> {
@@ -181,7 +182,7 @@ impl Status {
         let (code, message) = line.split_once(',')?;
         Some(Self {
             code: code.parse().ok()?,
-            message: message.trim(),
+            message: format!("{}:", message.trim()).into(),
             image: ImageState::NotRequested,
         })
     }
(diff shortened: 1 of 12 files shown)


DJMcNab:
The status of this is that it does now work - I'm following up today with a more focused example (reusing the dataset from `http_cats`).

xilem/examples/virtual_cats.rs
```diff
@@ -0,0 +1,201 @@
...
+        } else {
+            None
+        };
+        fork(
+            virtual_scroll(0..self.statuses.len() as i64, Self::virtual_item),
+            // TODO: This is really awful...
+            // Ultimately, this doesn't really make snse as a view?
+            task_raw(
+                move |proxy| {
+                    let mut rx = rx.as_ref().unwrap().lock().unwrap().take().unwrap();
```
jaredoconnell, line 103: This is a messy line. I don't see a better alternative to it through without adding code elsewhere.

jaredoconnell (commented):
First pass of comments. Will review more later.
The example works great! I'm glad that this is being worked on. I've been hoping for this since I joined the Linebender community when Druid was still supported.

xilem/examples/virtual_cats.rs
```diff
@@ -53,17 +51,46 @@ impl VirtualCats {
...
+                                );
+                            }
+                        }
+                    }
+                },
+                move |state: &mut Self, image| {
+                    if let Some(status) = state.statuses.iter_mut().find(|it| it.code == code) {
+                        status.image = ImageState::Available(image);
+                    } else {
+                        // TODO: Error handling?
```
jaredoconnell, line 87: What happens right now in the event of a failure?
Would it make sense to load a stock failure image?

xilem/src/view/virtual_scroll.rs
```diff
@@ -0,0 +1,355 @@
...
+                        view_state.view_states.insert(
+                            idx,
+                            private::ChildState {
+                                state: child_state,
+                                requested_rebuild: false,
+                            },
+                        );
+                    }
+                });
+            }
```
jaredoconnell, line 262: I'd say the `if` statement doesn't clearly define what that condition is for, so I think a comment may be useful.

xilem/src/view/virtual_scroll.rs
```diff
@@ -0,0 +1,355 @@
...
+                            &mut state.state,
+                            ctx,
+                            widgets::VirtualScroll::child_mut(&mut element, *idx),
+                        );
+                    });
+                    state.requested_rebuild = false;
+                }
+            }
+        } else {
+            debug_assert!(view_state.cleanup_queue.is_empty());
```
jaredoconnell, line 210: It may be useful to add a comment that this section is for when the set of children has changed.

jaredoconnell (approved):
Looks good for an initial implementation. I have some minor comments.
I think it would make sense to create Github issues for some of the TODOs in here, since there are a few places that you mention have room for future improvements.

xilem/examples/virtual_cats.rs
DJMcNab: The spinner continues forever, as far as I remember. I don't think that loading a stock failure image is doable, because what if fail to load that!

jaredoconnell (approved):
Much better, thanks for the improvements! Should be good once the failures found in CI are addressed.

DJMcNab:
As far as I know, both lots of CI failures are waiting for upstream to fix them.

PR #922, opened 2025-04-11
Add multiline text input support to TextArea

alexjg:
The textarea widget currently doesn't support multiline input because the Enter key is used to trigger the TextEntered action. This commit makes this behavior configurable by adding masonry_core::widgets::InsertNewline.

### DIFF
--- a/masonry_core/src/widgets/text_area.rs
+++ b/masonry_core/src/widgets/text_area.rs
@@ -40,7 +40,8 @@ use cursor_icon::CursorIcon;
 ///
 /// This widget emits the following actions only when `USER_EDITABLE` is true:
 ///
-/// - `TextEntered`, which is sent when the enter key is pressed
+/// - `TextEntered`, which is sent when the enter key is pressed (this can be
+///   configured using [`with_insert_newline`](Self::with_insert_newline))
 /// - `TextChanged`, which is sent whenever the text is changed
 ///
 /// The exact semantics of how much horizontal space this widget takes up has not been determined.
@@ -101,6 +102,10 @@ pub struct TextArea<const USER_EDITABLE: bool> {
     /// Can be set using [`set_padding`](Self::set_padding).
     /// Immediate parent widgets should use [`with_padding_if_default`](Self::with_padding_if_default).
     padding: Padding,
+
+    /// What key combination should trigger a newline insertion.
+    /// If this is set to `InsertNewline::OnEnter` then `Enter` will insert a newline and _not_ trigger a `TextEntered` event.
+    insert_newline: InsertNewline,
 }
 
 // --- MARK: BUILDERS ---
@@ -148,6 +153,7 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
             // We use -0.0 to mark the default padding.
             // This allows parent views to overwrite it only if another source didn't configure it.
             padding: Padding::UNSET,
+            insert_newline: InsertNewline::default(),
         }
     }
 
@@ -276,6 +282,12 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
         if self.padding.is_unset() {
             self.padding = padding;
         }
+        self
+    }
+
+    /// Configures how this text area handles the user pressing Enter <kbd>↵</kbd>.
+    pub fn with_insert_newline(mut self, insert_newline: InsertNewline) -> Self {
+        self.insert_newline = insert_newline;
         self
     }
 
@@ -425,6 +437,12 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
         this.widget.editor.set_alignment(alignment);
 
         this.ctx.request_layout();
+    }
+
+    /// Configures how this text area handles the user pressing Enter <kbd>↵</kbd>.
+    pub fn set_insert_newline(this: &mut WidgetMut<'_, Self>, insert_newline: InsertNewline) {
+        this.widget.insert_newline = insert_newline;
+        this.ctx.request_accessibility_update();
     }
 
     #[doc(alias = "set_color")]
@@ -738,9 +756,12 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
                         edited = true;
                     }
                     Key::Enter => {
-                        // TODO: Multiline?
-                        let multiline = false;
-                        if multiline {
+                        let insert_newline = match self.insert_newline {
+                            InsertNewline::OnEnter => true,
+                            InsertNewline::OnShiftEnter => shift,
+                            InsertNewline::Never => false,
+                        };
+                        if insert_newline {
                             let (fctx, lctx) = ctx.text_contexts();
                             self.editor
                                 .driver(fctx, lctx)
@@ -979,8 +1000,10 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
 
     fn accessibility_role(&self) -> Role {
         if EDITABLE {
-            Role::TextInput
-            // TODO: Role::MultilineTextInput
+            match self.insert_newline {
+                InsertNewline::OnShiftEnter | InsertNewline::OnEnter => Role::MultilineTextInput,
+                _ => Role::TextInput,
+            }
         } else {
             Role::Document
         }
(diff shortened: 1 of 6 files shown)


masonry_core/src/widgets/text_area.rs
```diff
@@ -1119,4 +1152,78 @@ mod tests {
...
+            let area = widget.downcast::<TextArea<true>>().unwrap();
+            let text = area.widget.text().to_string();
+
+            if scenario.expect_text_entered_event {
+                let (action, widget_id) = harness.pop_action().unwrap();
+                assert_eq!(action, Action::TextEntered("hello world".to_string()));
+                assert_eq!(widget_id, text_id);
+                assert_eq!(text, "hello world");
+            } else {
+                assert_eq!(text, "\nhello world");
```
DJMcNab, line 1225: We should expect a `TextChanged` here, right?

masonry_core/src/widgets/text_area.rs
```diff
@@ -1119,4 +1152,78 @@ mod tests {
...
+                None,
+            ));
+
+            let widget = harness.try_get_widget(text_id).unwrap();
+            let area = widget.downcast::<TextArea<true>>().unwrap();
+            let text = area.widget.text().to_string();
+
+            if scenario.expect_text_entered_event {
+                let (action, widget_id) = harness.pop_action().unwrap();
+                assert_eq!(action, Action::TextEntered("hello world".to_string()));
```
DJMcNab, line 1221: And we should expect that there is no textchanged event here, I guess?

xilem/src/view/textbox.rs
```diff
@@ -73,7 +81,8 @@ impl<State: 'static, Action: 'static> View<State, Action, ViewCtx> for Textbox<S
         // TODO: Maybe we want a shared TextArea View?
         let text_area = widgets::TextArea::new_editable(&self.contents)
             .with_brush(self.text_brush.clone())
-            .with_alignment(self.alignment);
+            .with_alignment(self.alignment)
+            .with_insert_newline(self.insert_newline);
```
DJMcNab, line 85: This needs to be handled in rebuild as well; that is, in `rebuild`, we should call `set_insert_newline`.

That does raise questions about how to handle text which already has a newline in it (and `InsertNewline::OnEnter`)

masonry_core/src/widgets/text_area.rs
```diff
@@ -1018,16 +1034,33 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+pub enum InsertNewline {
+    /// Insert a newline when the user presses Enter
+    ///
+    /// Note that if this is enabled, then the text area will never emit a `TextEntered` event
+    OnEnter,
+    /// Insert a newline when the user presses Shift+Enter
+    OnShiftEnter,
+    /// Never insert a newline
+    Never,
+}
```
DJMcNab, line 1048: I think this is a reasonable-ish set of options. It feels like longer-term you'd want something more customisable, but just getting something landed is worth doing.

DJMcNab (commented):
Some notes. The main thing which needs to change is the handling in Xilem.

I'd also consider using this new functionality in an example. Either `mason`, which is the generic playground, or maybe `todomvc`?

xilem/src/view/textbox.rs
alexjg: I've wired up the `rebuild`. When you say "text with newlines already in it" do you mean if someone changes the `InsertNewLine` behavior and then re-renders? It seems to me that in that case we should do nothing, the behavior - whether creating a newline or firing a TextEntered action - seems to me like it should be triggered by the enter key, not by the presence of newlines in the text?

masonry_core/src/widgets/text_area.rs
alexjg: I've added an assertion that only one action is fired.

masonry_core/src/widgets/text_area.rs
alexjg: I added such an  assertion

alexjg:
I've added an example to both `masonry` and `to_do_mvc`. The `to_do_mvc` example was a good place to show the `shift-enter` interaction.

masonry_core/src/widgets/text_area.rs
```diff
@@ -101,6 +101,10 @@ pub struct TextArea<const USER_EDITABLE: bool> {
     /// Can be set using [`set_padding`](Self::set_padding).
     /// Immediate parent widgets should use [`with_padding_if_default`](Self::with_padding_if_default).
     padding: Padding,
+
+    // What key combination should trigger a newline insertion
+    // If this is set to `InsertNewline::OnEnter` then `Enter` will insert a newwline and _not_ trigger a `TextEntered` event
```
RagibHasin, line 106: Shouldn't this be a doc comment?

Also

> will insert a newwline and

has a typo in "newwline". Is this a `typos` failure @DJMcNab ?

alexjg: :facepalm: fixed

DJMcNab: Thanks for catching this. When we catch a type, we like to contribute it to https://github.com/crate-ci/typos - there's a monthly issue.
(It's always worth sanity checking with a search engine that there isn't a reason to use the typo version, before submitting)

masonry_core/src/widgets/text_area.rs
```diff
@@ -279,6 +284,12 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
         self
     }
 
+    /// Set the `InsertNewline` behavior
```
DJMcNab, line 287: Our style is to use full stops at the ends of doc comment sentences (usually; I see the one directly below this doesn't do so :sweat_smile: ). This applies to all the doc comments in this PR.

xilem/src/view/textbox.rs
alexjg: Ah I see. That feels like it's a slightly different kind of configuration to do with what characters are allowed in the textbox? The same problem would arise when pasting text for example, or when trying to enter non-numeric characters into a numeric field of some kind.

masonry_core/src/widgets/text_area.rs
```diff
@@ -979,8 +998,10 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
 
     fn accessibility_role(&self) -> Role {
         if EDITABLE {
-            Role::TextInput
-            // TODO: Role::MultilineTextInput
+            match self.insert_newline {
+                InsertNewline::OnShiftEnter | InsertNewline::OnEnter => Role::MultilineTextInput,
+                _ => Role::TextInput,
+            }
```
DJMcNab, line 1004: (this is additional to/the same thing as my previous comment)

This won't be automatically rerun when you call `set_insert_newline`, which means that until the accessibility is updated by another means, the accessibility kind will be out-of-date.

DJMcNab (changes requested):
This is great. There's one blocking concern, which is the accessibility handling of the new `set_insert_newline`.

There are a few other nits I'd also like to see fixed before merging. Whoever does the merge can apply these, so these aren't blocked on your (although if you did apply them, that would be great!)

masonry_core/src/widgets/text_area.rs
```diff
@@ -1018,16 +1039,33 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
     }
 }
 
+/// When to insert a newline in a text area
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum InsertNewline {
```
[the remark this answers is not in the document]
alexjg: It's the `shift+enter` combo which makes all these names bad. `EnterHandling` would be fine if it was create newline _or_ emit text entered event, but now we have this other thing which is "insert a newline if the key combo was shift+enter" so now we're not handling `Enter` exactly.

Other candidate names:

* `NewlineAction`
* `EnterActions`
* `NewlineOrTextEntered`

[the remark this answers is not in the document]
DJMcNab: Maybe something like `SubmitCondition { AnyEnter, ShiftEnter, Never }`

alexjg:
I think I've handled everything apart from the naming of `InsertNewline`, which we're still thinking about.

xilem/src/view/textbox.rs
DJMcNab: Indeed. When we've had this discussion before, pasting has been the reason. Adding this into the mix just makes things even more complicated!

PR #923, opened 2025-04-14
Fix bounding rectangles in accessibility nodes for text runs

mwcampbell:
It's no longer correct to add the window-relative origin of the widget to the bounding rectangles of the text runs, since we now apply a transform on each widget's AccessKit node.


xorgy (approved):
Thanks!

PR #925, opened 2025-04-15
masonry: Scroll by 120 logical pixels per line scroll increment.

xorgy:
This is the default behavior in Chromium (when no configuration is provided) and is a better default than individual logical pixels.

### DIFF
--- a/masonry/src/event_loop_runner.rs
+++ b/masonry/src/event_loop_runner.rs
@@ -557,11 +557,12 @@ impl MasonryState<'_> {
                 }
             }
             WinitWindowEvent::MouseWheel { delta, .. } => {
-                // TODO - This delta value doesn't quite make sense.
-                // Figure out and document a better standard.
                 let delta = match delta {
+                    // 120 pixels is the default in Chrome and various other software
+                    // (when not overridden by configuration).
+                    // This is arbitrary, but better than nothing.
                     winit::event::MouseScrollDelta::LineDelta(x, y) => {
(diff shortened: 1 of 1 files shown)


masonry/src/event_loop_runner.rs
```diff
@@ -557,11 +557,12 @@ impl MasonryState<'_> {
                 }
             }
             WinitWindowEvent::MouseWheel { delta, .. } => {
-                // TODO - This delta value doesn't quite make sense.
-                // Figure out and document a better standard.
                 let delta = match delta {
+                    // 120 pixels is the default in Chrome and various other software
+                    // (when not overridden by configuration).
+                    // This is arbitrary, but better than nothing.
```
DJMcNab, line 563: Is this something we should be getting from the OS? That is, would this config be app specific?

I ask *only* to make this comment clearer (and maybe we should also link to an issue?)

DJMcNab (approved):
The updated comment isn't especially actionable, but otherwise this is reasonable

masonry/src/event_loop_runner.rs
xorgy: The ultimate solution is to allow the direct descendant of the portal to configure this at runtime, if true line deltas are desired, but for scrolling arbitrary content with unknown line/module sizes, 120 is a reasonable default.

xorgy: @DJMcNab I've opened #926 for further discussion on this matter.

PR #927, opened 2025-04-15
Masonry: remove scrolling factor

Azorlogh:
After https://github.com/linebender/xilem/pull/925, that 10x scrolling speed factor is no longer needed

### DIFF
--- a/masonry_core/src/widgets/portal.rs
+++ b/masonry_core/src/widgets/portal.rs
@@ -274,8 +274,6 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
         _props: &mut PropertiesMut<'_>,
         event: &PointerEvent,
     ) {
-        const SCROLLING_SPEED: f64 = 10.0;
-
         let portal_size = ctx.size();
         let content_size = ctx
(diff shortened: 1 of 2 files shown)


DJMcNab:
Welcome! This seems reasonable in theory, as the amount does now reasonably represent pixels.

The scroll amount for using my trackpad was previously quite good. This effectively makes any "pixel delta" 10 times slower, which also doesn't seem ideal. But I do agree that 10 lines per scroll increment definitely isn't good!

xorgy (approved):

xorgy:
@Azorlogh this is a good fix, the failing tests need to be adjusted because they depend on scrolling to a certain position (so you can probably just multiply some quantities by 10 in the tests and they should pass).

DJMcNab (approved):
Agreed, we should land this. It's only blocked on updating the tests.

Azorlogh:
@xorgy Thanks for updating the tests! They work on my machine :ok_hand:

xorgy:
It seems that at random sometimes they don't segfault. I can run the tests without SIGSEGV about one in nine times.

PR #929, opened 2025-04-17
Streamline Widget trait using trait upcasting

PoignardAzur:
Bump MSRV to 1.86

### DIFF
--- a/masonry_core/src/testing/helper_widgets.rs
+++ b/masonry_core/src/testing/helper_widgets.rs
@@ -18,7 +18,6 @@ use smallvec::SmallVec;
 use tracing::trace_span;
 use vello::Scene;
 
-use crate::AsAny;
 use crate::core::{
     AccessCtx, AccessEvent, BoxConstraints, ComposeCtx, EventCtx, LayoutCtx, PaintCtx,
     PointerEvent, PropertiesMut, PropertiesRef, QueryCtx, RegisterCtx, TextEvent, Update,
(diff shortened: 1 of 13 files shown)


PoignardAzur:
Not sure where the `cargo doc` ICE is coming from. It's not a fluke, I was triggered again when I re-ran the job, but running the same command on my machine yields no error.

Ideally I'd like to send a bug report, but I have no idea how to get the `rustc-ice-<...>.txt` file.

masonry_core/src/testing/helper_widgets.rs
```diff
@@ -734,12 +725,4 @@ impl<W: Widget> Widget for Recorder<W> {
     fn short_type_name(&self) -> &'static str {
         "Recorder"
     }
-
-    fn as_any(&self) -> &dyn std::any::Any {
-        self.child.as_any()
```
DJMcNab, line 739: This seems to be different semantics, right?

DJMcNab (approved):
Seems reasonable to me

DJMcNab:
The way to get the file is to add uploading it as an artifact (or `cat`ting it) as a step in CI.

masonry_core/src/testing/helper_widgets.rs
PoignardAzur: True, it makes downcasting code "skip" the Recorder wrapper.

These downcasting tricks became mostly redundant since we removed `impl Widget for Box<dyn Widget>`, and they were a source of headaches before then.

DJMcNab:
It's strange that you can't reproduce it even on the same nightly version that CI uses though.

PR #930, opened 2025-04-17
Rework find_widget_at_pos method

PoignardAzur:
Rename to find_widget_under_pointer
Make function body more readable.
Make default implementation use only public items.

### DIFF
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -354,13 +354,13 @@ pub trait Widget: AsAny + AsDynWidget {
     ///
     /// **pos** - the position in global coordinates (e.g. `(0,0)` is the top-left corner of the
     /// window).
-    fn find_widget_at_pos<'c>(
+    fn find_widget_under_pointer<'c>(
         &'c self,
         ctx: QueryCtx<'c>,
         props: PropertiesRef<'c>,
         pos: Point,
     ) -> Option<WidgetRef<'c, dyn Widget>> {
-        find_widget_at_pos(
+        find_widget_under_pointer(
             &WidgetRef {
                 widget: self.as_dyn(),
                 properties: props,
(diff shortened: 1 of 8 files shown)


masonry_core/src/core/widget.rs
```diff
@@ -410,38 +410,43 @@ pub trait Widget: AsAny + AsDynWidget {
...
+        return None;
+    }
+    if ctx.is_stashed() {
+        return None;
+    }
+
+    let local_pos = ctx.window_transform().inverse() * pos;
+
+    if let Some(clip) = ctx.clip_path() {
+        if !clip.contains(local_pos) {
```
Philipp-M, line 429: I guess we can soon combine these in one `if` (as let chains are just getting stabilized) :)

Philipp-M (approved):
I got no strong opinion on the name `find_widget_under_pointer` vs `find_widget_at_pos` but I think it's general enough to not just be a "pointer" but a (global) window position, and is more accurately described by "position" vs "pointer".

Otherwise LGTM.

masonry_core/src/core/widget.rs
PoignardAzur: Fingers crossed.

PoignardAzur:
> but I think it's general enough to not just be a "pointer" but a (global) window position, and is more accurately described by "position" vs "pointer".

The rename is deliberate: since the code specifically avoids returning ids of children with `accepts_pointer_events: false`, the name explicitly reflect that it's pointer-specific.

Or I guess we could add a predicate argument.

Philipp-M:
> Or I guess we could add a predicate argument.

sounds like a good idea. But that may also be done in a follow-up.

xStrom:
Some docs need updating.

```
error: unresolved link to `Widget::find_widget_at_pos`
   --> masonry_core/src/core/widget.rs:412:11
    |
412 | /// See [`Widget::find_widget_at_pos`] for more details.
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Widget` has no associated item named `find_widget_at_pos`
    |
    = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`

error: unresolved link to `Widget::find_widget_at_pos`
   --> masonry_core/src/core/widget_ref.rs:186:11
    |
186 |     /// [`Widget::find_widget_at_pos`] to descend the widget tree. If `self` does not contain the
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Widget` has no associated item named `find_widget_at_pos`
```

Issue #933, opened 2025-04-18
Each `Label::layout` pass stacks another text alignment offset additively.


eugenesvk:
After rebasing this example https://github.com/linebender/xilem/pull/864 off the latest main:

- Initial launch
<img width="581" alt="Image" src="https://github.com/user-attachments/assets/b9bec9f4-77c2-418b-9d00-7af1c5427809" />

- Maximized window - text moves outside
<img width="617" alt="Image" src="https://github.com/user-attachments/assets/6c282885-7b18-4266-afc1-0c16d2332e57" />

- After a few window max/restore: thetext moves on each window max and eventually it disappears off screen

<img width="589" alt="Image" src="https://github.com/user-attachments/assets/604c60fb-7b98-44c1-8dd3-3c5a506225d1" />

xStrom:
Thanks for the report! I created a bit more cut down repro from your example:

<details>
<summary>Click to see code</summary>

```rust
// Copyright 2025 the Xilem Authors
// SPDX-License-Identifier: Apache-2.0

//! Debugging layout

use masonry::dpi::LogicalSize;
use winit::window::Window;

use masonry::core::ArcStr;
use masonry::peniko::Color;
use winit::error::EventLoopError;
use xilem::view::{
    Axis, GridExt, Label, Padding, flex, grid, label, portal, sized_box,
};
use xilem::{EventLoop, TextAlignment, WidgetView, Xilem, palette::css};

const LABEL_COLOR: Color = css::ROYAL_BLUE;

struct AppState {
    l1i1: TextAlignment,
    l1i2: TextAlignment,
    l1i3: TextAlignment,
    l1i4: TextAlignment,
    l1i5: TextAlignment,
    l1i6: TextAlignment,
    l2i1: TextAlignment,
    l2i2: TextAlignment,
    l2i3: TextAlignment,
    l2i4: TextAlignment,
    l2i5: TextAlignment,
    l2i6: TextAlignment,
}
impl Default for AppState {
    fn default() -> Self {
        Self {
            l1i1: TextAlignment::Start,
            l1i2: TextAlignment::Left,
            l1i3: TextAlignment::Middle,
            l1i4: TextAlignment::End,
            l1i5: TextAlignment::Right,
            l1i6: TextAlignment::Justified,
            l2i1: TextAlignment::Start,
            l2i2: TextAlignment::Left,
            l2i3: TextAlignment::Middle,
            l2i4: TextAlignment::End,
[...]

PR #935, opened 2025-04-25
Fix alignment invalidation in `Label::layout`.

xStrom:
While investigating stacking alignment offsets as reported in #933 I noticed that the alignment invalidation logic in `Label::layout` is missing some key parts. This PR addresses that.

* Rename `alignment_changed` to `needs_alignment` to more accurately represent what is being tracked.
* Reset `needs_alignment` back to `false` after aligning.
* Check whether `available_width` has changed since last `layout`, as that requires new alignment.
* Clamp `available_width` to zero as negative values don't make sense but always happen due to the padding subtraction.

### DIFF
--- a/masonry_core/src/widgets/label.rs
+++ b/masonry_core/src/widgets/label.rs
@@ -55,8 +55,10 @@ pub struct Label {
 
     line_break_mode: LineBreaking,
     alignment: Alignment,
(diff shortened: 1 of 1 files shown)


Philipp-M (approved):
Looks good, thanks.

PR #936, opened 2025-04-25
Update to latest Parley version.

### LINKED ISSUE #933: Each `Label::layout` pass stacks another text alignment offset additively.

xStrom:
In order to deal with #933 I think we need some Parley changes. However, we're currently using `parley` v0.3.0 which is a bit different than what `main` looks like. So in anticipation of soon needing a Parley update I wanted to split the work and first make a PR that imports all the unrelated changes.

In addition to Parley being updated to the latest `main` revision, this PR also updates Winit and AccessKit to match what Parley expects.

The breaking AccessKit changes were trivial, just a few extra arguments to pass in.

Parley changes required a bit more work. Especially notable is that [vertical metrics are no longer rounded](https://github.com/linebender/parley/pull/297). The vertical metrics change alone caused half of Masonry's tests to fail. However, by doing glyph baseline rounding in Masonry [as insisted by Chad](https://github.com/linebender/parley/pull/297#issuecomment-2751450235) I got the failing test count down to 9. Those are caused by line spacing changes, which is expected behavior.

Here's an example of how baseline rounding improves pixel alignment (*1x scale factor original output, GIF has been been nearest neighbor upscaled*).

change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,15 +20,28 @@ checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
 name = "accesskit"
-version = "0.17.1"
+version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d3d3b8f9bae46a948369bc4a03e815d4ed6d616bd00de4051133a5019dc31c5a"
+checksum = "becf0eb5215b6ecb0a739c31c21bd83c4f326524c9b46b7e882d77559b60a529"
+
+[[package]]
+name = "accesskit_android"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0af881ef4f4996d3cfc7333622e4f2f059c6b5a6749e286b6d7dae7b1e00ad72"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "jni",
+ "log",
+ "once_cell",
+]
 
 [[package]]
 name = "accesskit_atspi_common"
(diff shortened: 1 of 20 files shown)


PoignardAzur (approved):
I'm not familiar with the APIs involved, but these changes all seem minor.

LGTM.

mwcampbell:
I'm going to need the AccessKit update for my own work, so would like to get this merged quickly. Thanks.

mwcampbell:
One question, now that I'm looking more closely: Why update accesskit_winit only to 0.24 and not the latest, 0.26?

xStrom:
This PR here probably won't be merged. I'm driving Parley in a different direction in [parley#344](https://github.com/linebender/parley/pull/344) and if that lands we won't need many of the changes I had to make in this PR here.

However, we can update Parley to the March 25 version, which will still get us the AccessKit update. I opened #941 to do that.

Regarding `accesskit_winit`, I did the bare minimum in this (and in #941) that was needed for Parley to limit the amount of work and the review burden. Once the update to 0.24 is done we can definitely see about updating to 0.26.

theoparis:
This branch also fixes https://github.com/linebender/parley/issues/340 for me on MacOS. Without this PR, xilem's main branch with masonry and xilem causes objc crashes 👀

xStrom:
As Parley changed I did it in a new PR - #953.

DJMcNab:
Closing this PR then, as it's been superseded. Obviously feel free to re-open if that's wrong

PR #937, opened 2025-04-27
Change filename scheme of screenshot tests

PoignardAzur:
Have all screenshot tests store their reference file in `<CRATE_ROOT>/screenshots/<TEST-NAME>.png`.
Move all existing screenshot tests accordingly.
Add namespace to most test names to avoid filename collisions.
Document the changes.
Remove serde dependency from `masonry_core`, as previously it was only used to look up the path to store the test in.


PoignardAzur:
This should help unblock #896, and get us closer to kompari integration.

PoignardAzur:
Also, note that I did a lot of regex renaming for expediency, so some of the test names might be a little redundant, although I fixed the worst offenders. That can be addressed case-by-case in follow-up PRs.

masonry_core/src/testing/harness.rs
```diff
@@ -139,24 +138,18 @@ pub struct TestHarnessParams {
 /// Assert a snapshot of a rendered frame of your app.
 ///
 /// This macro takes a test harness and a name, renders the current state of the app,
-/// and stores the render as a PNG next to the test, in a `./screenshots/` folder.
+/// and stores the render as a PNG next to the test, in a `<CRATE_ROOT>/screenshots/` folder.
```
DJMcNab, line 141: I don't think "next to" here is right.
```suggestion
/// and compares against the PNG file with the name in the folder `<CRATE_ROOT>/screenshots/` folder.
```

DJMcNab (approved):
One docs nit, but otherwise this looks good to me.

PR #938, opened 2025-04-28
Add Property trait and better document Property concept.

PoignardAzur:


masonry_core/src/doc/04b_widget_properties.md
DJMcNab, line 1: I'm surprised that this doesn't mention the `Property` trait now

masonry_core/src/core/properties.rs
```diff
@@ -3,8 +3,17 @@
...
+///
+/// See [properties documentation](crate::doc::doc_04b_widget_properties) for
+/// a full explanation of the general concept.
+///
+/// Note that if a type `Foobar` implements Property, that tells you that Foobar is meant
+/// to be a property of *some* widget, but it doesn't tell you *which* widget accepts Foobar
+/// as a property.
+/// That information is deliberately not encoded in the type system.
+/// We might change that in a future version.
+pub trait Property: Default + 'static {}
```
DJMcNab, line 16: Why `Default`? That superbound seems to be unused.

DJMcNab (commented):
Small nits, but this looks good, thanks.

masonry_core/src/core/properties.rs
PoignardAzur: Future changes.

DJMcNab: I don't think it's good practise to add it in this PR, but I won't block on it.

DJMcNab (approved):
With reluctance because of the semantic change.

masonry_core/src/core/properties.rs
```diff
@@ -3,8 +3,17 @@
 
 use anymap3::{AnyMap, Entry};
 
-// TODO - Add PropertyValue wrapper struct that implements receiver trait.
-// Return PropertyValue<T> instead of Option<T> from methods.
+/// A marker trait that indicates that a type is intended to used as a widget's property.
```
waywardmonkeys, line 6: "to be used"

PR #939, opened 2025-04-28
[WIP] Add Action associated type to Widget trait

PoignardAzur:


masonry_core/src/core/action.rs
```diff
@@ -24,10 +29,40 @@ pub enum Action {
...
+///
+/// Used by widgets which never emit actions.
+///
+/// Will be replaced with `!` once stable.
+pub enum NoAction {}
+
+/// Trait alias for traits [`Debug`] and [`Any`].
+///
+/// Used to be able to declare `Box<dyn DebugAndAny>`.
+pub trait DebugAndAny: Debug + Any {}
```
DJMcNab, line 45: I'd consider calling this `AnyAction`. That opens the door to future trait bounds without needing to rename all uses of this.

(That name suggestion is based on this being exactly the same idea as https://github.com/linebender/xilem/blob/825cfd88ebba3c7c838921a9755fe233c19eec50/xilem_core/src/message.rs#L44-L123; you may as well import similar tests and utility methods as well)

masonry_core/src/core/action.rs
```diff
@@ -2,18 +2,23 @@
...
+#[derive(Debug)]
+pub struct Action(pub Box<dyn DebugAndAny + Send>, pub WidgetId);
+
 // TODO - TextCursor changed, ImeChanged, EnterKey, MouseEnter
 #[non_exhaustive]
 /// Events from UI elements.
 ///
 /// Note: Actions are still a WIP feature.
-pub enum Action {
+pub enum DefaultAction {
```
DJMcNab, line 21: Can we document that this is an intermediate state. I think for reviewing, it wouldn't be unreasonable to leave this as-is, and then split it out as a follow-up

DJMcNab (commented):
Super excited for this!

masonry_core/src/core/action.rs
PoignardAzur: Yeah, the point where I saw AnyMessage was when I decided to just commit what I had and stop.

My plan is to create a clean slate PR that moves AnyMessage to Masonry, and when it's merged, rebase this PR on top of it.

PoignardAzur:
For the record, this is *extremely* incomplete code (I just committed what I had when I stopped working) and it's not super high-priority, so I might leave it as it for a while.

PoignardAzur:
Superseded by #1212.

PR #940, opened 2025-04-29
Update to `keyboard-types` 0.8

waywardmonkeys:
With W3C specs, I think we should be mapping meta / super differently than we were before. This will change again with winit 0.31 where the names of those keys on the winit side will have changed. (And also, much of this particular mapping won't be necessary.)

The bulk of this change is that in `keyboard-types` 0.8, the `Key` enumeration has changed to have a separate enumeration for `NamedKey` values.


masonry/src/convert_winit_event.rs
```diff
@@ -115,9 +114,11 @@ pub(crate) fn winit_physical_key_to_kbt_code(physical_key: WinitPhysicalKey) ->
         WinitPhysicalKey::Unidentified(_) => Code::Unidentified,
         WinitPhysicalKey::Code(key_code) => match key_code {
             // Variants that don't match 1:1
+            // With winit 0.31, these are renamed to swap the meta and super to match the W3C specs.
+            #[expect(deprecated, reason = "Map even the deprecated keys.")]
             WinitKeyCode::Meta => Code::Super,
-            WinitKeyCode::SuperLeft => Code::Super,
-            WinitKeyCode::SuperRight => Code::Super,
+            WinitKeyCode::SuperLeft => Code::MetaLeft,
+            WinitKeyCode::SuperRight => Code::MetaRight,
```
waywardmonkeys, line 121: @madsmtm This part is the main part that I'm sure about here.

madsmtm: I think it's correct. Note that `WinitKeyCode::Meta` isn't actually emitted in Winit v0.30.x as far as I know.

See also https://github.com/rust-windowing/winit/pull/4018.

masonry/src/convert_winit_event.rs
```diff
@@ -318,316 +321,320 @@ pub(crate) fn winit_physical_key_to_kbt_code(physical_key: WinitPhysicalKey) ->
...
+            WinitNamedKey::NumLock => Key::Named(NamedKey::NumLock),
+            WinitNamedKey::ScrollLock => Key::Named(NamedKey::ScrollLock),
+            WinitNamedKey::Shift => Key::Named(NamedKey::Shift),
+            WinitNamedKey::Symbol => Key::Named(NamedKey::Symbol),
+            WinitNamedKey::SymbolLock => Key::Named(NamedKey::SymbolLock),
+            WinitNamedKey::Meta => Key::Named(NamedKey::Meta),
+            #[expect(deprecated, reason = "Map even the deprecated keys.")]
+            WinitNamedKey::Hyper => Key::Named(NamedKey::Hyper),
+            #[expect(deprecated, reason = "Map even the deprecated keys.")]
+            WinitNamedKey::Super => Key::Named(NamedKey::Super),
```
madsmtm, line 342: This part is wrong probably? Should be:
```suggestion
            WinitNamedKey::Meta => Key::Named(NamedKey::Super),
            #[expect(deprecated, reason = "Map even the deprecated keys.")]
            WinitNamedKey::Hyper => Key::Named(NamedKey::Hyper),
            #[expect(deprecated, reason = "Map even the deprecated keys.")]
            WinitNamedKey::Super => Key::Named(NamedKey::Meta),
```

(If you want to match Winit `v0.31` that is)

waywardmonkeys: But we're still using winit 0.30 ... so I'm not sure. Once 0.31 hits and we update to that ... well, then this won't be needed ideally since we'd both be using `keyboard-types`.

madsmtm: I'm not that familiar with how Xilem handles keyboard input, but I guess, it kinda depends on how you want your users to experience things? I think you'd want them to use `keyboard_types::NamedKey::Meta` in their code, right?

xorgy (approved):

PR #941, opened 2025-04-29
Update to Parley 2025-03-25.

xStrom:
With #936 stalling but Matt wanting AccessKit changes I decided to split out parts of #936 into this PR. This brings in Parley from March 25th, 2025 and that gets us the AccessKit update.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,9 +92,11 @@ tree_arena = { version = "0.1.0", path = "tree_arena" }
 vello = "0.4.0"
 wgpu = "23.0.1"
 kurbo = "0.11.1"
-parley = { version = "0.3.0", features = ["accesskit"] }
+parley = { git = "https://github.com/linebender/parley", rev = "eabfb8175848183dde2516d562c6b7eab7f0df84", features = [
+    "accesskit",
+] }
 peniko = "0.3.1"
-winit = "0.30.4"
+winit = "0.30.9"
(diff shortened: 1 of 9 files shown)


DJMcNab (approved):
Seems reasonable

Cargo.toml
```diff
@@ -92,18 +92,20 @@ tree_arena = { version = "0.1.0", path = "tree_arena" }
...
 smallvec = "1.13.2"
 hashbrown = "0.15.2"
 dpi = "0.1.1"
 image = { version = "0.25.2", default-features = false }
 web-time = "1.1.0"
 bitflags = "2.6.0"
-accesskit = "0.17.0"
-accesskit_winit = "0.23.0"
+accesskit = "0.18.0"
+accesskit_winit = "0.24.0"
```
DJMcNab, line 108: This update means that we no longer support `NativeActivity`, an unfortunate regression. We panic at this code.
https://github.com/AccessKit/accesskit/blob/10f4bd1344b98ebb2baf969e62afdd767f5ff1d1/platforms/winit/src/platform_impl/android.rs#L31-L39. It looks like the non-support for NativeActivity is intentional here.

cc @mwcampbell what do you suggest doing in this case? Is there a no-op AccessKit winit backend we can use, or is the only solution to conditionally compile out the accessibility support on Android if we want to support Native Activity?

PR #942, opened 2025-04-29
Remove default padding from Prose widget.

PoignardAzur:
The codebase currently has too many redundant layers of padding around text, spread across multiple widgets.

This PR is the first step towards centralizing how padding is handled.


masonry_core/screenshots/prose_clipping.png
DJMcNab, line 1: Either this snapshot was always wrong, or is newly wrong. But the null hypothesis is that it's newly wrong without a good explanation otherwise.

masonry_core/screenshots/prose_alignment_flex.png
DJMcNab, line 1: This change is also confusing - I'm not sure what changing this achieves

masonry_core/screenshots/prose_clipping.png
PoignardAzur: My bad, I shouldn't have missed that.

The problem is that SizedBox's size is overridden by parent constraints, and when I removed the flex the min constraint became "the test window's size" instead of zero.

masonry_core/screenshots/prose_alignment_flex.png
PoignardAzur: I was adding padding for visibility, but in retrospect, I think having no padding at all makes the behavior clearer.

DJMcNab (approved):
I'm not hugely happy with these changes - I think it makes what we're exporting harder to use with unclear benefit. But if the properties experiment is to have any chance at succeeding, it needs to keep moving.

PR #943, opened 2025-04-29
Fix textbox test names

PoignardAzur:


DJMcNab (approved):
I think I'd prefer if we went the other way. That is, these should be `textbox_textbox_selection_textbox_textbox` /s

PoignardAzur:
Jokes aside, getting rid of the insta naming scheme has made browsing through screenshot files so much easier. It's night and day.

PR #944, opened 2025-04-29
Change how TextBox handles its padding

PoignardAzur:
Remove `with_padding_if_default` method and all related code.
Remove concept of "unset" padding.

Depends on #942.


DJMcNab (approved):
Approving as in the other PR. The same review comments from there also stand

PR #945, opened 2025-04-30
Refactor example tests to run more efficiently by default

PoignardAzur:


masonry/Cargo.toml
```diff
@@ -55,33 +55,5 @@ accesskit.workspace = true
...
-# This actually enables scraping for all examples, not just this one.
-# However it is possible to set doc-scrape-examples to false for other specific examples.
-doc-scrape-examples = true
-
-[[example]]
-name = "to_do_list"
-test = true
-
-[[example]]
-name = "two_textboxes"
```
DJMcNab, line 86: This could have a non-executable crate type, I think

masonry/examples/tests.rs
```diff
@@ -0,0 +1,23 @@
...
+
+fn main() {
+    println!("This example is only used to compile other examples.");
+}
+
+#[cfg(test)]
+#[allow(dead_code, reason = "We don't need to run the main functions.")]
+#[allow(missing_docs, reason = "Example code doesn't need docs.")]
+#[path = "."]
+pub mod others {
```
DJMcNab, line 17: If you run `cargo nextest run --examples`, do the versions of the tests in this example race with the versions in the main examples.

That is, is `assert_render_snapshot` written to be process safe, in both the success and failure cases?

masonry/examples/tests.rs
DJMcNab, line 1: I'm still not a massive fan of this being an example, and would strongly prefer it to be an integration test. The fact that we have a null main function belies that this is not meaningful as an example

DJMcNab (approved):
A few details, none of them sufficiently blocking. Approving reluctantly, because otherwise we'll be at an impasse with regards to the integration test thing.

masonry/examples/tests.rs
DJMcNab: Yeah, I don't think it's too important. This is not a blocking concern, just something that comes up

PoignardAzur: I've changed the code structure, this should no longer be a problem.

DJMcNab (approved):

PR #946, opened 2025-04-30
Lay the ground for multi-window feature

PoignardAzur:
Shuffle RenderRoot fields around.
Create WindowRef and WindowMut types.


PoignardAzur:
Leaving it as a draft for now because I'm not satisfied with the code structure, and the added complexity.

PoignardAzur:
The more I think about it, the more I believe the one-RenderRoot-per-window architecture is better. Closing this.

PR #947, opened 2025-05-02
Rewrite marker comments across multiple lines.

PoignardAzur:
This is because VSCode now interprets dashes as part of the marker title, which makes markers more likely to overflow the minimap width.

### DIFF
--- a/masonry/examples/calc_masonry.rs
+++ b/masonry/examples/calc_masonry.rs
@@ -424,7 +424,9 @@ fn main() {
     .unwrap();
 }
 
-// --- MARK: TESTS ---
+// ---
+// MARK: TESTS
+// ---
 #[cfg(test)]
 mod tests {
     use masonry::assert_render_snapshot;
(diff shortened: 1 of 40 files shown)


PoignardAzur:
Last time we had a markers-related PR, people were asking if markers were really useful for browsing code.

After a year of using them in Masonry, my takeaway is, yes, they're genuinely useful with VSCode's minimap.

DJMcNab (commented):
Is an explanation available for why these need to be bulked out so much? I don't like the empty noise of 250 repetitions of `// ---`

PoignardAzur:
Fair enough, that's an aesthetic choice on my part.

I think having big `// ---` blocks separating huge blocks of code makes the code easier to read on e.g. Github or a terminal pager.

xStrom:
I'm also not a fan of having more decorative lines everywhere.

I would be much more in favor of just addressing the issue directly and not adding more lines.

```diff
- // --- MARK: BUILDERS ---
+ // --- MARK: BUILDERS
```

Removing the suffix dashes seems to be enough to fix the minimap in my VSCode. The prefix dashes can remain to make the markers more easily recognizable.

PR #950, opened 2025-05-07
Use ui-events for keyboard and pointer events.

xorgy:
- Remove file hover/drop events from `PointerEvent` because they were never mapped and could never be mapped (since they relied on pointer positions which are not available in file hover on winit.

 - Remove modifier change events from `TextEvent`, since nobody uses them anywhere, and new `PointerEvent` and `KeyboardEvent` include them.

 - Remove pointer state from events that are unlikely to have new state when they occur (enter/leave/cancel).

 - Remove redundant default widget event handlers and imports.


masonry/examples/calc_masonry.rs
```diff
@@ -15,16 +15,17 @@
 use accesskit::{Node, Role};
 use masonry::app::{AppDriver, DriverCtx};
 use masonry::core::{
-    AccessCtx, AccessEvent, Action, BoxConstraints, EventCtx, LayoutCtx, PaintCtx, PointerEvent,
-    PropertiesMut, PropertiesRef, QueryCtx, RegisterCtx, StyleProperty, TextEvent, Update,
-    UpdateCtx, Widget, WidgetId, WidgetPod,
+    AccessCtx, AccessEvent, Action, BoxConstraints, EventCtx, LayoutCtx, PaintCtx, PropertiesMut,
```
waywardmonkeys, line 18: I think it would be better for consumers of the crate if they got `masonry::core::PointerEvent` and such still and that we didn't change that re-export. Having examples (and user apps) have to separately depend on `ui_events` doesn't seem great here.

(True for several other places below.)

masonry_core/src/core/action.rs
```diff
@@ -15,7 +15,7 @@ use crate::core::PointerButton;
 /// Note: Actions are still a WIP feature.
 pub enum Action {
     /// A button was pressed.
-    ButtonPressed(PointerButton),
+    ButtonPressed(Option<PointerButton>),
```
waywardmonkeys, line 18: This needs an explanation in the doc comment as to what it means if this is `None`.

masonry/examples/calc_masonry.rs
xorgy: I think the problem with that is the number of different things off of `pointer` you would want with `PointerEvent`... do these all go in the core namespace as well?

waywardmonkeys: I think for right now, you aren't using those things, so the first pass is to just keep things as they are and do the bare conversion.

afterwards, I think there's value in basically re-exporting `ui_events::pointer` as `masonry::core::events::pointer` and same for keyboard and so on ... then the types are "natively" available within Masonry.

cc: @DJMcNab @PoignardAzur

PoignardAzur: I'm in favor of anything that keeps the diff short. It's a pretty large PR.

On the long term, I don't know. I think Masonry and Xilem so far have had a "batteries included" approach where we re-export of a lot of things, and I'm not sure it's always the best for developer experience; but as long as we're not changing that, re-exporting ui_events stuff from Masonry would make sense.

(I would avoid deeply nested modules, though.)

xorgy: Alright, I'll re-export with existing names where it reduces diff size.

xorgy: I've gone ahead and re-exported the most common pointer and keyboard stuff right on core, and then re-exported the `pointer` and `keyboard` modules on there as well.

PR #952, opened 2025-05-08
Remove screenshots from online docs export.

PoignardAzur:


masonry_core/src/testing/screenshots.rs
```diff
@@ -5,7 +5,9 @@
 
 use image::{GenericImageView as _, Pixel as _, Rgb, RgbImage};
 
-#[cfg(docsrs)]
+// TODO - Re-enable this once we find a way to load screenshots that doesn't go against our
+// storage quotas.
+#[cfg(FALSE)]
```
DJMcNab, line 10: https://github.com/rust-lang/rust/pull/138632

`cfg(false)` should be stable in 6 weeks!

DJMcNab (approved):
Seems like a reasonable solution. We should check that this doesn't stop docs building (our docs.rs check doesn't actually work for this macro, as I've discovered by building the docs locally)

DJMcNab:
We should document this in ARCHITECTURE.md as well

masonry_core/src/testing/screenshots.rs
```diff
@@ -19,6 +21,21 @@ macro_rules! include_screenshot {
     };
 }
 
+#[cfg(docsrs)]
```
xStrom, line 24: Should make this unconditional now.

```suggestion
```

DJMcNab (approved):

PR #957, opened 2025-05-08
Remove mentions of "leading"/"trailing" and references to RTL handling.

PoignardAzur:
Rename "leading" to "left", "trailing" to "right", remove any mentions of RTL, etc.
See https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Remove.20bidi.20references.20from.20Masonry.20code/with/515690264 for rationale.

Merge Padding structs into one.


DJMcNab (commented):
I still think we need some indication we can point to that handling RTL "properly" is in-scope, but has been deferred.

DJMcNab (approved):
LGTM.
To answer:

> I haven't gone deeply into this issue, but I agree with Daniel. "Leading" and "trailing" are usually the correct concepts. I understand YAGNI and trying to keep things simple, but a goal to keep in mind right now is to minimize the pain and disruption for when we do properly support BiDi.

We went through all the stages of this in the linked thread. Adding the mentioned note (and maybe opening a GitHub issue...), was the conclusion we reached. Essentially, we don't think that these newly added fields would be deprecated/removed when we start to handle BiDi, only expanded, and we don't handle them properly at the moment anyway. So indicating that BiDi *is* absolutely in-scope, but isn't yet supported is the right way forward, which is what this PR does

PR #959, opened 2025-05-08
Remove type parameter from RootWidget

PoignardAzur:


DJMcNab (approved):
Approving, assuming that the Xilem changes have been properly validated

PR #960, opened 2025-05-08
Add shadow property.

PoignardAzur:


jaredoconnell (approved):
Seems like a reasonable first implementation. The TODOs definitely need to be addressed at some point. It may make sense to have some of the TODOs moved to help wanted issues on Github.
I hope that we don't end up needing to do much widget-specific implementation of each property. This property will be useful in many widgets.

PR #961, opened 2025-05-08
Add default property values and a default theme.

PoignardAzur:
Add method in Property trait that returns a static reference to a default value.
Make property getters return non-optional values.

Create DefaultProperties type.
Create default set of properties in `theme` module.
Add options to override that set of properties by passing DefaultProperties at startup.
Add some properties to RootWidget.
Change background color in calc example.

### DIFF
--- a/masonry/src/core/widget_mut.rs
+++ b/masonry/src/core/widget_mut.rs
@@ -52,23 +52,25 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
     }
 
     /// Returns `true` if the widget has a property of type `T`.
-    pub fn get_prop<T: Property>(&self) -> Option<&T> {
-        self.ctx.properties.get::<T>()
-    }
-
-    /// Get value of property `T`, or `None` if the widget has no `T` property.
+    ///
+    /// Does not check default properties.
     pub fn contains_prop<T: Property>(&self) -> bool {
         self.ctx.properties.contains::<T>()
     }
 
-    /// Get value of property `T`, or `None` if the widget has no `T` property.
-    pub fn get_prop_mut<T: Property>(&mut self) -> Option<&mut T> {
-        self.widget
-            .property_changed(&mut self.ctx.update_mut(), TypeId::of::<T>());
-        self.ctx.properties.get_mut::<T>()
+    /// Get value of property `T`.
+    ///
+    /// Returns Some if either the widget or the default property set has an entry for `T`.
+    /// Returns `None` otherwise.
+    pub fn get_prop<T: Property>(&self) -> Option<&T> {
+        self.ctx.properties.get::<T>()
     }
 
     /// Set property `T` to given value. Returns the previous value if `T` was already set.
+    ///
+    /// Does not affect default properties.
+    ///
+    /// This also calls [`Widget::property_changed`] with the matching type id.
     pub fn insert_prop<T: Property>(&mut self, value: T) -> Option<T> {
         self.widget
             .property_changed(&mut self.ctx.update_mut(), TypeId::of::<T>());
@@ -76,6 +78,10 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
     }
 
     /// Remove property `T`. Returns the previous value if `T` was set.
+    ///
+    /// Does not affect default properties.
+    ///
+    /// This also calls [`Widget::property_changed`] with the matching type id.
     pub fn remove_prop<T: Property>(&mut self) -> Option<T> {
         self.widget
             .property_changed(&mut self.ctx.update_mut(), TypeId::of::<T>());
--- a/masonry/src/core/widget_mut.rs
+++ b/masonry/src/core/widget_mut.rs
@@ -67,26 +67,28 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
         self.ctx.properties.get::<T>()
     }
 
-    /// Set property `T` to given value. Returns the previous value if `T` was already set.
+    /// Set property `T` to given value. Returns the previous value if `T` was already set locally.
     ///
     /// Does not affect default properties.
     ///
     /// This also calls [`Widget::property_changed`] with the matching type id.
     pub fn insert_prop<T: Property>(&mut self, value: T) -> Option<T> {
+        let value = self.ctx.properties.insert(value);
         self.widget
             .property_changed(&mut self.ctx.update_mut(), TypeId::of::<T>());
-        self.ctx.properties.insert(value)
+        value
     }
 
-    /// Remove property `T`. Returns the previous value if `T` was set.
+    /// Remove property `T`. Returns the previous value if `T` was set locally.
     ///
     /// Does not affect default properties.
     ///
     /// This also calls [`Widget::property_changed`] with the matching type id.
     pub fn remove_prop<T: Property>(&mut self) -> Option<T> {
+        let value = self.ctx.properties.remove::<T>();
         self.widget
             .property_changed(&mut self.ctx.update_mut(), TypeId::of::<T>());
-        self.ctx.properties.remove::<T>()
+        value
     }
 
     /// Set the local transform of this widget.
--- a/masonry/src/core/widget_mut.rs
+++ b/masonry/src/core/widget_mut.rs
@@ -51,7 +51,7 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
         }
     }
 
-    /// Returns `true` if the widget has a property of type `T`.
+    /// Returns `true` if the widget has a local property of type `T`.
     ///
     /// Does not check default properties.
     pub fn contains_prop<T: Property>(&self) -> bool {
--- a/masonry/src/core/widget_mut.rs
+++ b/masonry/src/core/widget_mut.rs
@@ -60,9 +60,10 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
 
     /// Get value of property `T`.
     ///
-    /// Returns Some if either the widget or the default property set has an entry for `T`.
-    /// Returns `None` otherwise.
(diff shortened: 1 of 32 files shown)


DJMcNab (commented):
This concerns me - setting global defaults is of course not at all composable. What problem is this trying to solve? 

To be clear, having properties themselves have inherent defaults seems reasonably uncontroversial, and I'm happy enough to have that

PoignardAzur:
> setting global defaults is of course not at all composable.

To be clear, this is what we were already doing; except we were setting global defaults per-widget with no way to override them.

Long term, the idea would be to have full cascading with classes, pseudo-classes, etc so that you can deeply customize the look of your app.

As a stopgap solution, this essentially gives us cascading between only three layers: local value, global theme, and `Default::default()`.

masonry/examples/calc_masonry.rs
```diff
@@ -22,10 +24,13 @@ use masonry::core::{
 use masonry::dpi::LogicalSize;
 use masonry::kurbo::{Point, Size};
 use masonry::peniko::Color;
+use masonry::properties::{Background, Padding};
+use masonry::theme::default_property_set;
 use masonry::widgets::{Align, CrossAxisAlignment, Flex, Label, RootWidget, SizedBox};
 use smallvec::{SmallVec, smallvec};
 use tracing::{Span, trace, trace_span};
 use vello::Scene;
+use vello::peniko::color::AlphaColor;
```
waywardmonkeys, line 33: This should be `use masonry::peniko::color::AlphaColor`.

masonry_core/src/app/render_root.rs
```diff
@@ -170,6 +173,9 @@ pub enum WindowSizePolicy {
 
 /// Options for creating a [`RenderRoot`].
 pub struct RenderRootOptions {
+    /// Default values that Properties will have if not defined per-widget.
```
waywardmonkeys, line 176: I suspect that `Properties` needn't be capitalized (or it should be a valid, linked identifier).

masonry_core/src/core/properties.rs
```diff
@@ -71,38 +91,55 @@ impl Properties {
 
 impl PropertiesRef<'_> {
     /// Returns `true` if the widget has a property of type `P`.
```
waywardmonkeys, line 93: Sounds like we should have a term for this that we use consistently, like "local property" (this is a term in another system so it would be a good choice).

masonry_core/src/core/properties.rs
```diff
@@ -71,38 +91,55 @@ impl Properties {
...
 
-    /// Get value of property `P`, or `None` if the widget has no `P` property.
-    pub fn get<P: Property>(&self) -> Option<&P> {
-        self.map.get::<P>()
+    /// Get value of property `P`.
+    ///
+    /// If the widget has an entry for `P`, returns that entry.
+    /// If the default property set has an entry for `P`, returns that entry.
+    /// Otherwise returns [`Property::static_default()`].
+    pub fn get<P: Property>(&self) -> &P {
```
waywardmonkeys, line 105: This is more of a chained lookup than a `get`. Maybe rename.

masonry_core/src/core/properties.rs
```diff
@@ -71,38 +91,55 @@ impl Properties {
...
+        } else {
+            P::static_default()
+        }
     }
 }
 
 impl PropertiesMut<'_> {
     /// Returns `true` if the widget has a property of type `P`.
+    ///
+    /// Does not check default properties.
```
waywardmonkeys, line 119: Again, "local property"

masonry_core/src/core/properties.rs
```diff
@@ -112,26 +149,46 @@ impl PropertiesMut<'_> {
...
     pub fn reborrow_mut(&mut self) -> PropertiesMut<'_> {
         PropertiesMut {
             map: &mut *self.map,
+            default_map: self.default_map,
         }
     }
 }
+
+impl DefaultProperties {
+    /// Create an empty set with no default values.
```
waywardmonkeys, line 171: This isn't a "set".

masonry_core/src/core/widget_ref.rs
```diff
@@ -88,20 +87,23 @@ impl<'w, W: Widget + ?Sized> WidgetRef<'w, W> {
     }
 
     /// Returns `true` if the widget has a property of type `T`.
-    pub fn get_prop<T: Property>(&self) -> Option<&T> {
```
waywardmonkeys, line 91: A local property, right?

waywardmonkeys (commented):
A start on some comments ...

masonry_core/src/core/properties.rs
PoignardAzur: Any name suggestions?

PoignardAzur:
@DJMcNab Do you have blocking concerns re: global style default (e.g. we shouldn't provide them)? I think otherwise we're good to merge.

DJMcNab:
I'm not a huge fan of it in a "from first principles way". But no, no blocking concerns. I don't have time to review this now, though.

masonry/src/core/widget_mut.rs
```diff
@@ -51,31 +51,38 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
...
+    ///
+    /// If the widget has an entry for `P`, returns that entry.
+    /// If the default property set has an entry for `P`, returns that entry.
+    /// Otherwise returns [`Property::static_default()`].
+    pub fn get_prop<T: Property>(&self) -> &T {
+        self.ctx.properties.get::<T>()
     }
 
     /// Set property `T` to given value. Returns the previous value if `T` was already set.
+    ///
```
waywardmonkeys, line 71: This is only returning a previous value if it was set locally, right?

masonry/src/core/widget_mut.rs
```diff
@@ -51,31 +51,38 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
...
+    /// Does not affect default properties.
+    ///
+    /// This also calls [`Widget::property_changed`] with the matching type id.
```
waywardmonkeys, line 74: It is confusing that this calls the notification prior to the value being set ...

waywardmonkeys: XAML's `DependencyObject` has a `DependencyPropertyChangedEventArgs` which gets the old and new values.

masonry/src/core/widget_mut.rs
```diff
@@ -51,31 +51,38 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
...
         self.widget
             .property_changed(&mut self.ctx.update_mut(), TypeId::of::<T>());
         self.ctx.properties.insert(value)
     }
 
     /// Remove property `T`. Returns the previous value if `T` was set.
+    ///
+    /// Does not affect default properties.
+    ///
+    /// This also calls [`Widget::property_changed`] with the matching type id.
```
waywardmonkeys, line 85: Same thing with the args to the callback. I know that's not part of this PR.

masonry/src/properties/border_color.rs
```diff
@@ -13,17 +13,22 @@ pub struct BorderColor {
...
 // TODO - The default border color in CSS is `currentcolor`,
 // the color text is displayed in.
 // Do we want to implement that?
 
 impl Default for BorderColor {
     fn default() -> Self {
-        Self {
-            color: AlphaColor::from_rgba8(0, 0, 0, 0),
-        }
+        *Self::static_default()
```
waywardmonkeys, line 31: Why's this different from the one for background? (I mean different in terms of a clone vs a `*`.)

masonry/src/properties/background.rs
```diff
@@ -19,12 +19,17 @@ pub enum Background {
     Gradient(Gradient),
 }
 
-impl Property for Background {}
+impl Property for Background {
+    fn static_default() -> &'static Self {
+        // This matches the CSS default.
+        const DEFAULT: Background = Background::Color(AlphaColor::TRANSPARENT);
```
waywardmonkeys, line 25: Elsewhere this is a `static` and here a `const` ... I think this should be `static` ?

waywardmonkeys: but also, this is a super cheap value, so I don't know what the gain here is.

masonry/src/properties/box_shadow.rs
```diff
@@ -77,8 +82,20 @@ impl BoxShadow {
         ctx.request_layout();
     }
 
+    /// Returns false if the shadow can be safely treated as non-existent.
```
waywardmonkeys, line 85: Should have backticks around "false" like we do in other places.

masonry/src/properties/border_color.rs
PoignardAzur: Background doesn't implement Copy.

masonry/src/properties/background.rs
PoignardAzur: > Elsewhere this is a static and here a const ... I think this should be static ?

`const` and `static` are essentially equivalent here, since the compiler will store `const` values in statics anyway. But yeah, it worth keeping the code consistent.

> but also, this is a super cheap value, so I don't know what the gain here is.

Basically, the line of reasoning is:

- We want `Properties[/Ref/Mut]::get()` to return a `T`, not an `Option<T>`.
- There is no `Default` impl for `&T`. So we need a `Property::static_default()` method that will return a `&'static T` we can use anywhere.
- The way we do that is by creating a static inside each impl, and returning a reference to that static.

I guess I could document that.

masonry/src/core/widget_mut.rs
waywardmonkeys: Wouldn't help. Some things need the old value, some need the new. This is a future concern though.

waywardmonkeys (approved):
I think that this is an improvement over a static set of globals. It isn't the end destination by any means and there's lots to improve, but this is a step along the way.

PR #962, opened 2025-05-08
Implement PartialEq, Send and Sync for all Properties

PoignardAzur:
Create alias for `anymap3::Map<dyn Any + Send + Sync>`.

PR #964, opened 2025-05-08
Add traits for setting widget properties from xilem elements

PoignardAzur:
Add Style, HasProperty and PropertyTuple trait.

Have Buttom element using Button widget styling properties.
Add some styles to calc example.


PoignardAzur:
Builds on top of #961 and #962.

xilem/src/property_tuple.rs
```diff
@@ -0,0 +1,122 @@
...
+                };
+                prop.downcast_mut().unwrap()
+            }
+
+        }
+
+    };
+}
+
+// impl_property_tuple!(P0, 0);
```
DJMcNab, line 111: This needs a comment explaining why it's been brought out seperately

xilem/src/style.rs
DJMcNab, line 1: I'm finding it hard to parse out what's going on here. My best interpretation of what's going on is that you've re-invented struct fields using a linear search through a tuple.

Is there some generic functionality you're planning to build on top of this infrastructure, to make use of the added abstraction?

xilem/src/style.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use masonry::core::Property;
+use masonry::properties::types::Gradient;
+use masonry::properties::{Background, BorderColor, BorderWidth, BoxShadow, CornerRadius, Padding};
+use vello::peniko::Color;
+
+use crate::property_tuple::PropertyTuple;
+
+/// Marker trait implement by elements to signal that a given property can be set on them.
+pub trait HasProperty<P: Property> {}
```
DJMcNab, line 14: I feel like the natural implementation would have this trait contain a method to get an exclusive reference to the property's value. It's trivial to write a macro to do that, although the DX is a bit ugly without the index metavariable.

xilem/src/view/button.rs
```diff
@@ -117,6 +153,8 @@ where
         ctx: &mut ViewCtx,
         mut element: Mut<Self::Element>,
     ) {
+        self.properties
+            .rebuild_properties(&prev.properties, &mut element);
```
DJMcNab, line 157: I like how hard this makes it to miss a property.

xilem/src/view/button.rs
```diff
@@ -89,10 +94,40 @@ pub struct Button<F> {
...
+        &mut self.properties
+    }
+}
+
+impl<F> HasProperty<Background> for Button<F> {}
+impl<F> HasProperty<BorderColor> for Button<F> {}
+impl<F> HasProperty<BorderWidth> for Button<F> {}
+impl<F> HasProperty<BoxShadow> for Button<F> {}
+impl<F> HasProperty<CornerRadius> for Button<F> {}
+impl<F> HasProperty<Padding> for Button<F> {}
```
DJMcNab, line 129: I do worry that it would be very easy to miss a property here. My suggested alternative does resolve that concern, if the final impl were macro based.

DJMcNab (commented):
Initial reaction: I think the tuple based approach is fantastic; it makes it really hard for view implementations to miss new properties.

However, I'm not convinced by use of linear search for fields, especially when the indices are known statically. I get the appeal of wanting to reduce boilerplate here, but we do also at least want to have an eye towards performance.
I think I'd prefer an implementation which makes slightly different tradeoffs here, of requiring more detail in the impls to significantly cut down on the amount of code gen and required operations.

PoignardAzur:
I'll admit this design is very unprincipled, and was just the result of me trying to get something, anything to let Xilem elements use properties without copy-pasting everything a hundred times.

I agree that the tuple-based approach ends up great for the job, and I'm mostly satisfied with it.

I agree that a macro-based solution would be less error-prone, so I'll push one today. I'm not super happy about that, because asking users to call macros when they implement elements is bad DX, but at least the macro won't do anything you couldn't do by hand.

PoignardAzur:
I hate Rust macros so much, but at least now it works.

xilem/src/property_tuple.rs
```diff
@@ -0,0 +1,182 @@
...
+    )
+    =>
+    {
+        impl $( <$($Args,)*> )? $crate::style::HasProperty<$Type> for $Self $( <$($Args),*> )? {
+            fn property(&mut self) -> &mut Option<$Type> {
+                    &mut self.properties().$idx
+            }
+        }
+    };
+}
```
DJMcNab, line 182: My macro-foo isn't great - I'm not sure why these two macros exist. I get the impression that they both could be inlined?

xilem/src/style.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Trait implemented by most elements that lets you set some styling properties.
+///
+/// Which methods you can use will depend on which marker traits the element implements,
+/// which matches which [`Properties`](masonry_winit::core::Properties) the underlying widget handles.
+pub trait Style: Sized {
+    /// The tuple type used by the element to store properties.
+    type Props: PropertyTuple;
+
+    /// Return a mutable reference to the element's property storage.
+    fn properties(&mut self) -> &mut Self::Props;
```
DJMcNab, line 30: If I were designing this, I probably wouldn't have had this, and instead provided the tuple field to the macro. But I won't block on that; there are tradeoffs either way.

xilem/src/style.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn property(&mut self) -> &mut Option<P>;
+}
+
+/// Trait implemented by most elements that lets you set some styling properties.
+///
+/// Which methods you can use will depend on which marker traits the element implements,
+/// which matches which [`Properties`](masonry_winit::core::Properties) the underlying widget handles.
+pub trait Style: Sized {
+    /// The tuple type used by the element to store properties.
+    type Props: PropertyTuple;
```
DJMcNab, line 27: I don't think you actually use that this is `PropertyTuple` anywhere

DJMcNab (commented):
Some comments; nothing critically broken any more. Not approving until after the rebase

xilem/src/style.rs
PoignardAzur: Overall I want to do as little as I can inside the macro. Macros are bad DX in general. Compare:

```rust
crate::declare_property_tuple!(
    ButtonProps;
    Button<F>;
    self.properties;

    Background, 0;
    BorderColor, 1;
    BorderWidth, 2;
    BoxShadow, 3;
    CornerRadius, 4;
    Padding, 5;
);
```

with

```rust
impl<F> Style for Button<F> {
    type Props = ButtonProps;

    fn properties(&mut self) -> &mut Self::Props {
        &mut self.properties
    }
}
```

Someone reading the code will have an easier time understanding what `self.properties` means and why it's there in the trait version. Error messages will be clearer, etc.

xilem/src/property_tuple.rs
DJMcNab: Hmm. I'm not going to claim that I completely follow, but I believe your claim. (I'm struggling to follow what two things you're looping over, I only see the one loop), but I haven't dug into it thoroughly.

In this kind of case, I'd expect them to just be different branches of the same macro. E.g.:
```rust
macro_rules! impl_for_all {
    ([$($ty:ty),*], $second_arg:tt) => {
        $(
            impl_for_all!(@macro_implementation_detail: $ty, $second_arg);
        )*
    }
    (@macro_implementation_detail: $ty:ty, [$($name:ident),*]) => {
            $(
                impl $name<$ty> {}
            )*
    }
}
```

At the very least, I'd expect that to happen for the two completely internal macros. (I can't remember if you can hide entire branches from the docs.)

DJMcNab:
I'm not sure why you've requested my review - is this not still blocked on #961?

PoignardAzur:
Oh, that's what you meant by "Not approving until after the rebase". Sorry.

fivemeyestoreserver:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemeyestore:
The performance optimizations here are quite clever. Nice engineering work!

eyesgodcode:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

myeyestore:
Impressive project structure! The architecture really demonstrates solid design principles.

fivemhostingserver:
Interesting solution to this problem! I've been working on similar challenges recently.

godofcodeeyes:
The performance optimizations here are quite clever. Nice engineering work!

eyesaider:
This is a well-structured codebase. I appreciate the clean separation of concerns.

eyestorefivem:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

eyestoreoye:
Very clean code with good test coverage. This is a great example of best practices.

fivemeyestoreserver:
Very clean code with good test coverage. This is a great example of best practices.

fivemserverhostings:
Excellent contribution to open source! The community will benefit greatly from this.

eyestorefivemstore:
Impressive project structure! The architecture really demonstrates solid design principles.

scriptforfivem:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemservereyestore:
Great implementation of this feature. I particularly like the approach to error handling.

ayestoro:
The documentation in this repository is excellent! Very clear and comprehensive.

xilem/src/property_tuple.rs
DJMcNab: My experience with reading other crates is that the usual pattern is to use an `@` prefix marker for internal branches - e.g. https://docs.rs/bitmask/latest/bitmask/macro.bitmask.html.

But I won't block on it.

xilem/examples/calc.rs
```diff
@@ -3,16 +3,21 @@
 
 //! A simple calculator example
 #![expect(clippy::cast_possible_truncation, reason = "Deferred: Noisy")]
+#![expect(
+    clippy::shadow_unrelated,
+    reason = "Seems to prevent patterns we'd want to use."
+)]
```
DJMcNab, line 9: See #978, which would obviate this addition. I'd like that to land before this PR

xilem/examples/calc.rs
```diff
@@ -267,22 +276,37 @@ fn display_label(text: &str) -> impl WidgetView<Calculator> + use<> {
 /// they take up all available space in flex containers.
 fn expanded_button(
     text: impl Into<Label>,
+    is_digit: bool,
```
DJMcNab, line 279: I don't really hugely like the API of this component - can we either make a function which both this and `digit_button` calls, or make an enum for this field, or just inline this into `digit_button`.

The lowest lift is just to make it an enum.

xilem/src/lib.rs
```diff
@@ -335,11 +339,17 @@ impl<W: Widget + FromDynWidget> Pod<W> {
...
             id: WidgetId::next(),
             transform: Affine::default(),
+            properties: Properties::new(),
         }
     }
 }
 
 impl<W: Widget + FromDynWidget + ?Sized> Pod<W> {
+    /// Builder-style method to set [`Properties`] on a new widget.
+    pub fn with_props(self, properties: Properties) -> Self {
```
DJMcNab, line 349: It might be nice to warn if the old properties in self weren't empty

xilem/src/property_tuple.rs
```diff
@@ -0,0 +1,182 @@
...
+impl_property_tuple!(P0, 0; P1, 1; P2, 2; P3, 3; P4, 4; P5, 5; P6, 6; P7, 7; P8, 8; P9, 9);
+impl_property_tuple!(P0, 0; P1, 1; P2, 2; P3, 3; P4, 4; P5, 5; P6, 6; P7, 7; P8, 8; P9, 9; P10, 10);
+impl_property_tuple!(P0, 0; P1, 1; P2, 2; P3, 3; P4, 4; P5, 5; P6, 6; P7, 7; P8, 8; P9, 9; P10, 10; P11, 11);
+
+// ---
+
+// We expect to use the ${index} metavariable here once it's stable
+// https://veykril.github.io/tlborm/decl-macros/minutiae/metavar-expr.html
+
+/// Macro used to declare a list of properties for an element.
```
DJMcNab, line 94: It's really great to get comprehensive docs here - thanks!

xilem/src/property_tuple.rs
```diff
@@ -0,0 +1,182 @@
...
+///     OtherProp, 1;
+///     AnotherProp, 2;
+///     YetOtherProp, 3;
+///     // ...
+/// );
+/// ```
+///
+/// This will declare a type `MyWidgetProp` as an alias to `(Option<SomeProp>, Option<OtherProp>, ...)`.
+///
+/// This will also implement [`HasProperty<Prop>`](crate::style::HasProperty) for `MyWidget` with each of the listed properties.
```
DJMcNab, line 113:
```suggestion
/// This will also implement [`HasProperty<Prop>`](crate::style::HasProperty) for `MyWidget` with each of the listed properties.
/// This enables using the corresponding extension method for each of them in the [`Style`](crate::Style) trait.
```

xilem/src/style.rs
```diff
@@ -0,0 +1,112 @@
...
+//! Traits used to set custom styles on elements.
+
+use masonry_winit::core::Property;
+use masonry_winit::properties::types::Gradient;
+use masonry_winit::properties::{
+    Background, BorderColor, BorderWidth, BoxShadow, CornerRadius, Padding,
+};
+use vello::peniko::Color;
+
+/// Marker trait implement by elements to signal that a given property can be set on them.
```
DJMcNab, line 13:
```suggestion
/// Marker trait implement by elements to signal that a given property can be set on them.
///
/// This should generally be implemented using [`declare_property_tuple!`](crate::declare_property_tuple).
/// This should only be used when authoring `View`s.
```

xilem/src/style.rs
```diff
@@ -0,0 +1,112 @@
...
+pub trait HasProperty<P: Property>: Style {
+    /// Return a mutable reference to the specific property.
+    fn property(&mut self) -> &mut Option<P>;
+}
+
+/// Trait implemented by most elements that lets you set some styling properties.
+///
+/// Which methods you can use will depend on which marker traits the element implements,
+/// which matches which [`Properties`](masonry_winit::core::Properties) the underlying widget handles.
```
DJMcNab, line 22:
```suggestion
/// which matches which [`Properties`](masonry_winit::core::Properties) the underlying widget handles.
///
/// If you're implementing your own Property types, you should create a trait like this for your styles.
```

xilem/src/style.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /// Set the element's background gradient.
+    fn background_gradient(mut self, gradient: Gradient) -> Self
+    where
+        Self: HasProperty<Background>,
+    {
+        *self.property() = Some(Background::Gradient(gradient));
+        self
+    }
```
DJMcNab, line 55: Maybe we should document that these overwrite each other.

xilem/src/style.rs
DJMcNab, line 13: `s/elements/views/`

Being consistent about terminology helps understanding. This applies to this entire PR.

DJMcNab (commented):
It will be good to keep moving here. This infrastructure largely matches my vision here.

There are some thoughts on docs, some of them blocking.
Otherwise, the interaction with #978 is also blocking.

xilem/src/lib.rs
PoignardAzur: In practice this can only be used in a new widget, so I don't expect that problem to come up too often.

PoignardAzur:
Mhhh... `cargo miri` is giving us this error:

> fatal error: cross-interpreting doctests is not currently supported by Miri.

That's pretty weird.

not-my-profile:
With this addition the last paragraph of the `Property` trait documentation becomes a bit outdated:

```rs
/// Note that if a type `Foobar` implements Property, that tells you that Foobar is meant
/// to be a property of *some* widget, but it doesn't tell you *which* widget accepts Foobar
/// as a property.
pub trait Property: Default + Send + Sync + 'static {
```

Here it would be nice to reference the new `HasProperty` trait you're introducing to Xilem but I think the problem with referencing things from a crate the current crate doesn't depend on is that you cannot link it as far as I know?

xilem/src/property_tuple.rs
```diff
@@ -0,0 +1,183 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use masonry_winit::core::{Properties, Property, Widget, WidgetMut};
+
+/// Helper trait implemented for all tuples of `Option<SomeProperty>` up to 12 items.
+pub trait PropertyTuple {
+    /// Helper method for [`xilem_core::View::build`].
+    ///
+    /// Returns list of properties to be set on the created widget.
+    fn build_properties(&self) -> Properties;
```
DJMcNab, line 11: So concretely, my suggestion is to make this be something like:
```suggestion
    fn build_properties(&self, props: &mut Properties);
```

xilem/src/style.rs
```diff
@@ -0,0 +1,115 @@
...
+use masonry_winit::properties::types::Gradient;
+use masonry_winit::properties::{
+    Background, BorderColor, BorderWidth, BoxShadow, CornerRadius, Padding,
+};
+use vello::peniko::Color;
+
+/// Trait implemented by views to signal that a given property can be set on them.
+///
+/// You should almost only implement this trait using [`declare_property_tuple!`](crate::declare_property_tuple)
+/// when authoring views.
```
DJMcNab, line 16:
```suggestion
/// You should in most cases only implement this trait using [`declare_property_tuple!`](crate::declare_property_tuple)
/// when authoring views.
```
I think "almost only" is ungrammatical.

xilem/src/property_tuple.rs
```diff
@@ -0,0 +1,183 @@
...
+// https://veykril.github.io/tlborm/decl-macros/minutiae/metavar-expr.html
+
+/// Macro used to declare a list of properties for an element.
+///
+/// The expected way to invoke this macro is:
+///
+/// ```ignore
+/// declare_property_tuple!(
+///     MyWidgetProps;
+///     MyWidget;
```
DJMcNab, line 101: You're declaring this for a view, not a widget. The use of the name `MyWidget` is I think somewhat misleading.
It is technically correct that we tend to give the views the same names as the widgets, but I don't think that this kind of "cleverness" is helpful for learning.

DJMcNab (changes requested):
One straightforward but slightly major refactor needed, and some docs tweaks. Otherwise, this is looking pretty good.

ofluffydev:
Can we get this merged please :sob:

xilem/src/property_tuple.rs
PoignardAzur: The `with_props` method has been removed, so now the code does `elem.properties = build_properties()`, which I think is more readable than a mutable parameter.

I'd rather merge the API as-is and maybe consider a version with a mutable param later. I don't think it changes much either way.

PR #965, opened 2025-05-09
Rename `masonry_core`, `masonry` to `masonry`, `masonry_winit`.

xStrom:
[Matt proposed](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Proposed.20change.20to.20the.20split.20between.20core.20and.20winit.20integrat/with/516903214) renaming the Masonry crates again. The proposal received positive feedback and so now is the time to do it, before we publish.

The proposal goes deeper than just renaming. However, in the interest of time and review scope, this PR implements just the renaming part. Getting the crate names right for publishing is paramount. API changes can happen later as follow-up work and delays there won't block publishing.


DJMcNab (approved):
I think just getting this landed is important.

I think probably splitting it into two PRs: one renaming the packages and one moving the files would have been better, but I don't think going back to redo that is worthwhile

xStrom:
I originally had them as separate commits, but later squashed due to habbit. 🤦‍♂️ Didn't want to redo it.

DJMcNab:
I think with the reflog, you should be able to recover the separate commits. I've unqueued, because I'm worried about the impact this will have on blame, if the separated commits are recoverable.

The only blame that's really clobbered is the Cargo.tomls, but those are files I've occasionally had to blame dive in before, so it would be nice to avoid that

PoignardAzur:
For the record, I'm not super happy about changing the crate names (*again*) without an ecosystem-wide discussion. I know I said I liked this pattern better on Zulip, but I was kind of assuming we'd switch *all* our projects to the same pattern.

I get the rationale about doing it before the next release, though, so I guess we'll have to bite the bullet.

xStrom:
> I think with the reflog, you should be able to recover the separate commits.

Good call, indeed possible. I've opened #966 as the first part of this split-up.

> For the record, I'm not super happy about changing the crate names (again) without an ecosystem-wide discussion. I know I said I liked this pattern better on Zulip, but I was kind of assuming we'd switch all our projects to the same pattern.

It's a bit ridiculous to be renaming so much, indeed. I don't have a strong personal take here, I just went based on the Zulip discussion where people were in favor with varying enthusiasm.

Regarding ecosystem-wide naming, this gets into the batteries included aspect. That is, for Xilem we want `xilem` to be the batteries included, easy-to-use crate - because it is intended to be used by app developers. Masonry, however, can be allowed to require more  care because it is targeted towards lower level framework developers.

PR #966, opened 2025-05-09
Rename `masonry_core`, `masonry` to `masonry`, `masonry_winit`.

xStrom:
[Matt proposed](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Proposed.20change.20to.20the.20split.20between.20core.20and.20winit.20integrat/with/516903214) renaming the Masonry crates again. The proposal received positive feedback and so now is the time to do it, before we publish.

The proposal goes deeper than just renaming. However, in the interest of time and review scope, this PR implements just the renaming part. Getting the crate names right for publishing is paramount. API changes can happen later as follow-up work and delays there won't block publishing.

---

For git blame benefit we wanted to split #965 into two parts. This PR is the first part that renames the crates. A follow-up PR will rename the directories.


DJMcNab (approved):
A quick scan indicates that this only does renames. I've not carefully verified everything, but if the tests still work I'm happy to land this.

PR #972, opened 2025-05-10
Put `Cargo.toml` sections into consistent order.

xStrom:
* Ordering of sections: `package`, `features`, `dependencies`, `lints`, `profile`, `example`.
* Ordering of `package` fields: `name`, `version`, `description`, `keywords`, `categories`, `edition`, `rust-version`, `license`, `repository`, `homepage`, `exclude`.

I also removed the `homepage` field from non-Xilem crates, some stale exclusions, and the example scraping from `masonry` as it doesn't have examples.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,30 @@ license = "Apache-2.0"
 repository = "https://github.com/linebender/xilem"
 homepage = "https://xilem.dev/"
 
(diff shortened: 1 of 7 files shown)


waywardmonkeys (approved):
Wish we could alphabetize some sections .. but maybe not today.

PR #973, opened 2025-05-10
Prepare to release 0.3.0.

xStrom:
We skip Xilem 0.2.0 and go straight to 0.3.0 in order to synchronize the version with its companion packages that are expected to be updated in lock-step for now.

[The one exception is `tree_arena`](https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Xilem.20version.20sync/with/517174026), which will retain its own version pacing, because it's likely to see fewer updates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1863,7 +1863,7 @@ dependencies = [
 
 [[package]]
 name = "masonry"
-version = "0.2.0"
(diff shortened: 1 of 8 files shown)


waywardmonkeys (approved):

xStrom:
It will be shortly.

PR #974, opened 2025-05-10
Fix visiblity of a property function

melix99:
The `prop_changed` function of the other properties are public, so I thought this was a mistake.


xStrom (approved):
I agree with your assessment. Thanks!

PR #982, opened 2025-05-20
masonry: Fix dpi handling in `Split` widget.

xorgy:


xorgy:
This was broken in #950 because coordinate conversion was only done in `::Down` and not in `::Move`

waywardmonkeys (approved):
Overall, this code doesn't really feel readable to me ... and it would be nice to have some actual tests ... but at least this is correct now. I think?

xorgy:
Agreed on testing. There are other bugs in this widget that I am not addressing here, I may write a couple issues/specs.

PR #984, opened 2025-05-21
Implement reparenting in tree_arena

not-my-profile:
This commit firstly implements root_ids and child_ids iterators so that the reparenting implementation can be properly tested.

Reparenting is then implemented in both the safe and the unsafe tree. For the unsafe tree no new unsafe code is introduced.

To make the tests more readable a new Node struct is introduced in the test code so that the tree after the reparenting can be compared with a single assert_eq!.

### DIFF
--- a/tree_arena/src/tree_arena_safe.rs
+++ b/tree_arena/src/tree_arena_safe.rs
@@ -140,6 +140,11 @@ impl<T> TreeArena<T> {
         }
     }
 
+    /// An iterator visiting all root ids in arbitrary order.
+    pub fn root_ids(&self) -> impl Iterator<Item = NodeId> {
+        self.roots.keys().copied()
+    }
+
     /// Returns a handle giving access to the roots of the tree.
     ///
     /// Using [`insert`](ArenaMutList::insert) on this handle
@@ -230,6 +235,11 @@ impl<T> ArenaRef<'_, T> {
         self.children
             .parent_id
             .expect("ArenaRefList always has a parent_id when it's a member of ArenaRef")
+    }
+
+    /// An iterator visiting all child ids in arbitrary order.
+    pub fn child_ids(&self) -> impl IntoIterator<Item = NodeId> {
+        self.children.children.keys().copied()
     }
 }
 
--- a/tree_arena/src/tree_arena_safe.rs
+++ b/tree_arena/src/tree_arena_safe.rs
@@ -193,6 +193,51 @@ impl<T> TreeArena<T> {
         };
         parents_map.get_id_path(id.into(), None)
     }
+
+    /// Moves the given child (along with all its children) to the new parent.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the parent is actually a child of the to-be-reparented node, or
+    /// if the to-be-reparented node is a root node, or
+    /// if either node id cannot be found, or
+    /// if both given ids are equal.
+    pub fn reparent(&mut self, child: impl Into<NodeId>, new_parent: impl Into<NodeId>) {
+        let child_id = child.into();
+        let new_parent_id = new_parent.into();
+
+        assert_ne!(
+            child_id, new_parent_id,
+            "expected child to be different from new_parent but both have id #{child_id}"
+        );
+        assert!(
+            !self.get_id_path(new_parent_id).contains(&child_id),
+            "cannot reparent because new_parent #{new_parent_id} is a child of the to-be-reparented node #{child_id}"
+        );
+        assert!(
+            !self.roots.contains_key(&child_id),
+            "reparenting of root nodes is currently not supported"
+        );
+
+        let old_parent_id = self
+            .parents_map
+            .get(&child_id)
+            .unwrap_or_else(|| panic!("no node found for child id #{child_id}"))
+            .unwrap();
+        let child_node = self
+            .find_mut(old_parent_id)
+            .unwrap()
+            .children
+            .children
+            .remove(&child_id)
+            .unwrap();
+        self.parents_map.insert(child_id, Some(new_parent_id));
+        self.find_mut(new_parent_id)
+            .unwrap_or_else(|| panic!("no node found for new_parent id #{new_parent_id}"))
+            .children
+            .children
+            .insert(child_id, child_node);
+    }
 }
 
 impl<T> TreeNode<T> {
--- a/tree_arena/src/tree_arena_unsafe.rs
+++ b/tree_arena/src/tree_arena_unsafe.rs
@@ -121,13 +121,7 @@ impl<T> DataMap<T> {
         }
     }
 
-    /// Find an item in the tree.
-    ///
-    /// Returns a shared reference to the item if present.
-    ///
-    /// Time Complexity O(1)
-    fn find_inner(&self, id: NodeId) -> Option<ArenaRef<'_, T>> {
-        let parent_id = *self.parents.get(&id)?;
+    fn find_node(&self, id: NodeId) -> Option<&TreeNode<T>> {
         let node_cell = self.items.get(&id)?;
 
         // SAFETY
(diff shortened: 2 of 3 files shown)


tree_arena/src/tree_arena_safe.rs
```diff
@@ -188,6 +198,62 @@ impl<T> TreeArena<T> {
         };
         parents_map.get_id_path(id.into(), None)
     }
+
+    // Moves the given child (along with all children) to the new parent.
```
DJMcNab, line 202: This should probably be a doc comment?

DJMcNab:
cc @KGrewal1, would you be willing to give this a review?

tree_arena/src/tree_arena_safe.rs
```diff
@@ -188,6 +198,62 @@ impl<T> TreeArena<T> {
...
+            .expect("no node found for new_parent id")
+            .children
+            .children
+            .insert(child_id, child_node);
+    }
+}
+
+/// An iterator over the children ids of a [`TreeArena`] in arbitrary order.
+#[derive(Debug)]
+pub struct RootIdIter<'a>(vec::IntoIter<NodeId>, PhantomData<&'a ()>);
```
DJMcNab, line 249: What's the value of this lifetime parameter? I presume it's to match the unsafe version?
It might be worth adding a single line comment to that effect

tree_arena/src/tree_arena_unsafe.rs
```diff
@@ -276,6 +286,64 @@ impl<T> TreeArena<T> {
     pub fn get_id_path(&self, id: impl Into<NodeId>) -> Vec<NodeId> {
         self.data_map.get_id_path(id.into(), None)
     }
+
+    // Moves the given child (along with all children) to the new parent.
```
DJMcNab, line 290: This should also be a doc comment

tree_arena/src/tree_arena_unsafe.rs
```diff
@@ -276,6 +286,64 @@ impl<T> TreeArena<T> {
...
+    /// If the parent is actually a child of the to-be-reparented node.
+    /// If the to-be-reparented node is a root node.
+    /// If both given IDs are equal.
+    pub fn reparent(&mut self, child: impl Into<NodeId>, new_parent: impl Into<NodeId>) {
+        let child_id = child.into();
+        let new_parent_id = new_parent.into();
+
+        assert_ne!(
+            child_id, new_parent_id,
+            "expected child to be different from new_parent"
```
DJMcNab, line 303:
```suggestion
            "expected child ({child_id}) to be different from new_parent ({new_parent_id})"
```

As a way to give a bit more context. It doesn't add all that much value, but it shouldn't really hurt either.
This applies to most of the asserts added here

DJMcNab (approved):
This is looking really good. I'd like to get at least one more set of eyes on this, as I'm not overly familiar with the tree arena code

tree_arena/src/tree_arena_safe.rs
not-my-profile: Oh yes, thanks good catch!

tree_arena/src/tree_arena_safe.rs
not-my-profile: Yes indeed. I've added such a comment :)

KGrewal1:
Happy to give this a review, will look at it over this weekend!

tree_arena/src/tree_arena_safe.rs
KGrewal1: I think the alternative would be storing `self.roots.keys().copied()` directly in the struct as the ugly type `Copied<hashbrown::hash_map::Keys<'_, u64, tree_arena_safe::TreeNode<T>>>`  and then having PhantomData<T> in the unsafe API - advantage is that this avoids the call to collect which isn't really necessary (as call `.copied().collect().into_iter()`, disadvantage is that it now has a generic type T associated which is not ideal (can't see a way to store a less concrete type)

KGrewal1: Though as I believe the safe arena is still generally used, maybe is worth caring about the extra allocation in that path?

KGrewal1: Or does this actually need a concrete type at all and just be a function like:
```
    /// An iterator visiting all root ids in arbitrary order.
    pub fn root_id_iter(&self) -> impl Iterator<Item = NodeId> + use<'_, T> {
        self.roots.keys().copied()
    }
  ```
  
  for the safe version and
  ```
      /// An iterator visiting all root ids in arbitrary order.
    pub fn root_id_iter(&self) -> impl Iterator<Item = NodeId> + use<'_, T> {
        self.roots.iter().copied()
    }
    ```
    
    for the unsafe?

Comment (not-my-profile) [reply tree_arena/src/tree_arena_safe.rs]: Oh yes indeed just `-> impl Iterator<Item = NodeId>` suffices ... it doesn't even need `+ use<'_, T>`.

Comment (KGrewal1) [reply tree_arena/src/tree_arena_safe.rs]: I think similar could be done with the `ArenaRef` into iter pattern too:

for safe:

```
 pub fn iter_children(&self) -> impl Iterator<Item = NodeId> {
        self.children.children.keys().copied()
    }
 ```

for unsafe:

```
    pub fn iter_children(&self) -> impl Iterator<Item = NodeId> {
        self.children
            .parent_arena
            .find_node(self.id())
            .into_iter()
            .flat_map((|c: &TreeNode<T>| &c.children))
            .copied()
    }
  ```

It avoids some messy types like `Copied<hashbrown::hash_map::Keys<'a, u64, TreeNode<T>>>` and `FlatMap<option::IntoIter<&'a TreeNode<T>>, &'a Vec<NodeId>, GetChildren<T>>`, takes `&self` instead of `self` and I think maintains least suprisal vs an `IntoIter` impl  on the `ArenaRef` iterating over specifically the `NodeId` - maybe both this and the function `root_ids` could have the same name too for ease of use?

KGrewal1:
On the unsafe side:
- split `find_inner` into `find_node` (w/ internal unsafe but same precondition as before) and `find_inner` calling this all seems great
- implemented reparenting logic - see comment re order of steps moving from one parent to the other but otherwise good

On the safe side new reparenting logic seems fine

And then see comments regarding the node iterator interface, but otherwise looks great

tree_arena/src/tree_arena_safe.rs
not-my-profile: I don't really mind messy types as long as they're hidden in a type but I agree that IntoIterator having specifically Item=NodeId is a bit unintuitive and an explicit `.child_ids()` also makes the calling code more readable.

Regarding the naming I prefer to name the iterator over the root ids `root_ids` rather than `child_ids` because the latter would imply that the roots have some parent node, which isn't the case.

not-my-profile:
@KGrewal1 Thanks, I addressed your feedback, except:

> see comment re order of steps moving from one parent to the other

I can't find that comment, perhaps you forgot to submit it?

KGrewal1:
It was this comment https://github.com/linebender/xilem/pull/984#discussion_r2103203816

not-my-profile:
Huh there appears to be some GitHub bug. While GitHub shows me a (cut off) preview of your comment when I hover over your link, when I click the link it just takes me to this PR page where I cannot see that comment anywhere. Could you please repeat it?

KGrewal1:
It was essentially moving the insertion into the parents hash map in the unsafe version to before the addition of the node to the new parent list instead of after so that it goes 


[old parent, old parent] -> [None, old parent] -> [None, new parent] -> [new parent, new parent]

instead of

[old parent, old parent]->[None, old parent] -> [new parent, old parent] -> [new parent, new parent]

and so avoiding the intermediate inconsistent state of [new parent, old parent]

not-my-profile:
I don't really follow. Moving the child takes 3 operations, no matter in which order you do them the state will always be inconsistent until all of them have been done. I think for readability it's nice to firstly update the `child_arr` and then the parent reference.

KGrewal1:
I think in terms of consistency having it in [None, New] is preferable to [New, Old] as cannot access the child from the None parent (but can from the New) - this should only occur in case of an unwind in the middle of the sequence though so don't think this is a blocker

KGrewal1:
I think it's fine and so don't think it's a blocker to merge, but also not to big a change to stop it going to the [new, old] state either (in case that does change ever)

KGrewal1:
LGTM @DJMcNab @not-my-profile !

DJMcNab:
Fwiw, something not being `UnwindSafe` is not something we can rely upon for safety. That is, unwind safety isn't talking about Rust's memory safety guarantees, unfortunately. That is, if:
1) it's possible for this code to panick partway through; AND
2) That panic leaves us in an inconsistent state; AND
3) that inconsistent state allows UB for any later operation

Then some remedial action needs to be taken (such as aborting when this code panicks, or something else).
This isn't high enough on my priority list for me to determine if that's the case myself. If @KGrewal1 asserts than any one (ideally any two!) of the conditions above are false, that's good enough for me.

DJMcNab:
Given that:
1) We *think* that the first condition doesn't apply
2) We don't actually ever use the unsafe tree arena

I'm happy to land this and accept that improved robustness as a follow-up, unblocking the work of integrating this in Masonry. Thanks @KGrewal1 for your excellent review.

PR #986, opened 2025-05-23
masonry: re-export accesskit and smallvec

not-my-profile:
The types of them are needed to implement masonry::core::Widget.


waywardmonkeys (approved):

DJMcNab:
Actually, we should replace any uses of AccessKit or Smallvec which should use these imports - for example, in Xilem or in Masonry's examples. I won't block on this, but I don't think there's any immediate need to merge unless and until we decide that's out-of-scope for this PR.

not-my-profile:
I've updated the examples. I'd defer the other issues I described in https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/re-export.20types.20for.20Widget to followup PRs.

PR #988, opened 2025-05-24
add lfs-to-blobs.sh

not-my-profile:
This is an idea to work around or LFS quota problem (#851) that prevents us from embedding our screenshots in our docs.

The idea is to have a `screenshots` branch (starting from scratch, not based on our existing branches) and for every release we run the script introduced in this PR to update that branch, converting all Git LFS objects into regular Git blobs.

To try this out:
```
git switch --orphan screenshots
git commit --allow-empty -m init
git switch -
./docs/lfs-to-blobs.sh v0.3.0
```

This creates a new screenshots-v0.3.0 branch based on the screenshots branch, with a new commit adding all LFS objects of the given git tag into a directory named after the tag.

### DIFF
--- a/docs/lfs-to-blobs.sh
+++ b/docs/lfs-to-blobs.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+# Embedding screenshots from Git LFS in our documentation via
+# https://media.githubusercontent.com/ depletes our GitHub LFS quota
+# so we instead convert our LFS objects into regular Git blobs
+# in the branch configured below on release so that we can embed them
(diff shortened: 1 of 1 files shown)


xStrom:
Can you give an example HTML tag that would actually show an image? Because I'm not sure that you can embed the raw links.

```html
<!-- runs into CORS limitations -->
<img src="https://raw.githubusercontent.com/linebender/xilem/refs/heads/main/docs/assets/xilem-layers.svg" />
```

Also, even if we could there are risks with abusing this, because the doc image links should preferably work for a very long time.

not-my-profile:
No idea. For me it worked on the first try.

PR #989, opened 2025-05-25
Improve accessibility of ProgressBar and Spinner

DataTriny:
- It is better to explicitly set min and max values rather than relying on defaults.
- We can then directly set the numeric value without doing any calculation.
- Setting text value on progress bar is not recommended unless there is strong reason to do so, the indeterminate case lacks translation anyway.
- A spinner can be considered as a progress indicator with indeterminate value.


PoignardAzur (approved):
LGTM.

waywardmonkeys:
Also, I'm really happy to see you (@DataTriny) here with PRs like this. :)

PR #992, opened 2025-05-26
xilem: support multiple windows

not-my-profile:
Writing tests for the driver logic will have to wait for winit 0.31 which introduces
traits like `ActiveEventLoop` and `Window` that we need for testing.

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -27,7 +27,7 @@
 //! ```rust,no_run
 //! use winit::error::EventLoopError;
 //! use xilem::view::{button, flex, label};
-//! use xilem::{EventLoop, WidgetView, Xilem};
+//! use xilem::{EventLoop, WindowOptions, WidgetView, Xilem};
 //!
 //! #[derive(Default)]
 //! struct Counter {
@@ -42,8 +42,8 @@
 //! }
 //!
 //! fn main() -> Result<(), EventLoopError> {
-//!     let app = Xilem::new(Counter::default(), app_logic);
-//!     app.run_windowed(EventLoop::with_user_event(), "Counter app".into())?;
+//!     let app = Xilem::new_simple(Counter::default(), app_logic, WindowOptions::new("Counter app"));
+//!     app.run_in(EventLoop::with_user_event())?;
 //!     Ok(())
 //! }
 //! ```
@@ -131,18 +131,20 @@
 #![expect(clippy::allow_attributes_without_reason, reason = "Deferred: Noisy")]
 
 use std::collections::HashMap;
+use std::iter::Once;
 use std::sync::Arc;
 
 use masonry::core::{
     DefaultProperties, FromDynWidget, Properties, Widget, WidgetId, WidgetMut, WidgetOptions,
     WidgetPod,
 };
-use masonry::dpi::LogicalSize;
 use masonry::theme::default_property_set;
-use masonry_winit::app::{MasonryUserEvent, WindowId};
+use masonry_winit::app::MasonryUserEvent;
 use view::{Transformed, transformed};
+use window_options::WindowCallbacks;
 use winit::error::EventLoopError;
-use winit::window::{Window, WindowAttributes};
+use winit::window::WindowAttributes;
+use xilem_core::map_state;
 
 use crate::core::{
     AsyncCtx, MessageResult, Mut, RawProxy, SuperElement, View, ViewElement, ViewId,
@@ -154,7 +156,8 @@ pub use masonry::parley::style::FontWeight;
 pub use masonry::peniko::{Blob, Color};
 pub use masonry::widgets::{InsertNewline, LineBreaking};
 pub use masonry::{dpi, palette};
-pub use masonry_winit::app::{EventLoop, EventLoopBuilder};
+pub use masonry_winit::app::{EventLoop, EventLoopBuilder, WindowId};
+pub use window_options::WindowOptions;
 pub use xilem_core as core;
 
 /// Tokio is the async runner used with Xilem.
@@ -166,6 +169,8 @@ mod any_view;
 mod driver;
 mod one_of;
 mod property_tuple;
+mod window_options;
+mod window_view;
 
 pub mod style;
 pub mod view;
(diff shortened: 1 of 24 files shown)


PoignardAzur:
> `parley::font::FontContext` is meant to be app-wide, can we move it out of RenderRoot? but all the contexts require mutable access to it

That seems like a pretty thorny problem. I think it's fine to do one-context-per-window for now, maybe with a big TODO comment for later.

> why should DriverCtx be publicly constructible? I removed it for now since it wasn't used and not clear to me

I think I already did that once and someone (Bruce?) said they were using it for stuff. I would leave it as-is for now.

PoignardAzur:
Also, this is minor compared to the rest, but given that this PR has a pretty large diff, I'd consider keeping the `AppDriver::on_action` method signature the same, and just add a `WindowId` to `DriverCtx`, and then change `AppDriver::on_action` in a follow-up PR. It's more work for you, so I'm not going to insist on it, but it *would* make the diff much shorter.

masonry_winit/src/app_driver.rs
```diff
@@ -1,59 +1,58 @@
...
+
 use crate::app::{MasonryState, RenderRoot};
 use crate::core::{Action, WidgetId};
+use crate::event_loop_runner::WindowState;
 
 /// Context for the [`AppDriver`] trait.
 ///
 /// Currently holds a reference to the [`RenderRoot`].
-pub struct DriverCtx<'a> {
+pub struct DriverCtx<'a, WindowKey> {
```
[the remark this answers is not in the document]
not-my-profile: I thought it would be neat to handle events but indeed it's not that helpful because just because you have a value of your type doesn't mean the window is still open and in Xilem we have window-specific callbacks anyway for event handling. I replaced it with a WindowId as you suggested :)

masonry/src/util.rs
```diff
@@ -38,7 +38,7 @@ macro_rules! debug_panic {
 
 // ---
 
-pub(crate) type AnyMap = anymap3::Map<dyn Any + Send + Sync>;
+pub(crate) type AnyMap = anymap3::Map<dyn CloneAny + Send + Sync>;
```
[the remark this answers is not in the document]
not-my-profile: The `RenderRoot` contains `DefaultProperties`, which contains an `AnyMap`. `MasonryState` is initialized with `DefaultProperties`, which I'm currently cloning when creating a new window.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -28,113 +31,144 @@ use crate::peniko::Color;
...
-    mut loop_builder: EventLoopBuilder,
-    // In future, we intend to support multiple windows. At the moment though, we only support one
-    window_attributes: WindowAttributes,
-    root_widget: impl Widget,
-    app_driver: impl AppDriver + 'static,
-) -> Result<(), EventLoopError> {
+    mut loop_builder: EventLoopBuilder<Key>,
+    windows: Vec<Window<Key>>,
+    app_driver: impl AppDriver<Key> + 'static,
+) -> Result<(), EventLoopError>
```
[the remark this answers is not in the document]
PoignardAzur: I'm not sure the `create_initial_windows` API is better? There's a kind of simplicity in just, like... creating the initial windows (and passing the window_attributes) from the initial constructor.

But I don't want to start backseat designing, and your proposed API is fine for now. I might revise it in a follow-up PR, FWIW.

masonry/src/util.rs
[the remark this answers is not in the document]
PoignardAzur: This feels like one of the rare cases where an Arc might be the idiomatic solution.

But alright, cloning is fine for now. We're bound to refactor the cascading system sooner than later anyway.

RagibHasin:
Didn't we [(tentatively) decide](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Plan.20for.20multi.20window) that an specialization of the `View` trait with Winit windows as element would be the way to handle reactivity in the Xilem side?

PoignardAzur:
I'm neutral on using the View trait to represent the set of open windows. If the first version of this PR doesn't do it or doesn't do it perfectly, that's fine.

masonry_winit/src/event_loop_runner.rs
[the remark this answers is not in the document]
not-my-profile: I think it's better because the `on_action` implementation is strongly tied to what happens in `create_initial_windows`. `on_action` is bound to do WidgetMut downcasts that depend on which widget hierarchy was created when setting up the windows initially. And I also think it's neat that windows are always created the same. If you look at `hello_masonry` which only creates windows on startup it's very intuitive how to create windows in `on_action` ... you just copy'n'paste the code a few lines down. Also if we add more documentation to `create_window` we have it in both places.

waywardmonkeys:
@PoignardAzur It should be a goal to get this driven to completion before the underlying code moves around too much and makes this overly painful.

masonry/src/core/contexts.rs
```diff
@@ -1192,51 +1192,39 @@ impl_context_method!(
         /// Moves the window with the left mouse button until the button is released.
         pub fn drag_window(&mut self) {
             trace!("drag_window");
-            self.global_state
-                .signal_queue
-                .push_back(RenderRootSignal::DragWindow);
+            (self.global_state.signal_sink)(RenderRootSignal::DragWindow);
```
PoignardAzur: At this point, I think I'd rather have these methods call `RenderRootState::emit_signal()`.

masonry/src/core/properties.rs
```diff
@@ -19,7 +19,7 @@ use crate::core::Widget;
 /// as a property.
 /// That information is deliberately not encoded in the type system.
 /// We might change that in a future version.
-pub trait Property: Default + Send + Sync + 'static {
+pub trait Property: Default + Send + Sync + Clone + 'static {
```
PoignardAzur: I *think* you don't need that bound anymore now that you're using Arc?

masonry_winit/README.md
```diff
@@ -40,21 +40,49 @@ Lots of things need improvements, e.g. text input is janky and snapshot testing
...
+            .with_title("To-do list")
+            .with_resizable(true)
+            .with_min_inner_size(window_size);
+
+        ctx.create_window(
+            self.window_id,
+            RootWidget::new(main_widget),
+            window_attributes,
+        );
+    }
```
PoignardAzur: I really don't like that most of the work now happens in `fn create_initial_windows()`.

I think it makes these examples less intuitive for newcomers and overall harder to read, especially in the single window case. I won't block the PR on it, but, fair warning, if that part does get merged then I'm likely to rewrite it down the line.

masonry_winit/src/app_driver.rs
```diff
@@ -1,59 +1,129 @@
...
     }
 
-    /// Returns `true` if something happened that requires a rewrite pass or a re-render.
-    pub fn content_changed(&self) -> bool {
-        self.render_root.needs_rewrite_passes()
+    /// Access the [`WindowHandle`] of the given window.
+    pub fn window_handle(&self, window_id: WindowId) -> &WindowHandle {
+        let WindowState::Rendering { handle: window, .. } = &self.state.window(window_id).state
+        else {
+            panic!("window is not not in rendering state");
```
PoignardAzur:
```suggestion
            panic!("window is not in rendering state");
```

(Although this error message is not super descriptive. Not sure what a better one would be.)

masonry_winit/src/app_driver.rs
```diff
@@ -1,59 +1,129 @@
...
+    }
+
+    /// Access the [`WindowHandle`] and [`RenderRoot`] of the given window.
+    pub fn window_handle_and_render_root(
+        &mut self,
+        window_id: WindowId,
+    ) -> (&WindowHandle, &mut RenderRoot) {
+        let window = self.state.window_mut(window_id);
+        let WindowState::Rendering { handle, .. } = &window.state else {
+            panic!("window is not not in rendering state");
```
PoignardAzur:
```suggestion
            panic!("window is not in rendering state");
```

masonry_winit/README.md
not-my-profile: We already started discussing this in https://github.com/linebender/xilem/pull/992#discussion_r2117888658. I'd prefer to keep the discussion in one place. I'd be curious what other people think (perhaps we should open a Zulip topic about it?). Doing it in one place does simplify masonry quite a bit. Also I'd think that most newcomers would be using xilem rather than masonry?

masonry/src/core/properties.rs
not-my-profile: Good catch! Removed.

masonry_winit/src/event_loop_runner.rs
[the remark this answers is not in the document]
not-my-profile: Ok my main concern that the on_action impl depends on the widget hierarchy can be addressed with #1031.

xilem/src/lib.rs
```diff
@@ -180,25 +184,115 @@ pub struct Xilem<State, Logic> {
...
+        Box<
+            dyn FnMut(
+                &mut ExitOnClose<State>,
+            ) -> std::iter::Once<(
+                WindowId,
+                WindowOptions<ExitOnClose<State>>,
+                Box<AnyWidgetView<ExitOnClose<State>>>,
+            )>,
+        >,
+    >
```
PoignardAzur: That type is way, way too long. In principle you could fix that with a type alias, but I think it's a symptom of a deeper problem. You should probably change this code some.

xilem/src/window_view.rs
```diff
@@ -0,0 +1,121 @@
...
+use masonry_winit::{app::RenderRoot, widgets::RootWidget};
+use winit::window::{Window, WindowAttributes};
+use xilem_core::{AnyViewState, View, ViewElement, ViewMarker};
+
+use crate::{AnyWidgetView, ViewCtx, WindowOptions};
+
+pub(crate) struct WindowView<State> {
+    options: WindowOptions<State>,
+    root_widget_view: Box<AnyWidgetView<State, ()>>,
+}
```
PoignardAzur, line 14: So I'm noticing that `WindowView` is more or less an analogue to `WidgetView`, but whereas `WidgetView` is a trait, `WindowView` is a type.

I think a lot of this code could get cleaner if you went the full way and created a parallel trait hierarchy:

- `trait WindowView { ... }`
- `trait WindowViewSequence { ... }`
- `struct Window { ... }`

xilem/src/lib.rs
```diff
@@ -180,25 +184,115 @@ pub struct Xilem<State, Logic> {
...
+/// The trait [`Xilem::new`] expects to be implemented for the state.
+///
+/// [`Xilem::new_simple`] does not use this trait implementation.
+pub trait AppState {
+    /// Returns whether the application should keep running or exit.
+    ///
+    /// Is currently only checked after a close request.
+    // TODO: check this after every state mutation
+    fn keep_running(&self) -> bool;
+}
```
PoignardAzur, line 269: I suspect you could make your code a lot simpler by removing that trait and moving the method to `WindowView`.

PoignardAzur (changes requested):

xilem/src/lib.rs
not-my-profile: I introduced a `WindowTuple` type alias. I don't think it's symptomatic of any problems. If you use `Xilem::new` the `Logic` type is inferred. It's just for this specialized `new_simple` that the type that has to be specified is more complex since it has to use `map_state`.

xilem/src/lib.rs
not-my-profile: The intention behind this is that any callback (that can mutate the state) could quit the app so e.g. also the click callback of a button. I think deriving the "keep running" boolean from the app state is a nicer solution to this than passing some additional context to every single callback. With this in mind I don't think moving it into `WindowView` makes sense.

xilem/src/window_view.rs
not-my-profile: I don't quite see how introducing more traits would clear a lot up (see also discussion below regarding `keep_running`), could you elaborate?. Regarding ViewSequence I think that's quite unnecessary if an iterator suffices.

PoignardAzur:
I don't agree with the provided rational, but there will always be time to change it later.

I'll merge this so it doesn't block other work.

PoignardAzur (approved):

PR #998, opened 2025-05-27
Reorder logic for reparent

KGrewal1:
As discussed in https://github.com/linebender/xilem/pull/984#event-17825988627 - also make the comment for insert consistent between the safe and unsafe version of the tree


DJMcNab (approved):
Seems reasonable. Updating the parent whilst the item is outside of the tree feels much lower risk to me. Thanks

PR #1004, opened 2025-05-29
Ensure that IME events interact with focus as expected; remove bubbling for IME events

DJMcNab:
Bubbling of IME events interacted extremely poorly with focus, especially around widget removal, but also potentially re-parenting in future.

This PR is as a result of #999.

The new test passes on main, but it does output the error added in that PR.

### DIFF
--- a/masonry/src/core/events.rs
+++ b/masonry/src/core/events.rs
@@ -32,6 +32,8 @@ pub enum TextEvent {
     /// A keyboard event.
     Keyboard(KeyboardEvent),
     /// An IME event.
+    ///
+    /// These events do not bubble.
     Ime(Ime),
     /// The window took or lost focus.
     WindowFocusChange(bool),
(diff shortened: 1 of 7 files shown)


PoignardAzur:
I *think* I get the idea behind this PR: it's hard to make bubbling work correctly for an event that widgets expect to receive when a child is deleted, and if bubbling is enabled for these events, users might make the wrong assumptions and use that events in ways that can't work.

But it adds a lot of documentation for one special case. I'd rather either make the special case work (for example, when the focused widget is removed, automatically set a flag on its parent and send the `Ime::Disabled` event to the widget with that flag), or have a wider discussion about removing bubbling altogether.

PoignardAzur (approved):

DJMcNab:
Oops - I forgot to update the title. Oh well

PR #1005, opened 2025-05-29
Use existing `LayoutContext` in `custom_widget` example.

xStrom:
Creating a new `LayoutContext` in `paint` is an anti-pattern. Instead we should promote the usage of the `text_contexts` method.


PoignardAzur (approved):
Good catch.

PR #1009, opened 2025-06-02
Add properties affecting hovered/disabled/active widgets

PoignardAzur:
Ideally, we'd want a system with full pseudo-class handling of properties, but until we implement that, this is a reasonable stopgap that lets the user customize *some* interactive behavior.

### DIFF
--- a/masonry/src/properties/background.rs
+++ b/masonry/src/properties/background.rs
@@ -19,6 +19,16 @@ pub enum Background {
     Gradient(Gradient),
 }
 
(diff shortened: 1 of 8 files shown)


jaredoconnell (approved):
Tested. It works.
With the latest design we may want to change the grid spacing in the calc example from 2 to 0. But that's just a preference.

PoignardAzur:
Yeah, we'll probably want to tinker with most of the examples once we have more styling options (which is pretty soon).

PR #1010, opened 2025-06-02
Rewrite marker comments across multiple lines.

PoignardAzur:
This is because VSCode now interprets dashes as part of the marker title, which makes markers more likely to overflow the minimap width.

Supersedes #947.


PoignardAzur:
The review feedback last time was that the PR added too much bloat around markers. This version just removes the trailing `// ---`.

waywardmonkeys (approved):

PR #1011, opened 2025-06-02
[Draft] Refactor WidgetMut to remove parent_state reference

PoignardAzur:
This lets us remove the Drop impl for WidgetMut, which makes it easier to use without making the borrow checker mad.


PoignardAzur:
This is something I've thought about during RustWeek; smart references are much easier to use when they don't have a Drop impl (because the borrow checker doesn't assume that the reference will be mutated when dropped).

I'm not sure about the implementation I just pushed, though. I'm thinking the flags should probably be updated in the `update_widget_tree` pass. Documenting that might be a pain.

PR #1014, opened 2025-06-02
Split off most of Xilem into xilem_masonry crate

PoignardAzur:
This decouples "using Xilem with Masonry" from "using Xilem with winit".

The composition root is still Xilem, whose public API is mostly unchanged.

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -13,7 +13,7 @@
 //! * Rendering is provided by [Vello][masonry::vello], a high performance GPU compute-centric 2D renderer.
 //! * GPU compute infrastructure is provided by [wgpu][masonry::vello::wgpu].
 //! * Text layout is provided by [Parley][masonry::parley].
-//! * Accessibility is provided by [AccessKit][] ([docs][accesskit_docs]).
+//! * Accessibility is provided by [AccessKit][] ([docs][::masonry::accesskit]).
 //! * Window handling is provided by [winit][].
 //!
 //! Xilem can currently be considered to be in an alpha state. Lots of things need improvements (including this documentation!).
@@ -88,7 +88,6 @@
 //! * [`lens`][crate::core::lens]: an adapter for using a component from a field of the current state.
 //! * [`memoize`][crate::core::memoize]: allows you to avoid recreating views you know won't have changed, based on a key.
 //!
-//! [accesskit_docs]: ::masonry::accesskit
 //! [AccessKit]: https://accesskit.dev/
 //! [Druid]: https://crates.io/crates/druid
 //! [Fontique]: https://crates.io/crates/fontique
--- a/xilem/README.md
+++ b/xilem/README.md
@@ -58,7 +58,7 @@ Masonry itself is built on top of a wide array of foundational Rust UI projects:
 * Rendering is provided by [Vello][masonry::vello], a high performance GPU compute-centric 2D renderer.
 * GPU compute infrastructure is provided by [wgpu][masonry::vello::wgpu].
 * Text layout is provided by [Parley][masonry::parley].
-* Accessibility is provided by [AccessKit][] ([docs][accesskit_docs]).
+* Accessibility is provided by [AccessKit][] ([docs][::masonry::accesskit]).
 * Window handling is provided by [winit][].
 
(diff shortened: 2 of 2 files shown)


PoignardAzur:
There's the issue with libdeflate again. I know it had something to do with hardcoded crate paths, but I can't find them again. I wish we documented that footgun better.

waywardmonkeys:
It would be nice to have feedback from @DJMcNab on this for the Xilem perspective.

But what's your long term goal here?

PoignardAzur:
The medium-term goal is to have:

- **masonry:** uses vello_api.
- **masonry_testing:** uses masonry and vello_cpu.
- **masonry_widgets:** uses masonry (and in tests, masonry_testing).
- **masonry_winit:** uses masonry, vello_hybrid and winit.
- **xilem_core**.
- **xilem_masonry:** uses xilem_core and masonry_widgets.
- **xilem:** uses xilem_masonry and masonry_winit.

Basically, provide a bunch of libraries which only use the dependencies they need.

So in this case, that means being able to use xilem and masonry without depending on winit.

PoignardAzur:
Very likely. It might also be useful for unit tests.

xilem_masonry/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub id: WidgetId,
+    /// The options the widget will be created with.
+    ///
+    /// If changing transforms of widgets, prefer to use [`transformed`]
+    /// (or [`WidgetView::transform`]).
+    /// This has a protocol to ensure that multiple views changing the
+    /// transform interoperate successfully.
+    pub options: WidgetOptions,
+    pub properties: Properties,
+}
```
PoignardAzur: By the way, one of the things I'd like to do after this PR is merged is to split items like this into separate files. Right now Xilem basically implements half its features in its `lib.rs` file, which is terrible for readability.

xilem/src/lib.rs
```diff
@@ -88,7 +88,7 @@
 //! * [`lens`][crate::core::lens]: an adapter for using a component from a field of the current state.
 //! * [`memoize`][crate::core::memoize]: allows you to avoid recreating views you know won't have changed, based on a key.
 //!
-//! [accesskit_docs]: accesskit
+//! [accesskit_docs]: https://docs.rs/accesskit/latest/accesskit/
```
DJMcNab: This is an unfortunate regression; can we instead use `xilem_masonry::masonry::accesskit` or similar?

PoignardAzur: Good point.

PoignardAzur:
Anyway, this is blocked on #1065 and something like #1049.

xilem/src/lib.rs
```diff
@@ -144,27 +139,19 @@ pub use tokio;
 
 pub use winit;
 
-mod any_view;
 mod app;
 mod driver;
```
DJMcNab, line 143: I believe that the driver should also be in xilem_masonry; pretty much the only stuff in Xilem should be the high-level winit-based runner, the windowing integration, and types to make the windowing integration work.

xilem_masonry/src/view_ctx.rs
```diff
@@ -47,11 +47,11 @@ impl ViewCtx {
...
-    pub(crate) fn set_state_changed(&mut self, value: bool) {
+    pub fn set_state_changed(&mut self, value: bool) {
         self.state_changed = value;
     }
 
-    pub(crate) fn get_id_path(&self, widget_id: WidgetId) -> Option<&Vec<ViewId>> {
+    pub fn get_id_path(&self, widget_id: WidgetId) -> Option<&Vec<ViewId>> {
         self.widget_map.get(&widget_id)
     }
 }
```
DJMcNab, line 57: I won't approve making these be public API; this is part of the driver side, not the app side. Every external user of Xilem should be using the driver, which carefully encapsulates what a (potentially embedded) Xilem app needs to know about the outside world.

PoignardAzur: Yeah, I think making these `pub` is a symptom I got the separation wrong.

PoignardAzur:
Okay, so the main problem is that the driver *heavily* depends on Winit, and the main objective of this is to not make `xilem_masonry` depend on Winit.

There's probably a solution to be found, but it's hard to disentangle.

PR #1016, opened 2025-06-03
Split time computations out of RenderRoot

PoignardAzur:


masonry_winit/src/event_loop_runner.rs
```diff
@@ -508,7 +513,23 @@ impl MasonryState<'_> {
...
+                let elapsed = last.map(|t| now.duration_since(t)).unwrap_or_default();
+
+                self.render_root
+                    .handle_window_event(WindowEvent::AnimFrame(elapsed));
+
+                // If this animation will continue, store the time.
+                // If a new animation starts, then it will have zero reported elapsed time.
+                let animation_continues = self.render_root.needs_anim();
+                self.last_anim = animation_continues.then_some(now);
+
```
waywardmonkeys, line 532: I'm not crazy about this requiring code in each windowing integration. Why's this an improvement over the render root knowing about it?

PoignardAzur: It keeps all the side-effects in the integration, which is in line with how other events integrate with RenderRoot.

waywardmonkeys: I think the counter argument to what I said is something like a web_sys adapter where it is using requestAnimationFrame which supplies the time step...

PoignardAzur: My specific use-case was unit tests, but yeah, in general you do want to keep side-effects in the event loop code.

PR #1017, opened 2025-06-03
Make TestHarness not use private items

PoignardAzur:


masonry/src/testing/harness.rs
```diff
@@ -244,7 +243,8 @@ impl TestHarness {
             render_root: RenderRoot::new(
                 root_widget,
                 RenderRootOptions {
-                    default_properties: default_property_set(),
+                    // TODO - Pass the default property set as an input instead.
+                    default_properties: crate::theme::default_property_set(),
```
waywardmonkeys, line 247: Slightly confused about having this here, but okay.

waywardmonkeys (approved):

PR #1022, opened 2025-06-03
Refactor some layout methods to only use public items

PoignardAzur:
This also removes the call to `set_paint_insets` in `Flex::layout`, which was redundant since the layout pass already computes paint insets on its own.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -859,7 +859,7 @@ impl Iterator for Spacing {
             if self.n_children == 0 {
                 self.extra
             } else {
-                #[allow(clippy::match_bool)]
+                #[allow(clippy::match_bool, reason = "readability")]
                 match self.alignment {
                     MainAxisAlignment::Start => match self.index == self.n_children {
                         true => self.extra,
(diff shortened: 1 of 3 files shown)


masonry/src/widgets/flex.rs
```diff
@@ -1189,7 +1188,6 @@ impl Widget for Flex {
 
                     let child_pos: Point = self.direction.pack(major, child_minor_offset).into();
                     ctx.place_child(widget, child_pos);
-                    child_paint_rect = child_paint_rect.union(ctx.widget_state.paint_rect());
```
waywardmonkeys, line 1192: Was this (unused) `child_paint_rect` being accumulated for anything or is vestigial?

waywardmonkeys (approved):

waywardmonkeys:
It is a little confusing at times having to go to the context for things ... but that's the design, I guess.

masonry/src/widgets/flex.rs
PoignardAzur: It was used in the call to set_paint_insets, which itself was vestigial.

PR #1024, opened 2025-06-03
Change find_widget_under_pointer signature

PoignardAzur:
The old signature took a WidgetRef, which external crates cannot construct.


waywardmonkeys (approved):
Would be nice if it had a commit message explaining why.

PR #1026, opened 2025-06-03
Tweak error case for event pass

PoignardAzur:
Replaces the error log with a debug_panic.
This also moves the check outside the loop, since it was only ever going to fail in the first iteration anyway.


xStrom:
This seems more than just tweaking the message. Can it no longer happen to other widgets in the tree as it's now out of the loop?

PR #1027, opened 2025-06-03
Rewrite TestHarness constructors to take default properties as a param.

PoignardAzur:
This will be necessary once testing and theming are split into two crates.

### DIFF
--- a/masonry/src/widgets/button.rs
+++ b/masonry/src/widgets/button.rs
@@ -274,6 +274,7 @@ mod tests {
     use crate::core::{PointerButton, StyleProperty};
     use crate::testing::{TestHarness, TestWidgetExt, widget_ids};
     use crate::theme::PRIMARY_LIGHT;
+    use crate::theme::default_property_set;
     use crate::widgets::{Grid, GridParams, SizedBox};
 
     #[test]
@@ -282,7 +283,8 @@ mod tests {
         let widget = Button::new("Hello").with_id(button_id);
 
         let window_size = Size::new(100.0, 40.0);
-        let mut harness = TestHarness::create_with_size(widget, window_size);
+        let mut harness =
+            TestHarness::create_with_size(default_property_set(), widget, window_size);
 
         assert_debug_snapshot!(harness.root_widget());
         assert_render_snapshot!(harness, "button_hello");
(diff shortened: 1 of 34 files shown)


masonry/src/widgets/button.rs
```diff
@@ -274,6 +274,7 @@ mod tests {
     use crate::core::{PointerButton, StyleProperty};
     use crate::testing::{TestHarness, TestWidgetExt, widget_ids};
     use crate::theme::PRIMARY_LIGHT;
+    use crate::theme::default_property_set;
```
waywardmonkeys: Merge with the one above.

masonry/src/widgets/checkbox.rs
```diff
@@ -260,14 +260,16 @@ mod tests {
     use crate::core::StyleProperty;
     use crate::testing::{TestHarness, TestWidgetExt, widget_ids};
     use crate::theme::PRIMARY_LIGHT;
+    use crate::theme::default_property_set;
```
waywardmonkeys: Merge here too

masonry/src/widgets/label.rs
```diff
@@ -491,6 +491,7 @@ mod tests {
     use super::*;
     use crate::assert_render_snapshot;
     use crate::testing::TestHarness;
+    use crate::theme::default_property_set;
```
waywardmonkeys: Merge with the line below

masonry_winit/examples/calc_masonry.rs
```diff
@@ -445,12 +445,13 @@ fn main() {
 mod tests {
     use masonry_winit::assert_render_snapshot;
     use masonry_winit::testing::TestHarness;
+    use masonry_winit::theme::default_property_set;
```
waywardmonkeys, line 448: This bad pattern should be fixed of using masonry via re-exports from masonry_winit

PoignardAzur: From the discussions we've had, we'll probably stop re-exporting `masonry` items from `masonry_winit` soon anyway.

waywardmonkeys: Yeah, not saying it has to happen in this PR (it shouldn't), but it needs to happen soon or have an issue filed to track it.

PR #1029, opened 2025-06-03
Split modules of masonry crate into masonry_widgets and masonry_testing

PoignardAzur:


PoignardAzur:
Depends on #1027 and #1028.

PoignardAzur:
This ended up being a larger PR than I expect, even discarding the file renames.

Honestly though, I think I'd have a hard time making the diff smaller without spending one more week on it. And given how many files this PR touches, I'd rather expedite it and fix the problems afterwards.

All the changes in this PR are non-functional. The biggest changes are to docs: because some parts of the doc need to refer to items that are now part of a parent/sibling crate, I've replaced the "path links" with URL links. Doing so is *much* more fragile, and we should probably open an issue to examine different options, but it was already a problem before this PR (masonry needing to document masonry_winit).

waywardmonkeys:
The (dev-)dependency of `masonry_widgets` on `masonry_testing` means that we can't just do testing first and leave widgets where they are, right?

PoignardAzur:
Yup.

PR #1030, opened 2025-06-04
refactor: stop using RawProxy for driver constructor

not-my-profile:
MasonryUserEvent will gain a WindowId but RawProxy::send_event won't.

Note that we're not using EventLoopProxy instead because we want the driver to be constructible without an event loop for testing (currently not feasible with winit 0.30 but will be with 0.31 which introduces traits for ActiveEventLoop and Window).

### DIFF
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -3,9 +3,10 @@
 
 #![expect(missing_docs, reason = "TODO - Document these items")]
 
+use std::fmt::Debug;
 use std::sync::Arc;
 
-use masonry_winit::app::{AppDriver, EventLoopProxy, MasonryState, MasonryUserEvent};
+use masonry_winit::app::{AppDriver, MasonryState, MasonryUserEvent};
 use masonry_winit::core::WidgetId;
 use masonry_winit::peniko::Blob;
 use masonry_winit::widgets::RootWidget;
@@ -31,7 +32,9 @@ where
     pub(crate) fn new(
         mut state: State,
         mut logic: Logic,
-        proxy: Arc<dyn RawProxy>,
+        // TODO: narrow down MasonryUserEvent in event_sink once masonry_winit supports custom event types
+        // (we only ever use it to send MasonryUserEvent::Action with ASYNC_MARKER_WIDGET)
+        event_sink: impl Fn(MasonryUserEvent) -> Result<(), MasonryUserEvent> + Send + Sync + 'static,
         runtime: tokio::runtime::Runtime,
         fonts: Vec<Blob<u8>>,
     ) -> (Self, Pod<View::Widget>) {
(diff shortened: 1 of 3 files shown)


xilem/src/lib.rs
```diff
@@ -277,15 +279,19 @@ where
...
     }
 
+    /// Builds the root widget and [`MasonryDriver`].
+    ///
+    /// The given event sink function sends the given event to the event loop
+    /// and returns the given event as an error in case the event loop is stopped.
     pub fn into_driver(
         self,
-        proxy: Arc<dyn RawProxy>,
+        event_sink: impl Fn(MasonryUserEvent) -> Result<(), MasonryUserEvent> + Send + Sync + 'static,
```
not-my-profile, line 287: The `MasonryDriver` only ever uses this to send `MasonryUserEvent::Action`, it probably makes sense to narrow down the type here accordingly?

xilem/src/driver.rs
```diff
@@ -88,18 +88,15 @@ impl RawProxy for MasonryProxy {
         }
     }
     fn dyn_debug(&self) -> &dyn std::fmt::Debug {
-        self
+        #[derive(Debug)]
+        struct MasonryProxy;
+        &MasonryProxy
     }
```
PoignardAzur: Wait, what?

not-my-profile: The `dyn Fn` doesn't implement `Debug` ... but yeah I can see how this is confusing. Changed to a manual `Debug` impl with `finish_non_exhaustive`.

PoignardAzur: No, I mean, why is a struct defined in the function body? I would expect it to be at the top level. Sorry, I realize I wasn't specific.

not-my-profile: Well it was never used outside this function. No worries. An explicit Debug impl is better anyway because it has the `..` indicating that it's not exhaustive.

xilem/src/lib.rs
not-my-profile: I think this can only be done nicely once masonry_winit supports custom user event types. I added a respective TODO.

PoignardAzur (approved):
I don't quite understand the rationale for this, but I don't see anything wrong.

LGTM.

not-my-profile:
The rationale is in the description:

> MasonryUserEvent will gain a WindowId but RawProxy::send_event won't.

(RawProxy::send_event won't have the window id as a param because it's already used in the context of a window).

not-my-profile:
For introducing multiple window support I had to add a `WindowId` into the `MasonryUserEvent::Action` tuple. And yes this proxy/sink/whatever-you-wanna-call-it is now masonry_winit-specific because `WindowId` is also very much winit-specific.

PR #1031, opened 2025-06-04
feat: introduce WidgetMut::find_mut

not-my-profile:
Finding widgets by their id in the RenderRoot::edit_root_widget callback makes the code more maintainable because you can rearrange the widgets without having to update the callback. And with the unsafe tree arena find_mut from the root is O(1).

To make this pattern convenient this commit also introduces a Default impl for WidgetId that just returns WidgetId::next().

### DIFF
--- a/masonry/src/core/contexts.rs
+++ b/masonry/src/core/contexts.rs
@@ -257,6 +257,41 @@ impl MutateCtx<'_> {
         }
     }
 
+    /// Return a [`WidgetMut`] to a child widget by its id.
+    pub fn find_mut(&mut self, child_id: WidgetId) -> WidgetMut<'_, dyn Widget> {
+        let child_state_mut = self
+            .widget_state_children
+            .reborrow_mut()
+            .find_mut(child_id)
+            .expect("find_mut: child not found");
+        let child_mut = self
+            .widget_children
+            .reborrow_mut()
+            .find_mut(child_id)
+            .expect("find_mut: child not found");
+        let child_properties = self
+            .properties_children
+            .reborrow_mut()
+            .find_mut(child_id)
+            .expect("find_mut: child not found");
+        let child_ctx = MutateCtx {
+            global_state: self.global_state,
+            parent_widget_state: Some(&mut self.widget_state),
+            widget_state: child_state_mut.item,
+            widget_state_children: child_state_mut.children,
+            widget_children: child_mut.children,
+            properties: PropertiesMut {
+                map: child_properties.item,
+                default_map: self.properties.default_map,
+            },
+            properties_children: child_properties.children,
+        };
+        WidgetMut {
+            ctx: child_ctx,
+            widget: child_mut.item.as_mut_dyn(),
+        }
+    }
+
     pub(crate) fn reborrow_mut(&mut self) -> MutateCtx<'_> {
         MutateCtx {
             global_state: self.global_state,
--- a/masonry/src/core/widget.rs
+++ b/masonry/src/core/widget.rs
@@ -50,6 +50,12 @@ impl WidgetId {
     }
 }
 
+impl Default for WidgetId {
+    fn default() -> Self {
+        Self::next()
+    }
+}
+
 /// A trait to access a `Widget` as trait object. It is implemented for all types that implement `Widget`.
 pub trait AsDynWidget {
     fn as_box_dyn(self: Box<Self>) -> Box<dyn Widget>;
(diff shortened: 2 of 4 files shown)


masonry/src/core/contexts.rs
```diff
@@ -257,6 +257,41 @@ impl MutateCtx<'_> {
         }
     }
 
+    /// Return a [`WidgetMut`] to a child widget by its id.
+    pub fn find_mut(&mut self, child_id: WidgetId) -> WidgetMut<'_, dyn Widget> {
+        let child_state_mut = self
+            .widget_state_children
+            .reborrow_mut()
+            .find_mut(child_id)
+            .expect("find_mut: child not found");
```
PoignardAzur, line 266: So, correct me if I'm missing something, but the problem with that implementation is that WidgetMut will not correctly update the flags of any intermediary nodes between the current widget and the descendant widget.

Like, say you have a `A -> B -> C -> D -> E` chain and you do `A.ctx.find_mut(E)` and then update some stuff; the flags for A and E will be updated, but not the flags for B, C and D.

You need something like #1011 to make this method possible.

masonry/src/core/widget.rs
```diff
@@ -50,6 +50,12 @@ impl WidgetId {
     }
 }
 
+impl Default for WidgetId {
+    fn default() -> Self {
+        Self::next()
+    }
+}
+
```
PoignardAzur, line 58: Big no to this. `Default` implementations that return different things if you call them twice are a huge code smell.

masonry_winit/examples/calc_masonry.rs
```diff
@@ -283,10 +284,7 @@ impl AppDriver for CalcState {
...
 
         ctx.render_root().edit_root_widget(|mut root| {
-            let mut root = root.downcast::<RootWidget>();
-            let mut flex = RootWidget::child_mut(&mut root);
-            let mut flex = flex.downcast();
-            let mut label = Flex::child_mut(&mut flex, 1).unwrap();
+            let mut label = root.ctx.find_mut(self.ids.display);
             let mut label = label.downcast::<Label>();
             Label::set_text(&mut label, &*self.value);
         });
```
PoignardAzur, line 290: Note that if you already know the id of the widget you're trying to mutate, you don't need a new method:

```rust
         ctx.render_root().edit_widget(self.ids.display, |mut label| {
            let mut label = label.downcast::<Label>();
            Label::set_text(&mut label, &*self.value);
         };
```

PoignardAzur: By the way, my medium-term plan for this kind of use-case is to create a new kind of string id (something like `WidgetKey`) that could be assigned to a widget at construction and let you identify them, basically an equivalent of the DOM `id` attribute.

masonry_winit/examples/calc_masonry.rs
```diff
@@ -347,12 +345,17 @@ fn flex_row(
         .with_flex_child(w4, 1.0)
 }
 
-fn build_calc() -> impl Widget {
+#[derive(Default, Clone)]
+struct WidgetIds {
+    display: WidgetId,
+}
```
PoignardAzur, line 351: This doesn't need to implement Default. You could add a `new()` or `create()` method.

PoignardAzur (changes requested):
I get the idea behind this, and it's something we'll eventually support first-class, but right now it doesn't really work.

masonry/src/core/contexts.rs
not-my-profile: Oh, right yeah. Forgot about that.

masonry_winit/examples/calc_masonry.rs
not-my-profile: > you don't need a new method

Oh right yeah. I'll create a new PR to use that in the examples.

> my medium-term plan for this kind of use-case is to create a new kind of string id

I really don't see how that would be better than just using WidgetIds. If we have sth else I don't think it should be strings because of type-safety.

PoignardAzur:
Honestly, that doesn't matter. We're talking about doing two `O(Depth)` operations per user input. Even if the widget tree were huge (it very much isn't), that would be acceptable on principle.

PoignardAzur:
Okay, so having re-read the code:

- Mutate passes are `O(Depth)` and are okay to re-run.
- Most update passes will just be skipped since we're not doing anything to change their values.
- Layout passes currently basically recompute everything, so they're `O(N)`.

The examples do change layout (adding buttons or changing text counts), so doubling rewrite passes isn't great for performance. Of course they're small enough that performance doesn't matter, but I get that this isn't the pattern we want to encourage.

I'm thinking this definitely needs to be documented somewhere.

PR #1034, opened 2025-06-05
Fix focus chain code

PoignardAzur:
Fix bug where parts of the focus chain would disappear when widgets updated. Make Button accept focus.
Make Prose not accept focus.
Add TextEvent constructors.

### DIFF
--- a/masonry/src/widgets/button.rs
+++ b/masonry/src/widgets/button.rs
@@ -8,7 +8,7 @@ use std::any::TypeId;
 use accesskit::{Node, Role};
 use smallvec::{SmallVec, smallvec};
 use tracing::{Span, trace, trace_span};
-use ui_events::keyboard::Key;
+use ui_events::keyboard::{Key, NamedKey};
 use vello::Scene;
 use vello::kurbo::{Affine, Size};
 use vello::peniko::Color;
@@ -125,10 +125,12 @@ impl Widget for Button {
     ) {
         const SPACE: &str = " ";
         match event {
-            TextEvent::Keyboard(event)
-                if event.key == Key::Character(SPACE.into()) && event.state.is_up() =>
-            {
-                ctx.submit_action(Action::ButtonPressed(None));
+            TextEvent::Keyboard(event) if event.state.is_up() => {
+                if event.key == Key::Character(SPACE.into())
+                    || event.key == Key::Named(NamedKey::Enter)
+                {
+                    ctx.submit_action(Action::ButtonPressed(None));
+                }
             }
             _ => (),
         }


masonry/src/widgets/button.rs
```diff
@@ -226,12 +237,19 @@ impl Widget for Button {
...
         } else {
             props.get::<BorderColor>()
         };
 
+        // FIXME - Handle this properly
+        if ctx.is_focus_target() {
+            border_color = &BorderColor {
+                color: Color::WHITE,
+            };
+        }
```
not-my-profile, line 255: Drawing the focus outline should be done centrally in the `paint` pass. Not inside of the widget.

PoignardAzur: Disagree. I think it should ultimately be stylable, once we add pseudo-classes to the styling system.

not-my-profile: I strongly think widgets shouldn't have to do anything to get a focus outline. I think that's important for accessibility, or otherwise there is bound to be widgets where it has been forgotten. I think it could still be stylable even if provided outside of the Widget impl.

PoignardAzur: That's a valid point. I'd recommend opening an issue about it, because I don't think this needs to be resolved in this PR.

not-my-profile: My point is that globally providing this by default is more important than making it customizable, so I'd rather see this implemented in the `paint` pass now and implementing customization afterwards.

PoignardAzur: There's a bunch of complications in implementing "paint an outline by default on any focused widget" that I don't really want to deal with right now.

Seeing as that the status quo is that outline-by-default isn't implemented and this PR doesn't make it any less likely to get implemented, I'd rather avoid adding a drive-by requirement to it.

In other words, outline-by-default should be implemented in a separate PR. This PR fixes an important accessibility issue even without adding outline-by-default it.

not-my-profile: I had implemented this locally in 10 lines without noticing any complications (might not be perfect but still better than nothing) ... but sure I can make a PR for that.

masonry/src/widgets/button.rs
```diff
@@ -115,10 +117,19 @@ impl Widget for Button {
...
-        _ctx: &mut EventCtx,
+        ctx: &mut EventCtx,
         _props: &mut PropertiesMut<'_>,
-        _event: &TextEvent,
+        event: &TextEvent,
     ) {
+        const SPACE: &str = " ";
+        match event {
+            TextEvent::Keyboard(event)
+                if event.key == Key::Character(SPACE.into()) && event.state.is_up() =>
```
DJMcNab: Should we also support <kbd>↵</kbd> here?

masonry/src/widgets/text_area.rs
```diff
@@ -796,7 +796,7 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
     }
 
     fn accepts_focus(&self) -> bool {
-        true
+        EDITABLE
```
DJMcNab, line 800: Can you confirm that `accepts_focus` is *only* related to the focus chain? If so, I believe this behaviour to be correct.

The one complication is that it should ideally still be possible to select text inside a prose from only the keyboard, but I'm happy to defer that as a longer-term consideration.

DJMcNab (approved):
LGTM, thanks. A couple of minor questions

masonry/src/widgets/text_area.rs
PoignardAzur: > Can you confirm that accepts_focus is only related to the focus chain?

Yup. The value is read in only two places in the code, both of them related to tab-focus (building the focus chain, and enabling `accesskit::Action::Focus`).

masonry/src/widgets/button.rs
PoignardAzur: Quick test with Firefox suggests "yes".

masonry/src/widgets/text_area.rs
DJMcNab: Hmm... I wonder if we should allow accesskit focus, but not tab focus here

masonry/src/widgets/button.rs
```diff
@@ -117,10 +119,21 @@ impl Widget for Button {
...
-        _ctx: &mut EventCtx<'_>,
+        ctx: &mut EventCtx<'_>,
         _props: &mut PropertiesMut<'_>,
-        _event: &TextEvent,
+        event: &TextEvent,
     ) {
+        const SPACE: &str = " ";
+        match event {
+            TextEvent::Keyboard(event) if event.state.is_up() => {
+                if event.key == Key::Character(SPACE.into())
```
xorgy, line 129: When checking for this, it is better to use the code field, especially if you are not in a pattern match.

https://docs.rs/ui-events/latest/ui_events/keyboard/enum.Code.html#variant.Space

xorgy: Fixed in #1072 @PoignardAzur

PR #1035, opened 2025-06-05
Remove tests and docgen from the xilem examples

PoignardAzur:
None of them have any tests and the doc is just a title, but the test runner still took the time to visit them.


waywardmonkeys:
Would this have an impact on the rustdoc flag for extracting examples from the tests?

PoignardAzur:
What flag?

waywardmonkeys:
https://doc.rust-lang.org/rustdoc/scraped-examples.html

PR #1038, opened 2025-06-07
masonry_winit: support multiple windows

not-my-profile:


xilem/src/driver.rs
```diff
@@ -37,12 +38,13 @@ where
         event_sink: impl Fn(MasonryUserEvent) -> Result<(), MasonryUserEvent> + Send + Sync + 'static,
         runtime: tokio::runtime::Runtime,
         fonts: Vec<Blob<u8>>,
-    ) -> (Self, Pod<View::Widget>) {
+    ) -> (Self, (WindowId, Box<dyn Widget>)) {
```
PoignardAzur:
```suggestion
    ) -> (Self, WindowId, Box<dyn Widget>) {
```

masonry_winit/src/app_driver.rs
```diff
@@ -1,59 +1,161 @@
...
+    pub fn trace(self) -> DisplayValue<NonZeroU64> {
+        tracing::field::display(self.0)
+    }
+}
+
+/// Context for the [`AppDriver`] trait.
+pub struct DriverCtx<'a, 's> {
+    state: &'a mut MasonryState<'s>,
+    event_loop: &'a ActiveEventLoop,
+}
```
PoignardAzur, line 48: Types with two lifetime parameters are kind of a pain to use.

If there's any possibility to make this a one-lifetime type, I'd like that instead. If not, I'd want to at least document the reasons, maybe add a TODO comment.

PoignardAzur (changes requested):
Looks really good, aside from a few nits!

masonry_winit/src/app_driver.rs
not-my-profile: `&'a mut Something<'a>` is never what you want. Because of invariance the receiver is borrowed for the rest of its lifetime.

PoignardAzur: Right, but usually the way we avoiding it is by avoiding referencing something that's a reference type. In our case, that could mean splitting away `MasonryState::surfaces`, for example.

Regardless, I'm not going to hold up the PR on it.

not-my-profile: The context can be used to create a new window including it's surface so splitting this up wouldn't be completely trivial. But yeah I agree we can do this later.

PR #1040, opened 2025-06-07
Normalize imports according to usual coding style

PoignardAzur:
Replace `crate::vello/kurbo/parley` with `vello/kurbo/parley`.
Group imports with the usual std-imports-crate grouping.
Import `debug_panic` manually in each file instead of using syntactic macro lookup.

These changes will make the crate split easier.

### DIFF
--- a/masonry/src/app/render_root.rs
+++ b/masonry/src/app/render_root.rs
@@ -4,6 +4,7 @@
 use std::collections::{HashMap, VecDeque};
 
 use accesskit::{ActionRequest, TreeUpdate};
+use cursor_icon::CursorIcon;
 use parley::fontique::{self, Blob, Collection, CollectionOptions, SourceCache};
 use parley::{FontContext, LayoutContext};
 use tracing::{info_span, warn};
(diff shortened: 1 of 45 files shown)


PoignardAzur:
> We spend a lot of effort doing the opposite

How so?

Either way, I don't see a strong reason for the `use crate::vello/kurbo/parley` pattern, and it's applied very inconsistently throughout the project. If it's supposed to be the desired coding style, then it's not documented anywhere.

PoignardAzur:
After doing some git blames, I see:

- https://github.com/linebender/xilem/pull/848: Replaced a lot of uses of `crate::{Point, Size, etc}` with `crate::kurbo::{Point, Size, etc}`.
- https://github.com/linebender/xilem/pull/223: Replaced `kurbo::{Point, Size, etc}` with either `crate::kurbo::{Point, Size, etc}` or `vello::kurbo::{Point, Size, etc}` depending on files. The rationale was that we wanted to stop importing kurbo directly, which doesn't really apply here.

Overall, I think we just haven't officially decided on a standard between `crate::kurbo::{xxx}` and `vello::kurbo::{xxx}`, and we've used both throughout the codebase. I strongly prefer the latter.

waywardmonkeys:
I guess I don’t actually care enough to argue about it. It is all bad anyway.

waywardmonkeys (approved):

PR #1045, opened 2025-06-07
masonry_winit: Pass a EventLoopProxy

waywardmonkeys:
Instead of passing an `&EventLoop` to `MasonryState`, we pass an `EventLoopProxy` as that's all that it actually needs.

We also rename `proxy` to `event_loop_proxy`.


PoignardAzur (approved):
LGTM.

PR #1046, opened 2025-06-08
`masonry_winit`: Stop exporting `masonry`

waywardmonkeys:
People should use `masonry` for all of the general purpose stuff and only use `masonry_winit` for the Winit integration.

Further cleanups to the `masonry_winit` crate can come after this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4934,6 +4934,7 @@ dependencies = [
  "accesskit",
  "anyhow",
  "image",
+ "masonry",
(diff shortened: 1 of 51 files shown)


waywardmonkeys:
Marking as merge when ready, so if you don't want your approval to mean it merges, feel free to remove it from the queue. (I'm not around for a few hours and hope this is able to land.)

PoignardAzur (approved):

PR #1048, opened 2025-06-08
Split off testing into masonry_testing crate

PoignardAzur:


masonry_testing/README.md
DJMcNab, line 1: I think this should probably have a more complete README.

PoignardAzur: I can add a placeholder description, but for a proper README I'd rather wait until all the crate splits are done, so I can document everything at once and have a clear picture of what the complete documentation story looks like.

DJMcNab: (We do need this crate does need to have the license file, if nothing else)

masonry/Cargo.toml
```diff
@@ -23,17 +23,13 @@ tracy = ["masonry_core/tracy"]
 accesskit.workspace = true
 cursor-icon = "1.1.0"
 dpi.workspace = true
-futures-intrusive = "0.5.0"
-image.workspace = true
 masonry_core.workspace = true
-oxipng = { version = "9.1.5", default-features = false }
+masonry_testing.workspace = true
```
DJMcNab: This doesn't seem right. We plan to rename the current `masonry` to `masonry_widgets`, right? And so this crate itself largely only contains widget implementations; it should be therefore acting like any other widget library, by only depending on the test code as a dev-dependency, I think

masonry/src/lib.rs
```diff
@@ -47,8 +46,10 @@ pub use vello::{kurbo, peniko};
 pub use {cursor_icon, dpi, parley, vello};
 
 pub use masonry_core::{app, core, util};
+pub use masonry_testing as testing;
```
DJMcNab, line 50: Oh, I see. This state is just an interim state, right?
We should add a comment here to that effect

masonry_testing/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+
+[package.metadata.docs.rs]
+all-features = true
+# There are no platform specific docs.
+default-target = "x86_64-unknown-linux-gnu"
+targets = []
+
+[features]
+default = []
+tracy = ["masonry_core/tracy"]
```
DJMcNab: I'm not sure that this is valuable

masonry_testing/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+web-time.workspace = true
+
+[dev-dependencies]
+assert_matches = "1.5.0"
+float-cmp = { version = "0.10.0", features = ["std"], default-features = false }
+image = { workspace = true, features = ["png"] }
+insta = { version = "1.43.1" }
+
+# Make wgpu use tracing for its spans.
+profiling = { version = "1.0.16", features = ["profile-with-tracing"] }
```
DJMcNab: This is also a little bit surprising; this new crate doesn't have any examples, and I'm not sure it's all that meaningful for it to have self tests

DJMcNab: I think the only one of these which is used is `insta`, for docs. But that uses the deprecated (iirc) form of tests, of the not very useful debug snapshots

masonry_testing/src/recorder_widget.rs
```diff
@@ -19,24 +19,27 @@ use smallvec::SmallVec;
...
 
-use crate::core::{
+use masonry_core::core::{
     AccessCtx, AccessEvent, BoxConstraints, ComposeCtx, EventCtx, LayoutCtx, PaintCtx,
     PointerEvent, PropertiesMut, PropertiesRef, QueryCtx, RegisterCtx, TextEvent, Update,
     UpdateCtx, Widget, WidgetId, WidgetRef,
 };
 
+// TODO - Re-enable doc test.
+// Doc test is currently disabled because it depends on a parent crate.
```
DJMcNab, line 29: I *think* this is the same state as https://github.com/rust-lang/cargo/issues/4242, which is workaroundable; this needn't happen now

DJMcNab (commented):
Overall I like this pattern - a few things which I think need to be addressed before we can land this.

masonry/Cargo.toml
PoignardAzur: My current plan is to also split masonry_widgets into a separate crate, and to have masonry basically be a thin shell that re-exports everything else (expect maybe the tutorial docs).

Either way, I can make it an optional dependency.

masonry_testing/Cargo.toml
PoignardAzur: The PNG feature is also needed (actually, it should be in dependencies, not dev-dependencies), but otherwise, yes, much of this can be removed.

masonry_testing/src/recorder_widget.rs
PoignardAzur: Kind of, yes, but I'd like to split off masonry_widgets first.

masonry_testing/Cargo.toml
PoignardAzur: Mhh. Yeah, I'll remove it. Anyone who really wants to import `masonry_testing` but not `masonry` and also wants to use that feature is already determined enough to directly add it through `masonry_core`.

masonry/Cargo.toml
waywardmonkeys: I thought the plan after last week was to *not* have a `masonry_widgets` since it sends a bad message.

PoignardAzur: Well this is why I wanted to discuss this at Office Hours, but I forgot to bring it up last week.

But yeah, we need to find a consensus. I think it's possible to split the crates the way I described without sending the message you were worried about (that the default widgets are supposed to be throwaway code).

PoignardAzur: Either way, whether we have another crate split or not, a feature-gated dependency is the best option.

DJMcNab (commented):
I think code-wise, this is a very reasonable implementation. I do however want to understand the motivation for this. My understanding before the discussion in this PR was that the plan was:

```mermaid
graph LR
  subgraph Masonry Crates
    masonry
    masonry_widgets
    masonry_testing
    masonry_winit
  end
  subgraph Third Party
    masonry_material
  end
  my_app[End User App]
  
  masonry_widgets --> masonry
  masonry_testing --> masonry
  masonry_widgets -- (dev) --> masonry_testing
  masonry_material --> masonry
  masonry_material -- (dev) --> masonry_testing
  my_app -- (dev) --> masonry_testing
  my_app --> masonry_material
  my_app -- (if needed, e.g. for baseline layout widgets) --> masonry_widgets
  masonry_winit --> masonry
  my_app --> masonry_winit
```

Which is a neat user story of everyone always depending only on the things which they need.

But seeing:
> We can enable that with feature flags, but even then it's simpler to implement if all the testing code and dependencies are in a self-contained crates (rather than adding cfgs everywhere).

and https://github.com/linebender/xilem/pull/1048#discussion_r2137796498 makes it seem that the plan is:

```mermaid
graph LR
  subgraph Masonry Crates
    masonry
    masonry_core
    masonry_testing
    masonry_winit
    masonry_widgets
  end
  subgraph Third Party
    masonry_material
  end
  my_app[End User App]
  
  masonry --> masonry_winit
  masonry_widgets --> masonry_core
  masonry_widgets -- (dev) --> masonry_testing
  masonry --> masonry_core
  masonry --> masonry_widgets
  masonry -- (dev) --> masonry_testing
  masonry -- (with 'testing' feature) --> masonry_testing
  masonry_testing --> masonry_core
  masonry_material -- (dev)+testing --> masonry
  masonry_material --> masonry
  my_app -- (dev)+testing --> masonry
  my_app --> masonry_material
  my_app --> masonry
```

Where all external users solely depend on the `masonry` "octopus" crate, and other crate splits are entirely for Masonry's own internal needs? And so the only remaining motivation for this split is to avoid `cfg(any(feature="testing", test))` in a few places? If we do decide to have the `masonry_widgets` crate, then this does become well-motivated, but I'm also not entirely sure why we would have the `masonry_widgets` crate in this view of the world?

This is probably falling back to the question of "do we expect every end user to depend on masonry's built-in widgets?" Until very recently, the state was such that the answer was clearly no. I'm still not sure how `masonry_widgets` as designed can reasonably expect to be used in the context of a design language; that's the main reason I was working on the assumption that the answer was still no. But it seems possible that I'm missing something there, because the only way such an uber "masonry" crate makes sense is if we do expect that and so want to privilege the built-in widgets.

PoignardAzur:
> Where all external users solely depend on the `masonry` "octopus" crate, and other crate splits are entirely for Masonry's own internal needs? And so the only remaining motivation for this split is to avoid `cfg(any(feature="testing", test))` in a few places? If we do decide to have the `masonry_widgets` crate, then this does become well-motivated, but I'm also not entirely sure why we would have the `masonry_widgets` crate in this view of the world?

I'm not sure what the big issue is.

We would have `masonry_widgets` because we want to provide a set of widgets, properties and default styles separate from `masonry_core` and `masonry_testing`. 

We can call the crate providing those widgets and properties `masonry`, but I would like to keep `masonry` as "the octopus crate that re-exports everything else and doesn't implement anything on its own". The reason I want to keep "re-exporting everything" and "defining default widgets/props" as separate crate is so that each crate as a well-defined scope, and it's easy to communicate what importing any given crate does.

The reason I want an "octopus" crate is so that new users don't need to engage with the crate hierarchy. Tutorials can just mention a single `cargo add masonry` and then they can get started.

So in your example, I would expect the third-party developer of `masonry_material` to first make their crate depend on `masonry` and then, after they get some experience with the ecosystem and/or some of their users create issues about it, realize they can depend on `masonry_core` instead.

PoignardAzur:
Created a discussion here: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Plan.20for.20splitting.20off.20Masonry/with/523520561

masonry_testing/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "masonry_testing"
+version.workspace = true # We mimic Xilem's version
+description = "Traits and types of the Masonry toolkit."
```
DJMcNab: Is this true?

DJMcNab (approved):
Approving to unblock work. I don't like the re-export; it is the wrong level of abstraction, but as mentioned on Zulip, I won't block on it.

PR #1049, opened 2025-06-08
Split xilem lib.rs into multiple files

PoignardAzur:
Having all the main items in one file made that code especially hard to read and review. Splitting across files should hopefully make that easier.


xilem/src/any_view.rs
```diff
@@ -28,7 +28,7 @@ pub type AnyWidgetView<State, Action = ()> =
 
 impl<W: Widget + FromDynWidget + ?Sized> SuperElement<Pod<W>, ViewCtx> for Pod<DynWidget> {
     fn upcast(ctx: &mut ViewCtx, child: Pod<W>) -> Self {
-        ctx.new_pod(DynWidget {
+        ctx.create_pod(DynWidget {
```
DJMcNab, line 31: I don't like that you've done this rename in this PR; what other changes other than the described one are made in the big blocks of removed/added code?

DJMcNab (commented):
This seems like a good split - I can't approve until I find the will to check that there aren't any other unmentioned changes in the moved code.

xilem/src/any_view.rs
PoignardAzur: Fine, I'll undo the rename.

DJMcNab: That's not strictly necessary. I just want confirmation that you haven't made any other drive-by-changes in this PR.

xilem/src/view_ctx.rs
```diff
@@ -0,0 +1,108 @@
...
+    ///
+    /// This is useful for views whose current value depends on current app state.
+    /// (That is, currently only virtual scrolling)
+    pub fn state_changed(&self) -> bool {
+        self.state_changed
+    }
+
+    pub fn set_state_changed(&mut self, value: bool) {
+        self.state_changed = value;
+    }
```
DJMcNab, line 89: Right: this method is also new; it's also incorrectly pub (it should only be set by the driver)

xilem/src/view_ctx.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    pub fn teardown_leaf<W: Widget + FromDynWidget + ?Sized>(&mut self, widget: WidgetMut<W>) {
+        self.widget_map.remove(&widget.ctx.widget_id());
+    }
+
+    pub fn get_id_path(&self, widget_id: WidgetId) -> Option<&Vec<ViewId>> {
+        self.widget_map.get(&widget_id)
+    }
```
DJMcNab, line 97: This new method seems... hard to use correctly, because it only applies for widgets which have called `with_action_widget`. Again, I *think* just making it `pub(crate)` would be reasonable here.

Xilem is currently only designed for use with its concrete driver; trying to make this a public-ish API feels like a bad idea.

xilem/src/view_ctx.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    fn view_path(&mut self) -> &[ViewId] {
+        &self.id_path
+    }
+}
+
+#[expect(missing_docs, reason = "TODO - Document these items")]
+impl ViewCtx {
+    pub fn new(proxy: Arc<dyn RawProxy>, runtime: tokio::runtime::Runtime) -> Self {
```
DJMcNab, line 43: Similarly here, I think `pub(crate)` is more correct for now

PoignardAzur:
See #1065.

xilem/src/view_ctx.rs
```diff
@@ -0,0 +1,108 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use masonry::core::{FromDynWidget, Widget, WidgetId, WidgetMut};
+
+use xilem_core::{AsyncCtx, RawProxy, ViewId, ViewPathTracker};
+
+use crate::Pod;
+
+type WidgetMap = HashMap<WidgetId, Vec<ViewId>>;
```
DJMcNab, line 13: I don't really think that this type alias adds a huge amount of value - it's only used once, anyway

DJMcNab (commented):
I've now reviewed the new code which was changed for privacy, I think

PR #1051, opened 2025-06-08
Fill in all elided lifetimes in Xilem

PoignardAzur:


burrbull:
I think this https://github.com/rust-lang/rust/pull/138677

not-my-profile:
I see. I suspected that there's a lint. I think it would be nice if we could enforce it in the CI.

DJMcNab (approved):
Thanks!

PR #1053, opened 2025-06-08
Add properties to Textbox

PoignardAzur:
Remove default padding around textbox.
Add some padding to examples.


masonry_winit/screenshots/example_calc_masonry_initial.png
DJMcNab, line 1: Two thoughts:
1) I think that the mauve here does not look good
2) The fact that this shows up in the screenshot tests is excellent

DJMcNab (approved):
Approving semi-optimistically, to keep things moving

masonry_winit/screenshots/example_calc_masonry_initial.png
PoignardAzur: > I think that the mauve here does not look good

:pleading_face: :pleading_face: :pleading_face:

PR #1054, opened 2025-06-08
Add Properties for Checkbox

PoignardAzur:


masonry/src/properties/checkmark.rs
```diff
@@ -0,0 +1,104 @@
...
+impl Property for DisabledCheckmarkColor {
+    fn static_default() -> &'static Self {
+        static DEFAULT: DisabledCheckmarkColor = DisabledCheckmarkColor(CheckmarkColor {
+            color: AlphaColor::BLACK,
+        });
+        &DEFAULT
+    }
+}
+
+/// The with the stroke with which a checkbox's "check" icon is drawn.
```
not-my-profile:
```suggestion
/// The width of the stroke with which a checkbox's "check" icon is drawn.
```

masonry/src/properties/checkmark.rs
```diff
@@ -0,0 +1,104 @@
...
+#[expect(missing_docs, reason = "obvious")]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct CheckmarkWidth {
```
not-my-profile: Perhaps rather `CheckmarkStrokeWidth`? This could be mistaken as the horizontal length of the checkmark.

masonry/src/properties/checkmark.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use std::any::TypeId;
+
+use crate::core::{Property, UpdateCtx};
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color of a checkbox's "check" icon.
+#[expect(missing_docs, reason = "obvious")]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct CheckmarkColor {
```
not-my-profile, line 17: `EnabledCheckmarkColor` for consistency with `DisabledCheckmarkColor`? (Assuming properties always have a value and they cannot be unset, is that true?)

masonry/src/theme.rs
```diff
@@ -92,6 +92,27 @@ pub fn default_property_set() -> DefaultProperties {
...
+        checkbox_gradient.clone(),
+    )));
+    properties.insert::<Checkbox, _>(DisabledBackground(Background::Gradient(
+        checkbox_gradient.clone(),
+    )));
+    properties.insert::<Checkbox, _>(CheckmarkWidth { width: 2.0 });
+    properties.insert::<Checkbox, _>(CheckmarkColor { color: TEXT_COLOR });
+    properties.insert::<Checkbox, _>(DisabledCheckmarkColor(CheckmarkColor {
+        color: DISABLED_TEXT_COLOR,
+    }));
```
not-my-profile: Always having to repeat `properties.insert::<Checkbox, _>` here doesn't look nice. I think it would be neat if we could avoid that somehow.

PoignardAzur: I think some amount of copy-pasting here is fine for the time being. It's a fairly small file anyway, and logically very simple.

masonry/src/properties/checkmark.rs
PoignardAzur: Properties can be unset.

But also, the different `DisabledFoobar/HoveredFoobar/ActiveFoobar` properties are kind of a hack anyway, we're bound to get rid of them soon, so don't want to worry about them too much.

masonry/src/properties/checkmark.rs
PoignardAzur: Probably, yeah.

masonry/src/properties/checkmark.rs
PoignardAzur: Good catch!

masonry/src/properties/checkmark.rs
not-my-profile: But if they're unset then the default value for them is used? That is they always have a value?

PoignardAzur: Yeah, right now there's no cascading between these different properties.

masonry/src/properties/checkmark.rs
```diff
@@ -0,0 +1,104 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::TypeId;
+
+use crate::core::{Property, UpdateCtx};
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color of a checkbox's "check" icon.
+#[expect(missing_docs, reason = "obvious")]
```
DJMcNab:
```suggestion
#[expect(missing_docs, reason = "field names are self-descriptive")]
```

masonry/src/properties/checkmark.rs
PoignardAzur: I've added a TODO comment about this, but I'm still calling it `CheckmarkColor` for consistency with other properties. If we do decide to rename all baseline properties, I'd rather we do it in a dedicated PR.

DJMcNab (approved):
LGTM. Thanks.

IWBN to see a screenshot test or two of at least one of the new properties.

PoignardAzur:
I think I'll procrastinate on that :no_mouth:

PR #1055, opened 2025-06-08
Add Properties to Flex and Grid

PoignardAzur:
Add padding to to_do_mvc example.
Remove uses of sized_box from examples.
Fix flex baseline default alignment.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -996,7 +996,7 @@ impl Widget for Flex {
         }
     }
 
-    fn property_changed(&mut self, ctx: &mut UpdateCtx, property_type: TypeId) {
+    fn property_changed(&mut self, ctx: &mut UpdateCtx<'_>, property_type: TypeId) {
         Background::prop_changed(ctx, property_type);
         BorderColor::prop_changed(ctx, property_type);
         BorderWidth::prop_changed(ctx, property_type);
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -3,7 +3,10 @@
 
 //! A widget that arranges its children in a one-dimensional array.
 
+use std::any::TypeId;
+
 use accesskit::{Node, Role};
+use masonry_core::core::RegisterCtx;
 use smallvec::SmallVec;
 use tracing::{Span, trace_span};
 use vello::Scene;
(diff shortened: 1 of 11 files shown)


masonry/src/widgets/flex.rs
```diff
@@ -1240,11 +1263,25 @@ impl Widget for Flex {
...
+        ctx.set_baseline_offset(baseline);
         my_size
     }
 
-    fn paint(&mut self, ctx: &mut PaintCtx, _props: &PropertiesRef<'_>, scene: &mut Scene) {
+    fn paint(&mut self, ctx: &mut PaintCtx, props: &PropertiesRef<'_>, scene: &mut Scene) {
+        let border_width = props.get::<BorderWidth>();
+        let border_radius = props.get::<CornerRadius>();
+        let bg = props.get::<Background>();
+        let border_color = props.get::<BorderColor>();
```
DJMcNab, line 1276: I'm not convinced that a pure layout type like this *should* have layout/border support; again, it's fine if we want to re-invent the web, but I don't particularly care to do so in this piecemeal fashion.

That being said, this is idiomatic for the current incarnation of properties

masonry/src/widgets/flex.rs
DJMcNab, line 1: Why is the flex gap not a property?

masonry/src/widgets/flex.rs
```diff
@@ -1007,11 +1032,7 @@ impl Widget for Flex {
         let mut max_below_baseline = 0_f64;
         let mut any_use_baseline = false;
 
-        // indicates that the box constrains for the following children have changed. Therefore they
-        // have to calculate layout again.
-        let bc_changed = self.old_bc != *bc;
-        let mut any_changed = bc_changed;
-        self.old_bc = *bc;
+        let mut any_changed = bc_changed || ctx.needs_layout();
```
DJMcNab, line 1035: Is this the line which changed the screenshot? I find this a bit confusing, because my impression was that if we need layout, that's the reason that this call has been made?

xilem/examples/to_do_mvc.rs
```diff
@@ -75,6 +76,7 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<TaskList> + use<> {
         first_line,
         tasks,
     ))
+    .padding(50.)
```
DJMcNab: If this has 50dp of padding, do we still need the 40dp of spacing at the top for the Android spacing?

masonry/src/widgets/flex.rs
PoignardAzur: > because my impression was that if we need layout, that's the reason that this call has been made?

Nope, whenever a widget is re-laid out, all its children are re-laid out as well. Previously this wasn't really taken into account.

xilem/examples/to_do_mvc.rs
PoignardAzur: I guess not. I think that line wasn't there when I did the initial change.

masonry/src/widgets/flex.rs
PoignardAzur: I didn't think of it. We can always add it later.

masonry/src/widgets/flex.rs
xStrom: What would be the better solution to have a border around `Flex` then? Some sort of `BorderProvider` widget that wraps `Flex`?

PR #1056, opened 2025-06-08
Refactor SizedBox to use Properties instead of struct attributes.

PoignardAzur:
This introduces a regression: SizedBox can no longer have radial gradients, sweep gradients or images as background.

We expect to add them back shortly; the feature was mostly unused as far as I can tell anyway.


masonry/screenshots/sized_box_empty_box_with_gradient_background.png
DJMcNab, line 1: This seems wrong? Have we regressed support for sweep gradients?

DJMcNab (approved):
Mostly looks good. I'm just concerned about the sweep gradient test changing

masonry/screenshots/sized_box_empty_box_with_gradient_background.png
PoignardAzur: We have, right now Properties only support linear gradients.

Adding sweep gradients back shouldn't be too hard, but that PR was already getting huge and I wanted to move on.

I'd argue it's an acceptable regression for now. I don't think anyone was using SizedBox with sweep gradient backgrounds anywhere so far.

DJMcNab: Right, that's fine. I'm just surprised it wasn't mentioned in the PR description

PR #1057, opened 2025-06-09
`masonry_winit`: Don't re-export `masonry::app::*`

waywardmonkeys:
Continue down the path of making `masonry_winit` only contain what it actually provides rather than being some wrapper crate.

### DIFF
--- a/xilem/src/window_view.rs
+++ b/xilem/src/window_view.rs
@@ -1,8 +1,8 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use masonry::app::RenderRoot;
 use masonry::widgets::RootWidget;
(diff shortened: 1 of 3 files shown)


xilem/src/window_view.rs
```diff
@@ -1,8 +1,8 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use masonry::app::RenderRoot;
```
xorgy, line 4: is this similar enough?
 https://github.com/linebender/xilem/blob/f97d19157a1930a9c08d5c7b05d968154b53a4fd/xilem/src/driver.rs#L13

waywardmonkeys: Similar enough how? Those things live in `masonry_winit`.

xorgy (approved):

PR #1058, opened 2025-06-09
`masonry_winit`: Use `vello` via `masonry`

waywardmonkeys:
`masonry_winit` and its examples shouldn't have a direct dependency on `vello`. This is especially true for example code as it should be able to be copied and pasted without having to add its own dependency upon `vello`.

### DIFF
--- a/masonry_winit/examples/custom_widget.rs
+++ b/masonry_winit/examples/custom_widget.rs
@@ -15,15 +15,14 @@ use masonry::core::{
 };
 use masonry::kurbo::{Affine, BezPath, Point, Rect, Size, Stroke};
 use masonry::palette;
-use masonry::peniko::Color;
+use masonry::peniko::{Color, Fill, Image, ImageFormat};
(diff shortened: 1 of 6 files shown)


masonry_winit/examples/custom_widget.rs
```diff
@@ -17,13 +17,13 @@ use masonry::kurbo::{Affine, BezPath, Point, Rect, Size, Stroke};
 use masonry::palette;
 use masonry::peniko::Color;
 use masonry::smallvec::SmallVec;
+use masonry::vello::Scene;
+use masonry::vello::peniko::{Fill, Image, ImageFormat};
```
RagibHasin: Shouldn't the two `peniko` imports be merged?

waywardmonkeys: Yes, good catch. Fixed now.

xorgy (approved):

PR #1061, opened 2025-06-09
Take a WidgetPod for the root widget

not-my-profile:
This does enable properties to be set for the root widget but the motivation for making this change is that it paves the way for removing the RootWidget widget.
(WindowView uses an AnyWidgetView for the main widget currently and AnyWidgetView::build returns a Pod, so without this change we couldn't forward the generated widget id).


PoignardAzur:
(Do you have merge access yet? I don't remember.)

PR #1062, opened 2025-06-09
Remove RootWidget and fix text event target fallback

not-my-profile:
The first commit has been submitted as #1061, this PR will be rebased once that has been merged.

In e064ffdbd1e06f9ce1cf87bce24ea95ebe89981e I implemented that if no widget is
focused then text events are sent to the root widget only child.

I then accidentally broke this in Xilem when introducing multiple window support
because AnyWidgetView introduces a DynWidget so the widget hierarchy became:

* RootWidget
  * DynWidget
    * main widget

Rather than making the event pass fall back to the grandparent which would be
too much special casing this commit removes the RootWidget from the widget
hierarchy, which primarily existed to report the top-level Window accesskit
node, by injecting that node in the accessiblity pass.

This also comes with the advantage that masonry_winit users can no longer
accidentally have a top-level accesskit node with a wrong Role.

This commit also documents the text event target fallback in the
RenderRoot::new documentation (previously it was undocumented).


not-my-profile:
@PoignardAzur could you review this? (I already rebased #1061 but I'm waiting to hear your thoughts before rebasing this.)

masonry_core/src/passes/accessibility.rs
```diff
@@ -201,5 +201,9 @@ pub(crate) fn run_accessibility_pass(root: &mut RenderRoot, scale_factor: f64) -
     );
     root.rebuild_access_tree = false;
 
+    let mut window_node = Node::new(Role::Window);
+    window_node.set_children(vec![root.root.id().into()]);
+    tree_update.nodes.push((root.window_node_id, window_node));
```
PoignardAzur, line 208: I think we may want to customize the root node type for multi-window or similar cases. I'd appreciate a way to pass the root node type to the RenderRoot, even if it doesn't show up in the default API. Either way, won't block on it.

not-my-profile: I don't quite follow ... if you have multiple windows then the root nodes are still of role window, no?

PoignardAzur: I think you might want roles like `Dialog` or `AlertDialog`. And embedded use cases might not want the `Window` role at all.

Either way, it's fine for now.

masonry_winit/examples/calc_masonry.rs
```diff
@@ -459,6 +454,11 @@ fn main() {
         window_id: WindowId::next(),
     };
 
+    let mut default_properties = default_property_set();
+    default_properties
+        .insert::<Flex, _>(Background::Color(AlphaColor::from_str("#794869").unwrap()));
+    default_properties.insert::<Flex, _>(Padding::all(2.0));
+
```
PoignardAzur: This seems redundant with the `root_props()` function.

PoignardAzur:
Running `cargo run --example to_do_list` gives me:

```sh
thread 'main' panicked at /home/olivier-faure/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/accesskit_consumer-0.28.0/src/tree.rs:139:13:
TreeUpdate includes 1 nodes which are neither in the current tree nor a child of another node from the update: [#10]
```

(Which, by the way, makes me think we should probably test for that by default somehow.)

masonry_core/src/passes/accessibility.rs
not-my-profile: I'd assume that these roles should be used in combination with modal windows which we don't yet support? I added a respective TODO comment in the accessibility pass.

masonry_winit/examples/calc_masonry.rs
not-my-profile: Thanks, I forgot to remove it and it actually wasn't used at all anymore but somehow clippy didn't spot that.

not-my-profile:
> Running `cargo run --example to_do_list` [crashes]

Now this is very interesting to me because I cannot reproduce that; it works fine for me (tested on both Linux and macOS).

PoignardAzur:
I get similar crashes with other masonry and xilem examples, presumably most of them.

I can try later on my laptop or on my Windows partition to see if it changes anything.

PoignardAzur:
PopOS.

PoignardAzur:
> before I resort to installing PopOS in a VM

I suspect what's happening is AccessKit is just not enabled in your build; a good first step would be enabling VoiceOver on MacOS (`Option + Command + F5`); if the build still doesn't panic, but VoiceOver does pick up the window contents, then it means the problem is on my end.

PoignardAzur:
The problem is that you don't set the synthetic `Window` node as the root.

Here's the relevant parts of the log for the `two_textboxes` example:

```
TreeUpdate {
    nodes: [
        (
            #5,
            Node {
                role: GenericContainer,
                children: [
                    #2,
                    #4,
                ],
                ...
            },
        ),
        (
            #2,
            Node { ... },
        ),
        (
            #4,
            Node { ... },
        ),
        ...
        (
            #6,
            Node {
                role: Window,
                children: [
                    #5,
                ],
            },
        ),
    ],
    tree: Some(
        Tree {
            root: #5,
            toolkit_name: Some(
                "Masonry",
            ),
            toolkit_version: Some(
                "0.3.0",
            ),
        },
    ),
    focus: #5,
}
Initialising in parallel using 14 threads

thread 'main' panicked at /home/olivier-faure/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/accesskit_consumer-0.28.0/src/tree.rs:139:13:
TreeUpdate includes 1 nodes which are neither in the current tree nor a child of another node from the update: [#6]
```

Basically, `TreeUpdate::tree::root` is set to the root widget `#5` instead of the node `#6` you created with role `Window`.

PoignardAzur:
Specifically, this is the line that needs to change: https://github.com/linebender/xilem/pull/1062/files#diff-2611ae6f9abf2051ced31081ab7b8b253ed777bf91239d5e355f555a9d7684daR158

PoignardAzur (approved):

DJMcNab:
I don't see why this PR wouldn't follow our normal author merges policy. If I'm missing some context for that, then apologies.

PR #1064, opened 2025-06-10
`masonry_winit`: Remove most dev dependencies

waywardmonkeys:
Some of the dev dependencies were unused or could be made to be unused by using the re-exports from `masonry`.

### DIFF
--- a/masonry_winit/examples/calc_masonry.rs
+++ b/masonry_winit/examples/calc_masonry.rs
@@ -13,7 +13,7 @@
 
 use std::str::FromStr;
 
-use masonry::accesskit::{Node, Role};
+use masonry::accesskit;
 use masonry::core::{
     AccessCtx, AccessEvent, Action, BoxConstraints, EventCtx, LayoutCtx, PaintCtx, PointerEvent,
     PropertiesMut, PropertiesRef, QueryCtx, RegisterCtx, StyleProperty, TextEvent, Update,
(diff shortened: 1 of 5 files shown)


masonry_winit/examples/calc_masonry.rs
```diff
@@ -249,15 +249,15 @@ impl Widget for CalcButton {
 
     fn paint(&mut self, _ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, _scene: &mut Scene) {}
 
-    fn accessibility_role(&self) -> Role {
-        Role::Button
+    fn accessibility_role(&self) -> accesskit::Role {
+        accesskit::Role::Button
```
PoignardAzur, line 253: Any reason for this change? We usually go in the other direction.

PoignardAzur (approved):
LGTM, aside from the one nit.

masonry_winit/examples/calc_masonry.rs
DJMcNab: For the virtual scrolling, I chose to retain the prefix because the names are a bit oblique without the "`accesskit`" context. That being said, we should be consistent here, and reverting this is consistent with other versions.

waywardmonkeys: We should just change all of the others to be consistent with this.

PoignardAzur: I'm fine with merging this as-is and deciding on a coding style guideline later.

waywardmonkeys: The real reason seems to have been lost to me ... I'll revert this here for now.

waywardmonkeys: Ahh, it was because there was also a `accessKit::Action` usage in this file and I didn't want to have `use masonry::accesskit::{Action, Node, Role};` as that just seemed way too generic without the module name.

DJMcNab: I concur - we can decide later, and we already have inconsistencies landed, so let's merge this

PR #1065, opened 2025-06-10
Prepare to split Xilem's lib.rs into multiple files

PoignardAzur:
Change code to remove reliance on private items.
Rename `ViewCtx::new_pod` to `ViewCtx::create_pod` to avoid ambiguity with constructor.

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -498,14 +498,12 @@ impl<Seq, State, Action> WidgetViewSequence<State, Action> for Seq where
 {
 }
 
-type WidgetMap = HashMap<WidgetId, Vec<ViewId>>;
-
 /// A context type passed to various methods of Xilem traits.
 pub struct ViewCtx {
     /// The map from a widgets id to its position in the View tree.
     ///
     /// This includes only the widgets which might send actions
-    widget_map: WidgetMap,
+    widget_map: HashMap<WidgetId, Vec<ViewId>>,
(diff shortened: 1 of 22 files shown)


xilem/src/lib.rs
```diff
@@ -498,14 +498,12 @@ impl<Seq, State, Action> WidgetViewSequence<State, Action> for Seq where
...
 
-type WidgetMap = HashMap<WidgetId, Vec<ViewId>>;
-
 /// A context type passed to various methods of Xilem traits.
 pub struct ViewCtx {
     /// The map from a widgets id to its position in the View tree.
     ///
     /// This includes only the widgets which might send actions
-    widget_map: WidgetMap,
+    widget_map: HashMap<WidgetId, Vec<ViewId>>,
```
DJMcNab, line 506: Oh oops; I didn't realise I originally added this type. I suspect it was originally more complicated.

xilem/src/window_view.rs
```diff
@@ -57,7 +57,7 @@ where
     ) {
         self.options.rebuild(&prev.options, window);
 
-        ctx.state_changed = true;
+        ctx.set_state_changed(true);
```
DJMcNab, line 60: @not-my-profile what was the intention of this line? It is indicating that something has updated the app's state, which doesn't seem to have happened?

I'm a little bit surprised that virtual scrolling still works on main despite this.

PR #1072, opened 2025-06-10
masonry: Don't allocate a String in button for every text event.

xorgy:
The previous approach allocates a String to compare (or appears to), and this is unnecessary.

Additionally, this space to activate behavior is special to PC keyboards and similar, and is not really intended for any key press that happens to produce a space, therefore makes more sense to activate with `code` than with `key`.

This commit also introduces two new functions to update the test, and documents the existing functions a little more clearly.


masonry/src/widgets/button.rs
```diff
@@ -123,12 +123,9 @@ impl Widget for Button {
...
     ) {
-        const SPACE: &str = " ";
         match event {
             TextEvent::Keyboard(event) if event.state.is_up() => {
-                if event.key == Key::Character(SPACE.into())
-                    || event.key == Key::Named(NamedKey::Enter)
-                {
+                if event.code == Code::Space || event.key == Key::Named(NamedKey::Enter) {
                     ctx.submit_action(Action::ButtonPressed(None));
                 }
```
PoignardAzur: Are those actually equivalent? I'd expect the `.key` value to be the "canonical" one e.g. if the user has a remapped keyboard or something.

masonry/src/widgets/button.rs
```diff
@@ -123,12 +123,9 @@ impl Widget for Button {
         _props: &mut PropertiesMut<'_>,
         event: &TextEvent,
     ) {
-        const SPACE: &str = " ";
         match event {
             TextEvent::Keyboard(event) if event.state.is_up() => {
```
xorgy, line 127: @PoignardAzur  On a side note here, is there any interest in an `active` state analogous to the web? It seems that primary pointer capture is used as a proxy for it in the case of interacting with a button with the pointer; but that doesn't work for non-primary pointers, and it doesn't work for activating the button with the keyboard either.
Maybe the button should just have a field that holds the active state though.

masonry/src/widgets/button.rs
xorgy: They are not meant to be exactly equivalent, but you're right about mappings... I think the equivalent might be space character with normal position. Somebody somewhere has no doubt remapped space...

xorgy: Alright, did it the other way, this way should work with mappings.

masonry/src/widgets/button.rs
```diff
@@ -123,10 +123,10 @@ impl Widget for Button {
...
-                if event.key == Key::Character(SPACE.into())
+                if (matches!(&event.key, Key::Character(c) if c == " ")
+                    && event.location == Location::Standard)
```
PoignardAzur: Why standard only?

masonry/src/widgets/button.rs
PoignardAzur: Yeah, I expect to add the active status back.

That said, AFAICT keyboard interaction doesn't trigger active status on the web. (Though maybe we'd want to anyway.)

DJMcNab (commented):
Generally agree with this, but I also agree that the addition of a "standard" location check needs justification.

xorgy:
@DJMcNab I took out the standard location check because I cba to explain it right now, it's fine.

masonry/src/widgets/button.rs
xorgy: So it's messier than that, I found this test page in a WhatWG thread on this topic: 
https://jsbin.com/cifugag/1/edit?html,output
, and it seems:
 - In Chromium, activating a button or a radio button with the space bar produces the `:active` state.
 - In Gecko, it does not produce the `:active` state on buttons, but does produce it on radio buttons.
 - In WebKit2, the behavior is the same as Chromium (both `<button>` and `<input type=radio>` produce the `:active` state when activated with the keyboard.
 
 Here's the WhatWG thread on the topic: <https://github.com/whatwg/html/issues/7578>

PR #1074, opened 2025-06-12
Fix pointer capture code, and add check to catch errors earlier.

PoignardAzur:


masonry_core/src/passes/event.rs
```diff
@@ -207,6 +209,13 @@ pub(crate) fn run_on_pointer_event_pass(root: &mut RenderRoot, event: &PointerEv
         }
     }
 
+    if let Some(id) = target_widget_id {
+        if !root.widget_arena.has(id) {
+            debug_panic!("Trying to target non-existent widget {id}");
+            return Handled::No;
+        }
+    }
```
DJMcNab: How is this different to the check which happens at the very start of `run_event_pass`? If it is meaningfully so, why isn't this copied to the other callers?

masonry_core/src/passes/event.rs
```diff
@@ -21,7 +21,9 @@ fn get_pointer_target(
 ) -> Option<WidgetId> {
     // See the [pointer capture documentation](../doc/06_masonry_concepts.md#pointer-capture).
     if let Some(capture_target) = root.global_state.pointer_capture_target {
-        return Some(capture_target);
+        if root.is_still_interactive(capture_target) {
+            return Some(capture_target);
+        }
```
DJMcNab, line 26: Discussion: Should we remove the pointer capture target in the else here?

Determining that needn't block this PR

masonry_core/src/passes/event.rs
```diff
@@ -21,7 +21,9 @@ fn get_pointer_target(
 ) -> Option<WidgetId> {
     // See the [pointer capture documentation](../doc/06_masonry_concepts.md#pointer-capture).
     if let Some(capture_target) = root.global_state.pointer_capture_target {
-        return Some(capture_target);
+        if root.is_still_interactive(capture_target) {
```
DJMcNab: I'd expect this to be a "does exist" check, rather than an "is interactive" check, because of:
https://github.com/linebender/xilem/blob/e1603483aa7117e43cd7b0038450a86f5e5ac209/masonry_core/src/passes/update.rs#L739-L745

Either that, or I'd expect the linked code to also change in this PR.

I think making this a "does exist" check should be sufficient, because the rewrite passes all run between events.

DJMcNab (commented):
I think that ignoring the pointer capture if the widget doesn't exist is a reasonably principled fix here, thanks. More thoughts in the comments

masonry_core/src/passes/event.rs
PoignardAzur: Yeah, I think that check is completely redundant.

masonry_core/src/passes/event.rs
PoignardAzur: I'd rather not, since it's already removed elsewhere. Duplicating that logic might lead to hard-to-track bugs.

masonry_core/src/passes/event.rs
PoignardAzur: I think having both checks test the same condition is coherent.

DJMcNab: Can you clarify how this is coherent? My basis for making my claim is:
1) We check if the capture target is interactive. If it isn't, we call `run_on_pointer_event_pass` with a cancel event, intending it to get to the capture target.
2) In `run_on_pointer_event_pass`, we call `get_pointer_target`.
3) `get_pointer_target` again checks if the capture target is interactive. It isn't (as seen in step 1), so the event doesn't get routed to the capture target.

PoignardAzur: I see your point. I'm a bit worried that removing that check means we might still send events to non-interactive widgets for a frame before the event pass runs, but I don't see an obvious way to deal with it, and it should be rare in practice.

DJMcNab (approved):
Thanks!

masonry_core/src/passes/event.rs
DJMcNab: > I see your point. I'm a bit worried that removing that check means we might still send events to non-interactive widgets for a frame before the event pass runs, but I don't see an obvious way to deal with it, and it should be rare in practice.

I don't think it's for a frame; that event pass happens "immediately", so it's only until the `run_update_pointer_pass` runs, which should happen after every "real" event anyway.

PoignardAzur: I know. I was thinking there might be corner cases, but I can't think of one.

PR #1075, opened 2025-06-12
Create placeholder of hero app Mastodon client

DJMcNab:
This is a very rough starting setup, with some necessary infrastructure. Current screenshot:
image

I'm expecting to land things piecemeal, so this is a PR intended for landing/discussion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
       - name: Install cargo-rdme
         uses: taiki-e/install-action@v2
         with:
-          tool: cargo-rdme
+          tool: cargo-rdme@1.4.8
 
       - name: Run cargo rdme (masonry)
         run: cargo rdme --check --heading-base-level=0 --workspace-project=masonry


.github/workflows/ci.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       - name: Install cargo-rdme
         uses: taiki-e/install-action@v2
         with:
-          tool: cargo-rdme
+          tool: cargo-rdme@1.4.8
```
DJMcNab, line 106: See https://github.com/taiki-e/install-action/pull/757#discussion_r2142730327

PoignardAzur:
Regarding the name, we were thinking about `doomscroll` IIRC. Do we still want to use it?

DJMcNab:
I don't particularly like it as a name, and was hoping that a better suggestion would arise. It feels to me that it's only used because of the term "scroll" there, but the name itself is too negative for me to be particularly happy with it.

PoignardAzur:
It's pretty metal though.

PR #1077, opened 2025-06-12
Accessibility for `VirtualScroll`

mwcampbell:
This implements the right AccessKit role, properties, and actions for `VirtualScroll`.

### DIFF
--- a/masonry/src/widgets/virtual_scroll.rs
+++ b/masonry/src/widgets/virtual_scroll.rs
@@ -454,7 +454,7 @@ impl<W: Widget + FromDynWidget + ?Sized> VirtualScroll<W> {
         }
     }
 
-    /// A wrapper to use [`post_scroll`] in event methods.
+    /// A wrapper to use [`post_scroll`](Self::post_scroll) in event methods.
     fn event_post_scroll(&mut self, ctx: &mut EventCtx<'_>) {
         match self.post_scroll(ctx.size()) {
             PostScrollResult::Layout => {
@@ -465,7 +465,7 @@ impl<W: Widget + FromDynWidget + ?Sized> VirtualScroll<W> {
         ctx.request_compose();
     }
 
-    /// A wrapper to use [`post_scroll`] in update methods.
+    /// A wrapper to use [`post_scroll`](Self::post_scroll) in update methods.
     fn update_post_scroll(&mut self, ctx: &mut UpdateCtx<'_>) {
         match self.post_scroll(ctx.size()) {
             PostScrollResult::Layout => {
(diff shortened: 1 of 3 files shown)


PoignardAzur:
So how much of this PR depends on the WIP branch of accesskit? If it's just the exact enum variants used for the scrolling events, do you think you could do an initial version with the previous events?

masonry/src/widgets/virtual_scroll.rs
```diff
@@ -428,7 +428,7 @@ impl<W: Widget + FromDynWidget + ?Sized> VirtualScroll<W> {
         this.ctx.request_layout();
     }
 
-    fn post_scroll(&mut self, ctx: &mut EventCtx<'_>) {
+    fn post_scroll(&mut self, ctx: &mut UpdateCtx<'_>) {
         // We only lock scrolling if we're *exactly* at the end of the range, because
         // if the valid range has changed "during" an active scroll, we still want to handle
         // that scroll (specifically, in case it happens to scroll us back into the active
```
PoignardAzur: It looks like you're only creating this new UserUpdate event so you can share the post_scroll function between two passes.

What I'd recommend doing instead is:

- Remove the `ctx` parameter for post_scroll.
- Have post_scroll return a bool (true is request_layout is needed, false otherwise).
- Move request_compose and request_layout to the callers.

waywardmonkeys: not a particularly big fan of bools used like that…

PoignardAzur: Then it could be a private struct or a newtype or an enum.

masonry_core/src/passes/event.rs
```diff
@@ -352,6 +352,10 @@ pub(crate) fn run_on_access_event_pass(
         false,
         |widget, ctx, props, event| {
             widget.on_access_event(ctx, props, event);
+            if !ctx.is_handled() && event.action == accesskit::Action::ScrollIntoView {
+                ctx.request_scroll_to_this();
+                ctx.set_handled();
+            }
```
DJMcNab: I think that this should move into the same block as `// handle focus events` below. This way, any parent can handle a bubbling `ScrollIntoView`, not just the direct target.

Edit: This is done in #1214, so we should just merge this now

DJMcNab (approved):
This looks good. I have also pushed a commit removing the macro, as it only saved ~1 line of code, which is too little for a macro to be worth it.

One concern, but happy for this to land with our without that resolved, in case I've missed some reason for things to be done this way.

mwcampbell:
@DJMcNab I think your one concern has now been addressed.

PR #1078, opened 2025-06-13
xilem: Make Worker take a store_sender callback

not-my-profile:
Before the Worker view was initialized with a value and the value was only sent with the UnboundedSender if a change was detected via its PartialEq impl. This commit changes the API to instead have you store the UnboundedSender in your app state and directly use it, which has several advantages:

* you can send multiple values at once
* you can send the same value multiple times in a row without having to change it
* you can send from other views
* the code is more clear (explicit send calls vs value changes that may send)

### DIFF
--- a/xilem/examples/http_cats.rs
+++ b/xilem/examples/http_cats.rs
@@ -10,6 +10,7 @@ use std::sync::Arc;
 
 use masonry::properties::Padding;
 use masonry::widgets::{Alignment, LineBreaking};
+use tokio::sync::mpsc::UnboundedSender;
 use vello::peniko::{Blob, Image};
 use winit::dpi::LogicalSize;
 use winit::error::EventLoopError;
@@ -29,6 +30,8 @@ struct HttpCats {
     statuses: Vec<Status>,
     // The currently active (http status) code.
     selected_code: Option<u32>,
+    /// Send a status code to download the image for it.
+    download_sender: Option<UnboundedSender<u32>>,
 }
 
 #[derive(Debug)]
@@ -60,44 +63,24 @@ impl HttpCats {
             .padding(Padding::left(5.)),
         );
 
-        let (info_area, worker_value) = if let Some(selected_code) = self.selected_code {
+        let info_area = if let Some(selected_code) = self.selected_code {
             if let Some(selected_status) =
                 self.statuses.iter_mut().find(|it| it.code == selected_code)
             {
-                // If we haven't requested the image yet, make sure we do so.
-                let value = match selected_status.image {
-                    ImageState::NotRequested => {
-                        // TODO: Should a view_function be editing `self`?
-                        // This feels too imperative.
-                        selected_status.image = ImageState::Pending;
-                        Some(selected_code)
-                    }
-                    // If the image is pending, that means that worker already knows about it.
-                    // We don't set the requested code to `selected_code` here because we could have been on
-                    // a different view in-between, so we don't want to request the same image twice.
-                    ImageState::Pending => None,
-                    ImageState::Available(_) => None,
-                };
-                (OneOf3::A(selected_status.details_view()), value)
+                OneOf3::A(selected_status.details_view())
             } else {
-                (
-                    OneOf3::B(
-                        prose(format!(
-                            "Status code {selected_code} selected, but this was not found."
-                        ))
-                        .alignment(TextAlignment::Middle)
-                        .brush(palette::css::YELLOW),
-                    ),
-                    None,
+                OneOf3::B(
+                    prose(format!(
+                        "Status code {selected_code} selected, but this was not found."
+                    ))
+                    .alignment(TextAlignment::Middle)
+                    .brush(palette::css::YELLOW),
                 )
             }
         } else {
-            (
-                OneOf3::C(
-                    prose("No selection yet made. Select an item from the sidebar to continue.")
-                        .alignment(TextAlignment::Middle),
-                ),
-                None,
+            OneOf3::C(
+                prose("No selection yet made. Select an item from the sidebar to continue.")
+                    .alignment(TextAlignment::Middle),
             )
         };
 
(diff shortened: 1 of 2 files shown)


xilem/examples/http_cats.rs
```diff
@@ -226,6 +220,7 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
     let data = HttpCats {
         statuses: Status::parse_file(),
         selected_code: None,
+        download_sender: unbounded_channel().0, // initialize with closed sender, will be overwritten once the view is built
```
DJMcNab: The lifecycle handling here is pretty far from ideal. The only advantage of the prior solution, was that it did not allow failures from requesting an item too early, which this now permits. I think that making this an Option in the app state would be less bad - that makes it obvious that this value can be in an uninitialised state.

not-my-profile: I don't really see the benefit of:

```rust
state.some_sender.unwrap().send(value).unwrap();
```

vs

```rust
state.some_sender.send(value).unwrap();
```

because the latter will also fail if the receiver has already been dropped.

I mean yeah in `http_cats` we're only sending in one place, I don't really mind the additional `unwrap()` but this was a more complex app with many `send` calls, I'd rather initialize it with a dummy sender. I'm curious on your thoughts on that but as I said I'm ok to change it here.

DJMcNab: Because the two failures are different? Panicking if the thread stops unexpectedly is reasonable behaviour, whereas panicking because the user pressed a button too early isn't really a great user experience.

It's possible that the latter panick is impossible, but at least we make it clear that we're not dealing with it/think it should be impossible

not-my-profile: I'd say it's the same failure (closed channel) with different causes. I don't think introducing an `Option` here really adds anything. If `HttpCats::view` is changed that the main view is only sometimes shown then the teardown of the `Worker` will be called and you'll have `Some(..)` with a closed channel. So I'd say the bottom-line is that calling `send` can fail and if it happens you'll have to investigate why it failed (called too early or too late).

PoignardAzur:
(I'm not familiar enough with this code to review it beyond what Daniel said.)

xilem/examples/http_cats.rs
DJMcNab: It's the same failure only because you shoved it into being the same failure though (by creating an always closed channel). I don't think that's particularly comphrensible.

Is the alternative API where the client stores both the sender and receiver, and passes the receiver once to the worker thread (i.e. by [taking](https://doc.rust-lang.org/stable/core/option/enum.Option.html#method.take) it) workable? That cuts this issue off at the root, and has a much more sensible lifecycle.

---

I'm not going to attempt to block this PR on changing this - if someone else signs off on it then we can land this. But this is the reason I'm not approving it myself.

not-my-profile: > Is the alternative API where the client stores both the sender and receiver, and passes the receiver once to the worker thread (i.e. by [taking](https://doc.rust-lang.org/stable/core/option/enum.Option.html#method.take) it) workable?

I would like to avoid mutations in view functions and even think we should make them impossible since they can lead to hard-to-trace bugs.

> I don't think that's particularly comphrensible.

Fair enough, I've changed that.

xilem/examples/http_cats.rs
```diff
@@ -111,28 +94,28 @@ impl HttpCats {
...
+                                    );
                                 }
-                            });
-                        }
+                            }
+                        });
                     }
                 },
+                |state: &mut Self, sender| {
+                    state.download_sender = Some(sender);
```
DJMcNab, line 117: To be clear, I meant having this callback be of the form `state.download_receiver.take().expect("We only ever create one worker")`. 
An argument in terms of this being a different kind of mutation doesn't seem coherent to me, but maybe I'm misunderstanding.

xilem/src/view/worker.rs
```diff
@@ -58,88 +59,80 @@ where
...
-    fn build(&self, ctx: &mut ViewCtx, _: &mut State) -> (Self::Element, Self::ViewState) {
+    fn build(&self, ctx: &mut ViewCtx, app_state: &mut State) -> (Self::Element, Self::ViewState) {
         let path: Arc<[ViewId]> = ctx.view_path().into();
 
         let proxy = ctx.proxy();
+
         let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
-        // No opportunity for the channel to be closed.
-        tx.send(self.value.clone()).unwrap();
+        (self.store_sender)(app_state, tx);
```
DJMcNab, line 111: :fire:

xilem/src/view/worker.rs
```diff
@@ -58,88 +59,80 @@ where
...
+    V: Send + 'static,
     Fut: Future<Output = ()> + Send + 'static,
+    S: Fn(&mut State, UnboundedSender<V>) + 'static,
     H: Fn(&mut State, M) -> Action + 'static,
     M: AnyMessage + 'static,
 {
     type Element = NoElement;
 
-    type ViewState = WorkerState<V>;
+    type ViewState = JoinHandle<()>;
```
DJMcNab, line 103: There are reasonable arguments for making this still a struct, but I agree that we don't have any strong guidelines here.

DJMcNab (approved):
I still think that giving the user control over creating the full sender/receiver pair feels more intuitive, if we're going this way. But this change is still an improvement, so could land as-is. I'd prefer that the proposed changes were made, though.

xilem/examples/http_cats.rs
not-my-profile: I think there are use cases for changing the worker that's handling a channel and that's easier with a store_sender callback than a take_receiver callback since the latter requires the state to be modified beforehand to initialize the new channel.

(But yeah disregard my prior mutation comment, I misunderstood what you were suggesting.)

PR #1079, opened 2025-06-13
Xilem: Pass `app_state` as a parameter to `View::{build, rebuild, teardown}`

DJMcNab:
This change is both a major departure from the prior state, and the most natural thing in the world.

My proposing this change has come from thinking related to #1070, for what build/rebuild/message actually represent, , how distinct important the phases of the lifecycle actually are (they aren't at all), and the reason that the Xilem model actually works. The core insight behind Xilem is that if you can prove to the compiler that you only access a piece of state in one place at once, the compiler will let you access it exclusively. Everything else is following that to its logical conclusion; this just continues that train of thought.

Current state of this PR: 
- This PR is ready to be merged
- This PR makes a lot of changes, so if we want this, merging quickly would be helpful
- The only thing which is broken is the `adapt` view, which currently unconditionally panicks, to avoid making behaviour changes in this PR
- The follow-up changes which make use of this (in task, worker, virtual_scroll, memoize, the new use_state, run_once, etc.) have not been done, to make this PR plausible to review.

Tradeoffs:
- In theory, this makes multi-threaded rebuild harder, but note that this was already probably hard
- This is rejecting the premise that you don't need component local state; this is something I treated as an experiment, but which has probably held us back significantly.

There's some more work to be done here (in particular, I really want to experiment with exclusive references in the app state, especially around `use_state`, for event handling).

### DIFF
--- a/xilem_web/src/events.rs
+++ b/xilem_web/src/events.rs
@@ -129,6 +129,7 @@ fn build_event_listener<State, Action, V, Event>(
     capture: bool,
     passive: bool,
     ctx: &mut ViewCtx,
+    app_state: &mut State,
 ) -> (V::Element, OnEventState<V::ViewState>)
 where
     State: 'static,
@@ -138,7 +139,7 @@ where
 {
     // we use a placeholder id here, the id can never change, so we don't need to store it anywhere
     ctx.with_id(ON_EVENT_VIEW_ID, |ctx| {
-        let (element, child_state) = element_view.build(ctx);
+        let (element, child_state) = element_view.build(ctx, app_state);
         let callback =
             create_event_listener::<Event>(element.as_ref(), event, capture, passive, ctx);
         let state = OnEventState {
@@ -161,6 +162,7 @@ fn rebuild_event_listener<State, Action, V, Event>(
     prev_passive: bool,
     state: &mut OnEventState<V::ViewState>,
     ctx: &mut ViewCtx,
+    app_state: &mut State,
 ) where
     State: 'static,
     Action: 'static,
@@ -173,6 +175,7 @@ fn rebuild_event_listener<State, Action, V, Event>(
             &mut state.child_state,
             ctx,
             element.reborrow_mut(),
+            app_state,
         );
         let was_created = element.flags.was_created();
         let needs_update = prev_capture != capture || prev_passive != passive || was_created;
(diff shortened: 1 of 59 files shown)


not-my-profile:
In #1078 I found a use case for mutating the state in `View::build` but had to send a message so that I could do it in `View::message` instead to access the state. So at least for `View::build` :+1: from me. And if you set sth in build it only follows that you should also be able to tear it down.

> The only thing which is broken is the adapt view, which currently unconditionally panicks, to avoid making behaviour changes in this PR

I don't quite follow. With currently you mean before your PR?

not-my-profile:
PRs can be reviewed commit by commit, so it's not impossible but I'm fine with deferring adapt.

Philipp-M:
> because this pr would not be possible to review if it had actual API changes as well as dumb refactoring.

I don't think it would be that bad to also implement it here, the actual changes here are otherwise trivial, but I'm also fine with deferring it to a separate PR (which I think will be done soonish anyways).

My concern (apart of the already named ones) of having mutable app state within `View::build/rebuild/teardown` is that it's now easily possible to change the state and have a different/inconsistent representation of it by the element-tree.

E.g. when a parent view uses a `count` within the state and displays it in its element, and a child view increments it for whatever reason the element tree might show now the incremented and the unincremented `count`, unless it should explicitly be rebuild again and the child view doesn't change the state again (by requesting it, when mutated?).
This might be a superficial concern that can be tackled by explicitly recommending best practices when implementing views though and providing API for these kinds of cases.
We could also make the state non-mutable of course for these functions (though that has other (annoying) limitations I think, and still could be undermined by using interior mutability).
Another idea would be something similar as in [this](https://doc.rust-lang.org/std/borrow/enum.Cow.html#examples) example, i.e. having a container, to automatically request a rebuild with cycle detection (so that it's clear that there's an issue within the view tree).

xilem_web/src/events.rs
```diff
@@ -401,13 +410,14 @@ macro_rules! event_definitions {
 
             type Element = V::Element;
 
-            fn build(&self, ctx: &mut ViewCtx) -> (Self::Element, Self::ViewState) {
+            fn build(&self, ctx: &mut ViewCtx,app_state: &mut State) -> (Self::Element, Self::ViewState) {
```
Philipp-M: Nit:

```suggestion
            fn build(&self, ctx: &mut ViewCtx, app_state: &mut State) -> (Self::Element, Self::ViewState) {
```

PR #1082, opened 2025-06-17
Add realloc_inner_storage to surface potential memory errors.

PoignardAzur:
The backstory behind that change is that I was trying to remove a layer of boxing from the unsafe arena, and then realized that doing so would make the impl unsound upon reallocation.

Adding these tests enforces that the arena implementation is robust to reallocs.

### DIFF
--- a/tree_arena/src/tree_arena_safe.rs
+++ b/tree_arena/src/tree_arena_safe.rs
@@ -552,8 +552,13 @@ impl<'arena, T> ArenaMutList<'arena, T> {
     }
 
     /// No-op. Added for parity with unsafe implementation.
+    ///
+    /// This is an unstable API which can only be used in tests of the `tree_arena` crate itself,
+    /// and may change in any release.
     #[doc(hidden)]
-    pub fn realloc_inner_storage(&mut self) {}
+    pub fn realloc_inner_storage(&mut self) {
+        std::hint::black_box(());
+    }
 }
 
 impl ArenaMapRef<'_> {
--- a/tree_arena/src/tree_arena_unsafe.rs
+++ b/tree_arena/src/tree_arena_unsafe.rs
@@ -663,12 +663,21 @@ impl<'arena, T> ArenaMutList<'arena, T> {
     /// Used in tests to simulate a call to `Self::insert` or `Self::remove` that
     /// triggers a realloc.
     ///
-    /// This is used to surface potential Use-After-Free (UAF) errors in the code.
+    /// This is an unstable API which can only be used in tests of the `tree_arena` crate itself,
+    /// and may change in any release.
+    /// It is used to surface potential Use-After-Free (UAF) errors in the code.
     #[doc(hidden)]
     pub fn realloc_inner_storage(&mut self) {
-        self.parent_arena.items.shrink_to_fit();
-        self.parent_arena
-            .items
-            .reserve(self.parent_arena.items.len() + 32);
+        // By doubling the required capacity (plus a small constant for small capacities),
+        // we hopefully guarantee that a reallocation will happen no matter the original capabity.
+        let capacity = self.parent_arena.items.capacity();
+        let capacity = std::hint::black_box(capacity);
+        self.parent_arena.items.reserve(capacity + 32);
+
+        // We try to discard the extra memory.
+        // We use black_box to hide the fact that the above call to reserve could be elided.
+        if std::hint::black_box(true) {
+            self.parent_arena.items.shrink_to_fit();
+        }
     }
 }


tree_arena/src/tree_arena_unsafe.rs
```diff
@@ -659,4 +659,16 @@ impl<'arena, T> ArenaMutList<'arena, T> {
...
+    /// Used in tests to simulate a call to `Self::insert` or `Self::remove` that
+    /// triggers a realloc.
+    ///
+    /// This is used to surface potential Use-After-Free (UAF) errors in the code.
+    #[doc(hidden)]
+    pub fn realloc_inner_storage(&mut self) {
+        self.parent_arena.items.shrink_to_fit();
+        self.parent_arena
+            .items
+            .reserve(self.parent_arena.items.len() + 32);
```
DJMcNab: Currently, this trying to set the capacity to `2*len + 32`, with the (probably reasonable) hope that this would overflow the allocated capacity. How was this heuristic reached?

I would expect this function's body to be something like `reserve(self.parent_arena.items.capacity() - self.parent_arena.items.len() + $epsilon)`, which has the advantage that it *always* forces a reallocation, but with a bounded number of additional elements.

tree_arena/src/tree_arena_unsafe.rs
```diff
@@ -659,4 +659,16 @@ impl<'arena, T> ArenaMutList<'arena, T> {
             None
         }
     }
+
+    /// Used in tests to simulate a call to `Self::insert` or `Self::remove` that
+    /// triggers a realloc.
+    ///
+    /// This is used to surface potential Use-After-Free (UAF) errors in the code.
```
DJMcNab:
```suggestion
    /// This is an unstable API which can only be used in tests of the `tree_arena` crate itself,
    /// and may change in any release.
    /// It is used to surface potential Use-After-Free (UAF) errors in the code.
```

tree_arena/src/tree_arena_safe.rs
```diff
@@ -550,6 +550,10 @@ impl<'arena, T> ArenaMutList<'arena, T> {
         let node = node_children.get_mut(&id)?;
         Some(node.arena_mut(*parent_id, &mut *self.parents_map.parents_map))
     }
+
+    /// No-op. Added for parity with unsafe implementation.
```
DJMcNab, line 554:
```suggestion
    /// No-op. Added for parity with unsafe implementation.
    /// 
    /// This is an unstable API which can only be used in tests of the `tree_arena` crate itself,
    /// and may change in any release.
```

tree_arena/src/tree_arena_unsafe.rs
PoignardAzur: The heuristic is quite arbitrary.

In general containers start with no memory, then a small fixed-sized slab, then reserve exponentially more memory when overflowing. Often "exponentially" means "double every time", though I don't know what the Rust standard library does. I guess I should look it up.

So the hope is that by doing shrink-to-fit then doubling the usage, we pass an allocation boundary at least once.

DJMcNab: Okay, I've done some more research here. If we request items beyond the capacity, then we will definitely get a reallocation (at the VM level). However, it's possible that it won't actually perform a reallocation, so if your unsafe tricks aren't maintaining provenance of pointers to the values, then a failure possibly wouldn't be caught (although I *think* miri would still catch this?). Trying to force a true reallocation is probably still worthwhile, though, as we might catch unaligned pointers elsewhere. Can we instead replace this full method with `it.reserve(it.capacity() + 32)`? That absolutely ensures that a realloc occurs (we're requesting more than the capacity, and likely kicks us into the next size class on the platform allocator). It will likely significantly over-allocate, but it doesn't really matter since this is test only code.

(I suppose ideally, we'd have a unit test that this method actually does cause a reallocation/move, i.e. that the raw pointer has a different [address](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.addr) after the operation)

PoignardAzur: Actually, it occurs to me that `shrink_to_fit -> reserve` is probably the wrong order, and `reserve -> shrink_to_fit` would probably work better.

> (I suppose ideally, we'd have a unit test that this method actually does cause a reallocation/move, i.e. that the raw pointer has a different [address](https://doc.rust-lang.org/stable/core/primitive.pointer.html#method.addr) after the operation)

Ideally, we'd have a test that actually triggers a use-after-free, but that's hard to integrate in CI. (And it's not like we want to go through the trouble of implementing an unsound version of TreeArena.)

I think it's fine if the test is potentially useless, we can revisit it later. We'll just have very detailed comments in the meantime.

KGrewal1: Would using std::black_box as a optimisation boundary help here (just to try to make sure that the shrink / allocation isn't optimised out) - agree reserve shrink to fit seems more likely to cause relocation (as otherwise could end up relocating to storage you'd just had

KGrewal1:
Unfortunately can't think of where else a similar change is likely to lead to UB - definitely one of the harder problems there is in the unsafe world! This change is definitely helpful tho - I think blackbox may possibly help in guaranteeing relocation though maybe MIRI pessimises resizing and assumes relocation already in checking for UB

DJMcNab (approved):
Thanks! And thanks for your help again @KGrewal1!

PR #1085, opened 2025-06-17
Remove `Adapt` for expediency

DJMcNab:
The [adapt](https://docs.rs/xilem_core/latest/xilem_core/fn.adapt.html) `View` has use cases around message handling, especially around immutable data structures. See also [Raph's blog post](https://raphlinus.github.io/rust/gui/2022/05/07/ui-architecture.html#finer-grained-change-propagation-memoizing). However, the API currently committed is incompatible with #1079.

This PR adds `map_message`, a view which handles the use-case of handling every possible message (which as far as I can tell, was only an accidental benefit of adapt, but what most users were using it for). It's clear that our message model still needs work, but that isn't a priority at the moment.

Outside of this, the API of `adapt` post #1079 isn't obvious to me. None of our current examples are using its interesting power (which is accessing the state you passed to the child after the child runs). Because of this, I don't think redesigning it is worth the time it will take to do so for me at the moment, but having it in its panicky state isn't great.

My intention is to make an issue to redesign and restore this (which I don't plan to work on), if this is approved. (#1088)

### DIFF
--- a/xilem/examples/elm.rs
+++ b/xilem/examples/elm.rs
@@ -34,7 +34,8 @@ fn elm_counter<T: 'static>(count: i32) -> impl WidgetView<T, CountMessage> {
     ))
 }
 
-enum AdaptMessage {
+/// A View Action type recording how the counter changed in [`map_message_counter`].
+enum CounterChanged {
     Changed,
     Reset,
     Nop,
@@ -42,23 +43,23 @@ enum AdaptMessage {
 
 // `map_message` is the most flexible but also most verbose way to modularize the views by action.
 // It's very similar to `map_action`, but it also allows to change the `MessageResult` for the parent view
-fn map_message_counter(count: i32) -> impl WidgetView<i32, AdaptMessage> {
+fn map_message_counter(count: i32) -> impl WidgetView<i32, CounterChanged> {
     flex((
         flex((
-            label(format!("adapt count: {count}")),
+            label(format!("map_message count: {count}")),
             button("+", |count| {
                 *count += 1;
-                AdaptMessage::Changed
+                CounterChanged::Changed
             }),
             button("-", |count| {
                 *count -= 1;
-                AdaptMessage::Changed
+                CounterChanged::Changed
             }),
         )),
         flex((
-            button("reset all", |_| AdaptMessage::Reset),
+            button("reset all", |_| CounterChanged::Reset),
             button("do nothing (and don't rebuild the view tree)", |_| {
-                AdaptMessage::Nop
+                CounterChanged::Nop
             }),
         )),
     ))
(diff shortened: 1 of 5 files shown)


xilem/examples/elm.rs
```diff
@@ -39,9 +40,9 @@ enum AdaptMessage {
     Nop,
 }
 
-// `adapt()` is the most flexible but also most verbose way to modularize the views by state and action,
-// This is basically a combination of `map_state` and `map_action`, but it also allows to change the `MessageResult` for the parent view
-fn adapt_counter(count: i32) -> impl WidgetView<i32, AdaptMessage> {
```
Philipp-M, line 44: I'm not sure whether `AdaptMessage` should be named differently, but I don't think `MapMessage` or `MapMessageMessage` is a good name...

xilem/examples/elm.rs
```diff
@@ -39,9 +40,9 @@ enum AdaptMessage {
...
+// `map_message` is the most flexible but also most verbose way to modularize the views by action.
+// It's very similar to `map_action`, but it also allows to change the `MessageResult` for the parent view
+fn map_message_counter(count: i32) -> impl WidgetView<i32, AdaptMessage> {
     flex((
         flex((
             label(format!("adapt count: {count}")),
```
Philipp-M: Nit:

```suggestion
            label(format!("map_message count: {count}")),
```

xilem_core/src/views/map_message.rs
```diff
@@ -69,34 +69,69 @@ where
...
+    Message,
+    impl Fn(&mut State, MessageResult<ChildAction, Message>) -> MessageResult<ParentAction, Message>
+    + 'static,
+>
+where
+    State: 'static,
+    ParentAction: 'static,
+    ChildAction: 'static,
+    V: View<State, ChildAction, Context, Message>,
+    F: for<'a> Fn(&mut State, ChildAction) -> ParentAction + 'static,
```
Philipp-M: `'a` is never used

```suggestion
    F: Fn(&mut State, ChildAction) -> ParentAction + 'static,
```

xilem_web/src/lib.rs
```diff
@@ -40,6 +40,7 @@ use std::any::Any;
 use std::ops::Deref as _;
 
 use web_sys::wasm_bindgen::JsCast;
+use xilem_core::MapMessage;
```
Philipp-M: Nittiest nit:
```suggestion
```

xilem_web/src/lib.rs
```diff
@@ -88,7 +89,7 @@ pub use templated::{Templated, templated};
 
 pub use xilem_core as core;
 
-use core::{Adapt, AdaptThunk, AnyView, MapAction, MapState, MessageResult, View, ViewSequence};
+use core::{AnyView, MapState, MessageResult, View, ViewSequence};
```
Philipp-M: and here:
```suggestion
use core::{AnyView, MapMessage, MapState, MessageResult, View, ViewSequence};
```

Philipp-M (approved):
> My intention is to make an issue to redesign and restore this (which I don't plan to work on), if this is approved.

Yeah that would be good.

Apart from nits, looks good mostly.

The `README.md` still contains `Adapt` in an example in the section "Components" this should be changed.

I wonder whether we find even better ways to manage state/actions with `app_state` in build/rebuild/teardown, but maybe that's what you're intending with opening the issue.

PR #1086, opened 2025-06-18
test: make TestHarness validate TreeUpdates returned by redraw

not-my-profile:
If you don't have the accessibility feature enabled in your OS then you don't run the accesskit_consumer code which can lead to accessibility being broken without you noticing.

Of course you should still do manual accessibility testing but this change ensures that the accesskit_consumer code at least doesn't crash if your app is tested with the TestHarness.


PoignardAzur (approved):
Super happy to get this!

PR #1087, opened 2025-06-18
Placehero: Get initial list of statuses for first exploration

DJMcNab:
image

This is an exploration into using the Mastodon API, continuing work on the hero app.
The actual changes here won't be all that useful, in the long term, but it's helpful for getting an understanding of how to use the API.

One big, unfortunate surprise is that we really need some HTML parsing (although only as described [in the docs](https://docs.joinmastodon.org/spec/activitypub/#sanitization)). That is something I'm not looking forward to!

One discovery in this PR is that our logging isn't set-up for the first `build` in Xilem. That makes sense with how logging is owned, but it's also very unfortunate).

I do intend to land this, to keep progress moving in a piecemeal way, and potentially to allow collaboration. It might be that hero app PRs should be largely rubber-stamped. Note that there are "real" changes to Xilem in this PR (limited to new re-exports)

### DIFF
--- a/placehero/src/lib.rs
+++ b/placehero/src/lib.rs
@@ -187,6 +187,7 @@ fn load_statuses(
                         id,
                         Some(&GetAccountStatusesInputOptions {
                             exclude_reblogs: Some(true),
+                            exclude_replies: Some(true),
                             ..Default::default()
                         }),
                     )
--- a/placehero/src/lib.rs
+++ b/placehero/src/lib.rs
@@ -11,14 +11,21 @@
 
 use std::sync::Arc;
 
-use megalodon::{Megalodon, entities::Instance, mastodon};
+use megalodon::{
+    Megalodon,
+    entities::{Account, Instance, Status},
+    mastodon,
+    megalodon::GetAccountStatusesInputOptions,
+};
 use xilem::{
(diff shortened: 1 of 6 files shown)


placehero/src/lib.rs
```diff
@@ -36,54 +58,178 @@ impl Placehero {
...
-            OneOf::B(label("Not yet connected (or other unhandled error)"))
+            OneOf::B(prose("Not yet connected (or other unhandled error)"))
+        }
+    }
+
+    fn main_view(&mut self) -> impl WidgetView<Self> + use<> {
+        if self.statuses.is_empty() {
+            OneOf::A(prose("No statuses yet loaded"))
+        } else {
+            Either::B(portal(flex(
```
not-my-profile: nit: mixing `Either` and `OneOf` looks weird to me. (same above)

DJMcNab: I think I agree; this format is idiomatic for `OneOf3` and above (it means that if you add a new variant, you only need to update the count on the first), but for `Either`, I think it is right to use `Either` for both.

DJMcNab: Oh yes, this ordering is also... weird

PR #1089, opened 2025-06-19
Placehero: Strip simple Mastodon HTML of tags

DJMcNab:
This doesn't support any rich text formatting, and only turns them from HTML soup (#1087) into readable text. Current state:
image

See also [#xilem > Hero Mastodon Client @ 💬](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Hero.20Mastodon.20Client/near/524860368)

PR #1091, opened 2025-06-20
Placehero: Load avatars for statuses

DJMcNab:
Current state:
image

Note that this doesn't include #1089, as that is awaiting review.


Cargo.toml
```diff
@@ -63,6 +64,11 @@ accesskit = "0.19.0"
 accesskit_winit = "0.27.0"
 accesskit_consumer = "0.28.0"
 time = "0.3.41"
+reqwest = { version = "0.12.20", default-features = false, features = [
+    # We use rustls as Android doesn't ship with openssl
+    # and this is likely to be easiest to get working.
+    "rustls-tls",
```
Philipp-M, line 70: Probably for backwards compatibility(?) but I think this should be the default at this point anyway doesn't it?

placehero/src/avatars.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[derive(Default)]
+pub(crate) struct Avatars {
+    icons: HashMap<String, Option<Image>>,
+    requester: Option<UnboundedSender<AvatarRequest>>,
+    // TODO: lru: ...
+}
+
+impl Avatars {
+    pub(crate) fn worker(&mut self) -> impl View<Self, (), ViewCtx, Element = NoElement> + use<> {
```
Philipp-M, line 35: We probably want an trait "alias" for `NoElement` views to avoid this don't we?

nicoburns (approved):

Philipp-M (approved):
Also slightly optimistic approval, but on a skim all of it makes sense.

Cargo.toml
nicoburns: > but I think this should be the default at this point anyway doesn't it?

FWIW, I've run into quite a lot of websites that don't work with rustls (because they target an older version of TLS/SSL). So that probably depends on whether you have control over the servers you need to connect to or whether you accept arbitrary URLs.l

DJMcNab: Do you want us to remove this comment?

nicoburns: Comment seems good to me. Probably not necessary, but nice to know for anyone reading the `Cargo.toml`.

placehero/src/avatars.rs
DJMcNab: Opened #1093

PR #1092, opened 2025-06-20
Placehero: Perform a styling pass on the status UI

DJMcNab:
image

Note that this doesn't include #1091 and #1089, which are still in review.

Note also the artifact at the top right of the screenshot. Those are pre-existing, which I haven't taken the time to track down yet. I believe it's related to the scrollbar, but I'm not sure how. I'm also not sure where the grey colour is coming from.

### DIFF
--- a/placehero/src/avatars.rs
+++ b/placehero/src/avatars.rs
@@ -70,9 +70,11 @@ impl Avatars {
         &mut self,
         url: &str,
     ) -> impl WidgetView<State> + use<State> {
+        let width = 50.;
+        let height = 50.;
         if let Some(maybe_image) = self.icons.get(url) {
             if let Some(image_) = maybe_image {
-                return Either::A(image(image_));
+                return Either::A(sized_box(image(image_)).width(width).height(height));
             }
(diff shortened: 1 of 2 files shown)


xilem/src/view/spinner.rs
```diff
@@ -58,7 +58,11 @@ impl<State, Action> View<State, Action, ViewCtx> for Spinner {
     type ViewState = ();
 
     fn build(&self, ctx: &mut ViewCtx, _: &mut State) -> (Self::Element, Self::ViewState) {
-        let pod = ctx.create_pod(widgets::Spinner::new());
+        let mut spinner = widgets::Spinner::new();
+        if let Some(color) = self.color {
+            spinner = spinner.with_color(color);
+        }
+        let pod = ctx.create_pod(spinner);
```
DJMcNab, line 65: I stuffed this in here as it's such a small change, but happy to split into a separate PR if anyone would prefer that

Philipp-M: Yeah small/simple change, good for me in this PR.

PR #1094, opened 2025-06-20
Xilem: Refactor lens into its own view without need for `state` as param.

Philipp-M:
See [#xilem > App State mutation detection, "simplification" of lens](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/App.20State.20mutation.20detection.2C.20.22simplification.22.20of.20lens/with/524866242) for more information.

This removes the need of `state` within `lens(component, state, map_state)`, by building the `component` view within the `View` impl of (the new View) `lens`.

### DIFF
--- a/xilem_core/src/views/lens.rs
+++ b/xilem_core/src/views/lens.rs
@@ -133,14 +133,13 @@ where
 
     fn teardown(
         &self,
-        view_state: &mut Self::ViewState,
+        (child, child_view_state): &mut Self::ViewState,
         ctx: &mut Context,
         element: Mut<'_, Self::Element>,
         app_state: &mut ParentState,
     ) {
         let child_state = (self.access_state)(app_state);
-        let child = (self.child_component)(child_state);
-        child.teardown(&mut view_state.1, ctx, element, child_state);
+        child.teardown(child_view_state, ctx, element, child_state);
     }
 
     fn message(


xilem_core/src/views/lens.rs
```diff
@@ -0,0 +1,159 @@
...
+    access_state: F,
+    child_component: CF,
+    phantom: PhantomData<fn(ParentState) -> (ChildState, Action, Context, Message, V)>,
+}
+
+impl<CF, V, F, ParentState, ChildState, Action, Context, Message> Debug
+    for Lens<CF, V, F, ParentState, ChildState, Action, Context, Message>
+{
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Lens").finish_non_exhaustive()
```
Philipp-M: Does it make sense to provide more here?
I'm not sure what though.

DJMcNab: It probably doesn't really matter. I'd consider something like:
```
Lens {
    from: {type_name::<ParentState>()},
    to: {type_name::<ChildState>()}
}
```
But not having that is also fine.

xilem_core/src/views/lens.rs
```diff
@@ -0,0 +1,167 @@
...
+    const {
+        assert!(
+            size_of::<Component>() == 0,
+            "The component function needs to be a non capturing closure"
+        );
+        assert!(
+            size_of::<StateF>() == 0,
+            "The child state access function needs to be a non capturing closure"
+        );
+    }
```
DJMcNab: I don't think either of these checks have value. `worker` does its checks because the function will only ever be called once, and so having capturing has the potential for weird behaviour. But, at least in theory, this shouldn't have any especially weird behaviour.

xilem_core/src/views/lens.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    fn teardown(
+        &self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut Context,
+        element: Mut<'_, Self::Element>,
+        app_state: &mut ParentState,
+    ) {
+        let child_state = (self.access_state)(app_state);
+        let child = (self.child_component)(child_state);
```
DJMcNab: I don't think this is right; I think this should be `child = view_state.0` (there's more that could be said here, about the role of rebuild, etc. But that's not for now

xilem_core/src/views/map_state.rs
```diff
@@ -6,7 +6,7 @@ use core::marker::PhantomData;
 
 use crate::{MessageResult, Mut, View, ViewId, ViewMarker, ViewPathTracker};
 
-/// The View for [`map_state`] and [`lens`].
+/// The View for [`map_state`].
 ///
 /// See their documentation for more context.
```
DJMcNab: Nit: This being `their` is because `map_state` and `lens` are plural, so should probably be `its` now

DJMcNab (commented):
Generally looks good, but there are a couple of details I think need fixing

xilem_core/src/views/lens.rs
Philipp-M: Hmm you're probably right, maybe we want to introduce warnings though, hinting that this may lead to undesired or unexpected behavior? But I don't know if there's a way to do this without some hacks in Rust, as there's no `compiler_warn!` AFAIK.

xilem_core/src/views/lens.rs
Philipp-M: Hmm isn't it? I would expect, that the most recent view should be taken with the current state (i.e. defined by `self.child_component`), but it probably doesn't make a big difference. For me it makes semantically more sense, as `View::teardown` takes the current view (`self`), and I think it should also act on the current child view.

Philipp-M: Though on second thought, as rebuild should already have built the most recent view, you're probably right as well.

Philipp-M: I've used `child = view_state.0`.

PR #1095, opened 2025-06-20
xilem_web: Fix clippy lints (in particular `elided_lifetimes_in_paths` and `allow_attributes_without_reason`)

Philipp-M:


xilem_web/src/concurrent/interval.rs
```diff
@@ -58,7 +58,10 @@ where
     }
 }
 
-#[allow(unnameable_types)] // reason: Implementation detail, public because of trait visibility rules
+#[allow(
```
DJMcNab: Probably should be `expect`

xilem_web/src/events.rs
```diff
@@ -150,11 +153,14 @@ where
     })
 }
 
-#[allow(clippy::too_many_arguments)] // reason: This is only used to avoid more boilerplate in macros, also so that rust-analyzer can be of help here.
+#[allow(
+    clippy::too_many_arguments,
```
DJMcNab: Hmm, why is this still needed?

xilem_web/src/events.rs
```diff
@@ -559,8 +565,10 @@ pub struct OnResize<V, State, Action, Callback> {
 
 pub struct OnResizeState<VState> {
     child_state: VState,
-    // reason: Closures are retained so they can be called by environment
-    #[allow(unused)]
+    #[allow(
+        unused,
```
DJMcNab: Super nit: Using the specific unused lint could be more idiomatic, rather than the entire unused lint group.

In practise, it doesn't really matter.

I'd also maybe like this to be `expect` if it applies in all build configurations.

DJMcNab (approved):
Seems reasonable

xilem_web/src/concurrent/interval.rs
Philipp-M: Ah well, seems my Rust knowledge is a little bit outdated, just looked it up, and changed it everywhere. Thanks for the hint :).

xilem_web/src/events.rs
Philipp-M: Oh right, we allow more arguments now IIRC. I removed this.

PR #1096, opened 2025-06-20
masonry: Improve default style

### LINKED ISSUE #1081: Improve default styles
The default styling has the hallmarks of being designed by programmers.

We should improve the default styles so that people who try out making apps with Masonry/Xilem don't need to open a visual design course first.

Some possible inspirations:

- https://ant.design/

melix99:
An attempt to resolve #1081.

Here's a summary of the choices I made:
- Used cooler gray colors instead of fully neutral grays, which usually tend to look colder on monitors
- Made the general look of the widgets more coherent with each other
- Removed gradients from everywhere since they look too old school imho :)
- Increased button padding, which follows the more breathable style of modern UIs
- Used [Tailwind's](https://tailwindcss.com/docs/colors) zync colors variations

I also cleaned a lot of unused constants from the theme module, but it will need more cleanup later once more properties are added.

Old:
Screenshot From 2025-06-20 21-56-51

New:
Screenshot From 2025-06-22 10-43-42

### DIFF
--- a/masonry/screenshots/textbox_outline.png
+++ b/masonry/screenshots/textbox_outline.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:95755017c73f6c2d5913cbd6b333816e86cfb8d82f80706df2d9711c20af5c6f
-size 1535
+oid sha256:1a61d2ac78cd4590d9a3610b936ccd209ba5e8bdba207c3af2dc32de69fc878c
+size 1501
--- a/masonry/screenshots/textbox_selection.png
+++ b/masonry/screenshots/textbox_selection.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:70c716f6e07ffed00917fcf1d04c045c490da26a956844630e90b534fe627cbd
-size 1713
+oid sha256:7088ddb463090ccbf58e8411c6eec20311ff95e1df6e0d43714e4af7270d954e
+size 1671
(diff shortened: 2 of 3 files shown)


not-my-profile:
The buttons look too similar to the textbox. Personally I think the horizontal padding on the buttons is too big but I think it's a good idea to let buttons and text inputs have the same height so that they line up properly when put on the same line. The background color of the checkboxes is now quite similar to the app background, I think this contrast could pose problems in terms of accessibility.

ricvelozo:
GNOME Adwaita uses bold font and larger paddings for text buttons:

https://developer.gnome.org/hig/patterns/controls/buttons.html

https://developer.gnome.org/hig/patterns/controls/text-fields.html

melix99:
New attempt:

Screenshot From 2025-06-22 10-43-42

xb284524239:
Could you provide both dark and light default theme styles simultaneously?

PoignardAzur (changes requested):
Overall, I really like it!

Some feedback:

- I think textboxes should look more rectangular than buttons. The corner radius of a textbox should be pretty small.
- Overall all borders could stand to be thicker (for textboxes, buttons and checkboxes).
- It looks like Textbox lost its focus color. Button still has proper focus when tabbed to.

melix99:
> I think textboxes should look more rectangular than buttons. The corner radius of a textbox should be pretty small.

Done


> Overall all borders could stand to be thicker (for textboxes, buttons and checkboxes).

Not really a fan and it's definitely not common having more than 1px borders. But I can add if you prefer it:

image

But I agree that borders look a bit jagged in the original screenshot, especially in the corners, but I think the main culprit is the strong contrast with a pure black bg. For example this looks better already imho (#09090b):

image


> It looks like Textbox lost its focus color. Button still has proper focus when tabbed to.

It never had one actually, it always had a white border. I tried to implement it correctly by using `ctx.is_focus_target()`, but it always returns false for some reason.

nilsmartel:
just chiming in to say that I really enjoy thoughts and efforts about this being made, thank you!

waywardmonkeys:
I don't think there's anything that is a blocking issue where this shouldn't land right now and then have some followups if necessary.

waywardmonkeys:
(Anything other than the CI needing to be fixed.)

PR #1097, opened 2025-06-23
Create example to showcase variable length lists

nilsmartel:
Lists are frequently a pain point in ui toolkits.
In Xilem it's so trivial, that it can easily be overlooked as something worth communicating explicitly. This example is really simple, yet I feel like it provides a lot of value to someone who wants to check out the capabilities of xilem. source: I myself wondered about lists :)

### DIFF
--- a/xilem/examples/lists.rs
+++ b/xilem/examples/lists.rs
@@ -1,0 +1,33 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Displaying a variable length list of widgets is trivially done using any kind of Seq
+
+use masonry::widgets::MainAxisAlignment;
+use winit::error::EventLoopError;
+use xilem::view::{Axis, button, flex, prose};
+use xilem::{EventLoop, WidgetView, WindowOptions, Xilem};
+
+#[derive(Default)]
+struct AppState {
+    count: usize,
+}
+
+fn app_logic(state: &mut AppState) -> impl WidgetView<AppState> + use<> {
+    let list = (0..state.count)
+        .map(|n| prose(format!("item #{}", n)))
+        .collect::<Vec<_>>();
+
+    flex((
+        button("more", |appstate: &mut AppState| appstate.count += 1),
+        list,
+    ))
+    .direction(Axis::Vertical)
+    .main_axis_alignment(MainAxisAlignment::Start)
+}
+
+fn main() -> Result<(), EventLoopError> {
+    let app = Xilem::new_simple(AppState::default(), app_logic, WindowOptions::new("Lists"));
+    app.run_in(EventLoop::with_user_event())?;
+    Ok(())
+}


xilem/examples/lists.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Displaying a variable length list of widgets is trivially done using any kind of FlexSequence
```
DJMcNab:
```suggestion
//! Displaying a variable length list is achieved using a [Vec] `FlexSequence`.
```

I want to avoid the word "trivially" in user-facing docs, because it presumes a level of obviousness which I'm not confident assigning.

xilem/examples/lists.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[derive(Default)]
+struct AppState {
+    count: usize,
+}
+
+fn app_logic(state: &mut AppState) -> impl WidgetView<AppState> + use<> {
+    // Here's the magic
+    // Creating a list of widgets can be as easy as using a flex(some_vec)
+    // with the alignment described using the flex(...) widget.
```
DJMcNab:
```suggestion
    // A vector (Vec) of views can be used as part of a `flex`'s children, allowing collections of dynamic length to be displayed.
```

xilem/examples/lists.rs
```diff
@@ -0,0 +1,35 @@
...
+    let list = (0..state.count)
+        .map(|n| prose(format!("item #{}", n)))
+        .collect::<Vec<_>>();
+
+    flex((
+        button("more", |appstate: &mut AppState| appstate.count += 1),
```
DJMcNab, line 24:
```suggestion
        // Even if a `Vec` is used for the children, other widgets can be included by putting them in a
        // tuple of children alongside the vector.
        button("more", |appstate: &mut AppState| appstate.count += 1),
```

xilem/examples/lists.rs
```diff
@@ -0,0 +1,35 @@
...
+        list,
+    ))
+    .direction(Axis::Vertical)
+    .main_axis_alignment(MainAxisAlignment::Start)
```
DJMcNab:
```suggestion
    // We want all the children to be left-aligned.
    .main_axis_alignment(MainAxisAlignment::Start)
```

xilem/examples/lists.rs
nilsmartel: I really love that you caught that, and I mean it.

xilem/examples/lists.rs
nilsmartel: Kind of used this, but elaborated a little more, while I was at it

nilsmartel:
Thanks, I feel like you really put thought into this

DJMcNab (approved):
Thank you!

I think that evaluating whether this example is valuable can be done as part of a holistic review of our examples (including potentially splitting into folders), and I don't think that stalling this on such questions is respectful of anyone's time.

DJMcNab:
I think it's correct that this example would become redundant with better and more thorough documentation, but those docs just don't exist at the moment.

PR #1098, opened 2025-06-23
Use flag trickery to always build both versions of tree_arena when working in IDE.

PoignardAzur:


tree_arena/src/lib.rs
```diff
@@ -65,14 +65,17 @@
 //!
 //! [Masonry]: https://crates.io/crates/masonry
 
+#![cfg_attr(feature = "implement_both", allow(unused))]
+#![cfg_attr(feature = "implement_both", allow(unreachable_pub))]
+
 type NodeId = u64;
 
-#[cfg(not(feature = "safe_tree"))]
+#[cfg(any(feature = "implement_both", not(feature = "safe_tree")))]
```
DJMcNab: Could we make this a "bare" cfg, so you set this in your local `.cargo/config.toml`? That way, it doesn't need to be in Cargo.toml, as it isn't public API.

Keeping it as a feature would be fine, though

tree_arena/src/lib.rs
```diff
@@ -65,14 +65,17 @@
 //!
 //! [Masonry]: https://crates.io/crates/masonry
 
+#![cfg_attr(feature = "implement_both", allow(unused))]
+#![cfg_attr(feature = "implement_both", allow(unreachable_pub))]
```
DJMcNab: Hmm. This is unfortunate. Since this feature isn't a default, this doesn't matter *too* much, but still.

Can we re-export the unsafe items under a different (`doc(hidden)`) module to avoid this?

tree_arena/Cargo.toml
```diff
@@ -22,6 +22,9 @@ targets = []
 # The unsafe tree is not yet fully tested, and is not used by default.
 default = ["safe_tree"]
 safe_tree = []
+# This flag implements both versions but only exports one.
+# This is useful when developing the crate.
+implement_both = []
```
DJMcNab: Can we please document that this is only an internal feature? I'd like us to rename the feature to "__implement_both" to indicate that, maybe?

It's a shame that this will be visible in e.g. `cargo add` output; that's part of why I suggest using the other implementation strategy.

tree_arena/src/lib.rs
```diff
@@ -65,14 +65,17 @@
 //!
 //! [Masonry]: https://crates.io/crates/masonry
```
DJMcNab, line 67: Not for this PR, but maybe an issue: This crate doesn't seem to have the lib.rs lint set items

(Can it also be made `no_std`?)

tree_arena/src/lib.rs
PoignardAzur: Oh, I didn't realize that was an option. But yeah, good point.

PoignardAzur:
You know what, this is overengineering.

I just tried it and building a single module takes 0.11s on my machine by default, and building both modules takes 0.12s. I honestly don't think it's worth optimizing that much.

We can just build both versions by default and only export one.

PoignardAzur:
Actually, doing `cargo build` after removing the incremental folder (but not the deps folder) gives me a spread of 0.32s to 0.52s.

That's a little more overhead, but still small enough I'd consider just eating it.

PoignardAzur:
I think `doc(hidden)` already communicates that.

PR #1099, opened 2025-06-23
Implement more properties

melix99:
Implement HoveredBackground, ActiveBorderColor and DisabledBorderColor properties.

### DIFF
--- a/masonry/src/properties/background.rs
+++ b/masonry/src/properties/background.rs
@@ -20,6 +20,10 @@ pub enum Background {
     Gradient(Gradient),
 }
 
+/// The background color/gradient a widget takes when hovered by a pointer.
+#[derive(Clone, Debug, PartialEq)]
+pub struct HoveredBackground(pub Background);
+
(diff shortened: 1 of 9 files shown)


Philipp-M:
The issue is simply that `Default` is only [implemented up to twelve elements](https://doc.rust-lang.org/std/default/trait.Default.html#impl-Default-for-(T,)).

So I guess we need a different strategy for storing (more) properties.

PoignardAzur:
Oh yeah, this is... inconvenient. I think the immediate takeaway is that our strategy for implementing properties in Xilem isn't going to scale.

I don't have any non-painful short-term solution, sorry.

PoignardAzur:
I'm working on a fix for the "tuple too big" problem.

But also, I'd rather not provide the entire matrix of "modifier x property" combos. I've tried to add a small set of modified properties so users can start adding some visual interactivity to apps, but if we want to add the full spectrum, we'll need to implement a system for conditional properties. The existing properties like DisabledXxx, ActiveXxx, etc are a band-aid solution.

PR #1100, opened 2025-06-23
masonry_winit: fix libwayland-client crash due to wrong drop order

dbuch:
We need to garbage collect surfaces before window.

Tested on linux wayland.

Would be helpful if someone could test on windows and mac?


dbuch:
Thank you for testing. No worries. If it works for others, I might just add a comment to clarify the drop order :man_shrugging:

PR #1102, opened 2025-06-24
Add a flex row helper to Xilem

DJMcNab:
This is a little papercut that I often run into, that the flex direction being defined after the sequence means that it can be hard to understand what element a piece of code relates to.

This is especially prevalent in nested flexes (e.g. a vertical flex containing horizontal flexes), as the changing flow direction is important to understand.

### DIFF
--- a/placehero/src/components.rs
+++ b/placehero/src/components.rs
@@ -6,6 +6,7 @@ use xilem::WidgetView;
 use xilem::palette::css;
 use xilem::style::Padding;
 use xilem::style::Style;
+use xilem::view::flex_row;
 use xilem::view::portal;
 use xilem::view::{
     CrossAxisAlignment, FlexExt, FlexSpacer, MainAxisAlignment, flex, inline_prose, label, prose,
@@ -33,7 +34,8 @@ pub(crate) fn timeline_status(
     status: &Status,
 ) -> impl WidgetView<Placehero> + use<> {
     sized_box(flex((
-        flex((
+        // Account info/message time
+        flex_row((
             avatars.avatar(&status.account.avatar_static),
             flex((
                 inline_prose(status.account.display_name.as_str())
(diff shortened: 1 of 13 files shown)


not-my-profile (changes requested):
I think it's bad that our `flex` defaults to Vertical while flex on the web defaults to horizontal.

I think we should either change our default to horizontal (in which case we need a flex_column rather than a flex_row), or rename the current `flex` to `flex_column` and introduce a `flex_with_axis` for dynamic axes.

not-my-profile:
Rust APIs should use snake_case. I think spelling things out is more idiomatic than and preferable to single character abbreviations. With rust-analyzer you get autocompletion anyway.

DJMcNab:
> I think it's bad that our `flex` defaults to Vertical while flex on the web defaults to horizontal.

Maybe? I'm not sure that this is a useful cue to take from the web, because we also don't have "block" flow/the default web flow. On the web, thinks flow downwards unless you specify e.g. `display: flex` directly. That is, `flex` is just our default multi-item layout, and for UI, you probably do want that to default to vertical.

> I think we should either change our default to horizontal (in which case we need a flex_column rather than a flex_row), or rename the current `flex` to `flex_column` and introduce a `flex_with_axis` for dynamic axes.

1) I'm strongly opposed to your `flex_column`/`flex`.
2) `flex_column` is a very verbose name; if we want both `flex_column` and `flex_row`, my preference would be to drop the `flex_` prefices, and just call them `column` and `row`
3) I don't think renaming them both in the same PR which splits them out would be good practice.

I would approve a follow-up PR renaming `flex` to `column` and `flex_row` to `row`, with something like `flex(axis, seq)` available (but I'm not planning on making it myself). Does that alleviate your concern? Even if we want that follow-up, I believe that this is the right first step to get there.

not-my-profile (approved):

PR #1103, opened 2025-06-24
Update to ui-events and ui-events-winit 0.2.0.

xorgy:


masonry/src/widgets/button.rs
```diff
@@ -97,17 +97,17 @@ impl Widget for Button {
...
+            PointerEvent::Down(..) => {
                 if !ctx.is_disabled() {
                     ctx.capture_pointer();
                     // Changes in pointer capture impact appearance, but not accessibility node
                     ctx.request_paint_only();
                     trace!("Button {:?} pressed", ctx.widget_id());
                 }
             }
-            PointerEvent::Up { button, .. } => {
+            PointerEvent::Up(b) => {
```
DJMcNab: Can we either destructure the struct, or use a clearer name here (e.g. `event`). The name `b` isn't self-describing enough that it is the right name here.

masonry/src/widgets/portal.rs
```diff
@@ -274,8 +274,8 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
         let content_size = ctx.get_raw_ref(&mut self.child).ctx().size();
 
         match *event {
-            PointerEvent::Scroll { delta, .. } => {
-                let delta = match delta {
+            PointerEvent::Scroll(ref s) => {
```
DJMcNab: This is more marginal; I'd probably use `event` here, but `s` is reasonable.

masonry/src/widgets/virtual_scroll.rs
```diff
@@ -1005,11 +1002,7 @@ mod tests {
         drive_to_fixpoint::<ScrollContents>(&mut harness, virtual_scroll_id, driver);
         assert_render_snapshot!(harness, "virtual_scroll_moved");
         harness.mouse_move_to(virtual_scroll_id);
-        harness.process_pointer_event(PointerEvent::Scroll {
-            pointer: PRIMARY_MOUSE,
-            delta: ScrollDelta::PixelDelta(PhysicalPosition::<f64> { x: 0., y: 25. }),
-            state: PointerState::default(),
-        });
+        harness.mouse_wheel(Vec2 { x: 0., y: 25. });
```
DJMcNab, line 1125: Thanks for catching this!

DJMcNab (commented):
Mostly looks good (once this is released)

masonry/src/widgets/button.rs
xorgy: Fair enough, the token is large and I have seen the preference against destructuring in other reviews. I prefer destructuring personally, so I'll gladly reintroduce it.

xorgy:
Would like a yes/no on the destructuring-based version (second commit) vs the opaque version (first commit), as we're about to do an ui-events-0.2.0 so we need to choose one. :+ )

DJMcNab:
Destructuring is fine by me. It's certainly better than extremely short names.
The lines like `PointerEvent::Up(PointerButtonEvent { button, ..}) =>` having two instances of each of `Pointer` and `Event` isn't great, but I don't see that being avoidable (without inline structs or equivalent).

waywardmonkeys (approved):
Would still like to get feedback from @DJMcNab on this.

masonry/examples/layers.rs
```diff
@@ -74,10 +74,10 @@ impl Widget for OverlayBox {
         _props: &mut PropertiesMut<'_>,
         event: &PointerEvent,
     ) {
-        if let PointerEvent::Move(e) = event
+        if let PointerEvent::Move(PointerUpdate { current, .. }) = event
             && ctx.is_hovered()
         {
-            let position = ctx.window_origin() + ctx.local_position(e.current.position).to_vec2();
+            let position = current.logical_point();
```
DJMcNab, line 80: I wonder if the `logical_point` method will be an attractive nuisance, given that in almost all cases, you want a logical "local" point.
In this case, this is certainly better than the old version though, so no change is needed.

masonry/src/widgets/text_area.rs
```diff
@@ -488,42 +488,44 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
         _props: &mut PropertiesMut<'_>,
         event: &PointerEvent,
     ) {
-        if self.editor.is_composing() {
+        if ctx.is_disabled() || self.editor.is_composing() {
```
DJMcNab: Rebase artifact

masonry/src/widgets/text_area.rs
```diff
@@ -488,42 +488,44 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
-                        }
-                        _ => {
-                            if state.modifiers.shift() {
-                                drv.shift_click_extension(cursor_pos.x as f32, cursor_pos.y as f32);
-                            } else {
-                                drv.move_to_point(cursor_pos.x as f32, cursor_pos.y as f32);
-                            }
-                        }
+            PointerEvent::Down(PointerButtonEvent {
+                button: None | Some(PointerButton::Primary),
```
DJMcNab, line 497:
```suggestion
                // None here allows moving the cursor on touch screens.
                button: None | Some(PointerButton::Primary),
```
I realise the lack of comment is pre-existing, but we may as well apply it whilst we have the chance.

masonry/Cargo.toml
```diff
@@ -28,7 +28,7 @@ masonry_core.workspace = true
 masonry_testing = { workspace = true, optional = true }
 parley.workspace = true
 tracing = { workspace = true, features = ["default"] }
-ui-events.workspace = true
+ui-events = { workspace = true, features = ["kurbo"] }
```
DJMcNab: Is there a reason this feature isn't enabled at the workspace level? It seems like a poor DX for a contributor to try to use one of the methods this enables in Masonry Core, for it to then fail in CI.

Alternatively, can we also enable this feature in Masonry Core?

DJMcNab: (Indeed, it isn't even clear to me why `masonry` has this dependency, given that `masonry_core` re-exports `ui_events`)

waywardmonkeys: The re-export was probably something I was cleaning up before until I ran out of time.

masonry/src/widgets/text_area.rs
xorgy: Thanks!

masonry/src/widgets/text_area.rs
xorgy: I feel like this comment doesn't belong here, since this is a routine activity (match primary mouse button, or no buttons).

masonry/Cargo.toml
xorgy: I think I'll just enable it in the workspace, to avoid the issue you described.

masonry/src/widgets/text_area.rs
DJMcNab: I mention this here, because this being an unclear pattern has come up in #1329. The API isn't really self describing.

masonry/examples/layers.rs
xorgy: That is how this code used to work before ui-events was first wired through (you'd get logical coords and subtract the origin manually, rather than using a context method).

We have also added ui-input-state, which could be of some interest (though currently it limits pointer state to just the primary pointer, because everything else is much more complicated).

PR #1104, opened 2025-06-25
refactor: stop wrapping signal_sender in Arc<Mutex<..>>

not-my-profile:
std::sync::mpsc::Sender::send does not require &mut self and can be cloned.


DJMcNab (approved):
Makes sense to me.

Issue #664, opened 2024-10-09
Add `Widget::Action` associated type


PoignardAzur:
During various passes, widgets that want to signal a semantically meaningful event from the UI (eg a button that was clicked) can emit actions.

Actions are a centralized queue of messages intended for the root application, mostly for the app driver (eg Xilem) to handle. They are inspired by [Facebook's Flux architecture](https://facebook.github.io/flux/docs/in-depth-overview).

Actions are currently under-designed. Right now, they only store the emitter's id, plus a payload among a small set of types with `Box<dyn Any>` as a wildcard:

```rust
pub enum Action {
    /// A button was pressed.
    ///
    /// Some presses are triggered without a pointer button;
    /// for example, a touch screen does not exercise buttons.
    /// In these cases, `None` will be the value here.
    ButtonPressed(Option<PointerButton>),
    /// Text changed.
    TextChanged(String),
    /// Text entered.
    TextEntered(String),
    /// A checkbox was toggled.
    CheckboxToggled(bool),
    // FIXME - This is a huge hack
    /// Other.
    Other(Box<dyn Any + Send>),
}
```

Ideally, we'd want each widget to have its own action variants, so the `Widget` trait should have an `Action` associated type.

This associated type be bound of `Self: Sized` so that the trait remains dyn-safe.

DJMcNab:
To clarify, this would mean that the action type is always downcasted? That seems reasonable to me.

I think Xilem would be able to make use of this for a slight ease-of-use improvement.

PoignardAzur:
> To clarify, this would mean that the action type is always downcasted?

Basically yes.

Philipp-M:
> I'm also curious why Daniel says this could make stuff easier in Xilem.

I think it's because `View::message` basically already receives a `message: Box<dyn Any>` which may just be casted down to `View::Element::Action`, it's "just" type information what the action may be for the relevant widget.

not-my-profile:
`View::Element::Action` doesn't work because `View::Element` is bounded by [ViewElement](https://docs.rs/xilem_core/latest/xilem_core/trait.ViewElement.html) (rather than Widget).

I think if we have such an associated type it would be nice to constraint widget implementations to only emit actions of that type. That's not possible in a dyn-safe type but we could make Widget dyn-unsafe and automatically lift it to a dyn-safe trait via a blanket impl as such:

```rust
pub trait Widget {
    type Action: Any;

[...]

PR #1105, opened 2025-06-25
Introduce AnyWidget trait and Widget::Action associated type

not-my-profile:
These changes are currently in one PR for a review of the total changes but I would like to land them with one PR per commit to keep the git history readable.
(PR for the first commit is  #1104 which is just a simple refactor.)

Open questions:

* should the associated type be called `Action` or do we rather want to go with `Event`?
* downcasting in views is quite unergonomic (`message.downcast::<<widgets::Button as Widget>::Action>()`) do we want to provide a helper function `downcast_action<W: Widget>(...) -> Result<Box<W::Action>>`? `DynMessage` lives in xilem_core so xilem would have to implement this method via a trait which isn't ideal from ergonomics but would still improve readability.


not-my-profile:
Nope Rust doesn't let you do that you have to disambiguate associated types.

PoignardAzur:
I need to take the time to review this in-depth, but overall I'm generally opposed to splitting the Widget trait into multiple traits. It makes the API surface larger and less intuitive for new users. We already use too many helper traits (Widget + AsDynWidget + FromDynWidget).

I'd lean more towards emitting actions in context types as before, and having the type-checking done at runtime. This still adds some potential brittleness, but I think it would avoid a lot of complexity.

The action associated type would be declared as

```rust
type Action: AnyDebug where Self: Sized;
```

PoignardAzur:
> I need to take the time to review this in-depth, but overall I'm generally opposed to splitting the Widget trait into multiple traits.

(Sorry, I should probably have broadcast this much earlier.)

not-my-profile:
> It makes the API surface larger and less intuitive for new users.

I think this can mostly be an implementation detail that users don't have to worry about. Case in point even with all the contributions I've made to this date I didn't really have to look into AsDynWidget or FromDynWidget ... maybe we could simplify them but I'd have to look into them.

Associated types that you can rely on actually improve ergonomics. Note that this is not just about the Action but I think we'd also want an associated type to enable reparenting (to represent the type of the additional params needed to add a new child to a container, e.g. FlexParams for Flex). Being able to turn runtime into compile errors is one of the major benefits of Rust and I think we should take advantage of that.

any_debug/Cargo.toml
DJMcNab, line 1: (This crate will need a README and license file)

PoignardAzur:
After mulling over this, I have some thoughts:

- Even if we make the `Widget` trait non-dyn-safe, I would still want actions to be emitted through the non-generic `FoobarCtx` types. Yes, that means more potential for runtime errors, but I think we could shape the code in ways that would make the runtime errors rare.
- If we do split the `WidgetTrait` into a `DynWidget` bit, I would like to see if we can make it an implementation detail, e.g. do it in a way where users would still write `WidgetPod<dyn Widget>`. Obviously, I'm not sure how feasible that is.

I don't know if you're interested in writing some code that goes in that direction?

PoignardAzur:
> Additionally, the position taken on splitting the widget trait into two has changed (between https://github.com/linebender/xilem/issues/664#issuecomment-2889000976 and https://github.com/linebender/xilem/pull/1105#issuecomment-3004572025). So overall, whilst your implementation here is a good implementation of a more involved Action type, I agree with Olivier that the tradeoffs aren't worth it for the relatively small increase in correctness.

I think this is a big part of the problem and I should have communicated more of my thoughts here.

Overall, we need to use `dyn Widget` in a lot of places, and we'd also like the Widget trait to have associated types and generics. That makes it increasingly tempting to split the trait in two (or to add `where Self: Sized` bounds everywhere), but that has a DX cost that I'd rather not pay. I'd rather not have users learn "You use `dyn DynWidget` in pointers but you implement `Widget`".

Ideally, what I'd really like is for the Rust language to allow traits with non-dyn-compatible items to be dyn-compatible. There's a bunch of reasons they haven't done that yet, but I think they'll eventually get there. If they do add that possibility and we add non-dyn-compatible items to the trait, the main cost would be that `dyn Widget` wouldn't implement the `Widget` trait anymore, but we don't actually need that.

(I should probably document all of that somewhere in the code.)

DJMcNab:
(just a note that work on AnyDebug has started in [#rust ui > Anymore](https://xi.zulipchat.com/#narrow/channel/422907-rust-ui/topic/Anymore))

PoignardAzur:
Superseded by #1212.

PR #1106, opened 2025-06-25
refactor: introduce AnyWidget trait

not-my-profile:
We want the Widget trait to have an associated type of which actions the widgets can emit. That makes Widget not dyn-safe, so we need a dyn-safe version.


not-my-profile:
Closing this for now until the discussion in #1105 has been resolved.

PR #1112, opened 2025-06-27
Fix `clippy::uninlined_format_args` lints

waywardmonkeys:


waywardmonkeys:
This is enabled in Rust 1.88 by default.

waywardmonkeys:
> Have you validated that these don't push the format strings past the "breaks Rustfmt in annoying ways" threshold?

Nope.

PR #1113, opened 2025-06-27
Bump MSRV to 1.88 and use new features.

PoignardAzur:
Use let-chains in code.
Use `#[cfg(false)]`.
Fix new clippy lints.

### DIFF
--- a/masonry_core/src/app/render_root.rs
+++ b/masonry_core/src/app/render_root.rs
@@ -509,7 +509,7 @@ impl RenderRoot {
         Some(WidgetRef { ctx, widget })
     }
 
-    /// Checks if a Widget with the given id is in the tree.
+    /// Checks if a widget with the given id is in the tree.
     pub fn has_widget(&self, id: WidgetId) -> bool {
         self.widget_arena.has(id)
     }


Philipp-M (approved):

Cargo.toml
```diff
@@ -77,9 +77,6 @@ reqwest = { version = "0.12.20", default-features = false, features = [
 # (And cargo doesn't let us have platform specific lints here)
 rust.unsafe_code = "deny"
 
-# Intentional break from the lint set. Intended to be temporary
-rust.unexpected_cfgs = { level = "warn", check-cfg = ['cfg(FALSE)'] }
```
DJMcNab, line 81: I see that this is now warn by default, and so probably needn't be in the lint set. I'll make the PR to remove it

masonry_core/src/app/render_root.rs
```diff
@@ -509,6 +509,11 @@ impl RenderRoot {
         Some(WidgetRef { ctx, widget })
     }
 
+    /// Checks if a Widget with the given id is in the tree.
```
DJMcNab: I'd use either "`Widget`" or "widget" here, probably.

Cargo.toml
DJMcNab: In fact, I'd like to treat getting an answer here as blocking. If it's the case that this is warn by default, feel free to dismiss this review.

DJMcNab (changes requested):

Cargo.toml
DJMcNab: This is https://github.com/linebender/linebender.github.io/pull/117, and so the lint set version should be bumped to v6.1

masonry_core/src/app/render_root.rs
PoignardAzur:
```suggestion
    /// Checks if a widget with the given id is in the tree.
```

PR #1114, opened 2025-06-27
Placehero: Add support for showing a single thread

DJMcNab:
On the timeline (there is now a `View Replies` button on each post):
image

In a thread (i.e. once the button has been pressed):
image

Note that to get this image, I followed two "View Replies"; the first to get to [this post](https://mastodon.online/@raph/114530193780235287).

Note a few features:
- There is now a `⬅️ Back to Timeline` button in the sidebar. This jumps you back to Raph's timeline.
- The ancestors and descendants of the "thread" are now visible.
- There is a "replies" label after the "main" post in the new thread

A few caveats (which I'm treating as out of scope for this PR):
- The "continuity" of replies is not visible, i.e. which thread items are actually replies to the previous item or one before.
- The scroll position is not handled very well; every page change jumps to the top.
   - See also [#masonry > Scrolling on (tab) Focus](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Scrolling.20on.20.28tab.29.20Focus/with/526109923)
- The "replies" label is not great design

### DIFF
--- a/placehero/src/components/thread.rs
+++ b/placehero/src/components/thread.rs
@@ -46,7 +46,7 @@ pub(crate) fn thread(
         flex((
             ancestor_views,
             base_status(avatars, root_status),
-            label("Replies:"),
+            label("Replies:").flex(CrossAxisAlignment::Start),
             descendant_views,
         ))
         .padding(Padding {
@@ -57,6 +57,10 @@ pub(crate) fn thread(
     )
 }
 
+/// The component for a single post in a thread.
+///
+/// These are rendered without a containing box, and with an adjoining "reply indicator"
+/// (which is currently known to be terrible!).
 fn thread_ancestor(avatars: &mut Avatars, status: &Status) -> impl WidgetView<Placehero> + use<> {
     sized_box(
         flex_row((
--- a/placehero/src/components/thread.rs
+++ b/placehero/src/components/thread.rs
@@ -1,0 +1,75 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use megalodon::entities::{Context, Status};
+use xilem::WidgetView;
+use xilem::palette::css;
+use xilem::style::{Padding, Style};
+use xilem::view::{CrossAxisAlignment, FlexExt, flex, flex_row, label, portal, sized_box};
+
+use crate::components::base_status;
+use crate::{Avatars, Placehero};
+
+/// Display a status in the context of its thread.
+///
+/// Notes:
+/// 1) We don't try and do anything "fancy", i.e. we just display all the items as one thing.
+pub(crate) fn thread(
+    avatars: &mut Avatars,
+    root_status: &Status,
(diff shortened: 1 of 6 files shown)


placehero/src/components/thread.rs
```diff
@@ -0,0 +1,73 @@
...
+    let mut descendant_views = Vec::new();
+    for descendant in &thread.descendants {
+        descendant_views.push(thread_ancestor(avatars, descendant));
+    }
+
+    portal(
+        flex((
+            ancestor_views,
+            base_status(avatars, root_status),
+            label("Replies:"),
```
jaredoconnell: Being in the center makes it harder to nice. Left-aligning (or making it take up the full width with left align text) would be an improvement.
Not that it really matters at this time, though.

placehero/src/components/thread.rs
```diff
@@ -0,0 +1,73 @@
...
+        ))
+        .padding(Padding {
+            // Leave room for scrollbar
+            right: 20.,
+            ..Padding::all(5.0)
+        }),
+    )
+}
+
+fn thread_ancestor(avatars: &mut Avatars, status: &Status) -> impl WidgetView<Placehero> + use<> {
```
jaredoconnell, line 64: A comment on this function would be useful, unless you're going to replace it really soon.

placehero/src/lib.rs
```diff
@@ -84,21 +91,41 @@ pub fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
...
         }
     }
 
     fn main_view(&mut self) -> impl WidgetView<Self> + use<> {
-        if self.statuses.is_empty() {
-            Either::A(prose("No statuses yet loaded"))
+        if let Some(show_context) = self.show_context.as_ref() {
+            if let Some(context) = self.context.as_ref() {
+                // TODO: Display the status until the entire thread loads; this is hard because
+                // the thread's scroll position would jump.
```
jaredoconnell, line 120: Would using the virtual scroll widget help with that scenario?

DJMcNab: I think that's right, yes.

Philipp-M: Yeah long-term virtualized lists probably make sense for long threads anyways, but I guess we can delay that to future PRs.

placehero/src/lib.rs
```diff
@@ -122,6 +150,53 @@ fn app_logic(app_state: &mut Placehero) -> impl WidgetView<Placehero> + use<> {
     )
 }
 
+fn load_contexts(
+    mastodon: Mastodon,
+) -> impl View<Placehero, (), ViewCtx, Element = NoElement> + use<> {
+    worker_raw(
```
Philipp-M, line 156: Not (directly) related to this PR, but maybe we should include the app state in the init future, so that we don't have to use the "raw" worker (that often).
Maybe we can completely get rid of it as well then (or at least minimalize the need for it).

PR #1115, opened 2025-06-28
xilem_web: stroke-linejoin

burrbull:
This PR adds `stroke-linejoin` attribute for pathes.

But is not complete as requires `kurbo::Stroke` defaults to be changed to correspond to SVG defaults:

`Join::Miter` and `Cap::Butt`.

cc @Philipp-M

### DIFF
--- a/xilem_web/src/svg/common_attrs.rs
+++ b/xilem_web/src/svg/common_attrs.rs
@@ -4,6 +4,7 @@
 use std::fmt::Write as _;
 use std::marker::PhantomData;
 
+use peniko::kurbo::Join;
 use peniko::{Brush, kurbo};
 
 use crate::core::{MessageResult, Mut, View, ViewElement, ViewId, ViewMarker};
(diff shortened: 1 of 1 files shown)


Philipp-M:
Hmm interesting, I don't think it necessarily needs to be changed in kurbo (I'm not sure about the choices about the defaults, there might be a reason for `Join::Round`).
As this is "opinionated" around kurbo shapes I think it's fine to take whatever the defaults in kurbo are.
There's still `attr` to be used directly in case the "raw" svg API is preferred (or otherwise of course setting a different join in `kurbo::Stroke`).

burrbull:
> I'm not sure about the choices about the defaults, there might be a reason for `Join::Round`

Maybe.

In this case there should be another `new_svg` or `svg_default` constructor for `kurbo::Stroke` to prevent adding `linejoin` in path if it is missed.

Philipp-M:
You might want to start a zulip thread or a PR in kurbo changing the defaults to spark a discussion why these are the defaults.

I personally don't think it's too bad to have it added by default, when using the `Stroke` view. It's simple to change it to the svg defaults. But having a dedicated constructor in kurbo for svg defaults is likely useful in case the defaults chosen are deliberate.

PR #1116, opened 2025-06-30
Placehero: trim trailing whitespace in HTML (+error context)

DJMcNab:
Fix self-closing behaviour of `<br>`
Add context to each `error!` in `status_html_to_plaintext` Add an error for a non-empty stack at end
Fix too many trailing whitespaces
Make `info` of not handling links a `trace` rather than info

image

Note that this does not include #1114.


nicoburns (approved):
Seems reasonable, but want to note that the proper way to fix trimming trailing whitespace would be to implement it in Parley.

PR #1117, opened 2025-07-04
Make `DynMessage` non-`Send`, and use it as the message type in Xilem Web

DJMcNab:
This creates a new message type called `SendMessage`, which is `DynMessage`, but `Send`. The follow-up I want to do here is to remove the `Message` generic on `View` entirely.

Follow-up from https://github.com/linebender/xilem/pull/920


xilem_core/src/message.rs
```diff
@@ -58,7 +55,8 @@ impl<A, Message> MessageResult<A, Message> {
 ///
 /// If the message contains sensitive data, make sure this isn't output in its `Debug` implementation,
 /// as that may be called by the Xilem runtime (e.g. due to a bug meaning messages are redirected) or
-/// any parent views. (That is, views do not need to design themselves as if the Debug implementation is )
+/// any parent views. That is, views do not need to be designed as if the Debug implementation
```
Philipp-M:
```suggestion
/// any parent views. That is, views do not need to be designed as if the `Debug` implementation
```

xilem_core/src/message.rs
```diff
@@ -90,15 +88,66 @@ impl DynMessage {
...
+// ```
+// enum DynMessage {
+//     Special(Box<dyn AnyMessage>),
+//     Send(SendMessage)
+// }
+// ```
+// to let "maybe-threaded" message handling. That would be especially useful for
+// handling stale messages (i.e. reporting them back to the task which failed).
+// Probably not worth it, but would be (reasonably) non-breaking, at least.
+// Ooh, we could pass a `fn(DynMessage)->Option<SendMessage>`, which would assume/validate that
```
Philipp-M: "Ooh" intentionally? Or was "Or" meant?

I've no meaningful comment though yet on the comment itself...

Philipp-M (approved):
Looks good. Good to see more unification between xilem and xilem_web.

xilem_core/src/message.rs
DJMcNab: Yes, "ooh" was intentional here. It's a "chain of thought"/brainstorming comment.

I'll rephrase slightly

PR #1124, opened 2025-07-09
Checkbox support focus and keyboard interaction

### LINKED ISSUE #1118: `Checkbox` doesn't support focus or keyboard interaction
According to https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/ (among other things), the checkbox should support focus and toggling when the keyboard space character is pressed.

This is probably a matter of:

```
    fn on_text_event(
        &mut self,
        ctx: &mut EventCtx<'_>,
        event: &TextEvent,
    ) {
        match event {
            TextEvent::Keyboard(event) if event.state.is_up() => {
                if matches!(&event.key, Key::Character(c) if c == " ") {
                    self.checked = !self.checked;
                    ctx.submit_action(Action::CheckboxToggled(self.checked));
                    // Checked state impacts appearance and accessibility node
                    self.request_render();
                }
            }
            _ => (),
        }
    }

    fn accepts_focus(&self) -> bool {
        true
    }

    fn accepts_text_input(&self) -> bool {
        false
    }
```

There should also be some sort of visual indication of the focused state in the paint method (the `update` fn already checks for `FocusChanged` which wasn't necessary at that time since the above wasn't in place).

And then the tests should get update along the lines of what the button tests do for this stuff.

tannal:
Fix #1118


masonry/src/widgets/checkbox.rs
```diff
@@ -105,10 +107,31 @@ impl Widget for Checkbox {
...
+    }
+
+    fn accepts_focus(&self) -> bool {
+        // Checkbox can be tab-focused...
+        true
+    }
+
+    fn accepts_text_input(&self) -> bool {
+        // But they still aren't text areas.
+        false
```
DJMcNab: I think since #1090, we should probably just remove this method.

masonry/src/widgets/checkbox.rs
```diff
@@ -209,12 +232,17 @@ impl Widget for Checkbox {
...
         let border_rect = border_width.border_rect(size, border_radius);
 
-        let border_color = if is_hovered && !ctx.is_disabled() {
+        let mut border_color = if is_hovered && !ctx.is_disabled() {
             &props.get::<HoveredBorderColor>().0
         } else {
             props.get::<BorderColor>()
         };
 
+        if ctx.is_focus_target() {
```
DJMcNab, line 236:
```suggestion
        // FIXME - Handle this properly
        if ctx.is_focus_target() {
```
To match the button version

DJMcNab (approved):
This looks really good, thanks. It's great to see tests here.

I'd probably expect the "is focused" border to be around the entire checkbox, including the label. If you'd be willing to make that change, that could be helpful. However, this is still an improvement, and so I'd be happy to land this as-is. Let me know either way.

tannal:
> to be around the entire checkbox

I can follow up with a separate PR to address the border around the whole checkbox and label.

masonry/src/widgets/checkbox.rs
```diff
@@ -209,12 +227,17 @@ impl Widget for Checkbox {
...
+        let mut border_color = if is_hovered && !ctx.is_disabled() {
             &props.get::<HoveredBorderColor>().0
         } else {
             props.get::<BorderColor>()
         };
-
+        // FIXME - Handle this properly
+        if ctx.is_focus_target() {
+            border_color = &BorderColor {
+                color: Color::WHITE,
```
jaredoconnell, line 238: We probably want to create a property for focused border color instead of hard-coding it.

waywardmonkeys: That's copied from `Button` ... we should probably create an issue to track this to solve focus indication globally.

PR #1132, opened 2025-07-09
Cache tracing spans

DJMcNab:
This is implementing the idea discussed in https://github.com/linebender/xilem/pull/687#discussion_r1808472248

The reason I didn't do this before was that I thought you needed to get the parents right. However, having re-read tracing's docs since then, I've discovered that this isn't necessary.

As discussed in https://github.com/bevyengine/bevy/pull/9390, this can get you pretty close to removing all the overhead; it's possible that #687 could be undone.


waywardmonkeys (approved):
This looks good to me, but due to it changing things to not pass the properties around as much, I think it is best if someone else also reviews this.

PoignardAzur (approved):
Very welcome change!

PR #1138, opened 2025-07-10
Prepare for `TextAlign`

waywardmonkeys:
Parley is going to rename `Alignment` to `TextAlign`, so this prepares for that by:

* Renaming `alignment` functions and variables to be `text_alignment`..
* Rename the `xilem::TextAlignment` type alias to `xilem::TextAlign`.
* Exporting `TextAlign` and `TextAlignOptions` from `masonry` to match what happens with `xilem::TextAlign`.
* Switch all uses of Parley's `Alignment` over to these renamed imports so that everything is consistent.


PoignardAzur (approved):
LGTM.

PR #1139, opened 2025-07-10
Fix `TextInput` tab focus

melix99:
In #1096 I noticed that `TextInput` didn't handle the focus style properly. This is testable by running the `to_do_mvc` example and tab-focusing between widgets. This PR fixes that.

### DIFF
--- a/masonry/src/widgets/text_input.rs
+++ b/masonry/src/widgets/text_input.rs
@@ -8,6 +8,7 @@ use smallvec::{SmallVec, smallvec};
 use tracing::{Span, trace_span};
 use vello::Scene;
 use vello::kurbo::{Affine, Point, Rect, Size};
+use vello::peniko::Color;
 
 use crate::core::{
     AccessCtx, BoxConstraints, LayoutCtx, PaintCtx, PropertiesMut, PropertiesRef, QueryCtx,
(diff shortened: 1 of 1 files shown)


masonry/src/widgets/text_input.rs
```diff
@@ -117,12 +118,13 @@ impl Widget for TextInput {
...
 
-    fn update(
-        &mut self,
-        _ctx: &mut UpdateCtx<'_>,
-        _props: &mut PropertiesMut<'_>,
-        _event: &Update,
-    ) {
+    fn update(&mut self, ctx: &mut UpdateCtx<'_>, _props: &mut PropertiesMut<'_>, event: &Update) {
+        match event {
+            Update::ChildFocusChanged(_) => {
```
waywardmonkeys, line 125: Why is this needed here? It doesn't seem like other widgets are doing this.

melix99: Because the widget that receives the focus is not `TextInput`, but its child `TextArea`. So, we can't just check for `Update::FocusChanged()` to request a re-paint to update the focus style.

waywardmonkeys: Can you add a small comment noting that so that future people know too? (And if we change how this works, we can know to update it there...)

DJMcNab (approved):
I'd like to see a snapshot test added here, I think (see #1124 for how a test can test focus support).

But this can also land as-is; just let me know if you want to do so and we can merge it.

melix99:
> I'd like to see a snapshot test added here

Umh, I had a look at the current test and I actually expected the `text_input_selection` snapshot test to fail, because the widget should be focused and so it should have a white outline, but apparently it doesn't. I don't get what's wrong with it though, `ctx.has_focus_target()` is returning false during paint for some reason, even though the `ChildFocusChanged` update gets called.

I think it needs a bit of debugging in the focus chain code, but it's bit too unfamiliar for me to look at ATM :/

PR #1142, opened 2025-07-10
Initial support for an environment system

DJMcNab:
This is extremely incomplete, but hopefully gives a sense of some of the implementation strategies.


DJMcNab:
To clarify on the current state:
- I think this could merge as-is to keep progress moving.
- It is probably worth writing a few tests here, however, I plan to make significant changes here soon, so I think those tests would become quickly outdated. I'm happy to do so if it's needed for this to land, though.

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+pub struct Rebuild;
+
+// --- MARK: Environment
+
+#[derive(Debug)]
+struct EnvironmentItem {
+    value: Box<dyn AnyMessage>,
+    // TODO: Can we/do we want to make these share an allocation?
+    // TODO: How do we GC this?
+    change_listeners: Vec<Option<Arc<[ViewId]>>>,
```
jaredoconnell, line 25: Is there an instance where the current design will just not cleanup after itself, causing memory usage to grow?

xilem/examples/mason.rs
```diff
@@ -65,80 +78,89 @@ fn app_logic(data: &mut AppData) -> impl WidgetView<AppData> + use<> {
...
-                data.active = false;
-            }),
-            button("Unlimited Power", |data: &mut AppData| {
-                data.count = -1_000_000;
-            }),
-            without_elements(run_once(|| {
-                tracing::warn!("The pathway to unlimited power has been revealed");
-            })),
-        ))
+        provides(
```
jaredoconnell, line 152: Is this code also under the provider for the 120 context, but the 777 value is overriding it underneath it in the tree?

xilem/examples/mason.rs
```diff
@@ -20,10 +20,23 @@ use xilem::{
...
 
 Phasellus in viverra dolor, vitae facilisis est. Maecenas malesuada massa vel ultricies feugiat. Vivamus venenatis et nibh nec pharetra. Phasellus vestibulum elit enim, nec scelerisque orci faucibus id. Vivamus consequat purus sit amet orci egestas, non iaculis massa porttitor. Vestibulum ut eros leo. In fermentum convallis magna in finibus. Donec justo leo, maximus ac laoreet id, volutpat ut elit. Mauris sed leo non neque laoreet faucibus. Aliquam orci arcu, faucibus in molestie eget, ornare non dui. Donec volutpat nulla in fringilla elementum. Aliquam vitae ante egestas ligula tempus vestibulum sit amet sed ante. ";
 
+#[derive(Debug)]
+struct SomeContext(u32);
+impl Resource for SomeContext {}
+
+fn env_using() -> impl WidgetView<AppData> + use<> {
```
Philipp-M, line 35: As this is example code, maybe worth mentioning (in a doc-comment?), that this requires a `provides`, as it otherwise panics.

xilem/examples/mason.rs
Philipp-M: Yes it looks like that (and I suspect it's here to test for that exact reason).

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+/// [`environment`](ViewPathTracker::environment)  method of `ViewPathTracker`, which
+/// they all implement.
+#[derive(Debug)]
+pub struct Environment {
+    slots: Vec<Slot>,
+    // We use u32 here so that we could move to a generation
+    free_slots: Vec<u32>,
+    types: HashMap<TypeId, u32>,
+    // TODO: Think about how to handle this.
+    // queued_rebuilds: Vec<Arc<[ViewId]>>,
```
Philipp-M, line 53: Care to give some context to this? Is this about fine-grained rebuilds (after reading the rest of the code, I think that it is)?
Or is it about a similar issue as mentioned here: [#xilem > App State mutation detection, "simplification" of lens @ 💬](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/App.20State.20mutation.20detection.2C.20.22simplification.22.20of.20lens/near/524866242)

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+    fn create_slot_for_type<Context>(&mut self) -> u32
+    where
+        Context: Resource,
+    {
+        match self.types.entry(TypeId::of::<Context>()) {
+            Entry::Occupied(occupied_entry) => *occupied_entry.get(),
+            Entry::Vacant(vacant_entry) => {
+                if let Some(slot) = self.free_slots.pop() {
+                    debug_assert_eq!(
+                        self.slots[usize::try_from(slot).unwrap()].ref_count,
```
Philipp-M, line 81: Out of curiosity: is there an advantage over just: `slot as usize`? (Same for the other usages of this here).

xilem_core/src/environment.rs
Philipp-M: Yes, the `change_listeners` is only appended currently (AFAICS there's not even the possibility that it's a containing a `None` currently).

For more context check out the `WithContext` View implementation which contains some comments regarding this.

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+/// This context value can be read using the [`with_context`] view with the same
+/// `Context` type parameter within child.
+///
+/// Note that it is not currently fully supported to mutate `Resource` values.
+/// In particular, if these values are read inside memoising views (such as
+/// [`memoize`](crate::memoize)), the values won't be correctly updated.
+/// We intend to address this limitation in the future.
+///
+/// This is analogous to `Context.Provider` in React.
+pub fn provides<State, Action, Context, InitialContext, ChildView, Ctx>(
```
Philipp-M, line 143: Maybe it makes sense to name it such that it's more obvious for what each context is (maybe `EnvContext` as well)?
```suggestion
pub fn provides<State, Action, Context, InitialContext, ChildView, ViewContext>(
```

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+///
+/// The provided `initial_context` will be called as soon as the returned view is
+/// built, to get the initial value of the resource.
+/// This context value can be read using the [`with_context`] view with the same
+/// `Context` type parameter within child.
+///
+/// Note that it is not currently fully supported to mutate `Resource` values.
+/// In particular, if these values are read inside memoising views (such as
+/// [`memoize`](crate::memoize)), the values won't be correctly updated.
+/// We intend to address this limitation in the future.
```
Philipp-M, line 140: That's actually a good point, we should probably check whether other (potentially future) views have similar issues, since #1079.

It would be great to have the context optionally within callbacks similar as [#xilem > How does "UseState" work and what are its use cases? @ 💬](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/How.20does.20.22UseState.22.20work.20and.20what.20are.20its.20use.20cases.3F/near/375606015), so that it can be mutated similarly as the app state itself.
But the problem there is, that we would need to use the `State` in `View` for this and it would require non-static lifetimes in `View`.

Maybe we want to have something like bevy's systems that can take all kinds of contexts/state types in event handlers/callbacks (so likely more dynamic potentially panicky code...)

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+
+    fn build(&self, ctx: &mut Ctx, app_state: &mut State) -> (Self::Element, Self::ViewState) {
+        let path: Arc<[ViewId]> = ctx.view_path().into();
+        ctx.with_id(WITH_CONTEXT_CHILD, |ctx| {
+            let env = ctx.environment();
+            let pos = env.get_slot_for_type::<Context>();
+            let Some(pos) = pos else {
+                panic!(
+                    // TODO: Track caller for this view?
+                    "Xilem: Tried to get context for {}, but it hasn't been `Provided`.",
```
Philipp-M: Maybe worth it giving a hint to `provides`, also maybe `Provided` without backticks, as it doesn't "link" to `Provides`?

Something like:
```suggestion
                    "Xilem: Tried to get context for {}, but it hasn't been provided. Did you forget to wrap this view with `xilem_core::environment::provides`?",
```

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+                    core::any::type_name::<Context>()
+                );
+            };
+            let slot_idx = usize::try_from(pos).unwrap();
+            let slot = &mut env.slots[slot_idx];
+            // TODO: Should this be &mut or just a shared ref?
```
Philipp-M, line 408: Good question, but since app state is mutable in components, it kind of makes sense to have it mutable as well (but similar caveats/issues apply I think)

xilem_core/src/view.rs
```diff
@@ -144,6 +145,11 @@ impl ViewId {
 /// For example, this context could be used to store a mapping from the
 /// id of widget to view path, to enable event routing.
 pub trait ViewPathTracker {
+    /// Access the [`Environment`] associated with this context.
+    ///
+    /// I hope that we can remove the "context" generic entirely, and so this is here
+    /// on a temporary basis.
+    fn environment(&mut self) -> &mut Environment;
```
Philipp-M, line 152: I mean less complexity/headaches (in the form of generic type params) is likely good, but I'm not yet sure how, while preserving flexibility of implementations of xilem_core (like orphan views, but I don't think it's the only use-case of a generic Context).
One way forward might be to find a nice way to convert/compose between contexts (something like `xilem_web::ViewCtx::as_owned`?).
Then maybe have this as a separate trait then, or maybe rename this trait (when it's obvious that all implementations of xilem_core should have it)?

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+                    value.change_listeners[first_empty] = Some(path);
+                    Some(first_empty)
+                } else {
+                    let idx = value.change_listeners.len();
+                    value.change_listeners.push(Some(path));
+                    Some(idx)
+                }
+            } else {
+                None
+            };
```
Philipp-M, line 458: Looks like, that this isn't yet fleshed out (and not in use), but I think we can land as is, as it's implementation is planned soon anyway.
I'm not yet sure how change-tracking should work (when it's finished) though.

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+        &self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut Ctx,
+        element: crate::Mut<'_, Self::Element>,
+        app_state: &mut State,
+    ) {
+        if let Some(_listener_idx) = view_state.listener_index {
+            // TODO: Potentially garbage collect the listener.
+            // Note that it wouldn't be correct to just remove it here, because
+            // there could be multiple listeners under the same "memoization-style" view.
```
Philipp-M, line 519: I guess that is because we don't currently have a 1:1 relation of ids <-> view isn't it (and thus the view path can point to multiple views)?

xilem_core/src/environment.rs
```diff
@@ -0,0 +1,555 @@
...
+/// A message sent to Views to instruct them to rebuild themselves.
+///
+/// This will be sent when a value in the environment value is modified.
+pub struct Rebuild;
+
+// --- MARK: Environment
+
+#[derive(Debug)]
+struct EnvironmentItem {
+    value: Box<dyn AnyMessage>,
```
Philipp-M, line 22: Is there a particular reason for using `AnyMessage` here? Should it be sent to views via `View::message` or is it for having a `Debug` requirement on resources?

Philipp-M (approved):
Looks good, there's a bunch of comments/questions and suggestions, but nothing blocking I think. I suspect the usage of a custom slotmap implementation is to have less dependencies and more potential for specialized customization (and it's just little extra code) isn't it?

xilem_core/src/environment.rs
DJMcNab: Sure - this is thinking towards the next steps, where:
1) Once you have made a mutation, the listeners for the mutated context need to be rebuilt
2) Once you know that you'll need to queue rebuilds, you need to have some way to "launch" those rebuilds. The details of that is something I'm still reasoning about.

xilem_core/src/environment.rs
DJMcNab: For usize, not really, as it's definitely bigger than u32 on all reasonable targets.
However, `as` has so many footguns, that I try to avoid it in all cases where it is avoidable.

xilem_core/src/environment.rs
DJMcNab: Given how prevalent this is in this PR, I'll defer doing this for now.

You are right that these all need renaming, which is why I made the comment about it

xilem_core/src/environment.rs
DJMcNab: Not quite - this is subtle. Basically, you only need to track the path to the closest "memoization" style view (i.e. memoize or `Arc`), if you make those views handling the `Rebuild` message themselves be required. That improves efficiency massively for these kinds of context listeners, as if there is no memoization view, you don't do anything, and if there are multiple listeners under each, then they can all share a single rebuild.

This is all presuming that actually do a full rerun of the `app_logic` - we're moving towards a world where that wouldn't actually be needed, but that's an even more radical design shift.

xilem_core/src/environment.rs
DJMcNab: Just for the `Debug` requirement.

PR #1147, opened 2025-07-11
Re-export from `masonry_core`, use in `masonry_testing`

waywardmonkeys:
We have things that are deps of `masonry_core`, `masonry`, and `masonry_testing` and if you use `masonry_core`, you need to know that and use the right versions.

Instead, we should be exporting these things from `masonry_core` and using them from there.

This updates `masonry_testing` to use things from `masonry_core` instead and subsequent commits can further improve upon this with the `masonry` crate and so on.


xorgy (approved):
:+1:

PR #1148, opened 2025-07-11
`masonry_winit`: Depend on `masonry_core` not `masonry`

waywardmonkeys:
Now that we don't use `masonry::theme::default_property_set()` and we have the re-exports in place from `masonry_core`, we can remove our dependency on `masonry` itself.


waywardmonkeys:
I don't like `masonry_core::core::....` but this PR isn't the place to go about fixing that ...

xorgy (approved):
Just gotta fix that doc link.

DJMcNab (approved):

PR #1149, opened 2025-07-11
masonry: Get `wgpu` from `vello`

waywardmonkeys:
This simplifies updating things as it is one less sibling dependency version to keep in sync.

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -15,10 +15,10 @@ use masonry_core::kurbo::Affine;
 use masonry_core::peniko::Color;
 use masonry_core::util::Instant;
 use masonry_core::vello::util::{RenderContext, RenderSurface};
+use masonry_core::vello::wgpu;
 use masonry_core::vello::{AaConfig, AaSupport, RenderParams, Renderer, RendererOptions, Scene};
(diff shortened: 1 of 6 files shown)


masonry_winit/src/event_loop_runner.rs
```diff
@@ -379,7 +379,7 @@ impl MasonryState<'_> {
             handle.clone(),
             size.width,
             size.height,
-            PresentMode::AutoVsync,
```
PoignardAzur, line 382: Inconsistent import style.

PoignardAzur (approved):
I'm not sure how much this will fix our peer dependency problems (eg for crates that import wgpu through winit) but either way, I see the benefits.

masonry_winit/src/event_loop_runner.rs
waywardmonkeys: Every other use of `wgpu` in this file uses `wgpu::` prefixes on all usages, including of the same `PresentMode` enumeration.

waywardmonkeys:
> I'm not sure how much this will fix our peer dependency problems (eg for crates that import wgpu through winit) but either way, I see the benefits.

`winit` doesn't depend on `wgpu` and we only get `wgpu` via Vello now. Other application may still have their own problems.

Issue #1119, opened 2025-07-07
Panic with stashed textarea:


waywardmonkeys:
We're doing a `CollapsePanel` where we have a button in a header that collapses (hides) a child, like for use within an accordion panel.

This sets the child widget to be stashed in the layout function, but then we get this assert during startup:

```
thread 'main' panicked at /Users/bruce/.cargo/git/checkouts/xilem-420a6f61d3d10f60/906244d/masonry/src/widgets/text_area.rs:941:14:
We just performed a layout
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_display
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:269:5
   3: core::option::expect_failed
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/option.rs:2049:5
   4: core::option::Option<T>::expect
             at /Users/bruce/.rustup/toolchains/1.88-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:958:21
   5: <masonry::widgets::text_area::TextArea<_> as masonry_core::core::widget::Widget>::accessibility
             at /Users/bruce/.cargo/git/checkouts/xilem-420a6f61d3d10f60/906244d/masonry/src/widgets/text_area.rs:941:14
   6: masonry_core::passes::accessibility::build_accessibility_tree
             at /Users/bruce/.cargo/git/checkouts/xilem-420a6f61d3d10f60/906244d/masonry_core/src/passes/accessibility.rs:62:9
   7: masonry_core::passes::accessibility::build_accessibility_tree::{{closure}}
             at /Users/bruce/.cargo/git/checkouts/xilem-420a6f61d3d10f60/906244d/masonry_core/src/passes/accessibility.rs:84:13
   8: masonry_core::passes::recurse_on_children
             at /Users/bruce/.cargo/git/checkouts/xilem-420a6f61d3d10f60/906244d/masonry_core/src/passes/mod.rs:105:9
   9: masonry_core::passes::accessibility::build_accessibility_tree
[...]

PR #1150, opened 2025-07-11
Update code which handles stashing

PoignardAzur:
Avoids the panic in #1119.

### DIFF
--- a/masonry_core/src/passes/update.rs
+++ b/masonry_core/src/passes/update.rs
@@ -368,18 +368,15 @@ fn update_stashed_for_widget(
         state.item.is_stashed = stashed;
         state.item.needs_update_focus_chain = true;
 
-        // Stashing doesn't stop the accessibility node from being generated or updated
-        // but it changes an accessibility flag.
-        state.item.needs_accessibility = true;
-        state.item.request_accessibility = true;
-
         // Items may have been changed while they were stashed in ways that require a
-        // relayout and a repaint.
+        // relayout and a re-render.
         if !stashed {
-            state.item.request_paint = true;
-            state.item.needs_paint = true;
             state.item.needs_layout = true;
             state.item.request_layout = true;
+            state.item.request_paint = true;
+            state.item.needs_paint = true;
+            state.item.needs_accessibility = true;
+            state.item.request_accessibility = true;
         }
     }
 


masonry/src/widgets/text_area.rs
```diff
@@ -930,15 +930,20 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+            0.,
+            0.,
+        );
+
+        let Some(()) = updated else {
+            // FIXME - We should be able to panic here.
+            // The accessibility pass should always run after a layout pass.
+            // See https://github.com/linebender/xilem/issues/1119
+            return;
+        };
```
waywardmonkeys, line 902: This previously panicking was the only way we knew this was going wrong ... I don't know that we should let this silently get ignored like this.

masonry_core/src/passes/paint.rs
```diff
@@ -39,7 +44,7 @@ fn paint_widget(
         widget_children: widget.children.reborrow_mut(),
         debug_paint,
     };
-    if ctx.widget_state.request_paint {
+    if ctx.widget_state.request_paint && !is_stashed {
```
waywardmonkeys, line 47: Maybe there should be a debug assert to ensure that we aren't seeing something happen where `request_paint` is true but it was stashed.

masonry_core/src/passes/paint.rs
```diff
@@ -58,16 +63,17 @@ fn paint_widget(
     state.item.request_paint = false;
     state.item.needs_paint = false;
 
-    let clip = state.item.clip_path;
-    let has_clip = clip.is_some();
-    let transform = state.item.window_transform;
-    let scene = scenes.get(&id).unwrap();
+    let has_clip = state.item.clip_path.is_some();
+    if !is_stashed {
```
waywardmonkeys, line 67: I don't know ... it just feels like if the widget its stashed, maybe it shouldn't even be getting this far and wanting to do all of this work.

masonry_core/src/passes/update.rs
```diff
@@ -383,17 +383,20 @@ fn update_stashed_for_widget(
             .update(&mut ctx, &mut props, &Update::StashedChanged(stashed));
         state.item.is_stashed = stashed;
         state.item.needs_update_focus_chain = true;
-        // Note: We don't need request_repaint because stashing doesn't actually change
-        // how widgets are painted, only how the Scenes they create are composed.
-        state.item.needs_paint = true;
+
+        // Stashing doesn't stop the accessibility node from being generated or updated
+        // but it changes an accessibility flag.
```
waywardmonkeys: I thought it was supposed to stop it....

Am I better off just not including stashed children in the children_ids from a widget or something?

masonry_core/src/passes/paint.rs
PoignardAzur: I think that case is fine. For example, if you change a stashed widget's background color, it's going to request repaint, and the request will be ignored.

masonry_core/src/passes/paint.rs
PoignardAzur: In the end the only work it does is resetting `needs_paint` flags.

masonry_core/src/passes/update.rs
PoignardAzur: > I thought it was supposed to stop it....

So far the behavior we have is that stashed widgets are included in the accessibility tree, so this is just better documenting existing code.

But I've started a discussion on Zulip about just not generating accessibility nodes for stashed widgets, which should be its own PR.

masonry_core/src/passes/layout.rs
```diff
@@ -191,6 +191,29 @@ pub(crate) fn run_layout_on<W: Widget + ?Sized>(
...
+    state.item.request_layout = false;
+
+    let id = state.item.id;
+    recurse_on_children(
+        id,
+        widget.reborrow_mut(),
+        state.children,
+        properties.children,
+        |widget, state, properties| {
+            clear_layout_flags(widget, state, properties);
```
DJMcNab, line 212: Would it be correct to `debug_assert` that `state.item.is_stashed` is true here?

DJMcNab (approved):
Looks good, thanks!

masonry_core/src/passes/layout.rs
PoignardAzur: It would be correct, yeah.

PR #1152, opened 2025-07-11
Simplify make_trace_span method

PoignardAzur:


jaredoconnell (approved):

waywardmonkeys (approved):
Was there originally some design intent to include more stuff into the span?

If nothing other than the `id` is going to be added, then all we really need on each widget is some sort of "debug name" and then this can be even simpler?

waywardmonkeys:
I mean if we have an `fn debug_name(&self) -> &str` that returned "Label" and the like, then we could have a base impelemtation of this function that called that and then the work per class is much simpler. We could probably even just use `std::any::type_name_of_val(self)` as a default.

That's a decent default, no boilerplate, can be customized by implementors if / as needed.

I do use the tracing to just observe what's going on.

DJMcNab:
> Overall I'm pretty unhappy with the state of the tracing code. The per-widget spans were pertinent back when most of the passes were highly recursive and needed to be debugged that way; but I don't think I've actually used them for debugging in a year. Given infinite time I'd like to completely rework the tracing system and the debugging workflow, but it's not a priority for now.

This would be a shame to lose, because for observing general performance metrics, they're very helpful.

At first I thought this PR was a bit strange, but you're right in observing that it's not possible to include dynamic content in the spans, so making them accept `WidgetId` is fine. There are a couple of reasonable ways forward here:
- Making the trace spans invalidatable, so that they can include dynamic context (e.g. `Flex height=1024` or w/e)
- Making the trace spans automatically, which we can now do trivially because this method is only called once, so can perform somewhat expensive operations.

PR #1153, opened 2025-07-12
masonry_core: Add `BoxConstraints::bounded_or`

xorgy:
A common layout use case is to have an inherent size (not necessarily the minimum) when the layout is unbounded.
`BoxConstraints::bounded_or` handles this on a per-axis basis, which is desirable in a number of cases, but is often not done because it is more code.


PoignardAzur (approved):
LGTM.

PR #1155, opened 2025-07-13
Re-export `winit` from `masonry_winit`

waywardmonkeys:
This reduces the dependencies to use things and removes the need for a sibling dependency.

When we move the examples back into `masonry`, it will avoid the need for a direct dev dependency on `winit` from `masonry`.

### DIFF
--- a/masonry/src/doc/01_creating_app.md
+++ b/masonry/src/doc/01_creating_app.md
@@ -165,7 +165,7 @@ The last step is to create our Winit window and start our main loop.
 
 ```rust,ignore
     use masonry::dpi::LogicalSize;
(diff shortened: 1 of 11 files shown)


masonry/src/doc/01_creating_app.md
PoignardAzur, line 1: You should also remove the `cargo add winit` part at the beginning of the tutorial.

PoignardAzur (approved):

masonry/src/doc/01_creating_app.md
waywardmonkeys: Done in #1159 since this was set to merge when ready.

PR #1158, opened 2025-07-13
Move examples from `masonry_winit` back to `masonry`

### LINKED ISSUE #1144: Move `masonry_winit` examples to `masonry`
This choice of place for examples was written at a time when we thought that Masonry Winit would be the "root crate".

However, Masonry Winit is now just the event loop runner, which will be combined with Masonry in case users want to use Winit. As such, the correct place for people to go for docs and examples is the Masonry crate itself.

This should be achievable by adding a dev dependency on `masonry_winit` to Masonry. I believe that Cargo would support this feature.

waywardmonkeys:
Fixes #1144.


waywardmonkeys:
This needs one more PR and then to be rebased ... to move the masonry-winit README content into masonry.

PR #1161, opened 2025-07-13
Fix macros to allow an arbitrary number of properties in Xilem views

PoignardAzur:
Fixes the issues encountered in  #1099.

### DIFF
--- a/xilem/src/property_tuple.rs
+++ b/xilem/src/property_tuple.rs
@@ -83,9 +83,17 @@ macro_rules! __declare_property_tuple_loop {
         )+
     ) => {
 
-        #[expect(missing_docs)]
+        /// Properties used by the
+        #[doc = stringify!($Self)]
+        /// view.
+        ///
+        /// This type implements the `PropertyTuple` trait.
+        /// Its methods should be used when writing `View::build()`
+        /// and `View::rebuild()` methods.
         pub struct $Props($(Option<$Type>,)+);
 
+        // We implement Default directly instead of calling `#[derive(Default)]`.
+        // The code should compile slightly faster this way.
         impl ::std::default::Default for $Props {
             fn default() -> Self {
                 $Props {


PoignardAzur:
Philip suggested using type composition instead, which we might adopt. This is more of a short-term fix.

See discussion on zulip: [#xilem > Refactor property system to use a composition property view](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Refactor.20property.20system.20to.20use.20a.20composition.20property.20view/with/527362645).

Philipp-M (approved):
Looks good, should at least unblock more than 12 properties.

I don't think though the trait `PropertyTuple` is necessary anymore though is it?
Could just be normal methods on all the `Props` types.

xilem/src/property_tuple.rs
```diff
@@ -154,7 +82,45 @@ macro_rules! __declare_property_tuple_loop {
             $Type: ident, $idx: tt;
         )+
     ) => {
-        type $Props = ($(Option<$Type>,)+);
+
+        #[expect(missing_docs)]
+        pub struct $Props($(Option<$Type>,)+);
+
+        impl ::std::default::Default for $Props {
```
DJMcNab, line 97:
```suggestion
        // We don't use the `Default` derive as we expect this manual implementation to be slightly faster.
        impl ::std::default::Default for $Props {
```

xilem/src/property_tuple.rs
```diff
@@ -154,7 +82,45 @@ macro_rules! __declare_property_tuple_loop {
             $Type: ident, $idx: tt;
         )+
     ) => {
-        type $Props = ($(Option<$Type>,)+);
+
+        #[expect(missing_docs)]
+        pub struct $Props($(Option<$Type>,)+);
```
DJMcNab, line 93: There are reasonable arguments for making this a struct with named fields. But I don't think that's especially worthwhile

DJMcNab, line 93: Why is this `pub`? I don't think this type is going to be part of the public API of any widget, nor have any stability guarantees.
I think it should at least have some basic docs if you're going to make it so.
Or it should be `expect(unreachable_pub)`, if it's for the associated type on `Style`

(it probably should have some basic docs anyway, even just mentioning that you should call `rebuild_properties` in `rebuild` and `build_properties` in `build`, and initialise it with `Default`)

DJMcNab (changes requested):
This looks good - I'd like to see better handling of the new public API which this creates.

xilem/src/property_tuple.rs
DJMcNab: I think this comment still stands - I'd like to understand why this type is pub before approving. I see that you've added the requested docs - thanks!

PoignardAzur: Oh, right, forgot to answer. Sorry.

It needs to be pub because it's used as an associated type for the various implementations of the `Style` trait, and you're not allowed to do that with a private type. We could do a pub-in-private trick or something similar, but I think just making it a pub item works for now.

DJMcNab (approved):
Seems reasonable enough. I have a small follow-up I'd like to do

PR #1163, opened 2025-07-13
Add TextColor and DisabledTextColor properties.

PoignardAzur:
Add these properties to Label and TextArea.
Remove with_brush/set_brush methods.
Add TextColor and DisabledTextColor properties to Label view in Xilem.
Replace `.brush()` methods in TextEdit, Prose and VariableLabel views with `.text_color()` and `.disabled_text_color()` methods which set these properties.

Making TextColor and DisabledTextColor properties also means they can now be changed app-wide with DefaultProperties.

### DIFF
--- a/masonry/src/properties/text_color.rs
+++ b/masonry/src/properties/text_color.rs
@@ -8,8 +8,8 @@ use crate::peniko::color::{AlphaColor, Srgb};
 
 /// The color of a widget's text and text decorations.
 ///
-/// **IMPORTANT:** This property is only defined for Label and TextArea, *not*
-/// for widgets embedding them such as Button, Checkbox, TextEdit, Prose, etc.
+/// **IMPORTANT:** This property is only defined for `Label` and `TextArea`, *not*
+/// for widgets embedding them such as `Button`, `Checkbox`, `TextEdit`, `Prose`, etc.
 #[expect(missing_docs, reason = "field names are self-descriptive")]
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct TextColor {
--- a/masonry/src/properties/text_color.rs
+++ b/masonry/src/properties/text_color.rs
@@ -1,0 +1,82 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::TypeId;
+
+use crate::core::{Property, UpdateCtx};
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color of a widget's text and text decorations.
+///
+/// **IMPORTANT:** This property is only defined for Label and TextArea, *not*
+/// for widgets embedding them such as Button, Checkbox, TextEdit, Prose, etc.
+#[expect(missing_docs, reason = "field names are self-descriptive")]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct TextColor {
+    pub color: AlphaColor<Srgb>,
+}
+
+impl Property for TextColor {
+    fn static_default() -> &'static Self {
+        static DEFAULT: TextColor = TextColor {
+            color: AlphaColor::BLACK,
+        };
+        &DEFAULT
+    }
+}
+
+impl TextColor {
+    /// Create new `TextColor` with given value.
+    pub fn new(color: AlphaColor<Srgb>) -> Self {
+        Self { color }
+    }
+}
+
+/// The color of a widget's text and text decorations when disabled.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct DisabledTextColor(pub TextColor);
+
+impl Property for DisabledTextColor {
+    fn static_default() -> &'static Self {
+        static DEFAULT: DisabledTextColor = DisabledTextColor(TextColor {
+            color: AlphaColor::BLACK,
+        });
+        &DEFAULT
+    }
+}
+
+// ---
+
+impl Default for TextColor {
+    fn default() -> Self {
+        *Self::static_default()
+    }
+}
+
+impl TextColor {
+    /// Helper function to be called in [`Widget::property_changed`](crate::core::Widget::property_changed).
+    pub fn prop_changed(ctx: &mut UpdateCtx<'_>, property_type: TypeId) {
+        if property_type != TypeId::of::<Self>() {
+            return;
+        }
+        ctx.request_paint_only();
+    }
+}
+
+// ---
+
+impl Default for DisabledTextColor {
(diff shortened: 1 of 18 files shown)


waywardmonkeys:
Why not `TextBrush` and `DisabledTextBrush`? It isn't great to have color properties when there will be times we want to set a brush to use a gradient or image.

xilem/src/view/label.rs
```diff
@@ -30,12 +31,12 @@ use crate::{Color, MessageResult, Pod, TextAlign, View, ViewCtx, ViewId};
 pub fn label(label: impl Into<ArcStr>) -> Label {
     Label {
         label: label.into(),
-        text_brush: Color::WHITE.into(),
         text_alignment: TextAlign::default(),
         text_size: masonry::theme::TEXT_SIZE_NORMAL,
         weight: FontWeight::NORMAL,
         font: FontStack::List(std::borrow::Cow::Borrowed(&[])),
         line_break_mode: LineBreaking::Overflow,
+        properties: Default::default(),
```
DJMcNab:
```suggestion
        properties: LabelProps::default(),
```

xilem/src/view/prose.rs
```diff
@@ -20,6 +22,7 @@ pub fn prose(content: impl Into<ArcStr>) -> Prose {
         text_size: masonry::theme::TEXT_SIZE_NORMAL,
         line_break_mode: LineBreaking::WordWrap,
         weight: FontWeight::NORMAL,
+        properties: Default::default(),
```
DJMcNab:
```suggestion
        properties: ProseProps::default(),
```

xilem/src/view/prose.rs
```diff
@@ -85,23 +89,42 @@ fn line_break_clips(linebreaking: LineBreaking) -> bool {
...
-            .with_brush(self.text_brush.clone())
             .with_text_alignment(self.text_alignment)
             .with_style(StyleProperty::FontSize(self.text_size))
             .with_style(StyleProperty::FontWeight(self.weight))
             .with_word_wrap(self.line_break_mode == LineBreaking::WordWrap);
-        let widget_pod = ctx.create_pod(
-            widgets::Prose::from_text_area(text_area)
+
+        // TODO - Handle more elegantly
+        let text_area = WidgetPod::new_with_props(text_area, self.properties.build_properties());
```
DJMcNab: Hmm yes, this is unfortunate

xilem/src/view/variable_label.rs
```diff
@@ -63,10 +64,10 @@ impl VariableLabel {
         self
     }
 
-    /// Set the brush used to paint the text.
-    #[doc(alias = "color")]
-    pub fn brush(mut self, brush: impl Into<Brush>) -> Self {
-        self.label = self.label.brush(brush);
+    /// Set the color used to paint the text.
+    pub fn color(mut self, color: Color) -> Self {
+        // TODO - Rewrite more elegantly
```
DJMcNab: This isn't a very actionable comment. What's inelegant about this form?

PoignardAzur:
> Why not TextBrush and DisabledTextBrush? It isn't great to have color properties when there will be times we want to set a brush to use a gradient or image.

Text color is simpler to implement, and matches what CSS does.

Properly implementing text gradients/images would require work that he haven't done yet. For example, what coordinate system does the image use? The existing implementation kind of didn't specify one, which means the gradient would end up (I think) being relative to the window space, which is almost never what users would want. In any case, it was undocumented.

For now, I'd rather just use TextColor, and remove the other options, since nobody was using them anyway. If someone wants non-uniform text brushes, they can add it in the PR and handle the added complexity.

xilem/src/view/variable_label.rs
PoignardAzur: I don't like that we're mixing inherent methods and property-based trait methods.

On the other hand, it feels less criminal than what I did with TextArea, so maybe it's the approach I should run with.

PoignardAzur:
I think this PR is in a good enough state to merge without splitting. I don't think my reservations about the Xilem-side code should be blocking.

masonry/src/properties/text_color.rs
```diff
@@ -0,0 +1,82 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::TypeId;
+
+use crate::core::{Property, UpdateCtx};
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color of a widget's text and text decorations.
+///
+/// **IMPORTANT:** This property is only defined for `Label` and `TextArea`, *not*
```
DJMcNab: Should we intra-doc link to `Label` and `TextArea` here?

masonry/src/widgets/label.rs
```diff
@@ -67,17 +69,6 @@ pub struct Label {
...
-    ///
-    /// Requires a new paint if edited whilst `disabled_brush` is not being used.
-    brush: Brush,
-    /// The brush to use whilst this widget is disabled.
-    ///
-    /// When this is `None`, `brush` will be used.
-    /// Requires a new paint if edited whilst this widget is disabled.
-    disabled_brush: Option<Brush>,
-    /// Whether to hint whilst drawing the text.
-    ///
```
DJMcNab: Why remove this comment?

xilem/src/style.rs
```diff
@@ -32,6 +33,24 @@ pub trait Style: Sized {
...
+    /// Set the element's text color.
+    fn text_color(mut self, color: Color) -> Self
+    where
+        Self: HasProperty<TextColor>,
+    {
+        *self.property() = Some(TextColor { color });
+        self
+    }
+
+    /// Set the element's text color when disabled.
```
DJMcNab, line 45: It might be nice to link to something user facing for `disabled` here, but iirc there isn't any in current Xilem.

DJMcNab (changes requested):
Seems reasonable enough, but is blocked on the doc comment for the hint field

masonry/src/widgets/label.rs
PoignardAzur: Good catch. I think I just fat-fingered here.

xilem/src/style.rs
PoignardAzur: We could always link to the Masonry glossary doc.

masonry/src/properties/text_color.rs
PoignardAzur: Fair enough.

DJMcNab (approved):
Approving mostly to avoid this languishing - we know that this will need a rethink

Issue #1137, opened 2025-07-09
`Stack` widget


waywardmonkeys:
A `Stack` widget is a container that only shows a single child at a time.

This can be used for something like a tab stack but also other things where you want to swap between various "pages" of UI while keeping the state loaded.

Flutter calls this an `IndexedStack` which is nice as it leaves the term `Stack` for the various forms of stack.

PoignardAzur:
Done in #1168 .

PR #1168, opened 2025-07-15
Add IndexedStack widget

KiranWells:
- Add an IndexedStack widget based on #1137 to Masonry and Xilem
- Update the widget gallery example to use IndexedStack as a tab switcher

The current implementation uses `set_stashed` to skip layout passes on inactive children without removing them directly (see [#masonry > How to handle children that should not be rendered](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/How.20to.20handle.20children.20that.20should.20not.20be.20rendered/with/528888212)).


masonry/screenshots/indexed_stack_builder_new_widget.png
DJMcNab, line 1: I think you might not have git lfs set up correctly; these files should be stored in LFS (moving away from that is more compelling these days - given that this file is only 500 bytes, but we've not made that move yet)

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+
+// --- MARK: IMPL INDEXEDSTACK ---
+impl IndexedStack {
+    /// Create a new stack with no children
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Builder-style method to add a child widget.
+    pub fn with_child(self, child: impl Widget) -> Self {
```
DJMcNab, line 39: For consistency, I *think* that this would need to check that the index is zero if this was the first child.

Maybe we should check that in `WidgetAdded`?

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+        self.with_child_pod(WidgetPod::new_with_id(child, id).erased())
+    }
+
+    /// Builder-style method to add a child widget already wrapped in a [`WidgetPod`].
+    pub fn with_child_pod(mut self, widget: WidgetPod<dyn Widget>) -> Self {
+        self.children.push(widget);
+        self
+    }
+
+    /// Builder-style method to set the active child
```
DJMcNab: I think that we do want full stops at the end of each doc line/sentence.

It is annoying that clippy doesn't give any hint here; even the stupid "the end of the doc comment, and before each double newline should be a non-latin-letter character" would be a useful addition.#

https://github.com/rust-lang/rust-clippy/issues/8371

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+    /// See also [`with_child_id`](IndexedStack::with_child_id).
+    pub fn add_child_id(this: &mut WidgetMut<'_, Self>, child: impl Widget, id: WidgetId) {
+        let child_pod: WidgetPod<dyn Widget> = WidgetPod::new_with_id(child, id).erased();
+        Self::add_child_pod(this, child_pod);
+    }
+
+    /// Add a child widget already wrapped in a [`WidgetPod`].
+    ///
+    /// See also [`with_child_pod`](IndexedStack::with_child_pod).
+    pub fn add_child_pod(this: &mut WidgetMut<'_, Self>, widget: WidgetPod<dyn Widget>) {
```
DJMcNab, line 101: Should these methods return the index the child has been pushed to? I can see reasonable arguments either way...

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+/// This is useful for switching between multiple views while keeping
+/// state loaded, such as in a tab stack.
+///
+/// The indexed stack acts as a simple container around the active child.
+/// If there is no active child, it acts like a leaf node, and takes up
+/// the minimum space.
+#[doc = crate::include_screenshot!("indexed_stack_builder_new_widget.png", "Indexed stack element showing only the fourth element in its children.")]
+#[derive(Default)]
+pub struct IndexedStack {
+    children: Vec<WidgetPod<dyn Widget>>,
```
DJMcNab, line 26: Looking at the full API, I wonder if making this be `HashMap<u32, WidgetPod<dyn Widget>>` would be better.

At the moment, you have to manually keep track of the indices outside of the widget, which is especially messy when you are removing a child.
If we did that, I also think there would be reasonable arguments for making having an unfilled active index being defined as valid and equivalent to the stack being empty). But I've not thought carefully about that

However, it's not clear how you'd make that work in Xilem without having to manually specify the index of each item. But I guess you have to be aware of the index anyway?

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+            let my_baseline = ctx.size().height - ctx.baseline_offset();
+            let line = Line::new((0.0, my_baseline), (ctx.size().width, my_baseline));
+
+            let stroke_style = Stroke::new(1.0).with_dashes(0., [4.0, 4.0]);
+            scene.stroke(&stroke_style, Affine::IDENTITY, color, None, &line);
+        }
+    }
+
+    fn accessibility_role(&self) -> Role {
+        Role::GenericContainer
```
DJMcNab, line 278: There are reasonable arguments for TabPanel here, but ultimately this is going to need more passes to get right, so using GenericContainer here now is reasonable

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    fn make_trace_span(&self, id: WidgetId) -> Span {
+        trace_span!("IndexedStack", id = id.trace())
+    }
+}
+
+// --- MARK: TESTS
+#[cfg(test)]
+mod tests {
```
DJMcNab, line 300: These tests look great

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+        harness.edit_root_widget(|mut stack| {
+            let mut stack = stack.downcast::<IndexedStack>();
+            IndexedStack::add_child(&mut stack, button::Button::new("D"));
+        });
+        assert_render_snapshot!(harness, "indexed_stack_builder_removed_widget"); // Should not change
+
+        // Make it active
+        harness.edit_root_widget(|mut stack| {
+            let mut stack = stack.downcast::<IndexedStack>();
+            IndexedStack::set_active_child(&mut stack, 1);
```
DJMcNab, line 376: This is partly what I mean - the new item you've added being index 1 is clear when you follow the logic, but it's not at all the easiest to understand.

xilem/src/view/indexed_stack.rs
```diff
@@ -0,0 +1,321 @@
...
+        {
+            let mut splice = IndexedStackSplice::new(element.reborrow_mut());
+            self.sequence
+                .seq_rebuild(&prev.sequence, view_state, ctx, &mut splice, app_state);
+            debug_assert!(splice.scratch.is_empty());
+        }
+
+        // set the active child after updating the sequence to
+        // ensure the index remains consistent with the children list
+        if prev.active_child != self.active_child {
```
DJMcNab: This is weird, because if the length of the sequence has changed, but in a way that gets recorded as a removal, but the index stays the same, then the result from rebuild would become out-of-sync with the result from a fresh `build`.

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+    use super::*;
+    use crate::assert_render_snapshot;
+    use crate::testing::TestHarness;
+    use crate::theme::default_property_set;
+    use crate::widgets::button;
+
+    #[test]
+    fn test_indexed_stack_basics() {
+        let widget = IndexedStack::new();
+        let window_size = Size::new(200.0, 200.0);
```
DJMcNab: The window size for these examples seem larger than necessary - the smaller they are, the less overhead the test has, both in runtime and on the repository.

xilem/src/view/indexed_stack.rs
DJMcNab, line 1: The semantics of this are messy, because the indices are implicit, and removing items updates the active index in strange ways.
I can see a few ways forward:
1) Make each IndexedStack item have a `.active(bool)` method, which can only validly be set on one item
2) Just ignore the problems
3) Make sure that we are more clever about resetting the index when it is needed, to make the rebuild correctly match build in all cases

masonry/src/widgets/indexed_stack.rs
```diff
@@ -0,0 +1,382 @@
...
+    /// Panics if the index is larger than the number of children.
+    pub fn insert_stack_child_pod(
+        this: &mut WidgetMut<'_, Self>,
+        idx: usize,
+        child: WidgetPod<dyn Widget>,
+    ) {
+        this.widget.children.insert(idx, child);
+        if this.widget.active_child >= idx {
+            // adjust index to keep the same widget active
+            this.widget.active_child += 1;
```
DJMcNab, line 133: I sent my review a bit early; I think perhaps the most reasonable way to make this consistent is to never change the active index in `insert_stack_child` or `remove_stack_child`.

This will be hard to document, but will avoid hard to reason about cases - the index will then always be the value you have set it to.

masonry/src/widgets/indexed_stack.rs
KiranWells: Using `WidgetAdded` isn't foolproof, since someone could theoretically remove all children, change the active index, then add a new one using `WidgetMut`. For now, I'm going to consider `self.active_child == 0 if self.children.is_empty()` an invariant (and enforce it) - that way when the first child is added, it will always be active.

The main issue is deciding what `active_child` should be when there are no children, since any value is invalid based on the validation we use elsewhere. I considered using Option, but that would make some of the interfaces more clunky.

That ties in with the issue of managing the indices in the first place - I feel like there should be a better way, but any type-based method I can think of (e.g. generics or an enum) would require the size of the children to be known at compile time. To me, this seems like a similar issue to managing a vector - you need to make sure the pointer, length, and capacity are all in sync, and check all element accesses to ensure they don't go off the end of the vector. There really isn't a way to make a dynamic container that doesn't have to do that bookkeeping.

KiranWells: It might make sense to have a different version of this Stack that uses a static enum of children using generics to make that case (such as a static set of tabs) more idiomatic/reliable.

masonry/src/widgets/indexed_stack.rs
KiranWells: I don't think that is necessary, since the widget already has `len` - the caller can use that to know what the new index is. It might be useful to document explicitly that the new widget is added to the end of the list, though.

masonry/src/widgets/indexed_stack.rs
KiranWells: If we don't modify the active index, then how do we handle the case where the index becomes invalid in `remove_stack_child`? Should the semantics be changed to allow an invalid index, and just render as empty? That would reduce the "panic-y" behavior when application logic for managing the active index breaks, but it might lead to confusing behavior instead.

masonry/src/widgets/indexed_stack.rs
KiranWells: Makes sense. Should there be a larger one to use as the documentation screenshot?

KiranWells:
I've updated the logic to be correct with the current implementation. I agree that using an index is clunky and easy to get wrong. From what I can see, there are two main use cases:

1. The static case - switching tabs, or something similar to "routing" tools in web UI libraries. In this case, the number of views and their identifiers are statically known. The ideal solution for this case is to embed the mapping from index -> view in a type, such as by using an enum. At the Xilem level, I would intuitively reach for a switch statement based on some enum stored in the state, but that has issues with the different types each match arm has. If there was some way to erase the type of the WidgetView, that logic would work, but that would be at odds with the current architecture of Xilem.
2. The dynamic case - where the user wants to add or remove views at runtime, and must handle the management of which indexes are valid. This adds a lot of extra complexity and potential for bugs.

What might work for both use cases is to move to a `HashMap` like @DJMcNab suggested, but make it generic over the key:

```rs
pub struct IndexedStack<K> {
    children: HashMap<K, WidgetPod<dyn Widget>>,
    // if active_child is not in children, render as empty
    active_child: K,
}
```

That would allow the user to select which use-case they want: using an enum as `K` allows for a static list of tabs, but a more complex application could use integers or even strings for a more complicated use case.

The biggest challenges I see with that implementation are making sure the public interfaces don't become hard to use (especially for the static use case), and managing the state reconciliation in Xilem.

masonry/src/widgets/indexed_stack.rs
DJMcNab: Hmm, that would make sense. But any except for that should be as small as is reasonable

masonry/src/widgets/indexed_stack.rs
DJMcNab: > If we don't modify the active index, then how do we handle the case where the index becomes invalid in remove_stack_child?

I think that's fine, because you only need to know the active index *after* the `WidgetMut` that you're using has finished, and we move into Layout, etc.
That is, yes it's panic prone, but I think that the alternatives are worse.

masonry/screenshots/indexed_stack_builder_new_widget.png
DJMcNab: This is still the case; these files are still binary blobs, not LFS blobs.

KiranWells:
That's strange, I changed them to lfs in https://github.com/linebender/xilem/pull/1168/commits/a6b4d18d36c595fbbf4d487d21527652b9cb91bc, but https://github.com/linebender/xilem/pull/1168/commits/3677ae14db024b98d93e4f65c1b5fa5511e69819 seems to have reverted them back. I re-ran the lfs migrate, so they should be lfs now - at least that is what it shows on my end.

xilem/examples/widgets.rs
DJMcNab, line 1: The tab focus behaviour in this example is now a bit weird. For example, I can get into a state where pressing <kbd>Shift</kbd> + <kbd>Tab</kbd> here:

Checkbox tab, with the checkbox focused.

Makes the focus disappear (and pressing space does nothing), then pressing <kbd>Shift</kbd> + <kbd>Tab</kbd> again returns to the same position.

(I'm opening this to turn it into an issue)

PR #1170, opened 2025-07-16
Migrate Xilem Core and Masonry to Anymore

DJMcNab:
[Anymore](https://github.com/linebender/anymore) is a new crate which we're developing for use in Xilem and Masonry. This crate contains the `AnyDebug` trait, which is a supertrait of both Any and Debug. We are developing this because Xilem Core and Masonry both want support for dynamically typed messages with introspection for debugging, but those have to use the same trait to allow proper interoperability.

See also [#rust ui > Anymore](https://xi.zulipchat.com/#narrow/channel/422907-rust-ui/topic/Anymore/with/528892562)

This crate has not yet published to crates.io; I intend for us to do so once we have migrated Masonry's `Action` to being only a wrapper around `Box<dyn AnyDebug + Send>`. See also discussion in #1105

### DIFF
--- a/masonry_core/src/core/action.rs
+++ b/masonry_core/src/core/action.rs
@@ -9,10 +9,11 @@ use crate::core::PointerButton;
 // See https://github.com/linebender/xilem/issues/664
 
 // TODO - TextCursor changed, ImeChanged, EnterKey, MouseEnter
-#[non_exhaustive]
 /// Events from UI elements.
 ///
 /// Note: Actions are still a WIP feature.
+#[non_exhaustive]
+#[derive(Debug)]
 pub enum Action {
     /// A button was pressed.
     ///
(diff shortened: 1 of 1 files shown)


masonry_core/src/core/action.rs
```diff
@@ -28,7 +28,7 @@ pub enum Action {
     CheckboxToggled(bool),
     // FIXME - This is a huge hack
     /// Other.
-    Other(Box<dyn Any + Send>),
+    Other(Box<dyn AnyDebug + Send>),
```
waywardmonkeys, line 32: Oh. I thought the `Action` enum itself might go away.

jaredoconnell (approved):

Philipp-M (approved):
Xilem side also looks good.

masonry_core/src/core/action.rs
Philipp-M: There's more necessary to do this on the masonry side, which I think should be a separate PR anyway (and I don't think that `AnyDebug` is the blocker here, it's more likely: https://github.com/linebender/xilem/issues/664).

DJMcNab: That's a bigger change, which this is just one part of. But yes, the current plan is for that enum to go away.

My planned next step here is to automatically unwrap the `Other` variant inside Xilem.

DJMcNab, line 32: Self review comment for tomorrow: `Action` has a `Debug` implemenation, which needs to be updated to use this.

PR #1171, opened 2025-07-17
Unwrap `Action::Other` automatically in Xilem

DJMcNab:
Step towards removing `Action` being an enum.


Philipp-M (approved):
Nice, less double-boxing and less code (at least in Xilem). Looks good.

PR #1172, opened 2025-07-17
Update Parley and AccessKit

mwcampbell:


mwcampbell:
I need the latest AccessKit release for my `VirtualScroll` accessibility PR. I don't know how soon Parley 0.6.0 is going to be released, so I updated to git main for now.

Cargo.toml
```diff
@@ -49,7 +49,7 @@ tree_arena = { version = "0.1.0", path = "tree_arena" }
 anymore = { version = "0.1.0", git = "https://github.com/linebender/anymore", rev = "3537bdc3aa03c92e380f106d18b7957b934c9eb5" }
 vello = "0.5.0"
 kurbo = "0.11.2"
-parley = { version = "0.5.0", features = ["accesskit"] }
+parley = { git = "https://github.com/linebender/parley", features = ["accesskit"] }
```
DJMcNab: Linebender style is to always use a rev for git dependencies, i.e. we need to add `rev="587b7634ae8601c10de7f0361bfd56085a5b7b4e"` here

DJMcNab (dismissed):
I think this is reasonable. However, we need to check in with @PoignardAzur, because it would also put a Parley release on the critical path for the release he is planning.

In the long term, we might need to make `parley_accesskit` be a separate crate which can be independently versioned.

PoignardAzur:
I'm fine with adding adding a Parley release on the critical path. I suspect Parley is overdue for a release anyway.

PR #1174, opened 2025-07-17
Fix AccessKit focus fallback when there's no focused widget

mwcampbell:
When no widget is focused, it's important to return the window node as the AccessKit focus, not the root widget.


mwcampbell:
This happened to do the right thing when we had `RootWidget` at the root of the accessibility tree, but removing that and adding special code to create the window node accidentally broke this.

Philipp-M (approved):
Makes sense and looks good.

PR #1176, opened 2025-07-18
Refactor flex gap

PoignardAzur:


masonry/src/widgets/flex.rs
```diff
@@ -177,42 +177,17 @@ impl Flex {
...
     /// If `gap` is not a non-negative finite value.
     ///
     /// [gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap
     // TODO: Semantics - should this include fixed spacers?
-    pub fn gap(mut self, gap: f64) -> Self {
-        if gap.is_finite() && gap >= 0.0 {
-            self.gap = Some(gap);
-        } else {
-            panic!("Invalid `gap` {gap}, expected a non-negative finite value.")
+    pub fn gap(mut self, mut gap: f64) -> Self {
```
waywardmonkeys: Rename to `with_gap` while you're here like the other builder methods?

waywardmonkeys (approved):
Approved, but 2 things:

* It might be good to have a better explanation of spacer vs gap
* It seems like `theme` should be split into `theme` for things that are really part of a theme and `defaults` for things like the default gap.

PR #1178, opened 2025-07-18
Type erase Masonry Actions

DJMcNab:
This was partially extracted from #1105 - this means that I haven't completely reviewed all of the changes. In particular, there were some merge conflicts in tests which might have had the wrong results.

The docs also still need to be updated.


masonry_core/src/core/mod.rs
```diff
@@ -43,3 +42,8 @@ pub use ui_events::{ScrollDelta, keyboard, pointer};
 pub(crate) use widget_arena::WidgetArena;
 pub(crate) use widget_pod::CreateWidget;
 pub(crate) use widget_state::WidgetState;
+
+/// Actions are emitted by Masonry Widgets to inform the.
+///
+/// This type is used internally in Masonry for.
```
waywardmonkeys: Missing text.

waywardmonkeys (approved):
There are some ci failures and some truncated text ... but apart from that, looks like a great bit of forward motion.

PR #1180, opened 2025-07-18
Remove detail sections of MSRV notices.

PoignardAzur:
See [#linebender > Should we mention MSRVs in our crate READMEs?](https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Should.20we.20mention.20MSRVs.20in.20our.20crate.20READMEs.3F/with/523325015) for rationale.

Also removes extra `<div class="rustdoc-hidden">` tag, which isn't needed when we use `cargo rdme`.

### DIFF
--- a/xilem_web/README.md
+++ b/xilem_web/README.md
@@ -57,18 +57,6 @@ This version of Xilem Web has been verified to compile with **Rust 1.88** and la
 Future versions of Xilem Web might increase the Rust version requirement.
 It will not be treated as a breaking change and as such can even happen with small patch releases.
 
-<details>
-<summary>Click here if compiling fails.</summary>
-
-As time has passed, some of Xilem Web's dependencies could have released versions with a higher Rust requirement.
-If you encounter a compilation issue due to a dependency and don't want to upgrade your Rust toolchain, then you could downgrade the dependency.
-
-```sh
-# Use the problematic dependency's name and version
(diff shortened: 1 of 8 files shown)


xilem_web/README.md
```diff
@@ -57,18 +57,6 @@ This version of Xilem Web has been verified to compile with **Rust 1.88** and la
...
-If you encounter a compilation issue due to a dependency and don't want to upgrade your Rust toolchain, then you could downgrade the dependency.
-
-```sh
-# Use the problematic dependency's name and version
-cargo update -p package_name --precise 0.1.1
-```
-</details>
-
 <!-- We hide these elements when viewing in Rustdoc, because they're not expected to be present in crate level docs -->
 <div class="rustdoc-hidden">
```
DJMcNab, line 61: 👀

DJMcNab (approved):
There's an extra rustdoc-hidden. This approval assumes that you'll fix that before merging

xilem_web/README.md
PoignardAzur: I think that one is normal. We don't do `cargo rdme` for xilem_web. Hold on, let me check...

PoignardAzur: Yup, we don't.

DJMcNab: Oh, fair enough. Sorry for the confusion!

:shipit:

PR #1179, opened 2025-07-18
Remove stashed nodes from accessibility tree.

PoignardAzur:
Fixes #1119.

Issue #1175, opened 2025-07-18
Strange tab focus behaviour in Xilem's `widgets` example


DJMcNab:
As of #1168, the tab focus behaviour in the `widgets` example of Xilem is not what I'd expect. For example, I can get into a state where pressing <kbd>Shift</kbd> + <kbd>Tab</kbd> here:

<img width="655" height="503" alt="Checkbox tab, with the checkbox focused." src="https://github.com/user-attachments/assets/cb91c73b-7cdb-43e7-b222-efa88408a02b" />

Makes the focus disappear (and pressing space does nothing), then pressing <kbd>Shift</kbd> + <kbd>Tab</kbd> again returns to the same position.

I can get into other pathological cases through more exploration just with <kbd>Tab</kbd>, <kbd>Shift</kbd> + <kbd>Tab</kbd> and <kbd>Space</kbd>.

_Originally posted in https://github.com/linebender/xilem/pull/1168#discussion_r2215650674_

This issues tracks investigating and fixing this

PR #1181, opened 2025-07-18
Prevent adding nodes to focus chain when stashed

KiranWells:
This appears to fix the issue in #1175. I'm not very familiar with the accessibility system, but from what I understand the issue was that stashed nodes were being included in the focus chain used by `update_focus`. However, when they would be focused, that was failing in some way (likely due to them being marked as hidden in the accessibility tree, again due to being stashed). This caused the `update_focus` logic to break in strange ways, including periodically failing to focus on anything when `tab` or `shift-tab` is pressed.

I checked the branch in #1179 since it looks related, but the tab selection issue persists on that branch as well.

### Testing

I did not find any regressions in the other examples during my testing, but I'm not sure how to properly test this change.

### DIFF
--- a/masonry_core/src/passes/update.rs
+++ b/masonry_core/src/passes/update.rs
@@ -467,7 +467,7 @@ fn update_focus_chain_for_widget(
         );
     }
 
-    if !state.item.is_disabled {
+    if !state.item.is_disabled && !state.item.is_stashed {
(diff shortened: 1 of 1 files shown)


waywardmonkeys (commented):
This sounds plausible to me. Needs better review from someone else though.

PoignardAzur (approved):
LGTM.

PR #1186, opened 2025-07-19
Rename doc files and modules

PoignardAzur:


PoignardAzur:
`cargo doc` fails to build the zerocopy crate. That's a new one.

masonry/src/doc/mod.rs
```diff
@@ -15,25 +15,25 @@
...
-#[doc = include_str!("./04b_widget_properties.md")]
+#[doc = include_str!("./widget_properties.md")]
 /// <style> .rustdoc-hidden { display: none; } </style>
-pub mod doc_04b_widget_properties {}
+pub mod doc_05_widget_properties {}
 
-pub use masonry_core::doc::doc_05_pass_system;
-pub use masonry_core::doc::doc_06_masonry_concepts;
+pub use masonry_core::doc::internals_01_pass_system;
+pub use masonry_core::doc::internals_02_masonry_concepts;
```
DJMcNab, line 39: Have we considered something like  `use masonry_core::doc as internals;`?

DJMcNab, line 39: The "concepts used in Masonry's APIs" being marked as "internals" feels a bit weird, but I've not carefully reviewed the actual file to see how much of it is something that users of Masonry would want to know. Thinking about this needn't block this PR

PoignardAzur: I haven't considered it.

Giving a quick thought, I think I like the current setup better. There aren't that many docs, we don't them to split them into folders.

PoignardAzur:
Quick note for the future: the numbered titles are useful for displaying stuff in the doc, but they make doc links annoying (if you change a number you have to change all the links).

I think we should use aliases to get both numberless and numberful versions.

PR #1187, opened 2025-07-19
Standardize capitalization of the word "widget" in doc.

PoignardAzur:
Avoid capitalizing it as "Widget" unless explicitly referring to the trait.

### DIFF
--- a/masonry/src/doc/02_implementing_widget.md
+++ b/masonry/src/doc/02_implementing_widget.md
@@ -1,4 +1,4 @@
-# Creating a new Widget
+# Creating a new widget
 
 <!-- Copyright 2024 the Xilem Authors -->
 <!-- SPDX-License-Identifier: Apache-2.0 -->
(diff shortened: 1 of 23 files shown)


masonry/src/doc/02_implementing_widget.md
```diff
@@ -371,7 +371,7 @@ impl Widget for ColorRectangle {
 Most of the methods we've listed take a `props: &mut PropertiesMut<'_>` argument.
 
 We won't cover properties in this chapter.
-See [Reading Widget Properties](crate::doc::doc_04b_widget_properties) for more info.
+See [Reading widget Properties](crate::doc::doc_04b_widget_properties) for more info.
```
waywardmonkeys: Properties  should also be lowercase here then (and look in 04b to see what has to be updated there as well).

waywardmonkeys (approved):

PR #1189, opened 2025-07-19
Add active status

PoignardAzur:


masonry_core/src/core/contexts.rs
```diff
@@ -920,6 +915,23 @@ impl_context_method!(
             self.widget_state.has_hovered
         }
 
+        /// The "active" status of a widget.
+        ///
+        /// A widget is "active" when the user is "actively" interacting with it.
+        /// Currently, a widget is determined to be active when it has [captured] a pointer.
```
waywardmonkeys: Is there some anticipation that this might change?

masonry_core/src/doc/06_masonry_concepts.md
```diff
@@ -51,6 +52,15 @@ Masonry should force the widget to lose pointer capture when some events occur;
 Examples of use cases for pointer capture include selecting text, dragging a slider, or long-pressing a button.
 
 
+## Active
+
+An "active" widget is one that the user is currently interacting with.
+
+Currently, a widget is determined to be active when it has pointer capture, though that definition could change.
```
waywardmonkeys: Again ... is there some anticipation that this might change? Anything might change, so why does this need to call this out specifically?

waywardmonkeys (commented):
This needs a commit message / explanation in the PR.

Since the `is_pointer_capture_target` remains, should there be a mention of how it differs from checking for `active`?

Also, this differs slightly from the `active` pseudo class in CSS ... but ends up doing the same thing, I think. Should we call out the relationship?

waywardmonkeys:
> > Also, this differs slightly from the active pseudo class in CSS ... but ends up doing the same thing, I think. Should we call out the relationship?
> 
> How does it differ?

This time, I went and actually found out rather than just noticing the rough details.

https://drafts.csswg.org/selectors/#the-active-pseudo says:

> The :active pseudo-class applies while an element is being “activated” by the user, as defined by the host language; for example, while a hyperlink is being triggered.
> 
> In addition, the [:active](https://drafts.csswg.org/selectors/#active-pseudo) pseudo-class applies while any generated box of any element (or pseudo-element) is being actively indicated by a pointing device (in the “down” state), e.g. between the time the user presses the primary mouse button and releases it, or while a finger is pressing on a touchscreen.
> 
> NOTE: [[HTML5]](https://drafts.csswg.org/selectors/#biblio-html5) defines [specific conditions for HTML elements to be activated ](https://html.spec.whatwg.org/multipage/semantics-other.html#selector-active).
> 
> An element also matches [:active](https://drafts.csswg.org/selectors/#active-pseudo) if one of its descendants in the [flat tree](https://drafts.csswg.org/css-scoping-1/#flat-tree) (including non-element nodes, such as text nodes) matches the above conditions.

The details in the HTML5 spec are interesting!

waywardmonkeys (approved):

PR #1190, opened 2025-07-19
Add ChildrenIds type alias to avoid spelling out SmallVec size.

### LINKED ISSUE #1141: Masonry's `Widget::children_ids` needs improvement
To implement `Widget::children_ids`, you currently need to:

* `use masonry::smallvec::{smallvec, SmallVec};`
* Know that the length of the `SmallVec` is 16 because it is in the return value type: `SmallVec<[WidgetId; 16]>`

This should be improved in two ways:

* There should be types / import aliases so that the dependency on `smallvec` isn't so obvious.
* There should be a type alias so that you're returning something like `ChildrenWidgetIds` but with a better name.

PoignardAzur:
Use type alias in various places.
Avoid referencing smallvec from widget code.

Fixes #1141.

### DIFF
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -103,7 +103,7 @@ impl FromDynWidget for dyn Widget {
 /// A collection of widget ids, to be returned from [`Widget::children_ids`].
 ///
 /// Internally, this uses a small vector optimisation, but you should treat it as an append-only `Vec<WidgetId>`.
-/// You can use `ChildrenIds::from_slice` with an array to make a list of children ids of known size, 
+/// You can use `ChildrenIds::from_slice` with an array to make a list of children ids of known size,
 /// or use `ChildrenIds::new` then `push` to it.
 /// This type also implements [`FromIterator<WidgetId>`](core::iter::FromIterator).
 // TODO: Consider making our own wrapper type here, to make future breaking changes easier.?
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -100,9 +100,13 @@ impl FromDynWidget for dyn Widget {
     }
 }
 
-/// Alias for the type returned by [`Widget::children_ids`].
+/// A collection of widget ids, to be returned from [`Widget::children_ids`].
 ///
-/// Essentially an optimization for `Vec<WidgetId>`.
+/// Internally, this uses a small vector optimisation, but you should treat it as an append-only `Vec<WidgetId>`.
+/// You can use `ChildrenIds::from_slice` with an array to make a list of children ids of known size, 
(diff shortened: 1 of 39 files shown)


masonry/src/doc/02_implementing_widget.md
```diff
@@ -265,13 +265,12 @@ And last, we stub in some additional methods:
...
 };
-use masonry::smallvec::SmallVec;
 
 impl Widget for ColorRectangle {
     // ...
 
     fn register_children(&mut self, _ctx: &mut RegisterCtx<'_>) {}
-    fn children_ids(&self) -> SmallVec<[WidgetId; 16]> {
+    fn children_ids(&self) -> ChildrenIds {
         SmallVec::new()
```
waywardmonkeys: This line needs updating too.

waywardmonkeys (commented):
There are still other changes to happen for this ...

* `masonry_core` should not export `smallvec`
* `masonry` should not depend upon or export `smallvec`
* Do a grep as there are other uses that should be fixed as part of this.
* Rebase this on top of #1199 once that lands.

masonry/src/doc/02_implementing_widget.md
DJMcNab: (This still applies - it's probably worth doing another grep for SmallVec)

PoignardAzur: Done.

PoignardAzur:
(I've replaced `SmallVec` with `Vec` in the VirtualScroll code. I think that's fine: the cases where we want to use VirtualScroll are ones with a lot of children.)

masonry_core/src/core/widget.rs
```diff
@@ -100,6 +100,11 @@ impl FromDynWidget for dyn Widget {
     }
 }
 
+/// Alias for the type returned by [`Widget::children_ids`].
+///
+/// Essentially an optimization for `Vec<WidgetId>`.
+pub type ChildrenIds = SmallVec<[WidgetId; 16]>;
```
DJMcNab:
```suggestion
/// A collection of widget ids, to be returned from [`Widget::children_ids`].
///
/// Internally, this uses a small vector optimisation, but you should treat it as an append-only `Vec<WidgetId>`.
/// You can use `ChildrenIds::from_slice` with an array to make a list of children ids of known size, 
/// or use `ChildrenIds::new` then `push` to it.
/// This type also implements [`FromIterator<WidgetId>`](core::iter::FromIterator).
// TODO: Consider making our own wrapper type here, to make future breaking changes easier.?
pub type ChildrenIds = SmallVec<[WidgetId; 16]>;
```

DJMcNab (approved):
Some docs improvements, otherwise looks good! I've not carefully checked that all instances are missing, but I believe that Bruce's comments are resolved.

PR #1195, opened 2025-07-19
Refactor layout code and remove `WidgetState::synthetic`

PoignardAzur:
Make `run_layout_on` function more modular:
- Make it not take a LayoutCtx parameter.
- Make it not propagate state to parent state.
Make run_layout_pass no longer create a fake WidgetState parent to the root.
Remove `WidgetState::synthetic()` constructor since it's no longer needed.

Otherwise, doesn't change any behavior.


waywardmonkeys:
I think @jaredoconnell is the right person to review this.

PR #1197, opened 2025-07-19
Create WidgetArenaMut type and start refactor of pass internals.

PoignardAzur:
This is the first step of a refactor where widgets, widget states and widget properties will be stored in the same object. So we would end up with a type like this:

```rust
struct WidgetArenaNode {
    pub(crate) widget: Box<dyn Widget>,
    pub(crate) state: WidgetState,
    pub(crate) properties: AnyMap,
}
```

and WidgetArena might look like:

```rust
pub(crate) struct WidgetArena {
    pub(crate) nodes: TreeArena<WidgetArenaNode>,
}
```

Doing so without making an absolutely unreadable diff is tough, so I split the work into several PRs. This one just moves code around to make future diffs shorter.

### DIFF
--- a/masonry_core/src/core/widget_arena.rs
+++ b/masonry_core/src/core/widget_arena.rs
@@ -120,15 +120,8 @@ impl WidgetArena {
     }
 }
 
-#[cfg(false)]
-impl<'a> WidgetArenaRef<'a> {
-    pub(crate) fn child(&self, id: WidgetId) -> Option<WidgetArenaRef<'_>> {
-        todo!()
-    }
-}
-
 impl<'a> WidgetArenaMut<'a> {
-    #[allow(unused)]
+    #[allow(unused, reason = "May be used later")]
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
I think this could still give an explanation of what's going on and what the end goal is ... There's clearly some plan given how things are being named here.

masonry_core/src/core/widget_arena.rs
```diff
@@ -106,3 +119,29 @@ impl WidgetArena {
             .expect("get_state_mut: widget state not in widget tree")
     }
 }
+
+#[cfg(false)]
+impl<'a> WidgetArenaRef<'a> {
+    pub(crate) fn child(&self, id: WidgetId) -> Option<WidgetArenaRef<'_>> {
+        todo!()
```
DJMcNab: 😆 fair enough; I was going to suggest implementing this in this PR, but I'm not surprised that the lifetimes would fight this

PR #1198, opened 2025-07-19
Remove multiple top-level `#[expect]` annotations

PoignardAzur:


masonry/src/widgets/split.rs
```diff
@@ -99,8 +99,8 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
     /// The value must be greater than or equal to `0.0`.
     /// The value will be rounded up to the nearest integer.
     pub fn min_size(mut self, first: f64, second: f64) -> Self {
-        assert!(first >= 0.0);
-        assert!(second >= 0.0);
+        assert!(first >= 0.0, "minimum size must be 0.0 or greater");
+        assert!(second >= 0.0, "minimum size must be 0.0 or");
```
waywardmonkeys: missing "greater"

masonry_core/src/core/box_constraints.rs
```diff
@@ -208,10 +206,10 @@ impl BoxConstraints {
...
     #[track_caller]
     pub fn constrain_aspect_ratio(&self, aspect_ratio: f64, width: f64) -> Size {
-        assert!(aspect_ratio.is_finite());
-        assert!(width.is_finite());
-        assert!(aspect_ratio >= 0.0);
-        assert!(width >= 0.0);
+        assert!(aspect_ratio.is_finite(), "invalid aspect_ratio");
+        assert!(width.is_finite(), "invalid width");
+        assert!(aspect_ratio >= 0.0, "invalid aspect_ratio");
+        assert!(width >= 0.0, "invalid width");
```
waywardmonkeys: In other places, you give better assert messages. I think all 4 of these should be more specific.

PoignardAzur: Right, I'll tweak the error messages in the style of the others.

Honestly, this isn't my favorite lint, and I don't think the error messages *really* need to be that detailed. The assert error messages already prints the code of the check that fails, and something like `aspect_ratio.is_finite()` is already pretty explicit.

(Also, I'd rather we use unsigned integers rather than using floats and doing a ton of checks on them, but that'd be a broader change.)

PR #1199, opened 2025-07-20
Use `Dashes` for `dash_pattern`

waywardmonkeys:
This avoids peeking through at the underlying type and uses the type as it is defined within the API.


waywardmonkeys:
Alternatively, remove the 2 dash properties and use `..Default::default()`.

PoignardAzur (approved):

PR #1200, opened 2025-07-20
Replace RawWrapper concept with much simpler RawCtx

PoignardAzur:
The main differences are:
- get_raw and get_raw_mut now both give a mutable context.
- get_raw_ and get_raw_mut are no longer included in AccessCtx.
- No need for more layers of macro copy-pasting, instead everything is in RawCtx.

Currently RawCtx is only used in Portal widget.
Places where the code was gated behind an `if false` block have been removed.

### DIFF
--- a/masonry_core/src/core/contexts.rs
+++ b/masonry_core/src/core/contexts.rs
@@ -118,6 +118,7 @@ pub struct ComposeCtx<'a> {
     pub(crate) widget_state: &'a mut WidgetState,
     pub(crate) widget_state_children: ArenaMutList<'a, WidgetState>,
     pub(crate) widget_children: ArenaMutList<'a, Box<dyn Widget>>,
+    pub(crate) properties_children: ArenaMutList<'a, AnyMap>,
 }
 
 /// A context passed to [`Widget::paint`] method.
@@ -135,9 +136,17 @@ pub struct AccessCtx<'a> {
     pub(crate) widget_state: &'a WidgetState,
     pub(crate) widget_state_children: ArenaMutList<'a, WidgetState>,
     pub(crate) widget_children: ArenaMutList<'a, Box<dyn Widget>>,
-    pub(crate) properties_children: ArenaMutList<'a, AnyMap>,
     pub(crate) tree_update: &'a mut TreeUpdate,
-    pub(crate) rebuild_all: bool,
+}
+
+/// A context given when calling another context's `get_raw()` method.
+pub struct RawCtx<'a> {
+    pub(crate) global_state: &'a mut RenderRootState,
+    pub(crate) parent_widget_state: &'a mut WidgetState,
+    pub(crate) widget_state: &'a mut WidgetState,
+    pub(crate) widget_state_children: ArenaMutList<'a, WidgetState>,
+    pub(crate) widget_children: ArenaMutList<'a, Box<dyn Widget>>,
+    pub(crate) properties_children: ArenaMutList<'a, AnyMap>,
 }
 
 // --- MARK: GETTERS
@@ -151,7 +160,9 @@ impl_context_method!(
     ComposeCtx<'_>,
     PaintCtx<'_>,
     AccessCtx<'_>,
+    RawCtx<'_>,
     {
+        #[allow(dead_code, reason = "Copy-pasted for some types that don't need it")]
         /// The `WidgetId` of the current widget.
         pub fn widget_id(&self) -> WidgetId {
             self.widget_state.id
@@ -189,6 +200,7 @@ impl_context_method!(
             child_state_ref.item
         }
 
+        #[allow(dead_code, reason = "Copy-pasted for some types that don't need it")]
         /// The current (local) transform of this widget.
         pub fn transform(&self) -> Affine {
             self.widget_state.transform
@@ -202,6 +214,7 @@ impl_context_method!(
     UpdateCtx<'_>,
     LayoutCtx<'_>,
     ComposeCtx<'_>,
+    RawCtx<'_>,
     {
         /// Helper method to get a mutable reference to a child widget's `WidgetState` from its `WidgetPod`.
         ///
@@ -339,6 +352,7 @@ impl_context_method!(
     ComposeCtx<'_>,
     PaintCtx<'_>,
     AccessCtx<'_>,
+    RawCtx<'_>,
     {
         /// Get the Parley contexts needed to build and paint text sections.
         ///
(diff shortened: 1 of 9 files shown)


masonry_core/src/core/contexts.rs
```diff
@@ -151,7 +160,9 @@ impl_context_method!(
     ComposeCtx<'_>,
     PaintCtx<'_>,
     AccessCtx<'_>,
+    RawCtx<'_>,
     {
+        #[allow(dead_code, reason = "Copy-pasted for some types that don't need it")]
```
DJMcNab: I don't see how this can ever be dead_code; the types are all pub, and it's a pub method

masonry_core/src/core/contexts.rs
```diff
@@ -1178,6 +1197,80 @@ impl_context_method!(
...
+        ///
+        /// This context lets you set pass flags for the child widget, which may be tricky.
+        /// In general, you should avoid setting the flags of a pass that runs before the
+        /// pass you're currently in.
+        ///
+        /// See [pass documentation](crate::doc::doc_05_pass_system) for the pass order.
+        pub fn get_raw_mut<Child: Widget + FromDynWidget + AllowRawMut + ?Sized>(
+            &mut self,
+            child: &mut WidgetPod<Child>,
+        ) -> (&mut Child, RawCtx<'_>) {
```
DJMcNab, line 1209: Can this return a `WidgetMut` to the child instead? That way, `RawCtx` could become read only, preserving encapsulation.

masonry_core/src/core/contexts.rs
```diff
@@ -1178,6 +1197,80 @@ impl_context_method!(
...
+        ///
+        /// This context lets you set pass flags for the child widget, which may be tricky.
+        /// In general, you should avoid setting the flags of a pass that runs before the
+        /// pass you're currently in.
+        ///
+        /// See [pass documentation](crate::doc::doc_05_pass_system) for the pass order.
+        pub fn get_raw<Child: Widget + FromDynWidget + ?Sized>(
+            &mut self,
+            child: &mut WidgetPod<Child>,
+        ) -> (&Child, RawCtx<'_>) {
```
DJMcNab, line 1175: I would expect the raw context here to only be able to read flags.  This is the motivation for my suggestion about `WidgetMut` below.

masonry_core/src/core/contexts.rs
```diff
@@ -861,6 +875,11 @@ impl_context_method!(
         pub fn to_window(&self, widget_point: Point) -> Point {
             self.widget_state.window_transform * widget_point
         }
+
+        /// Get the given child's size.
+        pub fn child_size(&self, child: &WidgetPod<impl Widget + ?Sized>) -> Size {
+            self.get_child_state(child).size
+        }
```
DJMcNab: We should document the default value before layout has been called on this method.

For example, during `WidgetAdded`, this won't be meaningfuly defined.

masonry_core/src/core/contexts.rs
```diff
@@ -1178,6 +1197,80 @@ impl_context_method!(
...
+            let widget = Child::from_dyn(&**child_mut.item).unwrap();
+
+            (widget, child_ctx)
+        }
+
+        /// Get direct mutable reference to the stored child, and a context handle for that child.
+        ///
+        /// This context lets you set pass flags for the child widget, which may be tricky.
+        /// In general, you should avoid setting the flags of a pass that runs before the
+        /// pass you're currently in.
```
DJMcNab, line 1197:
```suggestion
        /// pass you're currently in. No stability guarantees are made about Masonry's behaviour
        /// if this rule is not followed, and it may cause panics.
        /// 
        /// This method is an escape hatch for cases where a parent widget completely
        /// controls their child, but needs it to be a separate widget for user interaction to 
        /// behave as expected.
        /// As such, the child widget must opt-in using the `AllowRawMut` trait.
```

DJMcNab (commented):
I like the motivation; the old code was pretty bad - I do worry that this introduces new footguns, and so have proposed a slight tweak to how it operates. In particular, I think we should make the return type of `get_raw` weaker.

masonry_core/src/core/contexts.rs
PoignardAzur: The thing is that `MutateCtx` is basically the most powerful context type, and can do things which I'd eventually like to make exclusive to the mutate pass (e.g. adding and removing children).

I could maybe go the other way around and make `get_raw()` return a `QueryCtx`, though.

masonry_core/src/core/contexts.rs
PoignardAzur: Yup, that was a copy-pasting error.

masonry_core/src/core/contexts.rs
PoignardAzur: > I could maybe go the other way around and make get_raw() return a QueryCtx, though.

Done.

Returning a QueryCtx means a bunch of context types now need to carry a `&DefaultProperties` around, which isn't a huge cost.

masonry_core/src/core/contexts.rs
PoignardAzur: I've just removed the method, and added code in Portal to do the caching manually.

Portal still has the same problem, but (1) that was the case before (2) the Portal widget is cursed anyway.

masonry_core/src/core/contexts.rs
PoignardAzur: In practice I think setting flags out of order isn't that bad. You can already do it with `mutate_later` and realistically, the worse it's going to do is saturate the `run_rewrite_passes()` loop every frame.

It's worth adding a warning, but not at the "undefined behavior might launch your nukes" level of severity.

But yeah, added some details based on your text.

masonry_core/src/core/contexts.rs
```diff
@@ -64,13 +65,24 @@ pub struct QueryCtx<'a> {
     pub(crate) widget_state: &'a WidgetState,
     pub(crate) properties: PropertiesRef<'a>,
     pub(crate) children: WidgetArenaRef<'a>,
+    pub(crate) default_properties: &'a DefaultProperties,
+}
+
+/// A context given when calling another context's `get_raw()` method.
```
DJMcNab:
```suggestion
/// A context given when calling another context's `get_raw_mut()` method.
```

masonry_core/src/doc/pass_system.md
```diff
@@ -52,7 +52,7 @@ The animation pass may be considered as a special event pass: it's not triggered
 ## Rewrite passes
 
 After an event pass, some flags may have been changed, and some values may have been invalidated and need to be recomputed.
-To address these invalidations, Masonry runs a set of **rewrite passes** over the tree:
+To address these invalidations, Masonry runs a set of **rewrite passes** over the tree, in the following order:
```
DJMcNab, line 55: ❤️

DJMcNab (approved):
LGTM with one blocking docs tweak. Thanks!

PR #1202, opened 2025-07-20
Add color_rectangle.rs file

PoignardAzur:
Add color_rectangle mod which mirror the code examples in the tutorial. Fix bitrot in tutorials surfaced by creating color_rectangle. Add snapshot files for new tests in color_rectangle.

### DIFF
--- a/masonry/src/doc/color_rectangle.rs
+++ b/masonry/src/doc/color_rectangle.rs
@@ -200,6 +200,9 @@ impl Widget for ColorRectangle {
 
 // Implementation from "Reading widget properties" tutorial.
 #[expect(unused, reason = "example code")]
+#[expect(clippy::use_self, reason = "example code")]
+#[expect(clippy::trivially_copy_pass_by_ref, reason = "example code")]
+#[expect(clippy::all, reason = "example code")]
 impl ColorRectangle {
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, props: &PropertiesRef<'_>, scene: &mut Scene) {
         let background = props.get::<Background>();
--- a/masonry/src/doc/color_rectangle.rs
+++ b/masonry/src/doc/color_rectangle.rs
@@ -145,7 +145,6 @@ impl Widget for ColorRectangle {
 
     // ---
 
-    #[cfg(false)] // We show two `paint` implementations; check that both parse.
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene) {
         let rect = ctx.size().to_rect();
         scene.fill(
@@ -182,10 +181,15 @@ impl Widget for ColorRectangle {
     fn children_ids(&self) -> ChildrenIds {
         ChildrenIds::new()
     }
+}
 
-    // ---
+// ---
 
-    fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene) {
+// Second implementation from "Creating a new widget" tutorial.
+#[expect(dead_code, reason = "example code")]
+#[expect(clippy::trivially_copy_pass_by_ref, reason = "example code")]
+impl ColorRectangle {
+    fn paint2(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene) {
         let rect = ctx.size().to_rect();
         let color = if ctx.is_hovered() {
             Color::WHITE
@@ -208,7 +212,7 @@ impl Widget for ColorRectangle {
 #[expect(dead_code, reason = "example code")]
 #[expect(clippy::trivially_copy_pass_by_ref, reason = "example code")]
 impl ColorRectangle {
-    fn paint(&mut self, ctx: &mut PaintCtx<'_>, props: &PropertiesRef<'_>, scene: &mut Scene) {
+    fn paint3(&mut self, ctx: &mut PaintCtx<'_>, props: &PropertiesRef<'_>, scene: &mut Scene) {
         let background = props.get::<Background>();
         let rect = ctx.size().to_rect();
         scene.fill(
--- a/masonry/src/doc/color_rectangle.rs
+++ b/masonry/src/doc/color_rectangle.rs
@@ -145,6 +145,7 @@ impl Widget for ColorRectangle {
 
     // ---
 
+    #[cfg(false)] // We show two `paint` implementations; check that both parse.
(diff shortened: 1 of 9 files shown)


masonry/src/doc/color_rectangle.rs
```diff
@@ -0,0 +1,306 @@
...
+// TODO - Find some way to check that code chunks in docs
+// are up to date with this file.
+
+#![expect(missing_docs, reason = "This is example code")]
+
+use crate as masonry;
+
+use masonry::kurbo::Size;
+use masonry::peniko::Color;
+// ---
```
DJMcNab, line 22: It would be nice to have an explanatory comment for these cryptic `// ---` lines. I don't think that they add enough value to be worth keeping (at least not in the imports; I can understand it outside of them)

masonry/src/doc/color_rectangle.rs
```diff
@@ -0,0 +1,306 @@
...
+        _ctx: &mut LayoutCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        bc: &BoxConstraints,
+    ) -> Size {
+        bc.constrain(self.size)
+    }
+
+    // ---
+
+    #[cfg(false)]
```
DJMcNab:
```suggestion
    #[cfg(false)] // We show two  `paint` implementations; check that both parse.
```

masonry/src/doc/color_rectangle.rs
```diff
@@ -0,0 +1,306 @@
...
+            Some(Affine::IDENTITY),
+            &rect,
+        );
+    }
+}
+
+// ---
+
+// Implementation from "Reading widget properties" tutorial.
+#[expect(unused, reason = "example code")]
```
DJMcNab: I *think* this should be `dead_code` here - we want to get unused warnings if something in the function is unused. But I've not ran clippy with that change, so I'm happy to defer here

masonry/src/doc/color_rectangle.rs
```diff
@@ -0,0 +1,306 @@
...
+            Fill::NonZero,
+            Affine::IDENTITY,
+            &background.get_peniko_brush_for_rect(rect),
+            Some(Affine::IDENTITY),
+            &rect,
+        );
+    }
+
+    fn set_bg() {
+        let mut color_rectangle_mut: WidgetMut<'_, ColorRectangle> = None.unwrap();
```
DJMcNab:
```suggestion
    fn set_bg(mut color_rectangle_mut: WidgetMut<'_, ColorRectangle>) {
```
I think that it's too "clever" to use `None.unwrap` as a way to spell `unimplemented!()`.

If we do have to make it its own expression, I think using something like `unimplemented!("We don't have a tree set up.")` would be better.

masonry/src/doc/mod.rs
```diff
@@ -15,6 +15,11 @@
 // These docs all use the .rustdoc-hidden trick described in
 // https://linebender.org/blog/doc-include/
 
+/// Contains the items implemented in "Creating a new widget" and other tutorials.
+pub mod color_rectangle;
+
+// TODO - Add vertical_stack module.
```
DJMcNab, line 22: Maybe mention which chapters include code related to this?

masonry/src/doc/color_rectangle.rs
PoignardAzur: I don't remember why `x = None.unwrap()` works better than `x = unimplemented!()`, but I'm fairly sure it does.

Either way, your solution is much smarter, thanks.

masonry/src/doc/mod.rs
PoignardAzur: Eh, I'd rather not, that tends to bitrot fast.

masonry/src/doc/color_rectangle.rs
PoignardAzur: I've added a comment explaining them.

masonry/src/doc/color_rectangle.rs
```diff
@@ -0,0 +1,312 @@
...
+    fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene) {
+        let rect = ctx.size().to_rect();
+        scene.fill(
+            Fill::NonZero,
+            Affine::IDENTITY,
+            self.color,
+            Some(Affine::IDENTITY),
+            &rect,
+        );
+    }
```
DJMcNab, line 158: I think that the pattern around line 210 might be worth repeating here, so that the impl compiling is checked, not just parsing

masonry/src/doc/mod.rs
```diff
@@ -15,6 +15,11 @@
 // These docs all use the .rustdoc-hidden trick described in
 // https://linebender.org/blog/doc-include/
 
+/// Contains the items implemented in "Creating a new widget" and other tutorials.
+pub mod color_rectangle;
```
DJMcNab, line 20: Should this be `pub`? At the very least, the `missing_docs` expect should be removed, because these are now explorable through the public API docs.

My preference would probably lean towards making it an `example`, rather than this.
We could use something like:
```
[[examples]]
name = "color_rectangle"
path = "src/doc/color_rectangle.rs"
```
or just stick it in the examples folder.

Alternatively, this can be `doc(hidden)`.

DJMcNab (commented):
One last thought about exposing the new module in the docs; I don't think it makes sense to resolve as a follow-up, as you're best placed to explain your thinking here

masonry/src/doc/mod.rs
PoignardAzur: Yeah, I'm not quite sure. I'd like the file to be linked *somewhere* in the doc so that users can do `View Source` and compare it with the examples.

But, on second thought, I guess there isn't a strong need for users to see the full source, so `doc(hidden)` is fine.

masonry/src/doc/color_rectangle.rs
PoignardAzur: I get your point, but also part of the pattern I'm following is to make it easy to update the tutorial examples by just copy-pasting from the file. Moving the `paint` code into a separate block would break that. I guess I could move the other paint block though.

DJMcNab: I think it would make sense for them both to use the same pattern, whether that's neither being checked or both being checked.
But whatever you think works; this is a non-blocking concern (where the `doc(hidden)` question was/is blocking). 

Alternatively, we could make this whole block not be an `impl Widget` block, but that risks missing something if new widget methods are added or a signature changes.

PoignardAzur:
Wait, I remember now why I included the second paint method in the trait.

It's so the hovered screenshot works.

PR #1203, opened 2025-07-20
Add a nix flake to support development on NixOS

### LINKED ISSUE #66: Development on nixos
Not sure where this goes, but just wanted to share that I got xilem building / working on nixos. You just need to jump into a `nix-shell` with the following `shell.nix` in your tree:

```nix
let
  pkgs = import <nixpkgs> {};

  fenix = pkgs.callPackage
    (pkgs.fetchFromGitHub {
      owner = "nix-community";
      repo = "fenix";
      # commit from: 2023-03-03
      rev = "e2ea04982b892263c4d939f1cc3bf60a9c4deaa1";
      hash = "sha256-AsOim1A8KKtMWIxG+lXh5Q4P2bhOZjoUhFWJ1EuZNNk=";
    })
    { };

in pkgs.mkShell rec {

  buildInputs = with pkgs; [
    libxkbcommon xorg.libxcb libclang
    wayland wayland-protocols vulkan-loader
  ];

  nativeBuildInputs = with pkgs; [
    # Note: to use stable, just replace `default` with `stable`
    fenix.default.toolchain

    pkgs.rustPlatform.bindgenHook
    wayland-scanner
    pkg-config
  ];

  RUST_SRC_PATH = "${fenix.complete.rust-src}/lib/rustlib/src/rust/library";
  LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath buildInputs;
  LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
}
```

Philipp-M:
Adds a basic nix flake with `devShells` that uses a rust environment to run on various targets, using the MSRV `rust-version` specified in the root `Cargo.toml`.

Fixes #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,12 @@ sudo apt-get install clang libwayland-dev libxkbcommon-x11-dev libvulkan-dev
 
 There's a Nix flake in `docs/` which may be used for developing on NixOS:
 
+> [!INFO]
+> 
+> This flake is provided as a starting point, and we do not routinely validate its correctness.
+> We do not require contributors to ensure that this accurately reflects the build requirements, as we expect most contributors (and indeed many maintainers) will not be using NixOS.
+> If it is out of date, please let us know by opening an issue or PR.
+
 ```sh
 # For all crates within this repo
 nix develop ./docs
--- a/README.md
+++ b/README.md
@@ -145,6 +145,17 @@ To install the remaining packages on Debian or Ubuntu, run:
 sudo apt-get install clang libwayland-dev libxkbcommon-x11-dev libvulkan-dev
 ```
 
+There's a Nix flake in `docs/` which may be used for developing on NixOS:
+
+```sh
+# For all crates within this repo
+nix develop ./docs
+# For a specific crate
+nix develop ./docs#xilem
+nix develop ./docs#masonry
+nix develop ./docs#xilem_web
+```
+
 ## Recommended Cargo Config
 
 The Xilem repository includes a lot of projects and examples, most of them pulling a lot of dependencies.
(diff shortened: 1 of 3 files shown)


PoignardAzur (approved):
I'm not familiar with NixOS flakes, but nothing seems egregiously wrong.

Giving this a rubber stamp, since this is what we agreed on with @DJMcNab.

README.md
```diff
@@ -145,6 +145,17 @@ To install the remaining packages on Debian or Ubuntu, run:
 sudo apt-get install clang libwayland-dev libxkbcommon-x11-dev libvulkan-dev
 ```
 
+There's a Nix flake in `docs/` which may be used for developing on NixOS:
+
```
DJMcNab, line 149:
```suggestion

> [!INFO]
> 
> This flake is provided as a starting point, and we do not routinely validate its correctness.
> We do not require contributors to ensure that this accurately reflects the build requirements, as we expect most contributors (and indeed many maintainers) will not be using NixOS.
> If it is out of date, please let us know by opening an issue or PR.

```

docs/flake.nix
```diff
@@ -0,0 +1,58 @@
+{
+  inputs = {
+    flake-utils.url = "github:numtide/flake-utils";
+    rust-overlay.url = "github:oxalica/rust-overlay";
```
DJMcNab, line 4: These look very third-party. Should we alert our users that they are responsible for catching any supply chain attacks launched by these authors, or is this an acceptable risk in the Nix ecosystem?

DJMcNab (approved):
Agreed, this looks good. I've added a suggested caveat explaining that it might be out of date and why.

docs/flake.nix
Philipp-M: They're both fairly standard and widely used flake inputs. I think the risk is acceptable, similar as accepting the risk of Rust dependencies.

PR #1207, opened 2025-07-20
Make `ObjectFit`, `LineBreaking` and `Spinner` color be properties

PoignardAzur:
This lets users change the app-wide default values for them.


masonry/src/properties/object_fit.rs
```diff
@@ -1,14 +1,16 @@
...
 
+use std::any::TypeId;
+
+use masonry_core::core::{Property, UpdateCtx};
 use vello::kurbo::{Affine, Size};
 
 // These are based on https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
 /// Strategies for inscribing a rectangle inside another rectangle.
-#[derive(Clone, Copy, Default, PartialEq)]
+#[derive(Clone, Copy, PartialEq)]
```
waywardmonkeys, line 13: Why not still have a default here? also, the default isn't documented.

masonry/src/widgets/image.rs
```diff
@@ -95,7 +84,10 @@ impl Widget for Image {
             return size;
         }
         let image_aspect_ratio = image_size.height / image_size.width;
-        match self.object_fit {
+
+        let object_fit = props.get::<ObjectFit>();
+
+        match object_fit {
```
waywardmonkeys, line 93: Just leaving it as `match props.get::<ObjectFit>()` would be easier to understand (since no local that isn't otherwise unused).

masonry/src/widgets/image.rs
```diff
@@ -37,29 +39,12 @@ impl Image {
...
     pub fn new(image_data: ImageBuf) -> Self {
-        Self {
-            image_data,
-            object_fit: ObjectFit::default(),
-        }
-    }
-
-    /// Builder-style method for specifying the object fit.
-    #[inline]
-    pub fn fit_mode(mut self, mode: ObjectFit) -> Self {
```
waywardmonkeys, line 47: I think we need to have documentation with widgets about what properties can be used to customize them.

masonry/src/properties/spinner_color.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::any::TypeId;
+
+use crate::core::{Property, UpdateCtx};
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color a spinner is painted with.
```
waywardmonkeys, line 9: Why does this need something custom rather than a color property? What part of the spinner is painted with this?

PoignardAzur: The spinning "spokes" are painted with this.

I guess we could merge SpinnerColor and TextColor, which would match CSS. I'm leaving that for a future PR.

masonry/src/widgets/image.rs
PoignardAzur: Overall I like to leave `props.get` calls on their own line to make them easier to find.

masonry/src/properties/object_fit.rs
PoignardAzur: Overall, I want properties `Default` impls to always return the same as `Property::static_default()` (until const traits are stabilized and we can have it the other way around).

I've changed the code to reflect that.

masonry/src/properties/spinner_color.rs
waywardmonkeys: But you could just not have SpinnerColor in that case... so it is a very simple thing.

xilem/src/view/image.rs
DJMcNab, line 1: Can we document why we've not used the usual properties pattern here?

xilem/src/view/label.rs
```diff
@@ -3,10 +3,8 @@
 
 use masonry::core::{ArcStr, StyleProperty};
 use masonry::parley::style::{FontStack, FontWeight};
-use masonry::properties::{DisabledTextColor, TextColor};
-use masonry::widgets::{
-    LineBreaking, {self},
-};
+use masonry::properties::{DisabledTextColor, LineBreaking, TextColor};
+use masonry::widgets::{self};
```
DJMcNab:
```suggestion
use masonry::widgets;
```

xilem/src/view/prose.rs
```diff
@@ -3,10 +3,8 @@
 
 use masonry::core::{ArcStr, Properties, StyleProperty, WidgetPod};
 use masonry::parley::FontWeight;
-use masonry::properties::{DisabledTextColor, TextColor};
-use masonry::widgets::{
-    LineBreaking, {self},
-};
+use masonry::properties::{DisabledTextColor, LineBreaking, TextColor};
+use masonry::widgets::{self};
```
DJMcNab:
```suggestion
use masonry::widgets;
```

xilem/src/view/spinner.rs
DJMcNab, line 1: And here as in `Image`

masonry/src/widgets/image.rs
waywardmonkeys: I've no idea what that means since `git grep props.get` works fine either way.

PoignardAzur: I just mean keeping them visually distinct where reading the code.

masonry/src/widgets/image.rs
```diff
@@ -24,9 +27,10 @@ use crate::core::{
 /// This currently uses bilinear interpolation, which falls down when the image is
 /// larger than its layout size (e.g. it is in a [sized box](super::SizedBox) smaller
 /// than the image size).
+///
+/// You change change the sizing of the image with the [`ObjectFit`] property.
```
DJMcNab:
```suggestion
/// You can change the sizing of the image with the [`ObjectFit`] property.
```
?

masonry/src/widgets/image.rs
DJMcNab: I would also prefer it to be in the match statement, but I guess I can't block on it.

masonry/src/widgets/label.rs
```diff
@@ -29,17 +29,6 @@ use crate::{TextAlign, TextAlignOptions, theme};
...
-    /// Lines are broken at word boundaries.
-    WordWrap,
-    /// Lines are truncated to the width of the label.
-    Clip,
-    /// Lines overflow the label.
-    Overflow,
-}
-
 /// A widget displaying non-interactive text.
 ///
```
DJMcNab, line 33: While we're here, we should document all the properties that apply to this type.

masonry/src/widgets/spinner.rs
```diff
@@ -28,7 +29,6 @@ use crate::util::include_screenshot;
 #[doc = include_screenshot!("spinner_init.png", "Spinner frame.")]
 pub struct Spinner {
```
DJMcNab, line 31: And here, we should document the applicable properties

xilem/src/view/image.rs
```diff
@@ -35,24 +37,41 @@ pub fn image(image: &vello::peniko::Image) -> Image {
...
-    object_fit: ObjectFit,
+    properties: ImageProps,
 }
 
 impl Image {
+    // Because this method is image-specific, we don't add it to the Style trait.
     /// Specify the object fit.
     pub fn fit(mut self, fill: ObjectFit) -> Self {
-        self.object_fit = fill;
+        self.properties.0 = Some(fill);
```
DJMcNab:
```suggestion
        *self.property() = Some(fill);
```
Is probably the more idiomatic way (in your design for integrating property methods) to write this method.

This is non-blocking

xilem/src/view/spinner.rs
```diff
@@ -33,36 +35,53 @@ use crate::{MessageResult, Pod, View, ViewCtx};
...
+        properties: SpinnerProps::default(),
+    }
+}
+
+impl Spinner {
+    // Because this method is spinner-specific, we don't add it to the Style trait.
+    /// Set the [`SpinnerColor`] property for this spinner.
+    pub fn color(mut self, color: impl Into<Color>) -> Self {
+        let color = SpinnerColor(color.into());
+        self.properties.0 = Some(color);
```
DJMcNab: (Same here)

xilem/src/view/image.rs
PoignardAzur: ... Huh. I'd completely forgotten about that method. Thanks!

PR #1208, opened 2025-07-20
Document window_node_id

PoignardAzur:


jaredoconnell (approved):

masonry_core/src/app/render_root.rs
```diff
@@ -53,6 +53,9 @@ pub struct RenderRoot {
     /// Root of the widget tree.
     pub(crate) root: WidgetPod<dyn Widget>,
 
+    /// The accessibility pass creates a wrapper node for the app with `Role::Window`.
+    ///
+    /// This is the id of that node.
```
waywardmonkeys, line 58: Is this something that people ever need to use much (within the crate obviously)?

PoignardAzur: Idk, but having an undocumented field creates uncertainty in itself.

That is, even if you don't need to use the field, if it's undocumented, you have to worry you *might* need to use it because you don't know what it's there for.

PR #1209, opened 2025-07-20
Add `access_tree_active` flag

PoignardAzur:
Change how accessibility updates are handled to a "ignore updates if tree isn't built" model.

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -299,6 +299,7 @@ impl TestHarness {
             title: String::new(),
         };
         harness.process_window_event(WindowEvent::Resize(window_size));
+        harness.process_window_event(WindowEvent::EnableAccessTree);
 
         harness
     }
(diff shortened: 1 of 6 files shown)


mwcampbell:
I understand the desire to have separate `rebuild_access_tree` and `access_tree_active` flags, but to avoid the extra loop to mark all the widgets as needing accessibility when a rebuild is requested, maybe it would be better to have an `AccessTreeState` enum with variants like `Inactive`, `Active`, and `Rebuild`. That avoids the problem one would have with two independent flags, where the illegal state of `!access_tree_active` and `rebuild_access_tree` is representable.

masonry_testing/src/harness.rs
```diff
@@ -336,9 +336,11 @@ impl<W: Widget> TestHarness<W> {
 
         // Set up the initial state, and clear invalidation flags.
         harness.process_window_event(WindowEvent::Resize(window_size));
+        harness.process_window_event(WindowEvent::EnableAccessTree);
```
DJMcNab, line 339: 👍

It might be good to make sure that behaviour before this signal is also tested in at least one test.

I'm not sure how to do that neatly.

PR #1212, opened 2025-07-20
Add Action associated type.

PoignardAzur:
Add a `Widget::Action` associated type, and enforce (at runtime) that all submitted actions match that type.

While in theory the runtime checks add yet another risk of panic to Masonry, in practice the API is designed in such a way that you really have to call it wrong on purpose to make it panic. As long as you call `submit_action::<Self::Action>(...)`, the call will never panic.


PoignardAzur:
I like how the main API shook out, but I *really* don't like the change to the `WidgetPod::new_*` methods.

Keeping this as a draft until I find a way to avoid the latter, probably as part of the lifecycle refactor.

PoignardAzur:
Some notes re: downcasting in Xilem:

Ideally we'd like the message type given to `View::message` to always be `Widget::Action`.

In practice, it's not that simple, for a few reasons:
- `View::message` is used both for the "tree leaves", e.g. the leaf views which have no children and need to get their matching widget's actions, and for tree nodes which need to dispatch the messages to their children.
- Some views (eg: TextInput) accept messages for a child of their associated widget.
- Any modification of the View trait needs to take xilem_web into account.

Overall, I'd say all of that is beyond the scope of this PR.

masonry/src/widgets/flex.rs
```diff
@@ -6,7 +6,7 @@
 use std::any::TypeId;
 
 use accesskit::{Node, Role};
-use masonry_core::core::RegisterCtx;
+use masonry_core::core::{NoAction, RegisterCtx};
```
waywardmonkeys: These should probably be part of the `use crate::core` block below?

masonry/src/widgets/grid.rs
```diff
@@ -4,7 +4,7 @@
 use std::any::TypeId;
 
 use accesskit::{Node, Role};
-use masonry_core::core::UpdateCtx;
+use masonry_core::core::{NoAction, UpdateCtx};
```
waywardmonkeys: And these...

masonry/src/widgets/image.rs
```diff
@@ -5,6 +5,7 @@
 //! Please consider using SVG and the SVG widget as it scales much better.
 
 use accesskit::{Node, Role};
+use masonry_core::core::NoAction;
```
waywardmonkeys: And this...

masonry/src/widgets/indexed_stack.rs
```diff
@@ -4,7 +4,7 @@
 use crate::debug_panic;
 use crate::util::{fill, stroke};
 use accesskit::{Node, Role};
-use masonry_core::core::WidgetMut;
+use masonry_core::core::{NoAction, WidgetMut};
```
waywardmonkeys: and this ... (I won't note the additional ones remaining)

masonry_core/src/core/contexts.rs
```diff
@@ -1178,22 +1178,27 @@ impl_context_method!(
...
-        pub fn submit_erased_action(&mut self, action: ErasedAction) {
+        /// For further details see [`ErasedAction`](crate::core::ErasedAction).
+        pub fn submit_action<Action: AnyDebug + Send>(&mut self, action: impl Into<Action>) {
             trace!("submit_action");
-            self.global_state
-                .emit_signal(RenderRootSignal::Action(action, self.widget_state.id));
+            let action = action.into();
+            if action.type_id() != self.widget_state.action_type {
+                debug_panic!(
+                    "Trying to emit action of incorrect type {}. Type should always be `<Self as Widget>::Action",
```
waywardmonkeys: Can this include the expected type as well?

masonry_core/src/core/widget.rs
```diff
@@ -126,6 +126,13 @@ impl FromDynWidget for dyn Widget {
 /// widget should only be mutated either during a method call or through a [`WidgetMut`](crate::core::WidgetMut).
 #[allow(unused_variables, reason = "Default impls don't use method arguments")]
 pub trait Widget: AsDynWidget + Any {
+    /// The action type that will widget will submit.
+    ///
+    /// If this widget never submits action, this can be an empty type.
```
waywardmonkeys: Should mention `NoAction` more specifically here.

masonry/src/widgets/flex.rs
PoignardAzur: Since the problem is already there before this PR, I'm gonna leave it as-is. I don't think it matters a lot.

masonry_core/src/core/contexts.rs
PoignardAzur: Done.

masonry/src/widgets/text_area.rs
```diff
@@ -641,7 +643,9 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
                                 .insert_or_replace_selection("\n");
                             edited = true;
                         } else {
-                            ctx.submit_action(TextAction::Entered(self.text().to_string()));
+                            ctx.submit_action::<Self::Action>(TextAction::Entered(
```
DJMcNab, line 632: I've wondered about splitting `TextAction` into `TextEntered` and `TextChanged`, say. 

I think the gains are marginal, but I do think that would be better code. However, obviously that's incompatible with this PR.

masonry_core/src/core/contexts.rs
```diff
@@ -1178,22 +1178,38 @@ impl_context_method!(
...
-        pub fn submit_action(&mut self, action: impl AnyDebug + Send) {
-            self.submit_erased_action(Box::new(action));
-        }
-
-        /// Submit an already boxed action.
+        /// The `Action` type parameter should always be the `Self::Action` associated type
+        /// of the widget you're calling this method from.
         ///
-        /// See `submit_action` for more details.
-        pub fn submit_erased_action(&mut self, action: ErasedAction) {
```
DJMcNab, line 1275: Where's this gone?

I think there's value in this method, for a couple of reasons:
1) It's useful to have an escape hatch for this check, imo (for example, for text, you can end up with `TextChanged` and `TextSubmitted` and `LinkClicked` and (say) `Copied`, all returned in different subsets by different widgets)
2) Less importantly, but avoiding allocations in certain circumstances could be possible

masonry_core/src/core/contexts.rs
```diff
@@ -1178,22 +1178,38 @@ impl_context_method!(
...
+                    debug_panic!(
+                        "Trying to emit action of incorrect type {}. Expected type is {}. Type should always be `<Self as Widget>::Action",
+                        action.type_name(),
+                        self.widget_state.action_type_name,
+                    );
+                }
+                error!(
+                    "Trying to emit action of incorrect type {}. Type should always be `<Self as Widget>::Action",
+                    action.type_name(),
+                );
```
DJMcNab: I can't find a reason for this pattern here. It's not to avoid reading `action_type_name`, because that's not `cfg(debug_assertions)` - plus we already access `action.type_name()`, so we know that the type name we want is almost certainly already in the binary, so making `action_type_name` `cfg(debug_assertions)` doesn't matter here.

xilem/src/widget_view.rs
```diff
@@ -46,6 +47,16 @@ pub trait WidgetView<State, Action = ()>:
...
     }
+
+    /// Transform a `DynMessage` into the `Widget::Action` associated type.
+    fn downcast_message(message: DynMessage) -> Box<<Self::Widget as Widget>::Action>
+    where
+        Self::Widget: Sized,
+    {
+        message
+            .downcast::<<Self::Widget as Widget>::Action>()
+            .unwrap_or_else(|_| unreachable!())
```
DJMcNab: Can you explain how this is unreachable?
Even if we decide we want to panic for parents make a routing mistake, rather than returning a stale message (which is a reasonable change, as you've noticed that our routing techniques changed when we moved to Xilem "Masonry"), I thought that it would still be possible that this a message for a child, and the view has incorrectly not noticed that the id path wasn't empty.

At the very least, this needs an actually good panic message.

masonry_core/src/core/contexts.rs
PoignardAzur: `action_type_name` is `cfg(debug_assertions)` in WidgetState. Given that we already store a lot of stuff in there, I think that's worth doing.

masonry_core/src/core/contexts.rs
PoignardAzur: > It's useful to have an escape hatch for this check, imo (for example, for text, you can end up with TextChanged and TextSubmitted and LinkClicked and (say) Copied, all returned in different subsets by different widgets)

I don't want an escape hatch: I'd rather each widget be 100% guaranteed to only emit one type of action.

For widgets which want to emit multiple types of actions (which I expect will eventually be all of them), I'd rather give them an enum as their `Action` associated type.

(We should probably discuss this on Zulip. Either way, I don't think it blocks this PR.)

waywardmonkeys: The more that I think about this, the more uncertain I am when you have more complicated widgets that combine others. Of course, we could just do this at a view level ... but it starts to potentially reduce the utility of Masonry itself.

masonry_core/src/core/contexts.rs
PoignardAzur: Wait, no, it's not `cfg(debug_assertions)`. I forgot to add the annotation.

PoignardAzur:
I've rebased the PR and removed the `downcast_message` method, which we can leave for a future PR.

Overall I'm pretty happy with the resulting API, but I realize from the discussions in this PR that we need to have a broader discussion about how actions should work.

masonry_core/src/core/contexts.rs
PoignardAzur: See [#masonry > How should Masonry actions be typed?](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/How.20should.20Masonry.20actions.20be.20typed.3F)

masonry_core/src/core/contexts.rs
```diff
@@ -1260,22 +1260,48 @@ impl_context_method!(
             (widget, child_ctx)
         }
 
-        /// Submit an Action, which indicates that this widget requires something be handled
+        /// Submit an action, which indicates that this widget requires something be handled
         /// by the application, such as user input.
         ///
+        /// The `Action` type parameter should always be the `Self::Action` associated type
+        /// of the widget you're calling this method from.
```
DJMcNab, line 1267:
```suggestion
        /// of the widget you're calling this method from.
        /// Masonry will validate this, and this method may panic if this isn't the case.
```
Or some equivalent rewording. Your choice if you make this a true `## Panics` section.

masonry_core/src/core/widget.rs
```diff
@@ -134,6 +134,14 @@ pub type ChildrenIds = SmallVec<[WidgetId; 16]>;
 /// widget should only be mutated either during a method call or through a [`WidgetMut`](crate::core::WidgetMut).
 #[allow(unused_variables, reason = "Default impls don't use method arguments")]
 pub trait Widget: AsDynWidget + Any {
+    /// The action type that will widget will submit.
+    ///
+    /// If this widget never submits action, this can be an empty type
+    /// such as [`NoAction`](crate::core::NoAction).
```
DJMcNab, line 140: This probably should have similar security docs as the Xilem Core Message type does. However, this doesn't need to block this PR.

DJMcNab (approved):
Approving if and only if the docs of submit_action are updated to mention that it panics if the type is wrong.

The other point could be deferred, although I'd also prefer to see it actioned, or an issue opened for it.

PR #1213, opened 2025-07-20
Add `Widget::post_paint()` method

PoignardAzur:
Adds a new `post_paint()` method which paints content after children have been painted, and outside of the clip rect.

This will be useful for painting web-standards-accurate shadows, and other things which don't really fit our current painting model.

It's still missing some work; for instance, we should consider letting some widgets declare a shape which is considered "above" the widget's children when processing pointer events.

### DIFF
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -257,10 +257,16 @@ pub trait Widget: AsDynWidget + Any {
     /// Paint the widget appearance.
     ///
     /// Container widgets can paint a background before recursing to their
-    /// children, or annotations (for example, scrollbars) by painting
-    /// afterwards. In addition, they can apply masks and transforms on
-    /// the render context, which is especially useful for scrolling.
+    /// children. To draw on top of children, see [`Widget::post_paint`].
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene);
+
+    // TODO - Implement that method in some widgets.
+    // Shadows would be a good use case.
+    /// Second paint method, which paints on top of the widget's children.
+    ///
+    /// This method paints outside of the clip defined in [`LayoutCtx::set_clip_path`].
+    fn post_paint(&mut self, ctx: &mut PaintCtx<'_>, props: &PropertiesRef<'_>, scene: &mut Scene) {
+    }
 
     /// Return what kind of "thing" the widget fundamentally is.
     fn accessibility_role(&self) -> Role;
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -260,8 +260,6 @@ pub trait Widget: AsDynWidget + Any {
     /// children. To draw on top of children, see [`Widget::post_paint`].
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene);
 
-    // TODO - Implement that method in some widgets.
-    // Shadows would be a good use case.
     /// Second paint method, which paints on top of the widget's children.
     ///
     /// This method is not constrained by the clip defined in [`LayoutCtx::set_clip_path`], and can paint things outside the clip.
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -264,7 +264,7 @@ pub trait Widget: AsDynWidget + Any {
     // Shadows would be a good use case.
     /// Second paint method, which paints on top of the widget's children.
     ///
-    /// This method paints outside of the clip defined in [`LayoutCtx::set_clip_path`].
+    /// This method is not constrained by the clip defined in [`LayoutCtx::set_clip_path`], and can paint things outside the clip.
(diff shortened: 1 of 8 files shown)


Philipp-M (commented):
Motivation in the PR/commit message would be nice (generally for all kinds of PRs, as it would help/motivate potentially reviewers), I can't read that (quickly) out of the code.

Philipp-M:
I also wonder whether we need a new widget method for this, as it may also be possible to have multiple paint passes instead of just "hardcoded" 2, so we may also implement that behaviour within `Widget::paint`, and allow potentially even more passes (within reason), like setting a flag (and maybe info when, e.g. before/after childrens `paint`) in the `PaintCtx`?

PoignardAzur:
I think having two hard-coded layers is both much easier to handle and covers 99% of realistic cases.

PoignardAzur:
> Motivation in the PR/commit message would be nice (generally for all kinds of PRs, as it would help/motivate potentially reviewers), I can't read that (quickly) out of the code.

Fair enough. In my defense, I've been churning these out as fast as I can write them.

masonry_core/src/core/widget.rs
```diff
@@ -257,11 +257,17 @@ pub trait Widget: AsDynWidget + Any {
...
-    /// afterwards. In addition, they can apply masks and transforms on
-    /// the render context, which is especially useful for scrolling.
+    /// children. To draw on top of children, see [`Widget::post_paint`].
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene);
 
+    // TODO - Implement that method in some widgets.
+    // Shadows would be a good use case.
+    /// Second paint method, which paints on top of the widget's children.
+    ///
+    /// This method paints outside of the clip defined in [`LayoutCtx::set_clip_path`].
```
DJMcNab: These docs are bit ambiguous. We should specify that it only draws outside "your" clip layer, not all of them.

Also, it's probably worth clarifying that the method *can* draw outside of the clip, not *must*. That is, the drawing is unconstrained by the specified clip path.

masonry_core/src/passes/paint.rs
```diff
@@ -37,6 +37,8 @@ fn paint_widget(
     // (See WidgetState doc.)
     let is_stashed = state.is_stashed;
 
+    // TODO - Check different flags for paint and post_paint.
```
DJMcNab: I think that we *probably* should do this now. There a reasonable amount of overhead in handling `post_paint`, even though most widgets won't use it.

masonry_core/src/passes/paint.rs
```diff
@@ -67,7 +71,7 @@ fn paint_widget(
     let has_clip = state.clip_path.is_some();
     if !is_stashed {
         let transform = state.window_transform;
-        let scene = scenes.get(&id).unwrap();
+        let scene = &mut scene_cache.entry(id).or_default().0;
```
DJMcNab: It's not clear to me when this `or_default` triggering wouldn't be a bug in Masonry. Can you clarify why you've made this change here?

masonry_testing/src/modular_widget.rs
```diff
@@ -200,6 +202,15 @@ impl<S> ModularWidget<S> {
         self
     }
 
+    /// See [`Widget::post_paint`]
```
DJMcNab, line 205: Missing full stop in doc comment. Although I see that this is the prevailing style in this file...

masonry_core/src/core/widget.rs
```diff
@@ -257,11 +257,17 @@ pub trait Widget: AsDynWidget + Any {
...
     ///
     /// Container widgets can paint a background before recursing to their
-    /// children, or annotations (for example, scrollbars) by painting
-    /// afterwards. In addition, they can apply masks and transforms on
-    /// the render context, which is especially useful for scrolling.
+    /// children. To draw on top of children, see [`Widget::post_paint`].
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, _props: &PropertiesRef<'_>, scene: &mut Scene);
 
+    // TODO - Implement that method in some widgets.
+    // Shadows would be a good use case.
```
DJMcNab: I'm not sure that this is really a good todo comment. We shouldn't add this method if we're not planning on doing this, and so it will probably become instantly stale.

DJMcNab (commented):
Broadly, the implementation seems reasonable, but I've got a few thoughts here.

masonry_core/src/passes/paint.rs
PoignardAzur: You're right, this is defensive programming. I think I should make it a debug panic instead.

Come to think of it, we should probably document our panicking philosophy somewhere if we haven't already.

masonry_core/src/core/widget.rs
PoignardAzur: Good points.

masonry_testing/src/modular_widget.rs
PoignardAzur: Yeah, problem for later.

masonry_core/src/passes/paint.rs
PoignardAzur: Alright, I've done an implementation.

I don't really like the complexity the new flag as, and the code I wrote feels a little brittle, but I don't want to stall this PR on finding a perfect solution.

PR #1214, opened 2025-07-20
Add `TestHarness::scroll_into_view` method.

PoignardAzur:
This may be useful in cases where the user wants the harness to click something that would otherwise be scrolled away.

### DIFF
--- a/masonry_core/src/passes/event.rs
+++ b/masonry_core/src/passes/event.rs
@@ -361,6 +361,14 @@ pub(crate) fn run_on_access_event_pass(
                 handled = Handled::Yes;
             }
         }
+        accesskit::Action::ScrollIntoView if !handled.is_handled() => {
+            let widget_state = root.widget_arena.get_state(target);
+            let rect = widget_state.item.layout_rect();
+            root.global_state
+                .scroll_request_targets
+                .push((target, rect));
+            handled = Handled::Yes;
+        }
         _ => {}
     }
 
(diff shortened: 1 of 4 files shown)


PoignardAzur:
(@DataTriny I don't know if you're interested in reviewing Masonry PRs?)

masonry_core/src/passes/event.rs
```diff
@@ -361,6 +361,14 @@ pub(crate) fn run_on_access_event_pass(
...
             }
         }
+        accesskit::Action::ScrollIntoView if !handled.is_handled() => {
+            let widget_state = root.widget_arena.get_state(target);
+            let rect = widget_state.item.layout_rect();
+            root.global_state
+                .scroll_request_targets
+                .push((target, rect));
+            handled = Handled::Yes;
+        }
```
DJMcNab, line 371: Indeed - if we land this, we can remove the equivalent code in #1077 (see https://github.com/linebender/xilem/pull/1077#discussion_r2219635181).

masonry_testing/src/harness.rs
```diff
@@ -576,6 +577,23 @@ impl TestHarness {
...
+    ///
+    /// [`RequestPanToChild`]: masonry_core::core::Update::RequestPanToChild
+    /// [`ScrollIntoView`]: masonry_core::accesskit::Action::ScrollIntoView
+    #[track_caller]
+    pub fn scroll_into_view(&mut self, id: WidgetId) {
+        self.render_root.handle_access_event(ActionRequest {
+            action: Action::ScrollIntoView,
+            target: id.to_raw().into(),
+            data: None,
+        });
```
DJMcNab, line 594: Oh, this is an interesting implementation approach. I like it!

PoignardAzur:
I've added a test to portal.

DJMcNab (approved):
Great, thanks! We'll have to be careful in #1077 - if you want to merge this before then, can you push the commit to remove the highlighted code to that PR (otherwise, we can wait and remove it in this PR instead)

mwcampbell (approved):

PR #1215, opened 2025-07-21
Create `NewWidget` type and refactor all methods that add a child widget to use that type

PoignardAzur:
This is a fairly large PR, with a few motivations:

- To introduce the concept of a `NewWidget` type which holds all the options a widget will be created with.
- To standardize how new widgets are added, and remove the code duplication in `with_child`, `with_child_id`, `with_child_pod`, `insert_child`, `insert_child_pod`, etc, methods.
- To make the Xilem-side code much simpler by taking care of the widget creation options on the Masonry side.
- To prepare the way for #1212.
- To prepare the upcoming lifecycle refactor, in which NewWidget will be the only way to add a widget to the widget tree.

Overall I think it makes the Masonry APIs more elegant, at the cost of making the builder-style "Create a widget tree with the children already filled in" widget construction more verbose.
I think that's a worthwhile trade-off because (1) we still expect most of our users to be Xilem-style frameworks which can tolerate the verbosity (2) I expect the builder-style constructors to be significantly altered in the lifecycle refactor (though I think this PR is still worth adding even if the lifecycle refactor doesn't come).


masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
...
+    pub properties: Properties,
+}
+
+/// The options a new widget will be created with.
+#[derive(Default, Debug)]
+pub struct WidgetOptions {
+    /// The transform the widget will be created with.
+    pub transform: Affine,
+    /// The disabled state the widget will be created with.
+    pub disabled: bool,
```
waywardmonkeys, line 45: If doing this, allowing a widget to be created in the stashed state would also be useful. That is probably a follow up issue, but mentioning it here for now.

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
...
+}
+
 // TODO - This is a simple state machine that lets users create WidgetPods
```
waywardmonkeys, line 48: Is this whole comment still true? It wasn't clear if it was referring to the bunch of stuff following it or just the `CreateWidget` struct which you're removing.

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
...
+            widget: Box::new(inner),
+            id,
+            options: WidgetOptions::default(),
+            properties: Properties::default(),
+        }
     }
 
-    /// Create a new widget pod with properties.
+    /// Create a new widget with properties.
     pub fn new_with_props(inner: W, props: Properties) -> Self {
```
waywardmonkeys, line 79: Also probably a follow up, but having a way to have this take `Into<Properties>` and allowing creating that from various things (including a single property) would improve some of these APIs.

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
...
+            widget: Box::new(inner),
+            id: WidgetId::next(),
+            options,
+            properties: Properties::default(),
+        }
     }
 
-    /// Create a new widget pod with custom options and a pre-set [`WidgetId`].
-    pub fn new_with_id_and_options(inner: Box<W>, id: WidgetId, options: WidgetOptions) -> Self {
+    /// Create a new widget with custom options and custom [`Properties`].
```
waywardmonkeys: link options if you link properties.  Also, should call out the custom `WidgetId` as well?

waywardmonkeys (commented):
I'm watching to see how this evolves and how things look afterwards. I'm concerned that it adds yet another new concept and perhaps more ceremony to something that people do all the time. But maybe it also lets us improve that code ... in some cases, it looks like it, in other cases, I'm not sure yet.

It also looks like this isn't as new as it seems since there was already the `CreateWidget` and other stuff, so this is perhaps more of code in motion than changes? (But removing things with static lifetime bounds is nice.)

masonry_core/src/core/widget_pod.rs
PoignardAzur: It's still true. The "state machine" is specifically the enum that follows.

masonry_core/src/core/widget_pod.rs
PoignardAzur: Yeah, I thought about it. This is something where variadic generics would really help, though we can always use the macro tricks.

masonry_core/src/core/widget_pod.rs
DJMcNab: I don't think that's the case, because the stashed state is *only* controlled by the parent.
That being said, there should be a way for a parent to add a widget as stashed.

It's the subtle difference between creating in the stashed state and added in the stashed state; I don't think it makes sense as a field of NewWidget (but I've not carefully studied this)

masonry/src/widgets/flex.rs
```diff
@@ -386,58 +352,36 @@ impl Flex {
...
-    /// Insert a non-flex child widget at the given index.
-    ///
-    /// # Panics
-    ///
-    /// Panics if the index is larger than the number of children.
-    pub fn insert_child(this: &mut WidgetMut<'_, Self>, idx: usize, child: impl Widget) {
-        Self::insert_child_pod(this, idx, WidgetPod::new(child).erased());
-    }
-
     /// Insert a non-flex child widget wrapped in a [`WidgetPod`] at the given index.
```
DJMcNab: This doc probably needs updating

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
     inner: WidgetPodInner<W>,
 }
 
+/// A container for a widget yet to be inserted.
+#[non_exhaustive]
+#[expect(missing_docs, reason = "Names are self-explanatory.")]
+pub struct NewWidget<W: ?Sized> {
+    pub widget: Box<W>,
+    pub id: WidgetId,
```
DJMcNab: Should this id be pub? If it can be changed, its hard to rely upon.

E.g. in Xilem as it stands, you could have a `WithWidgetId` view, which assigns a specific id to the element created by its child, and that would *seem* valid, but would fall down if any side channels had already stashed the id (i.e. for event handling, but maybe anything else as well?)

Obviously mixing pub/non-pub fields is annoying. It would be nice to at least warn here.

(And so obviously you can `mem::replace` or equivalent to achieve the same effect, but that's more obviously wrong)

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
...
+    pub options: WidgetOptions,
```
DJMcNab, line 33: We could do with an explanation of why this is a separate type here, because it's not obvious how NewWidget and WidgetOptions are different (or why e.g. the properties are not part of `WidgetOptions`)

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
     inner: WidgetPodInner<W>,
 }
 
+/// A container for a widget yet to be inserted.
+#[non_exhaustive]
+#[expect(missing_docs, reason = "Names are self-explanatory.")]
```
DJMcNab: I'll note again my general displeasure at this approach to docs here, but I won't block on adding fields.

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,89 +15,120 @@ pub struct WidgetPod<W: ?Sized> {
     inner: WidgetPodInner<W>,
 }
 
+/// A container for a widget yet to be inserted.
```
DJMcNab, line 18: This is a really oblique doc comment. Perhaps we can add something like:

> Widgets have several additional attributes which are controlled by the app, which are common to all widgets.
> `NewWidget` contains a widget which hasn't yet been inserted into the widget tree, and the initial values which will be used for those attributes when it is added.
> This allows different areas to coordinate setting these values (for example, in Xilem, the transform is set by a dedicated view)

masonry_core/src/core/widget.rs
```diff
@@ -391,6 +391,14 @@ pub trait Widget: AsDynWidget + Any {
             .last()
             .unwrap_or(name)
     }
+
+    /// Convenience method to create a `NewWidget` from this.
+    fn with_next_id(self) -> NewWidget<Self>
```
DJMcNab: I don't hugely like this method name, but I don't have a better suggestion. Maybe `with_auto_id` or `hydrate`?

Essentially, `with_next_id` says what it does, but not really why you have to use it.

masonry_core/src/core/widget_pod.rs
DJMcNab: We should probably also point to with_next_id here

masonry_testing/src/recorder_widget.rs
```diff
@@ -263,4 +263,11 @@ impl<W: Widget> Widget for Recorder<W> {
     fn short_type_name(&self) -> &'static str {
         "Recorder"
     }
+
+    fn with_next_id(self) -> NewWidget<Self>
+    where
+        Self: Sized,
+    {
+        NewWidget::new(self)
+    }
```
DJMcNab: I'm surprised that this necessary.

xilem/src/pod.rs
```diff
@@ -1,38 +1,27 @@
...
+///
+/// If changing transforms of widgets, make sure to use [`transformed`]
+/// (or [`WidgetView::transform`]).
+/// This has a protocol to ensure that multiple views changing the
+/// transform interoperate successfully.
+///
+/// [`transformed`]: crate::view::Transformed
+/// [`WidgetView::transform`]: crate::view::transformed
 #[expect(missing_docs, reason = "TODO - Document these items")]
 pub struct Pod<W: Widget + FromDynWidget + ?Sized> {
```
DJMcNab, line 23: We probably should implement `Deref<Target=NewWidget<W>> + DerefMut` here and add our own wrapping `to_pod` method here.

xilem/src/pod.rs
```diff
@@ -91,7 +54,9 @@ impl<W: Widget + FromDynWidget + ?Sized> ViewElement for Pod<W> {
 
 impl<W: Widget + FromDynWidget + ?Sized> SuperElement<Pod<W>, ViewCtx> for Pod<dyn Widget> {
     fn upcast(_: &mut ViewCtx, child: Pod<W>) -> Self {
-        child.erased()
+        Pod {
+            new_widget: child.new_widget.erased(),
+        }
```
DJMcNab: Why doesn't `child.erased()` still work here?

xilem/src/view/button.rs
```diff
@@ -151,9 +151,9 @@ where
             View::<State, Action, _>::build(&self.label, ctx, app_state)
         });
         ctx.with_leaf_action_widget(|ctx| {
-            let mut pod = ctx.create_pod(widgets::Button::from_label_pod(child.into_widget_pod()));
-            pod.properties = self.properties.build_properties();
-            pod.options.disabled = self.disabled;
+            let mut pod = ctx.create_pod(widgets::Button::from_label(child.new_widget));
+            pod.new_widget.properties = self.properties.build_properties();
```
DJMcNab, line 155: This is quite ugly, and I think that adding Deref would fix that

masonry_testing/src/recorder_widget.rs
PoignardAzur: There's a clippy lint to make sure we don't forget methods. In this case it's a little redundant.

xilem/src/view/button.rs
PoignardAzur: I don't think it's that ugly? It's only one more identifier per line.

masonry_core/src/core/widget_pod.rs
PoignardAzur: I'm leaning towards having them all be private and using builder methods to set them.

masonry_core/src/core/widget_pod.rs
PoignardAzur: I think the simplest solution might be to remove WidgetOptions, but I'd want to do that as a follow-up PR.

xilem/src/pod.rs
PoignardAzur: I'll leave that to a follow-up PR. I don't think the current syntax is so verbose that a `Deref` impl is desperately needed.

(And that `to_pod` method would only be used in two files, I think it'd be fine to skip it as well.)

xilem/src/pod.rs
```diff
@@ -1,38 +1,27 @@
...
+/// If changing transforms of widgets, make sure to use [`transformed`]
+/// (or [`WidgetView::transform`]).
+/// This has a protocol to ensure that multiple views changing the
+/// transform interoperate successfully.
+///
+/// [`transformed`]: crate::view::Transformed
+/// [`WidgetView::transform`]: crate::view::transformed
 #[expect(missing_docs, reason = "TODO - Document these items")]
 pub struct Pod<W: Widget + FromDynWidget + ?Sized> {
-    pub widget: Box<W>,
```
DJMcNab, line 23: Losing this field in Xilem is unfortunate, because it means that you can no longer have composing views for setting fields (other than transform/disabled/properties).
I see that since https://github.com/linebender/xilem/pull/837, this has been boxed, which interacts pretty poorly with the builder methods (you have to unbox and rebox at every layer).

masonry_core/src/core/widget_pod.rs
```diff
@@ -13,117 +15,142 @@ pub struct WidgetPod<W: ?Sized> {
...
+///
+/// In general, functions which take a new widget and add it to the tree
+/// (e.g. `FooBarContainer::add_child_widget()`) should take a `NewWidget` as
+/// a parameter.
+///
+/// `NewWidget` holds both the widget itself and additional metadata which will be stored
+/// alongside it once it's added to the tree.
+#[non_exhaustive]
+pub struct NewWidget<W: ?Sized> {
+    pub(crate) widget: Box<W>,
```
DJMcNab: It's not even clear to me why this one isn't public; really I think changing the id is the problematic operation - before the widget has been added, it's just plain data.

DJMcNab (commented):
The latest changes effectively revert https://github.com/linebender/xilem/pull/806.

The fact that #837 changing that wasn't caught until now is indicative, but I still don't think removing this capability from Xilem is a good thing.

PR #1216, opened 2025-07-21
Add doc aliases so we can renumber doc chapters without changing imports

PoignardAzur:


masonry/src/doc/mod.rs
```diff
@@ -35,5 +35,19 @@ pub mod doc_04_testing_widget {}
...
+pub use self::doc_02_implementing_widget as implementing_widget;
+#[doc(hidden)]
+pub use self::doc_03_implementing_container_widget as implementing_container_widget;
+#[doc(hidden)]
+pub use self::doc_04_testing_widget as testing_widget;
+#[doc(hidden)]
+pub use self::doc_05_widget_properties as widget_properties;
+
+pub use masonry_core::doc::masonry_concepts;
+pub use masonry_core::doc::pass_system;
```
DJMcNab: I'm taking on trust that you've checked that the `doc(hidden)` transfers over properly.

PoignardAzur: That's actually a "I did a find-and-replace to the entire codebase" error, good catch. I renamed them back.

PR #1217, opened 2025-07-21
Add WidgetTag

PoignardAzur:
Add a new WidgetTag type, which lets you access widgets without having to store their WidgetIds and lets you skip downcasting.
Tweak the to-do-list example to demonstrate how to use those tags.

Tags are similar to Selectors back in Druid and have a similar purpose: to carry type information between places in the code.

This is a first step towards adding a more generalized query system to Masonry.

### DIFF
--- a/masonry_core/src/core/widget_pod.rs
+++ b/masonry_core/src/core/widget_pod.rs
@@ -4,7 +4,7 @@
 use std::any::TypeId;
 use vello::kurbo::Affine;
 
-use crate::core::{Properties, Widget, WidgetId};
+use crate::core::{Properties, Widget, WidgetId, WidgetTag};
 
 /// A container for one widget in the hierarchy.
 ///
@@ -37,6 +37,8 @@ pub struct NewWidget<W: ?Sized> {
     pub options: WidgetOptions,
     /// The properties the widget will be created with.
     pub properties: Properties,
+
+    pub(crate) tag: &'static str,
 }
 
 // TODO - Remove this and merge it into NewWidget?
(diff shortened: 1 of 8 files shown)


jaredoconnell:
Will this be able to be utilized to assign focus programmatically in the future?

PoignardAzur:
> I guess the better question is if this would make it more user friendly, utilizing the tag instead of the widget itself.

I think it's still too simple for non-example use cases, but yeah, it's the direction I want to go.

PoignardAzur:
Having one unique-per-app string identifier is how web ids work. And while id collisions are certainly possible, my experience is they virtually never happen even in complex web apps.

I don't think this feature will be fragile in practice.

PoignardAzur:
> As such, I'd approve this PR if the change to to_do_list were reverted, and one or two unit tests of a widget were tweaked to use this mechanism.

I've pushed #1267 which changes a few unit tests to use this feature.

I'd rather not revert to_do_list unless you're very strongly opposed to the current version. If nothing else, I find the new code much more readable and less fragile (the code will no longer break every single time you e.g. add a spacer to the root flex).

masonry_core/src/core/widget_pod.rs
```diff
@@ -37,6 +37,8 @@ pub struct NewWidget<W: ?Sized> {
     pub options: WidgetOptions,
     /// The properties the widget will be created with.
     pub properties: Properties,
+
```
DJMcNab, line 40: Can we link to `WidgetTag` here?

DJMcNab (approved):
I still strongly don't think this is a good mechanism for user code - certainly not using globals for this. Like, the reason that there aren't issues with id collision in web code, in my experience, is that nothing actually uses ids (except maybe to access a root for client-side rendered DOM nodes).

But I'm not going to block on it, and it's only in the one example.

(Again, for tests I think this is a perfectly reasonable solution for now)

masonry_core/src/core/widget_pod.rs
PoignardAzur: We already do below, and I do want to keep the field private.

DJMcNab: I don't see how a field being private prevents it having docs though... this field is quite oblique. The same applies to the new `widget_tags`, incidentally, which doubles the number of undocumented fields in `RenderRootState`.

PR #1221, opened 2025-07-23
`masonry`: Import from `masonry_core` via `crate::core`

waywardmonkeys:


masonry/src/widgets/label.rs
```diff
@@ -421,12 +421,12 @@ impl Widget for Label {
 #[cfg(test)]
 mod tests {
 
-    use masonry_core::core::Properties;
     use masonry_testing::TestWidgetExt as _;
```
DJMcNab, line 424: This could technically use the `crate::testing` thing below, but IMO the form from `masonry_testing` would probably be more idiomatic

DJMcNab (approved):

masonry/src/widgets/label.rs
waywardmonkeys: I only did `masonry_core` here ... and I think that we should remove the `masonry::testing` thing and just have `masonry_testing` everywhere ... that'll make internal usage match external usage. But that'll be a separate PR.

DJMcNab: I had assumed both counts here!

PR #1223, opened 2025-07-23
Add layout caching to Masonry and remove `skip_layout()` method.

PoignardAzur:
Remove caching code from `Flex`.
Fix how layout handles stashing flags.

`skip_layout()` was used for stashed widgets (no longer necessary) and layout caching, which we now handle at the Masonry level.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -29,7 +29,6 @@ pub struct Flex {
     main_alignment: MainAxisAlignment,
     fill_major_axis: bool,
     children: Vec<Child>,
-    old_bc: BoxConstraints,
     gap: f64,
 }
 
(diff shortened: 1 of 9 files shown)


waywardmonkeys (commented):
I agree with @jaredoconnell that a review from @jaredoconnell is needed here.

jaredoconnell (approved):
I did some basic performance tests. Switching from Flex's old cache to the new built-in cache resulted in slightly more sporadic timings for layout. now it's 4-12ms instead of the steady 6 ms. However, that's in debug mode, and in release mode the layout passes take less than one millisecond, so that's not an issue in the scale of examples like http_cats. The performance is still fairly good, though, since I've messed up efficiency in some of my past experiments that made http_cats very laggy, and that didn't happen here.

The only time this would not perform well is measuring, since measuring entails checking multiple sizes (and therefore invalidating the cache several times), but that's not the fault of these changes, it's just that we need a measure function for that.

I tested most examples and did not find a situation where the cache didn't invalidate properly.
Looks good to me.

PoignardAzur:
For the record, I'm fine with adding any amount of caching to Masonry (and the LayoutCache type could be fairly open-ended), as long as we do as little caching as possible in widget code.

PR #1224, opened 2025-07-23
Better document and handle disabled status

PoignardAzur:


waywardmonkeys (commented):
This is changing some assumptions in the code since we're now checking `is_disabled` a lot less ... so it seems like this needs a comment in the commit message / PR description that explains what's going on here. It seems like existing widget code / authors need to understand these changes.

PoignardAzur:
No tests yet.

And yes, one of the goals of this PR is to cut back on the defensive code in widget methods.

jaredoconnell:
Is there any good way for us to test the functionality? I am less comfortable with this code without tests.

masonry_core/src/doc/masonry_concepts.md
```diff
@@ -33,6 +33,11 @@ When one of these statuses changes, the `update` method is called on the widget.
 However, `update` can be called for reasons other than status changes.
 
 
+## Hovered
+
+A widget is "hovered" when a pointer is placed in the widget's hitbox, but not in the hitbox of its children.
```
DJMcNab, line 38: Hitbox might need to be defined here. It's not obvious from this prose that a non-interactive `label`, say, doesn't impact hovering of its parent.

masonry_core/src/doc/masonry_concepts.md
```diff
@@ -90,9 +95,21 @@ In that case, we still mark the widget as focused, but with a different color to
...
+Its pointer icon will be the default one.
+
+<!-- TODO: What about accessibility events? -->
+
+(Note: The above is not how browsers handle disabled form inputs, but it matches how most frameworks handle disable widgets.)
+
+While a widget is marked as disabled, all its children are automatically considered disabled.
 
-It is usually grayed out, and can't receive pointer or text events.
+Interactive widgets (e.g. buttons) should have a way to indicate when they are disabled, usually by showing a grayed-out appearance.
```
DJMcNab, line 113: Would it be worth specifying that the disabled status is controlled/set by the driver/app?

masonry_core/src/passes/update.rs
```diff
@@ -744,6 +744,13 @@ pub(crate) fn run_update_pointer_pass(root: &mut RenderRoot) {
     {
         next_hovered_widget = None;
     }
+    // Check if hovered widget is disabled
+    if let Some(id) = next_hovered_widget {
+        let state = root.widget_arena.get_state(id);
+        if state.is_disabled {
+            next_hovered_widget = None;
+        }
+    }
```
DJMcNab, line 754: Do we need to move the hover "up" the tree? (That is, should this be handled in `find_widget_under_pointer`?).

This needn't be answered in this PR, but a TODO comment to think about this would be helpful, if we don't make a decision one way or the other.

masonry_core/src/doc/masonry_concepts.md
PoignardAzur: It isn't necessarily. Widgets can disable themselves.

masonry_core/src/passes/update.rs
PoignardAzur: I think handling it this way is fine. If you put your pointer over a disabled widget, it would be counter-intuitive for the parent to behave as if you were hovering it directly. Overall, putting a pointer on a disabled widget should always be a "nothing happens" situation.

masonry_core/src/doc/masonry_concepts.md
PoignardAzur: I don't think there's a localized edit we can make to make the semantics clear, because these semantics are loosely defined in the first place.

masonry_core/src/doc/masonry_concepts.md
DJMcNab: Ah, I guess a widget can document that it is invalid to disable it with a mutate context. That seems potentially fragile, but can work.

PoignardAzur: Oh, you mean we should document that the app *can* disable a widget?

In that case, yeah, we could document that, but I don't feel a strong need to.

PR #1225, opened 2025-07-23
Tweak some accessibility code

PoignardAzur:
The `ctx.target() == ctx.widget_id()` checks were added back when Masonry still used the "send event to every widget in the tree, recurse inside widget code, use bloom filters to skip subtrees" approach.

It doesn't make sense anymore, except maybe as a guard against bubbling, but button and checkbox don't have to worry about that.


masonry/src/widgets/checkbox.rs
```diff
@@ -309,6 +302,16 @@ impl Widget for Checkbox {
         ChildrenIds::from_slice(&[self.label.id()])
     }
 
+    fn accepts_focus(&self) -> bool {
+        // Checkboxes can be tab-focused...
+        true
+    }
+
+    fn accepts_text_input(&self) -> bool {
```
waywardmonkeys: This seems like a different fix, so let's address it in a separate PR ... (as with moving the `accepts_focus` method)

PoignardAzur: Note that it doesn't change behavior, it just make it more explicit.

DJMcNab: Is that not even more of a reason to split this into its own PR? Since #1090, not doing it this way is the prevailing style. I guess we could change back, but I wouldn't approve that PR.

Fwiw, I'm happy enough moving `accepts_focus` in this PR, but adding `accepts_text_input` does need to be its own thing.

PR #1226, opened 2025-07-23
Remove focus chain and refactor tab-focus

PoignardAzur:
The new behavior has a few advantages:
- Avoids building and storing O(N) arrays of WidgetIds per-widget.
- When the user clicks somewhere and then presses `Tab`, the focused widget is more likely to be close to whatever the user clicked.
- Overall more reliable than the old focus code.

Incidentally fixes a bug where a textbox sometimes woulnd't be shown as focused.

### DIFF
--- a/masonry_core/src/core/contexts.rs
+++ b/masonry_core/src/core/contexts.rs
@@ -1100,7 +1100,7 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, UpdateCtx<'_>, {
     pub fn children_changed(&mut self) {
         trace!("children_changed");
         self.widget_state.children_changed = true;
-        self.widget_state.needs_update_focus_chain = true;
+        self.widget_state.needs_update_focusable = true;
         self.request_layout();
     }
 
@@ -1127,6 +1127,11 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, UpdateCtx<'_>, {
             .remove(id)
             .expect("remove_child: child not found");
         self.global_state.scenes.remove(&child.id());
+
+        // If we remove the focus anchor, its parent becomes the anchor.
+        if self.global_state.focus_anchor == Some(id) {
+            self.global_state.focus_anchor = Some(self.widget_state.id);
+        }
 
         self.children_changed();
     }
--- a/masonry_core/src/doc/masonry_concepts.md
+++ b/masonry_core/src/doc/masonry_concepts.md
@@ -112,11 +112,14 @@ A widget is considered "interactive" if it can still get text and/or pointer eve
 Stashed and disabled widget are non-interactive.
 
 
-## Focus chain
+## Focus anchor
 
-Masonry stores an array of all the ids of widgets which can be focused.
+When the user presses `Tab` or `Shift+Tab`, Masonry will look for the closest sibling of the focus anchor which accepts focus, and focus it.
 
-This array is called the focus chain, and is used by Masonry to move focus when the user presses `Tab` or `Shift+Tab`.
+The focus anchor is generally either the focused widget, or the most recently clicked widget.
+If the focus anchor is removed from the tree, its parent becomes the anchor.
+
+This behavior exists so that when the user clicks somewhere and then presses `Tab`, the focused widget is more likely to be close to whatever the user clicked.
 
 
 ## Properties / Props
--- a/masonry_core/src/passes/update.rs
+++ b/masonry_core/src/passes/update.rs
@@ -1,7 +1,7 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 
 use cursor_icon::CursorIcon;
 use tracing::{info_span, trace};
@@ -293,7 +293,7 @@ fn update_disabled_for_widget(
         };
         widget.update(&mut ctx, &mut props, &Update::DisabledChanged(disabled));
         state.is_disabled = disabled;
-        state.needs_update_focus_chain = true;
+        state.needs_update_focusable = true;
         state.request_accessibility = true;
         state.needs_accessibility = true;
     }
@@ -378,7 +378,7 @@ fn update_stashed_for_widget(
         };
         widget.update(&mut ctx, &mut props, &Update::StashedChanged(stashed));
         state.is_stashed = stashed;
-        state.needs_update_focus_chain = true;
+        state.needs_update_focusable = true;
 
         // Items may have been changed while they were stashed in ways that require a
         // relayout and a re-render.
@@ -424,21 +424,12 @@ pub(crate) fn run_update_stashed_pass(root: &mut RenderRoot) {
 
 // ----------------
 
-// --- MARK: FOCUS CHAIN
+// --- MARK: FOCUSABLE
 
-// TODO https://github.com/linebender/xilem/issues/376 - Some implicit invariants:
-// - A widget only receives BuildFocusChain if none of its parents are hidden.
-
-// TODO - This logic was copy-pasted from WidgetPod code and may need to be refactored.
-// It doesn't quite behave like other update passes (for instance, some code runs after
-// recurse_on_children), and some design decisions inherited from Druid should be reconsidered.
-/// See the [passes documentation](../doc/05_pass_system.md#update-passes).
-fn update_focus_chain_for_widget(
-    global_state: &mut RenderRootState,
+fn update_focusable_for_widget(
     widget: ArenaMut<'_, Box<dyn Widget>>,
     mut state: ArenaMut<'_, WidgetState>,
     properties: ArenaMut<'_, AnyMap>,
-    parent_focus_chain: &mut Vec<WidgetId>,
 ) {
     let _span = enter_span(state.reborrow());
     let id = state.item.id;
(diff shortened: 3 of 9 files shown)


masonry_core/src/passes/update.rs
```diff
@@ -451,56 +442,119 @@ fn update_focus_chain_for_widget(
...
-        &mut root.global_state,
-        root_widget,
-        root_state,
-        root_properties,
-        &mut dummy_focus_chain,
-    );
+    update_focusable_for_widget(root_widget, root_state, root_properties);
+}
+
+pub(crate) fn find_next_in_focus_chain(root: &mut RenderRoot, forward: bool) -> Option<WidgetId> {
```
waywardmonkeys: Should this still use the term "focus_chain"?

masonry_core/src/doc/masonry_concepts.md
```diff
@@ -112,11 +112,14 @@ A widget is considered "interactive" if it can still get text and/or pointer eve
...
 
-## Focus chain
+## Focus anchor
 
-Masonry stores an array of all the ids of widgets which can be focused.
+When the user presses `Tab` or `Shift+Tab`, Masonry will look for the closest sibling of the focus anchor which accepts focus, and focus it.
 
-This array is called the focus chain, and is used by Masonry to move focus when the user presses `Tab` or `Shift+Tab`.
+The focus anchor is generally either the focused widget, or the most recently clicked widget.
+If the focus anchor is removed from the tree, its parent becomes the anchor.
```
DJMcNab: I wonder if this should be its next sibling, rather than the parent. Do we know where in the sibling order the removed widget was?

masonry_core/src/passes/update.rs
DJMcNab: Should we make this publicly accessible outside of Masonry Core itself (i.e. outside of the automatic tab handling)
That would be useful for e.g. textboxes, which sometimes want to customise their tab behaviour (say by making <kbd>ctrl</kbd> + <kbd>tab</kbd> move focus), or button groups, which want to make pressing tab in them go to the next widget outside of the group, and use arrow keys to control focus inside themselves.

That needn't block this PR

masonry_core/src/passes/update.rs
```diff
@@ -451,56 +442,119 @@ fn update_focus_chain_for_widget(
...
+}
+
+fn find_next_focusable_sibling(
+    root: &mut RenderRoot,
+    anchor: WidgetId,
+    forward: bool,
+) -> Option<WidgetId> {
+    let parent_id = root.widget_arena.parent_of(anchor)?;
+    let parent_widget = root.widget_arena.get_widget(parent_id);
+    let siblings = parent_widget.item.children_ids();
```
DJMcNab: We could technically short-circuit here is the parent widget doesn't have focusable descendants. However, I don't think that's worthwhile, since the current focus anchor will be "here" anyway.

masonry_core/src/passes/update.rs
```diff
@@ -451,56 +442,119 @@ fn update_focus_chain_for_widget(
...
+
+    let anchor_idx = siblings.iter().position(|id| *id == anchor).unwrap();
```
DJMcNab: Possibly we should say at least the name of the widget in this panic message (this unwrap is based on the assumption that `children_ids` is correctly implemented). OTOH, this isn't a blocking concern here.

masonry_core/src/passes/update.rs
```diff
@@ -451,56 +442,119 @@ fn update_focus_chain_for_widget(
...
-        &mut dummy_focus_chain,
-    );
+    update_focusable_for_widget(root_widget, root_state, root_properties);
+}
+
+pub(crate) fn find_next_in_focus_chain(root: &mut RenderRoot, forward: bool) -> Option<WidgetId> {
+    let focus_anchor = root.global_state.focus_anchor;
+
+    // If the anchor isn't already focused, try to find its first focusable descendant.
+    if focus_anchor != root.global_state.focused_widget
```
DJMcNab: What about if we are focused on the focus anchor (and are going forwards)? For a tree like:
```text
A
  - B [focused]
    - C
  - D
E
```
Where all five nodes are focusable, and the focus is currently on B, I would expect the focus order to be:
`B -> C -> D -> E`.

(This also needs care when going backwards)

masonry_core/src/core/contexts.rs
```diff
@@ -1128,6 +1128,11 @@ impl_context_method!(MutateCtx<'_>, EventCtx<'_>, UpdateCtx<'_>, {
             .expect("remove_child: child not found");
         self.global_state.scenes.remove(&child.id());
 
+        // If we remove the focus anchor, its parent becomes the anchor.
+        if self.global_state.focus_anchor == Some(id) {
+            self.global_state.focus_anchor = Some(self.widget_state.id);
+        }
```
DJMcNab, line 1142: What happens if the focus anchor isn't the direct child, but a descendant of that child? Do we revert to the root?

DJMcNab (commented):
I agree that this makes sense, and is a lot more comprehensible. A few thoughts:
1) It's not completely clear to me why/how this fixed the textbox test. Would you mind just briefly explaining that?
2) I don't think that this handles nested focusable widgets correctly. If we want to declare this out of scope, that's also fine-ish, but maybe we should warn if we detect it.
3) I agree with Bruce that it would be nice to see some comprehensive tests here. Personally, I wouldn't block this PR on that, but I also think making a tree with a few widgets and testing forward/back tab shouldn't take too much code.

masonry_core/src/doc/masonry_concepts.md
PoignardAzur: It's possible. I think it's not trivial to implement, and there might be some performance concerns, but neither are strong blockers.

Still, I'd rather leave that as future work.

masonry_core/src/passes/update.rs
PoignardAzur: If `children_ids()` is implemented wrong, I'd expect the user to find out in the `paint` pass or the `update_pointer` pass anyway.

masonry_core/src/passes/update.rs
PoignardAzur: Either forward or backward, if B has focusable siblings, the next loop will find them. `get_id_path(some_node)` includes `some_node`.

(Man, re-reading this code, I'm actually pretty proud of it, because I'm super smashed and I can still understand most of it. That wasn't the case at all the first time I tried to rewrite the focus code.)

masonry_core/src/passes/update.rs
PoignardAzur: Good catch. It shouldn't.

masonry_core/src/core/contexts.rs
PoignardAzur: Hmm... I actually think it might crash. I need to include a check.

Ideally, `remove_child` would run on children recursively first, which would naturally propagate the value up, but that's a bit beyond the scope of this PR.

masonry_core/src/passes/update.rs
PoignardAzur: Oh, wait, I just saw that you wrote C as a child of B.

In that case... Yeah, I guess it will skip B's children. And that would also be a problem backwards.

Alright, let me try again...

masonry_core/src/doc/masonry_concepts.md
DJMcNab: Leaving it as future work is good. I'd suggest creating an issue for it.

masonry_core/src/passes/update.rs
PoignardAzur: Alright, the new new algorithm gets it right, in both directions.

PoignardAzur:
Moments like that is when having a test suite pays off, because I've already caught a few bugs I would never have thought to look for on my own.

masonry_core/src/passes/update.rs
```diff
@@ -377,69 +377,147 @@ pub(crate) fn run_update_stashed_pass(root: &mut RenderRoot) {
...
+
+    let root_node = root.widget_arena.get_node_mut(root.root.id());
+    update_focusable_for_widget(root_node);
+}
+
+pub(crate) fn find_next_focusable(root: &mut RenderRoot, forward: bool) -> Option<WidgetId> {
+    let mut focus_anchor_id = root.global_state.focus_anchor;
+
+    if let Some(id) = focus_anchor_id
+        && !root.is_still_interactive(id)
```
DJMcNab, line 423: Do we need it to still be interactive, or do we just want it to still be in the tree?

I don't see that if I have a widget selected which becomes disabled, my focus anchor shouldn't remain at it.

Having `stashed` be treated the same as removing for the purposes of focus makes some sense, but again, we noted that we would rather removal return to the nearest sibling, not just reset it to the root.

masonry/src/tests/update.rs
```diff
@@ -1,13 +1,250 @@
...
+    let button1_id = harness.get_widget_with_tag(button1_tag).id();
+    let button2_id = harness.get_widget_with_tag(button2_tag).id();
+    let button3_id = harness.get_widget_with_tag(button3_tag).id();
+
+    harness.focus_on(Some(button2_id));
+    harness.edit_widget_with_tag(button2_tag, |mut button| {
+        button.ctx.set_disabled(true);
+    });
+
+    // The focus anchor should have reset to the parent flex, so we select button1
```
DJMcNab: This test could also succeed if the focus anchor was being deleted, because button1 is both the first focusable child of the root, and the first child of the parent.
To really test this, we need to also make the parent be focusable.

This isn't necessarily a blocker

masonry_core/src/passes/update.rs
DJMcNab, line 1: I can parse out the algorithm here, and it's very neat! However, I don't feel that the docs in here explain it sufficiently well.
The wording is very clumsy, and I don't think the comments are in the right places.

This doesn't block this PR, however.

masonry_core/src/passes/update.rs
PoignardAzur: Look, we're splitting hairs at an atomic level here. Nobody has strong opinions about "what should get focused when I press tab after the focused item was disabled", I don't even think browsers handle it the same way.

The whole focus anchor thing is a convenience, so that focus stays "close" to what you're currently looking at. It doesn't need to be watertight, and if we want to nail its semantics down, we can do it in a later PR.

masonry/src/tests/update.rs
PoignardAzur: Honestly, I mostly added that part as a bonus. The core part of the test is the "skip over disabled/stashed/removed widget" part. If we want to tightly test focus anchor behavior, we should add another test in a later PR.

masonry_core/src/passes/update.rs
PoignardAzur: This is already higher-effort doc than 95% of the code in linebender repos. I'm not gonna say it's impossible to do better, but I really don't feel motivated to, for reasons you can guess.

masonry_core/src/passes/update.rs
DJMcNab: Sure. But the comments on the test which you're adding (https://github.com/linebender/xilem/pull/1226#discussion_r2266885483) say that the expected behaviour is different to what appears to be implemented in the code. I'm trying to parse out what the behaviour here is/is expected to be.

I agree that the exact semantics can be nailed down later, and so this can be deferred.

masonry_core/src/core/contexts.rs
PoignardAzur: To be clear: what happens in the case where a grandparent of the anchor widget is removed is you end up with an outdated focus anchor.

The `!root.is_still_interactive(id)` check I've added in `find_next_focusable` will early-exit in that case. The focus anchor isn't used anywhere else, so it's fine.

Overall we could deal with all that more elegantly, but that's also the case with existing focus code, so this PR is at least a step in the right direction.

masonry_core/src/passes/update.rs
PoignardAzur: Fair enough. I think I'll remove that part of the test, and document that the focus anchor semantics are a bit loose.

PoignardAzur: Done.

masonry_core/src/passes/update.rs
DJMcNab: Yeah, that's completely fine. I don't expect to see it changed any time soon, I'm just noting my experience from reviewing it.

My word choice in this comment was unkind, and I'm sorry about that. It is really great that there is an attempted explanation.

PR #1228, opened 2025-07-23
Add placeholder option to text input

PoignardAzur:
It's still pretty basic (no properties, no way to set the color or the style or the font, etc), but hey, now we have placeholders!

### DIFF
--- a/masonry/src/properties/placeholder_color.rs
+++ b/masonry/src/properties/placeholder_color.rs
@@ -1,0 +1,38 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::core::Property;
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color of a `TextInput`'s placeholder text.
+///
+/// [`TextInput`]: crate::widgets::TextInput
+#[expect(missing_docs, reason = "field names are self-descriptive")]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct PlaceholderColor {
+    pub color: AlphaColor<Srgb>,
+}
+
+impl Property for PlaceholderColor {
+    fn static_default() -> &'static Self {
+        static DEFAULT: PlaceholderColor = PlaceholderColor {
+            color: AlphaColor::BLACK,
+        };
+        &DEFAULT
+    }
+}
+
+impl PlaceholderColor {
+    /// Create new `PlaceholderColor` with given value.
+    pub const fn new(color: AlphaColor<Srgb>) -> Self {
+        Self { color }
+    }
+}
+
+// ---
+
+impl Default for PlaceholderColor {
+    fn default() -> Self {
+        *Self::static_default()
+    }
+}
--- a/masonry/src/properties/placeholder_color.rs
+++ b/masonry/src/properties/placeholder_color.rs
@@ -4,7 +4,7 @@
 use crate::core::Property;
 use crate::peniko::color::{AlphaColor, Srgb};
 
-/// The color of a `TextInput`'s placeholder text.
+/// The color of a [`TextInput`]'s placeholder text.
 ///
 /// [`TextInput`]: crate::widgets::TextInput
 #[expect(missing_docs, reason = "field names are self-descriptive")]
--- a/masonry/src/widgets/text_input.rs
+++ b/masonry/src/widgets/text_input.rs
@@ -4,27 +4,22 @@
 use std::any::TypeId;
 
 use accesskit::{Node, Role};
-use masonry_core::core::{ArcStr, Properties};
 use tracing::{Span, trace_span};
 use vello::Scene;
 use vello::kurbo::{Affine, Point, Rect, Size};
 use vello::peniko::Color;
 
 use crate::core::{
-    AccessCtx, BoxConstraints, ChildrenIds, LayoutCtx, NewWidget, NoAction, PaintCtx,
-    PropertiesMut, PropertiesRef, RegisterCtx, Update, UpdateCtx, Widget, WidgetId, WidgetMut,
-    WidgetPod,
+    AccessCtx, ArcStr, BoxConstraints, ChildrenIds, LayoutCtx, NewWidget, NoAction, PaintCtx,
+    Properties, PropertiesMut, PropertiesRef, RegisterCtx, Update, UpdateCtx, Widget, WidgetId,
+    WidgetMut, WidgetPod,
 };
 use crate::properties::{
-    Background, BorderColor, BorderWidth, BoxShadow, CornerRadius, DisabledBackground, Padding,
-    TextColor,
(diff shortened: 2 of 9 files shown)


masonry/src/widgets/text_input.rs
```diff
@@ -4,6 +4,7 @@
 use std::any::TypeId;
 
 use accesskit::{Node, Role};
+use masonry_core::core::{ArcStr, Properties};
```
waywardmonkeys: Use `crate::core::...`

masonry/src/widgets/text_input.rs
```diff
@@ -15,9 +16,14 @@ use crate::core::{
...
     Background, BorderColor, BorderWidth, BoxShadow, CornerRadius, DisabledBackground, Padding,
+    TextColor,
 };
 use crate::util::{fill, stroke};
-use crate::widgets::TextArea;
+use crate::widgets::{Label, TextArea};
+
+// TODO - Make placeholder style customizable
+// Semi-transparent white.
+const PLACEHOLDER_COLOR: TextColor = TextColor::new(Color::from_rgba8(0xFF, 0xFF, 0xFF, 0x8F));
```
waywardmonkeys: We run in a light mode UI ... so this won't work for us and should be style/property controlled from the start.

masonry/src/widgets/text_input.rs
```diff
@@ -33,6 +39,7 @@ use crate::widgets::TextArea;
 /// This is because `TextInput` largely serves as a wrapper around a [`TextArea`].
 pub struct TextInput {
     text: WidgetPod<TextArea<true>>,
+    placeholder: WidgetPod<Label>,
```
waywardmonkeys, line 38: Why is this a `Label` instead of a `TextArea<false>`?

DJMcNab: Not the PR author, but from context.
Whether we like it or not, the usual experience of placeholder text is that you can't select within in, or copy/paste out of it. (This really annoys me, I feel like there must be a keyboard shortcut for "realise the placeholder text" which I just haven't learned)

This also makes it so that a click would be correctly routed to the actual editable textarea, so the cursor shows in the right point, and entering text works as expected.

PoignardAzur:
The main problem with that solution is that if then there's no way to set a default placeholder color in `DefaultProperties`, and the placeholder will use the default label color unless you manually override it.

masonry/src/properties/placeholder_color.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::core::Property;
+use crate::peniko::color::{AlphaColor, Srgb};
+
+/// The color of a `TextInput`'s placeholder text.
+///
+/// [`TextInput`]: crate::widgets::TextInput
```
DJMcNab: I'm surprised that rustdoc doesn't complain at this...

masonry/src/widgets/text_input.rs
```diff
@@ -150,6 +194,21 @@ impl Widget for TextInput {
...
         ctx.place_child(&mut self.text, pos);
 
+        let text_is_empty = {
+            ctx.get_raw(&mut self.text)
+                .0
+                .text()
+                .chars()
+                .next()
+                .is_none()
+        };
```
DJMcNab: It seems strange to exclude the IME contents from the detection for whether this is empty.

Image

masonry_winit/examples/to_do_list.rs
```diff
@@ -63,7 +63,12 @@ pub fn make_widget_tree() -> NewWidget<impl Widget> {
         Flex::column()
             .with_child(NewWidget::new_with_props(
                 Flex::row()
-                    .with_flex_child(TextInput::new("").with_auto_id(), 1.0)
+                    .with_flex_child(
+                        TextInput::new("")
+                            .with_placeholder("ex: 'Do the dishes', 'File my taxes', ...")
```
DJMcNab: Todomvc uses "What needs to be done?", which I think is quite nice

masonry/src/widgets/text_input.rs
DJMcNab, line 1: Have you validated that a child label is the right way to express this in AccessKit? I might have expected us to use the [`placeholder`](https://docs.rs/accesskit/latest/accesskit/struct.Node.html#method.placeholder) method here. cc @mwcampbell

xilem/src/view/text_input.rs
```diff
@@ -72,6 +74,18 @@ impl<State, Action> TextInput<State, Action> {
         self
     }
 
+    /// Set the string which is shown when the input is empty.
+    pub fn placeholder(mut self, placeholder_text: impl Into<ArcStr>) -> Self {
+        self.placeholder = placeholder_text.into();
+        self
+    }
+
+    /// Set the color of the text shown when the input is empty.
```
DJMcNab: Why does this not have the documentation like `Spinner::color` in #1207?

DJMcNab (commented):
I guess as an interim solution this is passable, if you can get the accessibility right.
It's an important functionality to have.

I would note that your argument about Placeholder default color applies equally well to buttons (and checkboxes), but I don't see those getting their own properties. This is another reason that I'm dubious of the default properties system (alongside the non-reactivity of it).

masonry/src/widgets/text_input.rs
PoignardAzur: So how do I properly detect an empty string?

DJMcNab: Probably using [`PlainEditor::raw_text`](https://docs.rs/parley/0.5.0/parley/layout/editor/struct.PlainEditor.html#method.raw_text)

masonry/src/properties/placeholder_color.rs
DJMcNab: (To clarify, that's because this doc link isn't used)

DJMcNab (commented):
I'm still not a fan of the placeholder used in the examples (but that isn't blocking). I now cannot test with IME pre-edit (my system is Linux, so that piece of functionality no longer works), but the change to how empty content works looks correct.

The previous comment about accessibility still stands.

masonry/src/properties/placeholder_color.rs
PoignardAzur: Oh, yeah, I forgot to ask what you meant. Good catch.

xilem/src/view/text_input.rs
```diff
@@ -33,6 +33,7 @@ where
         on_enter: None,
         text_color: None,
         disabled_text_color: None,
+        placeholder: ArcStr::default(),
```
DJMcNab, line 35: Does `// TODO: Allow setting a placeholder` still apply?

PoignardAzur:
@mwcampbell Any thoughts about the implications of this PR for text input accessibility?

PR #1231, opened 2025-07-24
DRAFT: Implement reparenting and disjoint access in tree_arena_safe

PoignardAzur:
Add `ArenaMutList::item_mut_disjoint`.
Add `ArenaMutList::reparent`.
Make `ArenaMapMut` an explicit part of the API.

Overall, the goal of this is to make reparenting possible while in the middle of iterating the widget tree.
This is a key part of the groundwork for the lifecycle refactor as described in [#masonry > Plan for widget lifecycle](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Plan.20for.20widget.20lifecycle).

### DIFF
--- a/tree_arena/src/tree_arena_safe.rs
+++ b/tree_arena/src/tree_arena_safe.rs
@@ -326,7 +326,7 @@ impl<T> ArenaMut<'_, T> {
 
 impl<'arena, T> ArenaRefList<'arena, T> {
     /// Returns `true` if the list has an element with the given id.
-    pub fn has(self, _parents_map: ArenaMapRef<'arena>, id: impl Into<NodeId>) -> bool {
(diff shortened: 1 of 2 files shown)


PoignardAzur:
This is, obviously, not remotely close to compiling, since it changes the arena APIs without updating the Masonry code that uses them.

@KGrewal1 I'm uploading this because I'm especially interested in your take on this API so far. How feasible do you think it would be to make a sound version of this API on the unsafe side?

KGrewal1:
apologies for taking so long to look at this 
- from looking the main benefit is the ability to get disjoint mutable access to the sub trees
- I do not think this is unfeasible though in practice though may require rethinking the current API a bit - not sure how hard this would  - the invariants for this should be stored in `parents_map` which is also in the safe version of the arena so the check for being disjoint in that should mirror over to disjoint in unsafe too, making this change easier

Issue #979, opened 2025-05-18
When the minimize button was clicked, the program crashed.


xb284524239:
```
PS D:\Workspace\VSCode_Workspace\learn_gui\learn_xilem> cargo r
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.50s
     Running `target\debug\learn_xilem.exe`
---
Writing full logs to C:\Users\paul\AppData\Local\Temp\masonry-0001747562638618-dense.log
---
10:03:58.980Z DEBUG accessibility: Running ACCESSIBILITY pass with rebuild_all
Initialising in parallel using 6 threads
10:03:59.177Z DEBUG dispatch_text_event: Running ON_TEXT_EVENT pass with WindowFocusChange
10:03:59.177Z DEBUG dispatch_text_event: ON_TEXT_EVENT finished handled=false
10:04:01.216Z DEBUG dispatch_pointer_event: Running ON_POINTER_EVENT pass with Cancel
10:04:01.217Z DEBUG dispatch_pointer_event: ON_POINTER_EVENT finished handled=false
10:04:01.217Z DEBUG dispatch_text_event: Running ON_TEXT_EVENT pass with WindowFocusChange
10:04:01.219Z DEBUG dispatch_text_event: ON_TEXT_EVENT finished handled=false

thread 'main' panicked at C:\Users\paul\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wgpu-24.0.3\src\backend\wgpu_core.rs:1423:26:
wgpu error: Validation Error

Caused by:
  In Device::create_texture
    Dimension X is zero


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\learn_xilem.exe` (exit code: 101)
```

DJMcNab:
This isn't necessarily a hard problem, just something we've not prioritised. Previous discussion is in https://github.com/linebender/vello/issues/371 and https://github.com/linebender/xilem/pull/771. A previous attempt to fix this didn't work, and neither wgpu nor winit give especially great docs on how you're meant to handle this (at least the last time I checked).
A contribution to continue the work in #771 would be very welcome!

PR #1232, opened 2025-07-24
Implement tests to check that widgets can handle a size of zero

PoignardAzur:
The tests re-use the examples, since they already use a variety of widgets.

Tweak harness renderer code to not give wgpu a size of zero in those cases.

Fix some problems that came up.

### DIFF
--- a/masonry_winit/examples/custom_widget.rs
+++ b/masonry_winit/examples/custom_widget.rs
@@ -11,8 +11,8 @@
 use masonry::accesskit::{Node, Role};
 use masonry::core::{
     AccessCtx, AccessEvent, BoxConstraints, ChildrenIds, ErasedAction, EventCtx, LayoutCtx,
-    NewWidget, ObjectFit, PaintCtx, PointerEvent, PropertiesMut, PropertiesRef, RegisterCtx,
-    TextEvent, Widget, WidgetId,
+    NewWidget, PaintCtx, PointerEvent, PropertiesMut, PropertiesRef, RegisterCtx, TextEvent,
+    Widget, WidgetId,
 };
 use masonry::kurbo::{Affine, BezPath, Point, Rect, Size, Stroke};
 use masonry::palette;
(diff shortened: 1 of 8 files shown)


DJMcNab:
Does this impact on https://github.com/linebender/xilem/issues/979? I presume not, because the hack around 0x0 is only done for masonry_testing.

masonry_winit/examples/custom_widget.rs
```diff
@@ -156,8 +158,7 @@ impl Widget for CustomWidget {
         // Let's burn some CPU to make a (partially transparent) image buffer
         let image_data = make_image_data(256, 256);
         let image_data = Image::new(image_data.into(), ImageFormat::Rgba8, 256, 256);
-        let transform = ObjectFit::Fill.affine_to_fill(ctx.size(), size);
-        scene.draw_image(&image_data, transform);
+        scene.draw_image(&image_data, Affine::IDENTITY);
```
DJMcNab: It's not clear why this change was necessary, since you have a fix in `affine_to_fill`

DJMcNab (commented):
I think I'd slightly prefer moving the zero size parameters to be an associated const of the harness, and inline each of these tests into their corresponding example - so that we don't cause confusion about why the creation functions are pub.
From my side, that isn't blocking, though

masonry_winit/examples/custom_widget.rs
PoignardAzur: I have no idea why I ever wrote the code that way. That line dates from the time Masonry moved from Piet to Vello, and I'm pretty sure it's wrong. It just asks for a ratio from `ctx.size()` to `ctx.size()`, which is going to be IDENTITY.

DJMcNab: Hmm, I'd probably expect something like `ObjectFit::affine_to_fill(ObjectFit::Fill, ctx.size(), Size::new(256., 256.))`, maybe?

PoignardAzur:
> I think I'd slightly prefer moving the zero size parameters to be an associated const of the harness, and inline each of these tests into their corresponding example - so that we don't cause confusion about why the creation functions are pub.

I'd rather have these specific tests be outside of the examples, because I think the tests themselves would be more confusing to newcomers. Having a few items be pub is pretty "discreet", so to speak.

masonry_winit/examples/custom_widget.rs
PoignardAzur: Done.

PR #1233, opened 2025-07-24
Cache RenderContext and VelloRenderer in TestHarness

PoignardAzur:
Constructing those types take a lot of time.
Caching them between render calls lets us considerably speed up tests with multiple snapshots. (Still doesn't help with single-snapshot tests.)

Re-enable a multi-screenshot test that was taking too long.

### DIFF
--- a/masonry/screenshots/portal_button_list_scroll_to_item_13.png
+++ b/masonry/screenshots/portal_button_list_scroll_to_item_13.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:9660536caa97bfdaee97ff987e7cf1dfe19afd3a8ac57d49c0e2e70ff88e834e
+size 2573
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -130,6 +130,8 @@ pub struct TestHarness {
     signal_receiver: mpsc::Receiver<RenderRootSignal>,
     render_root: RenderRoot,
     access_tree: Option<accesskit_consumer::Tree>,
+    render_context: Option<RenderContext>,
+    vello_renderer: Option<vello::Renderer>,
     mouse_state: PointerState,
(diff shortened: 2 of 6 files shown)


masonry/screenshots/portal_button_list_scroll_to_item_13.png
DJMcNab, line 1: The text display/centering here is a little bit janky...

masonry_testing/src/harness.rs
```diff
@@ -384,25 +388,32 @@ impl TestHarness {
...
-            },
-        )
-        .expect("Got non-Send/Sync error from creating renderer");
+
+        let mut renderer = self.vello_renderer.take().unwrap_or_else(|| {
+            vello::Renderer::new(
+                device,
+                vello::RendererOptions {
+                    // TODO - Examine this value
+                    use_cpu: true,
```
DJMcNab, line 408: Hmm... This *probably* isn't what I'd expect us to use, but maybe for cross-platform compat it's worth it.

But I also don't think the pre-fine stages are likely to introduce cross-platform issues.

Definitely shouldn't change in this PR, though

masonry/screenshots/portal_button_list_scroll_to_item_13.png
PoignardAzur: Oh yeah, I bumped into that before. I can rewrite the test.

masonry_testing/src/harness.rs
PoignardAzur: Honestly, given the endgame is removing wgpu entirely, I don't think we should take the time to fix this.

masonry/screenshots/portal_button_list_scroll_to_item_13.png
PoignardAzur: On second thought, I'll leave the test as-is, because I'm not actually sure what's going wrong here.

PR #1235, opened 2025-07-24
Add type parameter to TestHarness

PoignardAzur:
This lets us skip downcast calls around edit_root_widget and root_widget (though we don't really use the latter).


Philipp-M (approved):
Makes sense. Though as soon as the arbitrary self types is stabilized and used here, we likely need to specify the type anyway I guess.

PR #1236, opened 2025-07-24
Make a bunch of property functions const

PoignardAzur:
For no particular reason.
Some of the property constructors cannot be made const because they use trait methods.


DJMcNab (approved):
"For no particular reason"...

PoignardAzur:
The longer version would be "Because it is a generally good thing to have functions be const and it lets us call them from constants, and since non-constness is infectious, it's better to make functions const proactively, otherwise you bump into cases where if you want to make a function const, you have to change a dozen of them at once".

PR #1237, opened 2025-07-24
Add rudimentary clipboard support

PoignardAzur:
This only supports copy-pasting strings, not rich text. And the way Ctrl+C is detected is somewhat dubious.

But it does effectively add clipboard support to Masonry apps.

### DIFF
--- a/masonry/src/widgets/text_area.rs
+++ b/masonry/src/widgets/text_area.rs
@@ -489,35 +489,23 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
                     Key::Character(x)
                         if EDITABLE && action_mod && x.as_str().eq_ignore_ascii_case("x") =>
                     {
-                        edited = true;
-                        // TODO: use clipboard_rs::{Clipboard, ClipboardContext};
-                        // if let Some(text) = self.editor.selected_text() {
-                        //     let cb = ClipboardContext::new().unwrap();
-                        //     cb.set_text(text.to_owned()).ok();
-                        //     self.editor.drive(fcx, lcx, |drv| drv.delete_selection());
-                        // }
-                        // edited = true;
+                        if let Some(text) = self.editor.selected_text()
+                            && !text.is_empty()
+                        {
+                            let text = text.to_string();
+                            self.editor.driver(fctx, lctx).delete_selection();
+                            edited = true;
+                            ctx.set_clipboard(text);
+                        }
                     }
                     // Copy
                     #[cfg(any(target_os = "windows", target_os = "macos", target_os = "linux"))]
                     Key::Character(c) if action_mod && c.as_str().eq_ignore_ascii_case("c") => {
-                        // TODO: use clipboard_rs::{Clipboard, ClipboardContext};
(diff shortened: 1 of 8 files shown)


masonry/src/widgets/text_area.rs
DJMcNab, line 1: If you think that it would be feasible to add a small test for both pieces of functionality, that would be good.

But I don't think that's worth blocking this PR on. The rest of this code is insufficiently tested (i.e. most of the tests are in Parley).

masonry/src/widgets/text_area.rs
```diff
@@ -714,6 +704,23 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+            TextEvent::ClipboardPaste(text) => {
+                if EDITABLE {
+                    let (fctx, lctx) = ctx.text_contexts();
+                    self.editor
+                        .driver(fctx, lctx)
+                        .insert_or_replace_selection(text);
+
+                    // TODO - Factor out with other branches
+                    let new_generation = self.editor.generation();
+                    if new_generation != self.rendered_generation {
```
DJMcNab, line 717: Yeah; this whole generation thing is something we need to clean up eventually. This is probably for the best for now.

masonry_core/src/core/contexts.rs
```diff
@@ -1239,6 +1239,15 @@ impl_context_method!(
             self.widget_state.ime_area = None;
         }
 
+        /// Set the contents of the platform clipboard.
+        ///
+        /// For example, text widgets should call this for "cut" and "copy" user interactions.
```
DJMcNab, line 1244:
```suggestion
        /// For example, text widgets should call this for "cut" and "copy" user interactions.
        /// Note that we currently don't support the "Primary" selection buffer on X11/Wayland.
```

masonry_core/src/core/events.rs
```diff
@@ -33,7 +33,9 @@ pub enum TextEvent {
     Ime(Ime),
     /// The window took or lost focus.
     WindowFocusChange(bool),
-    // TODO - Add ClipboardPaste variant.
+    // TODO - Handle rich text copy-pasting
+    /// The user pasted content in
```
DJMcNab:
```suggestion
    /// The user pasted content in.
```

masonry_winit/src/event_loop_runner.rs
```diff
@@ -591,7 +597,25 @@ impl MasonryState<'_> {
             if let Some(wet) = window.event_reducer.reduce(&event) {
                 match wet {
                     WindowEventTranslation::Keyboard(k) => {
-                        window.render_root.handle_text_event(TextEvent::Keyboard(k));
+                        // TODO - Detect in Masonry code instead
```
DJMcNab, line 600: Broadly, the implementation strategy I'd expect for this would be a `request clipboard` signal (perhaps with a u32 id), then returning the `ClipboardPaste` when it arrives.

Because getting the clipboard contents is inherently async anyway (e.g. there might be a permissions dialog; it also is possibly requesting the content from another app which has to do async ops to get it, it can be streamed, etc.).

But that is very much not for this PR.

PR #1239, opened 2025-07-25
Add pixel snapping to layout pass

PoignardAzur:
Remove coordinates rounding code from Align and Flex.

Remove rounding from BoxConstraints: instead, widgets do layout under the assumption that they're in a full f64 coordinate space, and rounding only happens at the end of layout.

Document pixel snapping.
Pixel-snap the baseline as well.
Remove `invalid_screenshot_2` test (which relied on placing a child widget with sub-pixel boundaries to produce a slightly incorrect image, which we can no longer do).
Update all screenshot tests.

See [#masonry > Aligning layout boxes to pixel boundaries](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Aligning.20layout.20boxes.20to.20pixel.20boundaries) for details.


PoignardAzur:
The different is mostly invisible, but you can see it in the zstack screenshot tests (the "Foreground" box is now pixel-aligned, so its border is no longer slightly blurry).

masonry/screenshots/align_right.png
DJMcNab, line 1: Do you know why this image is now 7 times smaller?

masonry_core/src/doc/masonry_concepts.md
```diff
@@ -165,6 +165,17 @@ Handling writing modes is in-scope for Masonry in the long term, but is deferred
...
+
+Masonry currently handles pixel snapping for its widgets.
+
+The basic idea is that when widgets are laid out, Masonry takes their reported sizes and positions, and rounds them to integer values, so that the drawn shapes line up with pixels.
+
+This is done "at the end" of the layout pass, so to speak, so that widgets can lay themselves out assuming a floating point coordinate space, and without worrying about rounding errors.
+
+The snapping is done in a way that preserves relations between widgets: if one widget ends precisely where another stops, Masonry will pick values so that their pixel-snapped layout rects have no gap or overlap.
+
+
```
DJMcNab: We should document that this is intentionally wrong in the presence of scaling because we've deferred doing it properly.

masonry/screenshots/align_right.png
PoignardAzur: What do you mean? Github tells me they're both 400x400.

PoignardAzur: Or do you mean file weight? Probably I didn't run oxipng on it the first time around.

masonry_core/src/doc/masonry_concepts.md
PoignardAzur: Done.

masonry/screenshots/align_right.png
DJMcNab: Yeah, the file size.

DJMcNab (approved):
I think for this to be complete, we also need to do something similar in compose, or at least in Portal/VirtualScroll. The situation becomes more complex when there's a scroll ongoing, but we don't have support for smooth scroll yet.

I'd like to get Kaur's thoughts here, so it might be worth leaving this for a few days.

PoignardAzur:
> I think for this to be complete, we also need to do something similar in compose, 

Oh crap, I forgot to implement that part, good catch.

My plan regarding compose is:

- Arbitrary transforms may not be pixel-snapped, and that's fine, they're mostly for animations and fancy graphics.
- `scroll_translation` *must* be pixel-snapped, probably directly in `set_child_scroll_translation`.

We should also document those in more details.

DJMcNab:
> scroll_translation must be pixel-snapped.

From my discussions in Utrecht, I think that this is the case - except when an animated smooth scroll is ongoing. In that case, it's probably better to allow "off-grid" drawing. That is, the scroll being smooth is more important than the exact pixel alignment, IIUC

PoignardAzur:
Alright, then I'll leave it as a TODO for now. Making a distinction between animated and non-animated scroll offset is way beyond the scope of this PR.

PoignardAzur:
All considered, I've decided to leave a `set_animated_child_scroll_translation` method for future use-cases, and do scroll-snapping by default for now.

DJMcNab (approved):
Looks good to me. I can confirm that the scrolling behaviour does lead to an improvement (obviously only at 1x scale)

jaredoconnell (commented):
Looking at the screenshots, this does look like an improvement for unscaled displays. It's a downgrade for scaled displays, but that's out of the scope of this PR. I hope that's addressed soon, though.

PR #1240, opened 2025-07-25
Factor Axis, CrossAxisAlignment and MainAxisAlignment out of `flex.rs`

PoignardAzur:
These types are going to be used in other widgets (notably the future Stack widget) and should not exclusive to Flex.

### DIFF
--- a/masonry_core/src/core/axis.rs
+++ b/masonry_core/src/core/axis.rs
@@ -1,0 +1,105 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use vello::kurbo::{Point, Rect, Size, Vec2};
+
+use crate::core::BoxConstraints;
+
+/// An axis in visual space.
+///
+/// Most often used by widgets to describe
+/// the direction in which they grow as their number of children increases.
+/// Has some methods for manipulating geometry with respect to the axis.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Axis {
+    /// The x axis
+    Horizontal,
+    /// The y axis
+    Vertical,
+}
+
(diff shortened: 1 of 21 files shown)


masonry_core/src/core/axis.rs
DJMcNab, line 1: It's not clear to me why this file is in Masonry Core whereas the other is in Masonry itself.

xilem/src/view/mod.rs
```diff
@@ -3,6 +3,9 @@
 
 //! Views for the widgets which are built-in to Masonry. These are the primitives your Xilem app's view tree will generally be constructed from.
 
+// TODO - Remove
```
DJMcNab: Why? If so, why is it here?

PoignardAzur: Because I didn't want to rewrite the imports of a dozen examples, and anyway the reexport story of Xilem is kind of a mess.

DJMcNab: Sure. But in this kind of comment, please at least leave where you expect it to be moved to.

That is, it's fine deferring stuff, but it does need to be clear what we're actually deferring...

PoignardAzur: True.

I'll go into more detail.

masonry_core/src/core/axis.rs
PoignardAzur: I don't have an extremely strong reason, except to me `Axis` feels like a more "essential" type that we'd want to keep in `masonry_core` (or even `kurbo`), whereas alignment types feel very widget-specific.

DJMcNab (approved):
LGTM. I'm not convinced by the TODO comment, but it pointing to an issue is great!

PR #1241, opened 2025-07-25
Make TestHarness constructors take NewWidget

PoignardAzur:
Have WidgetExt methods return NewWidget instead of WrapperWidget.
Rewrite testing tutorial a bit.
Rewrite some tests to avoid relying on WidgetExt.

This overall makes code more consistent: everything which creates a widget takes `NewWidget`.

### DIFF
--- a/masonry/src/widgets/tests/safety_rails.rs
+++ b/masonry/src/widgets/tests/safety_rails.rs
@@ -4,7 +4,7 @@
 use vello::kurbo::{Point, Size};
 
 use crate::core::{ChildrenIds, PointerButton, Update, Widget, WidgetId, WidgetPod};
-use crate::testing::{ModularWidget, TestHarness, TestWidgetExt};
+use crate::testing::{ModularWidget, TestHarness};
 use crate::theme::default_property_set;
 use crate::widgets::Flex;
 
--- a/masonry/src/widgets/tests/safety_rails.rs
+++ b/masonry/src/widgets/tests/safety_rails.rs
@@ -26,9 +26,11 @@ fn make_parent_widget<W: Widget>(child: W) -> ModularWidget<WidgetPod<W>> {
 #[should_panic(expected = "not visited in method on_text_event")]
 #[test]
 fn check_forget_to_recurse_text_event() {
-    let widget = make_parent_widget(Flex::row()).text_event_fn(|_child, _ctx, _, _event| {
-        // We forget to call child.on_text_event();
-    });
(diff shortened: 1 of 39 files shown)


masonry/src/widgets/tests/safety_rails.rs
DJMcNab, line 1: Would you consider removing these `cfg(false)` tests (or rewriting them if appropriate)?

To be clear, that's explicitly not for this PR.

masonry_testing/src/lib.rs
```diff
@@ -31,24 +31,21 @@ pub trait TestWidgetExt: Widget + Sized + 'static {
     }
 
     // TODO - Move to `Widget` trait.
```
DJMcNab: Now seems a reasonable time to either action (or remove as unplanned) this TODO.

masonry_testing/src/lib.rs
```diff
@@ -31,24 +31,21 @@ pub trait TestWidgetExt: Widget + Sized + 'static {
...
-        WrapperWidget::new_pod(child)
+    /// Wrap this widget in a [`NewWidget`] with the given [`Properties`].
+    fn with_props(self, props: Properties) -> NewWidget<Self> {
+        NewWidget::new_with(self, WidgetId::next(), WidgetOptions::default(), props)
     }
 }
 
 impl<W: Widget + 'static> TestWidgetExt for W {}
 
+// TODO - Remove this
```
DJMcNab: And replace with...?

masonry/src/widgets/tests/safety_rails.rs
PoignardAzur: Yeah, I intend to, but I keep putting it off. Writing basic behavior tests for UI is dredge work.

masonry_testing/src/lib.rs
PoignardAzur: I should probably make a full issue explaining my reasoning.

The short version is, I'd like to stop pre-allocating IDs, and instead use `WidgetTag` (#1217) or and other methods to get a specific widget's id in tests.

The main reason to do that is that I'd like to remove `WidgetId::next()`, and instead have the `TreeArena` produce whatever ids it wants when adding new children. The main reason to do *that* is that, if we do, then we can stop using a `HashMap` as the backing storage for `TreeArena` and use a slotmap instead.

masonry_testing/src/lib.rs
PoignardAzur: Fair enough, yes.

PR #1242, opened 2025-07-25
Add `impl From<...> for Properties` with various tuples.

PoignardAzur:


DJMcNab (approved):
Seems reasonable. It's a little bit of a shame that there's no smooth off ramp.

PR #1247, opened 2025-07-26
DRAFT: Do box shadows in post_paint

PoignardAzur:


masonry/src/properties/box_shadow.rs
```diff
@@ -78,7 +80,7 @@ impl BoxShadow {
         if property_type != TypeId::of::<Self>() {
             return;
         }
-        // TODO - request_paint_only?
+        // TODO - Call request_post_paint instead, which should be lighter?
         ctx.request_layout();
```
DJMcNab: Given how localised this change would be, it's not clear why it's not been done in this PR.

Perhaps this should be replaced with "TODO: determine whether `request_post_paint` would be correct here, as it would be more performant".

masonry/src/properties/box_shadow.rs
```diff
@@ -4,10 +4,12 @@
...
-use vello::kurbo::{Affine, Insets, Point, RoundedRect};
+use vello::kurbo::{Affine, BezPath, Insets, Point, RoundedRect, Shape as _, Size};
+use vello::peniko::Mix;
 use vello::peniko::color::{AlphaColor, Srgb};
 
 use crate::core::{Property, UpdateCtx};
+use crate::properties::CornerRadius;
 
 // TODO - This is a first implementation of box shadows. A full version would need
 // to address the following points:
```
DJMcNab, line 15: Is this full section still correct? The first item at least looks no longer true.

DJMcNab (approved):
Seems reasonable to me. Approving if and only if the TODO comments are clarified/updated.

PR #1248, opened 2025-07-26
Add WidgetArenaNode type to centralize all data in a single TreeArena

PoignardAzur:
This reduces code duplication (a lot of code no longer has to be copy-pasted three times).
This is especially important for the upcoming widget refactor, which will add more code for each arena.

### DIFF
--- a/masonry_core/src/core/widget_arena.rs
+++ b/masonry_core/src/core/widget_arena.rs
@@ -1,39 +1,30 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use tree_arena::{ArenaMut, ArenaMutList, ArenaRef, ArenaRefList, TreeArena};
+use tree_arena::{ArenaMut, ArenaRef, TreeArena};
 
 use crate::core::{Widget, WidgetId, WidgetState};
 use crate::util::AnyMap;
 
 pub(crate) struct WidgetArena {
-    pub(crate) widgets: TreeArena<Box<dyn Widget>>,
-    pub(crate) states: TreeArena<WidgetState>,
-    pub(crate) properties: TreeArena<AnyMap>,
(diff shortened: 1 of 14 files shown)


masonry_core/src/core/widget_arena.rs
```diff
@@ -1,39 +1,30 @@
...
+use tree_arena::{ArenaMut, ArenaRef, TreeArena};
 
 use crate::core::{Widget, WidgetId, WidgetState};
 use crate::util::AnyMap;
 
 pub(crate) struct WidgetArena {
-    pub(crate) widgets: TreeArena<Box<dyn Widget>>,
-    pub(crate) states: TreeArena<WidgetState>,
-    pub(crate) properties: TreeArena<AnyMap>,
+    pub(crate) nodes: TreeArena<WidgetArenaNode>,
```
DJMcNab, line 10: Thoughts on `Deref<Target=TreeArena<WidgetArenaNode>> + DerefMut` here? I see that it's useful to have the wrapping struct, for the util methods. But the fact that nearly every single use of this is `arena.nodes` is unfortunate.

DJMcNab (approved):
There's too much to reasonably review here, so I'm taking it on trust/CI that this works.

This direction is where we want to go. The new ergonomics aren't great, but since this doesn't impact user code, it isn't too important

masonry_core/src/core/widget_arena.rs
PoignardAzur: Overall I *really* don't like using `Deref` for anything which isn't explicitly a smart pointer. It adds another hurdle when reading the code, where people need to go "Why is this code calling `foo.do_stuff()` even though `do_stuff()` is a method of `Bar` and not `Foo`?" and eventually "Oh, it's because Bar implements Deref".

I also think verbosity isn't inherently bad as long as it follows predictable patterns.

PR #1250, opened 2025-07-26
Add Length type, and have widgets use that type instead of floats.

PoignardAzur:
See [#masonry > Plan for scale factor, dpi, and layout units](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Plan.20for.20scale.20factor.2C.20dpi.2C.20and.20layout.20units) for rationale.

This PR adds a Length newtype which guarantees its value is always finite and non-negative.
Using a newtype should also make it easier to handle unit conversion in the future.

I don't really like this PR as it stands (especially not the added verbosity), so I'm keeping it as draft now.
I'd like to merge #1245 and a few others before getting back to this one.

### DIFF
--- a/masonry/src/properties/types/length.rs
+++ b/masonry/src/properties/types/length.rs
@@ -1,0 +1,65 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use masonry_core::debug_panic;
+
+/// A value representing a width, height, or similar distance value.
+///
+/// Its value is always finite and non-negative.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Length {
+    value: f64,
+}
+
+impl Length {
+    /// A length of zero.
+    pub const ZERO: Self = Length { value: 0. };
+
+    /// Create a length, in logical pixels.
+    ///
+    /// # Panics
+    ///
+    /// If debug assertions are on, this will panic in these cases:
+    ///
+    /// - `value` is NaN.
+    /// - `value` is infinite.
+    /// - `value` is negative.
+    ///
+    /// If debug assertions are off, this will return zero instead of panicking.
+    pub fn px(value: f64) -> Self {
+        if value < 0. || !value.is_finite() {
+            // TODO - Make const once const formatting is allowed.
+            // (aka see you in 2030)
+            debug_panic!("Invalid length value '{value}'");
+            return Self::ZERO;
+        }
+        Self { value }
+    }
+
+    /// Create a length, in logical pixels.
+    ///
+    /// Can be called from const contexts.
+    ///
+    /// # Panics
+    ///
+    /// This will always panic if value is negative or non-finite.
+    pub const fn const_px(value: f64) -> Self {
+        if value < 0. || !value.is_finite() {
+            panic!("Invalid length value");
+        }
+        Self { value }
+    }
+
+    /// Get the value, in logical pixels.
+    pub const fn value(self) -> f64 {
+        self.value
+    }
+
+    #[doc(hidden)]
+    // TODO - Remove once pixel-snapping is merged.
+    pub fn ceil(self) -> Self {
+        Self {
+            value: self.value.ceil(),
+        }
+    }
+}
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -15,6 +15,7 @@ use crate::core::{
     AccessCtx, BoxConstraints, ChildrenIds, LayoutCtx, NewWidget, PaintCtx, PropertiesMut,
     PropertiesRef, RegisterCtx, UpdateCtx, Widget, WidgetId, WidgetMut, WidgetPod,
 };
+use crate::properties::types::Length;
 use crate::properties::{Background, BorderColor, BorderWidth, CornerRadius, Padding};
 use crate::util::{debug_panic, fill, include_screenshot, stroke};
 
@@ -29,7 +30,7 @@ pub struct Flex {
     main_alignment: MainAxisAlignment,
     fill_major_axis: bool,
     children: Vec<Child>,
-    gap: f64,
+    gap: Length,
 }
 
 /// Optional parameters for an item in a [`Flex`] container (row or column).
@@ -114,7 +115,7 @@ enum Child {
         alignment: Option<CrossAxisAlignment>,
         flex: f64,
     },
-    FixedSpacer(f64, f64),
+    FixedSpacer(Length, f64),
     FlexedSpacer(f64, f64),
 }
 
@@ -128,7 +129,7 @@ impl Flex {
             cross_alignment: CrossAxisAlignment::Center,
             main_alignment: MainAxisAlignment::Start,
             fill_major_axis: false,
-            gap: crate::theme::WIDGET_PADDING,
+            gap: Length::px(crate::theme::WIDGET_PADDING),
         }
     }
 
@@ -172,17 +173,9 @@ impl Flex {
     /// Equivalent to the css [gap] property.
     /// This gap is also present between spacers.
     ///
-    /// ## Panics
-    ///
-    /// If `gap` is not a non-negative finite value.
-    ///
     /// [gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap
     // TODO: Semantics - should this include fixed spacers?
-    pub fn with_gap(mut self, mut gap: f64) -> Self {
-        if !gap.is_finite() || gap < 0.0 {
-            debug_panic!("Invalid gap value '{gap}', expected a non-negative finite value.");
-            gap = 0.0;
-        }
+    pub fn with_gap(mut self, gap: Length) -> Self {
         self.gap = gap;
         self
     }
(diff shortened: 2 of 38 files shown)


masonry/src/properties/types/as_unit.rs
```diff
@@ -0,0 +1,92 @@
...
+// https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.int-as-float
+
+/// Utility trait for wrapping numbers in logical units.
+pub trait AsUnit {
+    /// Create a length, in logical pixels.
+    ///
+    /// # Panics
+    ///
+    /// Panics if value is negative, infinite.
+    fn px(self) -> Length;
```
DJMcNab, line 20: Given that Jetpack compose uses Px for physical pixels (see e.g. <https://developer.android.com/reference/kotlin/androidx/compose/ui/unit/Density#(androidx.compose.ui.unit.Dp).toPx()>), I'm not convinced that this is the name we should use.

masonry/src/properties/types/as_unit.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn px(self) -> Length {
+        Length::px(self as f64)
+    }
+}
+
+impl AsUnit for usize {
+    fn px(self) -> Length {
+        Length::px(self as f64)
+    }
+}
```
DJMcNab: I'd maybe leave this and isize out?

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// # Panics
+    ///
+    /// If debug assertions are on, this will panic in these cases:
+    ///
+    /// - `value` is NaN.
+    /// - `value` is infinite.
+    /// - `value` is negative.
+    ///
+    /// If debug assertions are off, this will return zero instead of panicking.
+    pub fn px(value: f64) -> Self {
```
DJMcNab, line 42: Should this `track_caller`?

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /// Create a length, in logical pixels.
+    ///
+    /// Can be called from const contexts.
+    ///
+    /// # Panics
+    ///
+    /// This will always panic if value is negative or non-finite.
+    pub const fn const_px(value: f64) -> Self {
```
DJMcNab, line 60: Maybe same here?

DJMcNab (commented):
This seems broadly reasonable.

To gut-check my current understanding; this doesn't change how layout/paint work, just wraps (and unwraps) values which are logical pixels?

I do think that the use of `px` has potential for confusion; I think that the naming conventions in jetpack compose are pretty good, of using `Dp` here.

masonry/src/properties/types/length.rs
PoignardAzur: Right, I forgot that one. By the way, is track_caller allowed on trait methods?

PoignardAzur:
I'm fine with any name. I picked `px` because it matches CSS syntax.

masonry/src/properties/types/length.rs
DJMcNab: I believe that it works on the individual implementations, rather than the trait method itself? I forget the details

PoignardAzur:
@DJMcNab

Thinking about it, I'm actually partial in favor of `.px()`, as it matches CSS, and it's more "obvious" than `.dp()` for users not familiar with Jetpack Compose.

But I can switch to `dp` if you're strongly opposed to `px`.

masonry/src/properties/types/length.rs
PoignardAzur: The reference actually sounds like it also works with traits: https://doc.rust-lang.org/reference/attributes/codegen.html#r-attributes.codegen.track_caller

> When applied to functions and methods in trait declarations, the attribute applies to all implementations.

masonry/src/widgets/tests/layout.rs
```diff
@@ -15,16 +16,17 @@ fn layout_simple() {
     const BOX_WIDTH: f64 = 50.;
```
DJMcNab, line 16: Why is this not a Length itself?

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// Its value is always finite and non-negative.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Length {
+    value: f64,
+}
+
+impl Length {
+    /// A length of zero.
+    pub const ZERO: Self = Self { value: 0. };
```
DJMcNab, line 28: *Maybe* we should use `const_px` here, so that we show users what to use. This is non-blocking

masonry/src/widgets/flex.rs
```diff
@@ -58,7 +59,7 @@ enum Child {
         alignment: Option<CrossAxisAlignment>,
         flex: f64,
     },
-    FixedSpacer(f64, f64),
+    FixedSpacer(Length, f64),
```
DJMcNab, line 62: The second f64 here is very inscrutable. It ~~needn't~~ edit: shouldn't be in this PR, but we really need to have some docs here.

masonry/src/widgets/sized_box.rs
```diff
@@ -59,14 +60,14 @@ impl SizedBox {
     }
 
     /// Set container's width.
-    pub fn width(mut self, width: f64) -> Self {
-        self.width = Some(width);
+    pub fn width(mut self, width: Length) -> Self {
+        self.width = Some(width.value());
```
DJMcNab: We should document on `width/height` that they don't use Length to support `expand`

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use masonry_core::debug_panic;
+
+/// A value representing a width, height, or similar distance value.
+///
+/// Its value is always finite and non-negative.
+#[derive(Clone, Copy, Debug, PartialEq)]
```
DJMcNab: Can we make a nicer debug impl (i.e. `"{value}px"` or `{value} logical units`

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use masonry_core::debug_panic;
+
+/// A value representing a width, height, or similar distance value.
+///
+/// Its value is always finite and non-negative.
```
DJMcNab, line 8: Is it intentional that we don't talk about logical pixels here? Is this leaving the door open for this also representing a number of physical pixels (for example).

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,67 @@
...
+    #[track_caller]
+    pub const fn const_px(value: f64) -> Self {
+        if value < 0. || !value.is_finite() {
+            panic!("Invalid length value");
+        }
+        Self { value }
+    }
+
+    /// Get the value, in logical pixels.
+    pub const fn value(self) -> f64 {
```
DJMcNab: Maybe `to_logical` or `as_logical` here? 

Otherwise, I think `.get()` is the idiomatic name for this kind of method, see e.g. https://doc.rust-lang.org/stable/std/num/struct.NonZero.html#method.get (if we think this is always just a wrapper for a meaningful number of logical pixels)

DJMcNab (commented):
Seems reasonable to me, beside my previously overruled complaint. There are a couple of little details I'd like to think about

masonry/src/properties/types/length.rs
PoignardAzur: Yes, this or percentage values or similar.

masonry/src/widgets/tests/layout.rs
PoignardAzur: It's used as a value in a few places in the test. Also, I plan to rewrite that test later on.

masonry/src/properties/types/length.rs
PoignardAzur: Long-term, I'd like to replace this method with one taking a scale factor input, and maybe other thing params (like the available space, to resolve percent values against).

In the short term, I guess `get()` is the better name.

masonry/src/properties/types/length.rs
```diff
@@ -0,0 +1,71 @@
...
+///
+/// Its value is always finite and non-negative.
+#[derive(Clone, Copy, PartialEq)]
+pub struct Length {
+    value: f64,
+}
+
+impl std::fmt::Debug for Length {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self)
```
DJMcNab:
```suggestion
        write!(f, "{}px", self)
```
I don't think there's any advantage to this terseness here.

masonry/src/widgets/split.rs
```diff
@@ -143,17 +137,25 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
 }
 
 // --- MARK: INTERNALS
+// TODO - Remove this block, and remove pixel-snapping code from this file.
+impl Length {
+    #[doc(hidden)]
+    pub fn ceil(self) -> Self {
+        Self::px(self.get().ceil())
+    }
+}
```
DJMcNab: Hmm, I'm not hugely happy with this - I'd probably prefer it to be written out manually, or as a `ceil_length`, say.

(I didn't particularly mind it in the length.rs file; obviously it's not great, but here is not good at all; non-local impls can be confusing).

masonry/src/properties/types/length.rs
PoignardAzur: It's not terse, the Debug impl re-uses the Display impl. This is to avoid code duplication.

masonry/src/widgets/split.rs
PoignardAzur: I'm putting it here because `split.rs` is the only file that uses that impl (that and the xilem `split.rs`).

I guess `ceil_length` works too.

masonry/src/properties/types/length.rs
DJMcNab: Oops... Thanks!

PR #1253, opened 2025-07-28
Dynamic child for masonry `Button`

nixonyh:
Switching from `Button` to `Button<W>` so that it can contain other type of widgets as well!

Edit: Now switching to `Button` with a `dyn Widget` as the child instead!

### DIFF
--- a/masonry/src/widgets/button.rs
+++ b/masonry/src/widgets/button.rs
@@ -28,7 +28,7 @@ use crate::util::{fill, include_screenshot, stroke};
 
 use super::Label;
 
-/// A button with a text label.
+/// A button with a child widget.
 ///
 /// Emits [`ButtonPress`] when pressed.
 ///
@@ -40,6 +40,9 @@ pub struct Button {
 // --- MARK: BUILDERS
 impl Button {
     /// Create a new button with a child widget.
+    ///
+    /// The child widget probably shouldn't be interactive,
+    /// to avoid behaviour which might be confusing to the user.
     ///
     /// # Examples
     ///


PoignardAzur:
Thanks for the PR!

Overall, while adding an arbitrary child to Button is very probably something we'll want to do, we're kind of moving in the opposite direction regarding generic widget parameters: we've removed most of them, and I'm planning to remove the last ones.

This isn't set in stone, and we might want to start a discussion about when generic widget parameters are appropriate, but in the meantime, I think it'd be simpler to have `Button` keep a `WidgetPod<dyn Widget>` instead for this PR.

nixon-voxell:
@PoignardAzur Hey I made the changes for the switch! May I know how I can fix the formatting error?
It says `README is not up to date`.

masonry/src/widgets/button.rs
DJMcNab, line 1: I think it would make sense to add a test with a non-label child here (e.g. emulating something like #866, but something simpler would also be fine)

masonry/src/widgets/button.rs
```diff
@@ -14,71 +14,50 @@ use vello::peniko::Color;
...
 };
 use crate::properties::{
     ActiveBackground, Background, BorderColor, BorderWidth, BoxShadow, CornerRadius,
     DisabledBackground, HoveredBorderColor, Padding,
 };
 use crate::theme;
 use crate::util::{fill, include_screenshot, stroke};
-use crate::widgets::Label;
 
 /// A button with a text label.
```
DJMcNab: I don't think that this description is still accurate.

masonry/src/widgets/button.rs
```diff
@@ -14,71 +14,50 @@ use vello::peniko::Color;
...
 #[doc = include_screenshot!("button_hello.png", "Button with text label.")]
 pub struct Button {
-    label: WidgetPod<Label>,
+    child: WidgetPod<dyn Widget>,
 }
 
 // --- MARK: BUILDERS
 impl Button {
-    /// Create a new button with a text label.
+    /// Create a new button with a child widget.
```
DJMcNab, line 42: It *might* be worth documenting that this child widget probably shouldn't be interactive, to avoid behaviour which might be confusing to the user.

xilem/src/view/button.rs
DJMcNab, line 1: I think this should also be changed to be generic. We can even still have the `button` function accept a string, and implicitly return a button with just a label, at least as a first pass to make this easier to land.

DJMcNab (commented):
This looks good, and is the direction we want to go here.

There are a few things I'd like to see before this lands, mostly around docs.
If you don't feel able to make the Xilem change, let us know and someone else can do it.

masonry/src/widgets/button.rs
```diff
@@ -14,71 +14,50 @@ use vello::peniko::Color;
...
     /// # Examples
     ///
     /// ```
-    /// use masonry::widgets::Button;
-    ///
-    /// let button = Button::new("Increment");
-    /// ```
-    pub fn new(text: impl Into<ArcStr>) -> Self {
-        Self::from_label(Label::new(text).with_auto_id())
-    }
```
PoignardAzur: I think it'd make sense to keep that constructor as the "easy default".

masonry/src/widgets/button.rs
```diff
@@ -14,71 +14,50 @@ use vello::peniko::Color;
...
         }
     }
 }
 
 // --- MARK: WIDGETMUT
 impl Button {
-    /// Set the text.
-    pub fn set_text(this: &mut WidgetMut<'_, Self>, new_text: impl Into<ArcStr>) {
-        Label::set_text(&mut Self::label_mut(this), new_text);
-    }
```
PoignardAzur, line 77: I wish there was a way to keep that method. I don't see an obvious one, though.

masonry/src/widgets/button.rs
```diff
@@ -419,10 +400,22 @@ mod tests {
...
+                GridParams::new(1, 0, 1, 1),
+            )
+            .with_child(
+                Button::new(Label::new("C").with_auto_id()).with_auto_id(),
+                GridParams::new(0, 1, 1, 1),
+            )
+            .with_child(
+                Button::new(Label::new("D").with_auto_id()).with_auto_id(),
+                GridParams::new(1, 1, 1, 1),
+            );
```
PoignardAzur: For instance, you could keep these tests fairly short by using a `Button::with_text` constructor.

masonry/src/widgets/button.rs
nixonyh: We can do that, but not sure how would one edit the properties of the `Label` tho?
Edit: Ok, they can use the `new` constructor..

masonry/src/widgets/button.rs
nixonyh: This might be a little hard as we will need to make assumptions of the child widget type. We could do `try_downcast` and return a `Result` or a `bool` but I feel like it's not ergonomic to use.

masonry/src/widgets/button.rs
PoignardAzur: I agree that it would make sense, but also, I don't think it's strongly necessary. The existing tests already cover a lot of potential failure points.

PoignardAzur (approved):
LGTM.

I can think of potential improvements, but none of them are blocking.

xilem/src/view/button.rs
nixonyh: I'm doing changes on this, I think I would prefer a separate PR for this instead as a follow up.

DJMcNab (approved):

PR #1256, opened 2025-07-29
Placehero: Use the environment system for avatars

DJMcNab:
This has no rendering changes.

A few notes:
1) The APIs added here are obviously pretty janky. The intention is not for these to be the final API design, but for this to guide future direction.
2) I probably should write at least a handful of tests for `OnActionWithContext` before landing this.
3) This actually just worked first time, which is pretty good!

This doesn't handle memoization at all, at the moment.

### DIFF
--- a/placehero/src/avatars.rs
+++ b/placehero/src/avatars.rs
@@ -26,7 +26,7 @@ struct AvatarResponse {
     image: Image,
 }
 
-#[derive(Default, Debug)]
+#[derive(Debug)]
 pub(crate) struct Avatars {
     icons: HashMap<String, Option<Image>>,
     requester: Option<UnboundedSender<AvatarRequest>>,
@@ -36,6 +36,46 @@ pub(crate) struct Avatars {
 impl Resource for Avatars {}
 
 impl Avatars {
+    /// Get the avatar view/component for the given url, as a 50x50 pixel box.
+    ///
+    /// This will fetch the image data from the URL, and cache it.
+    /// If the image hasn't yet loaded, will show a placeholder,
+    ///
+    ///  Requires that this View is within a [`Self::provide`] call.
+    // TODO: ArcStr for URL?
+    pub(crate) fn avatar<State: 'static>(url: String) -> impl WidgetView<State> + use<State> {
+        with_context(move |this: &mut Self, _: &mut State| {
+            let width = 50.;
+            let height = 50.;
+            if let Some(maybe_image) = this.icons.get(&url) {
+                if let Some(image_) = maybe_image {
+                    return Either::A(sized_box(image(image_)).width(width).height(height));
+                }
+            } else if let Some(requester) = this.requester.as_ref() {
+                drop(requester.send(AvatarRequest {
+                    avatar_url: url.to_string(),
+                }));
+                this.icons.insert(url.to_string(), None);
+            } else {
+                // If the worker hasn't started yet, we have to wait until it does to do so.
+            }
+            Either::B(
+                sized_box(spinner().color(css::BLACK))
+                    .background_gradient(
+                        Gradient::new_linear(
+                            // down-right
+                            const { -45_f64.to_radians() },
+                        )
+                        .with_stops([css::YELLOW, css::LIME]),
+                    )
+                    .width(width)
+                    .height(height)
+                    .padding(4.0),
+            )
+        })
+    }
+
+    /// Provide support for Mastodon Avatar display to the `child` view.
     pub(crate) fn provide<State, Action, Child>(
         child: Child,
     ) -> impl WidgetView<State, Action, Element = Child::Element>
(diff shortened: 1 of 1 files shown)


placehero/src/avatars.rs
```diff
@@ -23,79 +26,114 @@ struct AvatarResponse {
...
+            on_action_with_context(
+                |_: &mut State, this: &mut Self, response| {
+                    let ret = this.icons.insert(response.url, Some(response.image));
+                    if !matches!(ret, Some(None)) {
+                        tracing::warn!("Potentially loaded or tried to load same avatar twice.");
+                    }
+                },
+                env_worker(
+                    |proxy: MessageProxy<AvatarResponse>,
+                     mut rx: UnboundedReceiver<AvatarRequest>| async move {
```
Philipp-M, line 112: (Not directly related to this PR)

I think a query based system similar to bevys systems could really help readability here (like in the callbacks of the `env_worker` maybe?)

And I think we should maybe look into refining the `OptionalAction` in `xilem_web` to maybe even `xilem_core` to avoid this `map_message`?

placehero/src/avatars.rs
```diff
@@ -23,79 +26,114 @@ struct AvatarResponse {
...
-    pub(crate) fn avatar<State: 'static>(
-        &mut self,
-        url: &str,
-    ) -> impl WidgetView<State> + use<State> {
-        let width = 50.;
-        let height = 50.;
-        if let Some(maybe_image) = self.icons.get(url) {
-            if let Some(image_) = maybe_image {
-                return Either::A(sized_box(image(image_)).width(width).height(height));
+    // TODO: ArcStr for URL?
```
Philipp-M: (Apart from general documentation) This could probably have a doc comment (stole this from the mason example):

```suggestion
    /// Requires the `Avatars` resource to be [provided](provides).
    // TODO: ArcStr for URL?
```

placehero/src/lib.rs
```diff
@@ -118,20 +116,20 @@ impl Placehero {
...
+            OneOf::C(timeline(&mut self.thread_statuses))
         } else {
             OneOf::D(prose("No statuses yet loaded"))
         }
     }
 }
 
 fn app_logic(app_state: &mut Placehero) -> impl WidgetView<Placehero> + use<> {
-    fork(
+    Avatars::provide(fork(
```
Philipp-M, line 132: Interesting pattern

xilem/src/view/worker.rs
```diff
@@ -33,13 +35,47 @@ pub fn worker<F, H, M, S, V, State, Action, Fut>(
...
+pub fn env_worker<F, H, M, S, V, Res, State, Action, Fut>(
+    init_future: F,
+    store_sender: S,
+    on_response: H,
+) -> Worker<F, H, M, S, V, Res>
+where
+    F: Fn(MessageProxy<M>, UnboundedReceiver<V>) -> Fut,
+    Fut: Future<Output = ()> + Send + 'static,
+    S: Fn(&mut State, &mut Res, UnboundedSender<V>),
+    H: Fn(&mut State, M) -> Action + 'static,
```
Philipp-M, line 76: I'm wondering whether it would be more consistent to also have `&mut Res` in this callback?

xilem_core/src/environment.rs
```diff
@@ -20,16 +20,20 @@ pub struct Rebuild;
 // --- MARK: Environment
 
 #[derive(Debug)]
-struct EnvironmentItem {
-    value: Box<dyn AnyDebug>,
+#[expect(missing_docs, reason = "Public on an interim basis")]
```
Philipp-M, line 23: Really on interim basis (same for the other changes here)?
I think since xilem (and other crates based on xilem_core) will likely have more views accessing the environment in specialized ways, that do not make sense(?) in xilem_core, that we can just leave this `pub`?

Or do you plan to further abstract this?

Philipp-M (approved):
Looks good + some thoughts, and rather cosmetic suggestions.

It probably also makes sense to have `on_action_with_context` as builder method in xilem (via extension trait), IMO it would feel more natural that way.

placehero/src/avatars.rs
DJMcNab: I agree - something system-like is where this seems to be pretty inevitably heading. You end up with two kinds of callbacks, `components` and actions. And they're similar in a lot of ways, but different in others.

I don't think I've looked into OptionalAction; it's not surprising that this is something you've ran into.

xilem/src/view/worker.rs
DJMcNab: Probably, but I think that would make more sense as part of a generalised query system.

The reason I only added it to `store_sender` here was because the general pattern I want to use (for now) is `on_action_with_context` and returning an explicit action. But that doesn't work for `store_sender`, because it's not actually a true action.

xilem_core/src/environment.rs
DJMcNab: I don't know! I don't expect this to be the final API for the environment.

That is, this PR is an exploration of the capabilities of the environment system, rather than the final state.

So specifically, I expect to end up with methods for similar capabilities, as one possible end state.

placehero/src/lib.rs
DJMcNab: I think it makes sense - provide the resource, and its interactivity at the same time.

PR #1258, opened 2025-07-31
Update to git vello for wgpu 25

waywardmonkeys:
Also, update Kurbo at the same time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -684,10 +684,11 @@ dependencies = [
 
 [[package]]
 name = "codespan-reporting"
-version = "0.11.1"
+version = "0.12.0"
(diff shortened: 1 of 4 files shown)


DJMcNab (approved):
I guess this is fine? The motivation here isn't overly clear, but if it helps the transition to the (upcoming-ish) Vello 0.6, then it seems reasonable enough

waywardmonkeys:
Partly, I want to update my own codebase to wgpu 25 on the way to doing the update to wgpu 26. Partly, I want the extra bit of validation that all is well with wgpu 25 in Masonry / Xilem before we update to 26. And this also helps to make sure that we'll end up publishing and updating Vello when we go to do a new Masonry release.

PR #1260, opened 2025-08-01
Xilem: Make virtual scroll implementation much saner

DJMcNab:
This is possible due to changes such as https://github.com/linebender/xilem/pull/1079 and indeed partially motivated that change.

One possibility here is to remove the `state_changed` boolean entirely. However, I think that would actually still be useful in things like `lens` (or indeed, in `virtual_scroll` itself, to avoid rerunning callbacks in response to only a request rebuild). But this is an area of semantics which probably needs more thought.

I have tested this in `virtual_cats`, and don't see any regressions.

### DIFF
--- a/masonry/src/widgets/virtual_scroll.rs
+++ b/masonry/src/widgets/virtual_scroll.rs
@@ -285,6 +285,17 @@ impl<W: Widget + FromDynWidget + ?Sized> VirtualScroll<W> {
         self
     }
 
+    /// The number of currently active children in this widget.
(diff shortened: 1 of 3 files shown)


jaredoconnell (approved):
This appears to be a nice simplification. I'm not familiar enough with the design to fully evaluate all of the design decisions that you made, but I can confirm that the example worked fine when I ran it.
Hopefully all robustness concerns and TODOs can be addressed soon.

PR #1262, opened 2025-08-03
Add check that run_rewrite_passes doesn't saturate during tests

PoignardAzur:
Fix `RenderRoot::needs_rewrite_passes`.
Fix `WidgetState::needs_rewrite_passes`.
Use both methods more consistently.


DJMcNab (approved):
Seems reasonable enough

Issue #1264, opened 2025-08-03
Tracking issue for scale factor / dpi handling


PoignardAzur:
See [#masonry > Plan for scale factor, dpi, and layout units](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Plan.20for.20scale.20factor.2C.20dpi.2C.20and.20layout.20units).

One of the big TODOs we've left unaddressed for a while is scale factor. We do wire the value from winit to RenderRoot, and we do in theory use it, but it's a mess and I'm not sure we got any of it correct. (And from the few times people have reported changing their scale factor, I don't think we have.)

Right now our handling of scale factor is pretty inconsistent. Some coordinates are scaled, some aren't. `RenderRoot::render()` returns an un-scaled scene, and then `masonry_winit` has to apply a transform to the entire scene.

Instead, I would like to make a radical change: `masonry_core` should only ever deal in physical pixels / units, and logical units would only be "accepted" in newtypes defined in `masonry`, such as the `Length` type introduced in #1250.

At the time I write this issue, the places in Masonry that take logical coordinates are:

- Flex widget (gap)
- Grid widget (spacing)
- SizedBox widget (width, height)
- Split widget (min sizes, bar size, min bar area)
- BorderWidget property
- CheckmarkStrokeWidth property
- CornerRadius property
- Padding property
- One of the sub-sub-sub types of the Background property (specifically, `RadialGradientShape::FixedCircle`).

So the plan is:

- Add a `Width` type and use it in `Flex`, `Grid`, `SizedBox`, etc.
- Have Properties use either `Width` or other newtypes.
- Add a `scale_factor` param to `Width` getter and to all properties returning coordinates.
- Document that all layout APIs (mostly in FooCtx types) now take physical units.
- Remove the existing conversion code.
- Go through the entire codebase and document which unit is used in cases where it isn't clear.

PoignardAzur:
One problem with the plan above is that adding a `scale` param to all the property getters makes widget layout code more verbose.

We need to actually try it to see how verbose it would be in practice.

PoignardAzur:
Note: Now that #1565 has been merged and `Length` is part of masonry_core, the above description is a little out of date.

Ideally the whole roadmap should get a minor rewrite, but right now it looks like the direction we're taking is "Any 'bare float' value in masonry_core should refer to physical pixels, newtypes may refer to logical pixels".

@xStrom Does that sound right?

xStrom:
[...]

PR #1265, opened 2025-08-03
Add/tweak comments keeping track of needed changes for scale factor

PoignardAzur:
This updates all the relevant comments to point to the same issue, for easier future reference.


PoignardAzur:
We can always change #1264 to not point at #1250 if that's a problem. The main point here is to have an easy-to-grep note next to every scale-related TODO comment.

Issue #1255, opened 2025-07-28
Stop pre-allocating IDs for Widgets


PoignardAzur:
I'd like to eventually make it impossible to "reserve" a WidgetId, to assign an id to a NewWidget and, in fact, remove `WidgetId::next()` so it becomes impossible to create a new id in user code.

New ids would then be created by Masonry when a widget it added to the arena. Those new ids would be returned by `TreeArena::insert_child`, meaning the arena would be allowed to choose what ids get returned.

The main motivation behind this is that I'd *really* like to have TreeArena be backed by a slotmap instead of a hashmap for performance reasons, but a slotmap only works if it gets to pick its own keys/ids.

Therefore I want to progressively remove code that relies of pre-generating ids, and replace it with other methods.

For tests, that means some of the code would use `WidgetTag` (https://github.com/linebender/xilem/pull/1217) and other methods to get a specific widget's id when necessary.

PR #1267, opened 2025-08-03
Portal widget tag

PoignardAzur:
Modify Portal unit tests to use WidgetTag

This is a follow-up to https://github.com/linebender/xilem/pull/1217.
Also helps prepare https://github.com/linebender/xilem/issues/1255.


masonry_core/src/core/widget_tag.rs
```diff
@@ -28,3 +29,28 @@ impl<W: Widget> WidgetTag<W> {
         }
     }
 }
+
+// Some of the impls could be derived, but then the bounds would be too restrictive.
+
+#[allow(clippy::non_canonical_clone_impl, reason = "We don't need `W: Clone`")]
```
DJMcNab: This reason doesn't jive with the docs on this lint. https://rust-lang.github.io/rust-clippy/stable/index.html#/non_canonical_clone_impl

It seems to be complaining that the body isn't just `*self`. Is there a reason you didn't make that the body?

PoignardAzur: I copy-pasted it from the WidgetRef impl. I'm thinking maybe *that* one didn't use `*self` for lifetime reasons, and both allow annotations are wrong.

PR #1269, opened 2025-08-03
Introduce "focus fallback" mechanism

PoignardAzur:
This introduces an explicit fallback mechanism, for text events rather than the undocumented "only child of root" method we currently use.

### DIFF
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -14,6 +14,7 @@ use masonry_winit::app::{
 };
 use winit::window::WindowAttributes;
 
+use crate::any_view::DynWidget;
 use crate::core::{
     AnyViewState, DynMessage, MessageContext, MessageResult, ProxyError, RawProxy, SendMessage,
     View, ViewId, ViewPathTracker,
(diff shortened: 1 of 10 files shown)


PoignardAzur:
Is there any example that used the "root is focused" behavior?

I've tried to run the to_do_mvc example, and it doesn't crash, but that's not the same as actually checking it works.

Other than that, PR is ready to merge.

xilem/src/driver.rs
```diff
@@ -338,6 +339,11 @@ where
         let fonts = std::mem::take(&mut self.fonts);
 
         for root in state.roots() {
+            if let Some(root_widget) = root.get_root_widget().downcast::<DynWidget>() {
+                let fallback = root_widget.inner().inner_id();
+                root.set_focus_fallback(Some(fallback));
+            }
```
DJMcNab, line 345: I'm not a huge fan of this behaviour. Unless someone has a stated need to use this with Xilem, I'd probably just leave it as a currently unsupported feature at the Xilem level.

But this doesn't block this PR.

PoignardAzur: I don't remember who wanted this, but yes, the feature that this PR replaces was added by a user who specifically needed it for their use-case.

PR #1276, opened 2025-08-04
Rewrite some Flex code to make it more readable.

PoignardAzur:
This PR tries to de-spaghettify the `Flex::layout()` method.

Follow-up to #1271.

Rename `total_major` to `bc_major`.
Rename `remaining` to `remaining_major`.
Remove the distinction between `minor` and `minor_dim`.
Remove some redundant comments.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -640,7 +640,7 @@ impl Widget for Flex {
 
         const MIN_FLEX_SUM: f64 = 0.0001;
         let gap_count = self.children.len().saturating_sub(1);
-        let bc_length = self.direction.major(bc.max());
+        let bc_major = self.direction.major(bc.max());
 
         // ACCUMULATORS
         let mut minor = self.direction.minor(bc.min());
@@ -692,7 +692,7 @@ impl Widget for Flex {
             }
         }
 
-        let remaining_major = (bc_length - major_non_flex).max(0.0);
+        let remaining_major = (bc_major - major_non_flex).max(0.0);
         let px_per_flex = remaining_major / flex_sum;
 
         // MEASURE FLEX CHILDREN
@@ -788,12 +788,12 @@ impl Widget for Flex {
             }
         }
 
-        if flex_sum > MIN_FLEX_SUM && bc_length.is_infinite() {
+        if flex_sum > MIN_FLEX_SUM && bc_major.is_infinite() {
             tracing::warn!("A child of Flex is flex, but Flex is unbounded.");
         }
 
         let final_major = if flex_sum > MIN_FLEX_SUM || self.fill_major_axis {
-            bc_length.max(major_non_flex)
+            bc_major.max(major_non_flex)
         } else {
             major_non_flex
         };


masonry/src/widgets/flex.rs
```diff
@@ -637,30 +638,27 @@ impl Widget for Flex {
         // we loosen our constraints when passing to children.
         let loosened_bc = bc.loosen();
 
-        // minor-axis values for all children
+        const MIN_FLEX_SUM: f64 = 0.0001;
+        let gap_count = self.children.len().saturating_sub(1);
+        let bc_major = self.direction.major(bc.max());
```
PoignardAzur, line 643: Rename `total_major` to `bc_major`.

masonry/src/widgets/flex.rs
```diff
@@ -694,25 +692,14 @@ impl Widget for Flex {
             }
         }
 
-        let total_major = self.direction.major(bc.max());
-        let remaining = (total_major - major_non_flex).max(0.0);
+        let remaining_major = (bc_major - major_non_flex).max(0.0);
```
PoignardAzur, line 695: Rename `remaining` to `remaining_major`.

masonry/src/widgets/flex.rs
```diff
@@ -736,29 +723,19 @@ impl Widget for Flex {
...
             // means we don't have any extra, unless dictated by our constraints.
             (self.direction.major(bc.min()) - (major_non_flex + major_flex)).max(0.0)
         };
-
-        // the actual size needed to tightly fit the children on the minor axis.
-        // Unlike the 'minor' var, this ignores the incoming constraints.
-        let minor_dim = match self.direction {
-            Axis::Horizontal if any_use_baseline => max_below_baseline + max_above_baseline,
-            _ => minor,
-        };
```
PoignardAzur, line 753: Remove the distinction between `minor` and `minor_dim`.

This distinction was only deliberately used in the `CrossAxisAlignment::Baseline` block below, where I moved this logic. Doing so lets us remove a bunch of code, including the `any_use_baseline` variable.

As a side-effect, this means that `minor` is used instead of `minor_dim` for the non-baseline blocks below, which I'm 99% sure will remove some corner-case jank (see further comments).

masonry/src/widgets/flex.rs
```diff
@@ -768,31 +745,28 @@ impl Widget for Flex {
...
                             let child_baseline = ctx.child_baseline_offset(widget);
                             let child_above_baseline = child_size.height - child_baseline;
                             extra_height + (max_above_baseline - child_above_baseline)
                         }
                         CrossAxisAlignment::Fill => {
                             let fill_size: Size = self
                                 .direction
-                                .pack(self.direction.major(child_size), minor_dim)
+                                .pack(self.direction.major(child_size), minor)
                                 .into();
```
PoignardAzur, line 760: Use `minor` instead of `minor_dim` as the size the child must fill.

This is the correct behavior, and the previous one was a bug: it meant that a given child with `CrossAxisAlignment::Fill` could end up with different size depending on whether or not a sibling had baseline alignment.

masonry/src/widgets/flex.rs
```diff
@@ -768,31 +745,28 @@ impl Widget for Flex {
...
-                            if ctx.old_size() != fill_size {
```
PoignardAzur, line 785: That block was there for layout caching, but (1) we now do that at the Masonry level, (2) I'm pretty sure it was wrong anyway. (It compared the previous size of the parent to the current size of a child.)

masonry/src/widgets/flex.rs
```diff
@@ -768,31 +745,28 @@ impl Widget for Flex {
...
+                            let child_bc = BoxConstraints::tight(fill_size);
+                            // TODO: This is the second call of layout on the same child,
+                            // which can lead to exponential increase in layout calls
+                            // when used multiple times in the widget hierarchy.
+                            ctx.run_layout(widget, &child_bc);
                             0.0
                         }
                         _ => {
-                            let extra_minor = minor_dim - self.direction.minor(child_size);
+                            let extra_minor = minor - self.direction.minor(child_size);
```
PoignardAzur, line 769: Use `minor` instead of `minor_dim` as the size of the box to align into.

This is the correct behavior, for the same reason as above.

masonry/src/widgets/flex.rs
```diff
@@ -801,32 +775,30 @@ impl Widget for Flex {
...
         let final_major = if flex_sum > MIN_FLEX_SUM || self.fill_major_axis {
-            total_major.max(major_non_flex)
+            bc_major.max(major_non_flex)
         } else {
             major_non_flex
         };
 
-        // my_size may be larger than the given constraints.
-        // In which case, the Flex widget will either overflow its parent
-        // or be clipped (e.g. if its parent is a Portal).
```
PoignardAzur, line 828: This comment seems fairly redundant, given that the same applies to any widget.

We should have a whole documentation chapter about overflow, but in the meantime, this paragraph doesn't really give any useful info.

PoignardAzur (commented):
Some notes:

PoignardAzur:
I *will* make that code readable even if it kills me.

masonry/src/widgets/flex.rs
jaredoconnell: It was a good place for the comment. Hopefully we can get the documentation chapter done soon.

jaredoconnell (approved):
The code looks fine, and with the lack of tests broken I think this is good enough to merge.
I like the segmented documented sections of code.

PR #1277, opened 2025-08-04
Rename some variables in Flex code to make it more readable.

PoignardAzur:
I've often been tripped up by the "major/minor" names in the Flex code.
As far as I can tell, it's not a well-known naming scheme (e.g. Taffy doesn't use it, other GUI projects don't use it), and it's not intuitive to a newcomer.

This uses more explicit names, and documents their meaning.

Also renames self.direction to self.axis, mostly for conciseness.

Follow-up to https://github.com/linebender/xilem/pull/1276.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -25,7 +25,7 @@ use crate::util::{debug_panic, fill, include_screenshot, stroke};
 ///
 #[doc = include_screenshot!("flex_col_main_axis_spaceAround.png", "Flex column with multiple labels.")]
 pub struct Flex {
-    direction: Axis,
+    axis: Axis,
     cross_alignment: CrossAxisAlignment,
     main_alignment: MainAxisAlignment,
     fill_major_axis: bool,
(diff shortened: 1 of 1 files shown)


PoignardAzur:
Another thing I'd like to do is move `Axis` to kurbo, so we could get some better getters and setters than `Axis.major()` and `Axis.minor()`.

PoignardAzur:
(Done in https://github.com/linebender/kurbo/pull/476)

PoignardAzur:
I'm aware of the CSS precedent for `direction`, and that's part of why I didn't rename `set_direction`. I still think renaming the internal value to `axis` makes the code more readable.

jaredoconnell (commented):
I honestly prefer the old names except for the rename to axis.

I think the best terminology would be main axis and cross axis. And in terms of the variable names like main_axis_size, and cross_axis_size would probably make sense.

I checked the web documentation and they use main and cross axis naming: https://www.w3.org/TR/css-flexbox-1/#justify-content-property

PR #1278, opened 2025-08-04
Add a `Prop` view to compose properties of widgets

Philipp-M:
This is the first and minimal implementation to get a composable `Prop` view working (see mason example), this is very far from being done.
In it's current form, it should serve as a base for discussion for this controversial direction, and includes the first of (likely) two controversial changes, by adding `changed_properties` in masonry (apart from the obvious goal of composing properties with views instead of having all properties on the widget views).

The likely second controversial change (that takes a lot more effort) would be to transfer `HasProperty` to masonry (as marker trait) and implement it directly on the widgets instead of the *views* of the widgets. Motivation for this, is my experience of developing xilem_web, where we first had a similar case on the views, but having it on the element (as it is now) is a lot cleaner and more idiomatic (e.g. being compatible with xilem_core views).

That can obviously also be done in xilem, but I think there's value of having it in masonry instead.
I'd like to include this as well here in this PR, to avoid churning code (and cleanup of the xilem side of this PR).

See [#xilem > Refactor property system to use a composition property view](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Refactor.20property.20system.20to.20use.20a.20composition.20property.20view) for prior discussion

### DIFF
--- a/masonry_core/src/core/properties.rs
+++ b/masonry_core/src/core/properties.rs
@@ -6,7 +6,7 @@ use std::collections::HashMap;
 use std::default::Default;
 
 use crate::core::Widget;
-use crate::util::AnyMap;
+use crate::util::{AnyMap, AnySet};
 
 /// A marker trait that indicates that a type is intended to be used as a widget's property.
 ///
@@ -17,7 +17,7 @@ use crate::util::AnyMap;
 /// as a property.
 /// That information is deliberately not encoded in the type system.
 /// We might change that in a future version.
-pub trait Property: Default + Send + Sync + 'static {
+pub trait Property: Default + Clone + Send + Sync + 'static {
     /// A static reference to a default value.
     ///
     /// Should be the same as [`Default::default()`].
@@ -55,6 +55,7 @@ pub struct PropertiesRef<'a> {
 pub struct PropertiesMut<'a> {
     pub(crate) map: &'a mut AnyMap,
     pub(crate) default_map: &'a AnyMap,
+    pub(crate) changed: &'a mut AnySet,
 }
 
 // TODO - Better document local vs default properties.
@@ -196,6 +197,7 @@ impl PropertiesMut<'_> {
     ///
     /// [`WidgetMut::insert_prop`]: crate::core::WidgetMut::insert_prop
     pub fn insert<P: Property>(&mut self, value: P) -> Option<P> {
+        self.changed.insert(TypeId::of::<P>());
         self.map.insert(value)
     }
 
@@ -207,7 +209,17 @@ impl PropertiesMut<'_> {
     ///
     /// [`WidgetMut::remove_prop`]: crate::core::WidgetMut::remove_prop
     pub fn remove<P: Property>(&mut self) -> Option<P> {
+        self.changed.insert(TypeId::of::<P>());
         self.map.remove::<P>()
+    }
+
+    /// Whether the property `P` has been modified in this pass.
+    ///
+    /// If you're using a `WidgetMut`, call [`WidgetMut::prop_has_changed`] instead.
+    ///
+    /// [`WidgetMut::prop_has_changed`]: crate::core::WidgetMut::prop_has_changed
+    pub fn has_changed<P: Property>(&self) -> bool {
+        self.changed.contains(&TypeId::of::<P>())
     }
 
     /// Get a `PropertiesMut` for the same underlying properties with a shorter lifetime.
@@ -215,6 +227,7 @@ impl PropertiesMut<'_> {
         PropertiesMut {
             map: &mut *self.map,
             default_map: self.default_map,
+            changed: &mut *self.changed,
         }
     }
 }
--- a/masonry_core/src/core/widget_mut.rs
+++ b/masonry_core/src/core/widget_mut.rs
@@ -58,6 +58,11 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
         self.ctx.properties.contains::<T>()
     }
 
+    /// Whether the property `P` of this widget has been modified in this pass.
+    pub fn prop_has_changed<P: Property>(&self) -> bool {
+        self.ctx.properties.has_changed::<P>()
+    }
+
     /// Get value of property `T`.
     ///
     /// If the widget has an entry for `P`, returns that entry.
--- a/masonry_core/src/util.rs
+++ b/masonry_core/src/util.rs
@@ -38,6 +38,10 @@ pub use crate::debug_panic;
 // ---
 
 pub(crate) type AnyMap = anymap3::Map<dyn Any + Send + Sync>;
+pub(crate) type AnySet = std::collections::HashSet<
+    std::any::TypeId,
+    std::hash::BuildHasherDefault<anymap3::TypeIdHasher>,
+>;
 
 // --- MARK: PAINT HELPERS
 
--- a/xilem/src/view/prop.rs
+++ b/xilem/src/view/prop.rs
@@ -1,0 +1,101 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::cmp::PartialEq;
+use std::marker::PhantomData;
+
+use masonry::core::Property;
+use xilem_core::{MessageContext, Mut};
+
+use crate::core::{View, ViewMarker};
+use crate::style::{HasProperty, Style};
+use crate::{Pod, ViewCtx, WidgetView};
+
+/// The view for [`property`].
+pub struct Prop<P, V, State, Action> {
+    pub(crate) property: P,
+    pub(crate) child: V,
+    pub(crate) phantom: PhantomData<(State, Action)>,
+}
+
+// TODO: (likely) Remove
+impl<P, V: Style, State, Action> Style for Prop<P, V, State, Action> {
+    type Props = V::Props;
+
+    fn properties(&mut self) -> &mut Self::Props {
+        self.child.properties()
+    }
+}
+
+// TODO: (likely) Remove
+impl<P: Property, Pe: Property, V: HasProperty<Pe>, State, Action> HasProperty<Pe>
+    for Prop<P, V, State, Action>
+{
+    fn property(&mut self) -> &mut Option<Pe> {
+        self.child.property()
+    }
+}
+
+impl<P, V, State, Action> ViewMarker for Prop<P, V, State, Action> {}
+impl<P, Child, State, Action> View<State, Action, ViewCtx> for Prop<P, Child, State, Action>
+where
+    P: Property + PartialEq,
+    Child: WidgetView<State, Action>,
+    // TODO implement this on the element/widget
+    // Child::Widget: HasProperty<P>,
+    State: 'static,
+    Action: 'static,
+{
+    type Element = Pod<Child::Widget>;
+    type ViewState = Child::ViewState;
+
+    fn build(&self, ctx: &mut ViewCtx, app_state: &mut State) -> (Self::Element, Self::ViewState) {
+        let (mut child_pod, child_state) = self.child.build(ctx, app_state);
+        child_pod
+            .new_widget
+            .properties
+            .insert(self.property.clone());
+        (child_pod, child_state)
+    }
+
+    fn rebuild(
+        &self,
+        prev: &Self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut ViewCtx,
+        mut element: Mut<'_, Self::Element>,
+        app_state: &mut State,
+    ) {
+        self.child.rebuild(
+            &prev.child,
+            view_state,
+            ctx,
+            element.reborrow_mut(),
+            app_state,
+        );
+        // If a child view changed the property, we know we're out of date.
+        if self.property != prev.property || element.prop_has_changed::<P>() {
+            element.insert_prop(self.property.clone());
+        }
+    }
+
+    fn teardown(
+        &self,
+        view_state: &mut Self::ViewState,
+        ctx: &mut ViewCtx,
+        element: Mut<'_, Self::Element>,
+        app_state: &mut State,
+    ) {
+        self.child.teardown(view_state, ctx, element, app_state);
+    }
+
+    fn message(
+        &self,
+        view_state: &mut Self::ViewState,
+        message: &mut MessageContext,
+        element: Mut<'_, Self::Element>,
+        app_state: &mut State,
+    ) -> xilem_core::MessageResult<Action> {
+        self.child.message(view_state, message, element, app_state)
+    }
+}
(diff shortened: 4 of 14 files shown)


PoignardAzur:
> The likely second controversial change (that takes a lot more effort) would be to transfer HasProperty to masonry (as marker trait) and implement it directly on the widgets instead of the views of the widgets. Motivation for this, is my experience of developing xilem_web, where we first had a similar case on the views, but having it on the element (as it is now) is a lot cleaner and more idiomatic (e.g. being compatible with xilem_core views).

Feel free to open a zulip thread to discuss it.

There are very compelling reasons to add *some* kind of type-level checking to Masonry, because the current system is very easy to get wrong. But there are also compelling use-cases for allowing arbitrary properties for every widget, so we should probably take the time to hash that out.

masonry_core/src/core/properties.rs
```diff
@@ -17,7 +17,7 @@ use crate::util::AnyMap;
 /// as a property.
 /// That information is deliberately not encoded in the type system.
 /// We might change that in a future version.
-pub trait Property: Default + Send + Sync + 'static {
+pub trait Property: Default + Clone + Send + Sync + 'static {
```
DJMcNab: This is quite a thing to place in here...

I get the impulse, but I think the fact that Masonry/Masonry Core itself never uses this bound suggests that it shouldn't be here.

Maybe we should just document that your property types should implement `Clone` for maximum flexibility for reactive layers.

masonry_core/src/util.rs
```diff
@@ -38,6 +38,10 @@ pub use crate::debug_panic;
 // ---
 
 pub(crate) type AnyMap = anymap3::Map<dyn Any + Send + Sync>;
+pub(crate) type AnySet = std::collections::HashSet<
+    std::any::TypeId,
+    std::hash::BuildHasherDefault<anymap3::TypeIdHasher>,
+>;
```
DJMcNab: I'm not sure that `AnySet` is a brilliant name for this type - it feels like it implies the same name semantics as `AnyMap`. Maybe something like `TypeSet`

xilem/src/view/prop.rs
```diff
@@ -0,0 +1,101 @@
...
+            &prev.child,
+            view_state,
+            ctx,
+            element.reborrow_mut(),
+            app_state,
+        );
+        // If a child view changed the property, we know we're out of date.
+        if self.property != prev.property || element.prop_has_changed::<P>() {
+            element.insert_prop(self.property.clone());
+        }
```
DJMcNab, line 63: Interesting. So if a child tries to overwrite a property, we just force it? I *feel* like there should/could be some debug reporting in that case, but maybe it's fine not to.

(Fwiw, if properties start getting set using the environment system, then this mechanism would be reasonably correct, so maybe it's fine)

masonry_core/src/core/widget_mut.rs
```diff
@@ -58,6 +58,11 @@ impl<W: Widget + ?Sized> WidgetMut<'_, W> {
         self.ctx.properties.contains::<T>()
     }
 
+    /// Whether the property `P` of this widget has been modified in this pass.
```
DJMcNab, line 61:
```suggestion
    /// Whether the property `P` of this widget has been modified in this pass.
    ///
    /// This is useful when composing units which might mutate a property value.
    /// In these cases, the "strongest" of these will want to check if it has been
    /// changed by a different unit, to overwrite this change.
```

Or something to that effect

DJMcNab (commented):
I'm still a little bit mixed on having the changed properties semantics be hardcoded into Masonry. But it's not something which I think it makes sense to block, and it does potentially allow composing different reactivity layers on top of Masonry.

I think I agree that having the `HasProperty` trait in Masonry Core would make sense, even without limiting which properties are supported.

masonry_core/src/core/properties.rs
Philipp-M: Yeah you're right, it can also just be a bound on the view, like the `PartialEq` but on the other hand it enforces something that I think should be the case anyway, and makes sure that future types are compatible with e.g. `Prop` (and possibly other views or possibly in the future something in masonry etc.), we could also reasonably add `PartialEq` here. But I don't have a strong opinion either way.

masonry_core/src/util.rs
Philipp-M: Nope it is not, I also had the same thought, right after I have comitted this,`TypeSet` sounds better.

xilem/src/view/prop.rs
Philipp-M: I think it's probably fine without logging I think it's fairly standard behaviour, we could also encode this in the type, like `OverwriteProp`

Philipp-M:
> I think I agree that having the `HasProperty` trait in Masonry Core would make sense, even without limiting which properties are supported.

Yeah I don't think it has to restrict, though I think in xilem it makes sense to restrict the view `Prop` to `HasProperty<P>`

masonry_core/src/core/properties.rs
PoignardAzur: I think my reasoning for not making it `Clone` was that I wanted to leave the door open for non-clone resources (heavy buffers, file descriptors) to be used as properties.

I guess we could still use them with a resource system.

Philipp-M:
> But it's not something which I think it makes sense to block, and it does potentially allow composing different reactivity layers on top of Masonry.

I would lean towards leaving it as is, as I have a similar feeling (different reactivity layers), and I think the alternatives in Xilem are (quite a bit?) more convoluted code, just to not touch masonrys code here. But if you have ideas to keep it in within Xilem without introducing a lot of code I'm open for it.


I've now drafted a basic version of how I think this should work when it's fully implemented with `border_color_`  and the `Button` widget. So if there's no major objection, I would start in roughly a week with the more complete and cleaner implementation of this.

masonry_core/src/core/properties.rs
Philipp-M: I've removed it for now, it's a simple change, introducing it (and `PartialEq`?), but I think this is orthogonal, and can be and should better be discussed in its own PR.

masonry/src/widgets/button.rs
```diff
@@ -302,6 +303,16 @@ impl Widget for Button {
...
 
+impl HasProperty<DisabledBackground> for Button {}
+impl HasProperty<ActiveBackground> for Button {}
+impl HasProperty<Background> for Button {}
+impl HasProperty<HoveredBorderColor> for Button {}
+impl HasProperty<BorderColor> for Button {}
+impl HasProperty<BorderWidth> for Button {}
+impl HasProperty<CornerRadius> for Button {}
+impl HasProperty<Padding> for Button {}
+impl HasProperty<BoxShadow> for Button {}
```
Philipp-M: I've put these below everything, they could reasonably be also directly below the widget structs. I think that would make sense if we're dedicating ourselves more towards a Struct of Arrays (Components of ECS) architecture.

xilem/src/widget_view.rs
```diff
@@ -46,6 +46,20 @@ pub trait WidgetView<State, Action = ()>:
     {
         transformed(self).transform(by)
     }
+
+    // TODO: Remove?
+    /// This is a temporary test for composing properties, this won't (?) be the final API
```
Philipp-M: Maybe we should keep this, I started using this throughout in this PR to reduce boilerplate, it could also be the general way to do this.
For example: `button().prop(Background::Color(color))` instead of `button().background_color(color)`.
I'm not sure though (more to write but uniform way to set props vs more code within Xilem in separate traits(?) but documented directly with these methods, so better for completion).

Philipp-M: I've decided for now to keep this and properly document it.

Philipp-M: As described in [#xilem > Refactor property system to use a composition property view @ 💬](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Refactor.20property.20system.20to.20use.20a.20composition.20property.20view/near/535082352)
we can add or even replace this with similar views that allow more advanced usage.

masonry/src/widgets/button.rs
DJMcNab: I think that they should be at least above the `Widget` implementation.

masonry/src/widgets/variable_label.rs
```diff
@@ -125,6 +125,13 @@ pub struct VariableLabel {
     weight: AnimatedF32,
 }
 
+impl<P> HasProperty<P> for VariableLabel
+where
+    P: Property,
+    Label: HasProperty<P>,
+{
+}
+
```
DJMcNab: This needs another look. I think that `VariableLabel` currently has no properties

masonry_core/src/core/properties.rs
```diff
@@ -207,14 +212,25 @@ impl PropertiesMut<'_> {
...
 
+    /// Whether the property `P` has been modified in this pass.
+    ///
+    /// If you're using a `WidgetMut`, call [`WidgetMut::prop_has_changed`] instead.
+    ///
+    /// [`WidgetMut::prop_has_changed`]: crate::core::WidgetMut::prop_has_changed
+    pub fn has_changed<P: Property>(&self) -> bool {
+        self.changed.contains(&TypeId::of::<P>())
+    }
+
```
DJMcNab: Can we remove this entirely, and so make it *only* a property of `WidgetMut`? I don't think that this value makes any meaningful sense *inside* a `Widget`.

masonry_core/src/core/properties.rs
```diff
@@ -207,14 +212,25 @@ impl PropertiesMut<'_> {
     ///
     /// [`WidgetMut::remove_prop`]: crate::core::WidgetMut::remove_prop
     pub fn remove<P: Property>(&mut self) -> Option<P> {
+        self.changed.insert(TypeId::of::<P>());
```
DJMcNab: To be clear, my proposal would also remove this

xilem/src/view/button.rs
```diff
@@ -53,58 +52,57 @@ use crate::{MessageResult, Pod, ViewCtx, ViewId};
...
 /// ```
-pub fn button<State, Action>(
-    label: impl Into<Label>,
+pub fn button<State, Action, V: WidgetView<State, Action>>(
+    child: V,
     callback: impl Fn(&mut State) -> Action + Send + 'static,
 ) -> Button<
     impl for<'a> Fn(&'a mut State, Option<PointerButton>) -> MessageResult<Action> + Send + 'static,
+    V,
 > {
```
DJMcNab: I'm not hugely happy with this part of this change. Like, on its own it's not unreasonable, but mixing it in with this change just adds so much noise.

I'd have liked to keep the "simple" constructor for the 95% of cases which use just a string with no styling, and add a new `any_button` or w/e constructor with this signature. That would make this PR much less noisy.
We can then make moving all uses of `button` to `any_button` a separate PR.

xilem/src/view/button.rs
```diff
@@ -114,47 +112,27 @@ impl<F> Button<F> {
 
 const LABEL_VIEW_ID: ViewId = ViewId::new(0);
```
DJMcNab: I don't think `LABEL_VIEW_ID` is really still the right name

xilem/src/view/prop.rs
```diff
@@ -0,0 +1,81 @@
...
+use std::cmp::PartialEq;
+use std::marker::PhantomData;
+
+use masonry::core::Property;
+use xilem_core::{MessageContext, Mut};
+
+use crate::core::{View, ViewMarker};
+use crate::{Pod, ViewCtx, WidgetView};
+
+/// A view that adds a property `P` or overrides a previously defined property `P`.
```
DJMcNab: We maybe should point to the `Style` trait here? And we definitely should point to however you construct this (as a non-Xilem crate)

xilem/src/widget_view.rs
```diff
@@ -46,6 +46,36 @@ pub trait WidgetView<State, Action = ()>:
     {
         transformed(self).transform(by)
     }
+
+    /// Set a [Property] on this view, when the underlying widget [supports](HasProperty) it.
```
DJMcNab: Either Property should be `Property` or shouldn't be capitalised, I think. See #1187

DJMcNab (commented):
Some thoughts here. I'm not going to *block* on the button change, but I don't like it being inside this PR; it makes it harder to review for actually interesting changes/consequences.

Overall, this seems a much more idiomatic direction than the previously implemented one, and I look forward to getting this landed.

masonry/src/widgets/variable_label.rs
Philipp-M: Yeah I'm happy to remove this, as this also prevents something like (which I think should be a follow-up):

```rust
impl<T: Widget> HasProperty<Transform> for T {}
```
And I think something like that is more useful anyways.

masonry_core/src/core/properties.rs
Philipp-M: Ok, done, I've left the `changed_properties` in the widget arena though.

xilem/src/view/button.rs
Philipp-M: Yeah that opens up a more general discussion whether such widget views should be more general or optimized towards the happy-path. I don't think `button(label())` is too bad and quite descriptive and allows the same constructor for all kinds of child widgets (and less headaches with the edge-cases where it's a styled label or other content), and it would be closer to a HTML `<button/>`, but I see your point.

I've done as you said, and will likely open this as a new PR then for separate discussion.

xilem/src/view/button.rs
Philipp-M: Hah, I have had changed it locally already for some time but for some reason forgot to push it.

xilem/src/view/prop.rs
Philipp-M: As it is now?

xilem/src/widget_view.rs
Philipp-M: I used `Property`.

masonry_core/src/core/properties.rs
```diff
@@ -70,6 +70,9 @@ pub struct DefaultProperties {
     pub(crate) dummy_map: AnyMap,
 }
 
+/// A marker trait that should be implemented for all properties that a widget supports
```
DJMcNab:
```suggestion
/// A marker trait indicating that the widget this is implemented for supports the property `P`.
///
/// You should implement this for your widget types, with each property the widget reads.
/// This is not used directly by Masonry Core, but is instead provided for the convenience of external crates.
```

DJMcNab (approved):
I suspect we'll want something like:
```rust
trait GlobalProperty {}
impl<P: GlobalProperty, W: Widget> HasProperty<P> for W {}
```
That could also be used for e.g. `FlexGrow` or w/e.

(Although my intuition about whether that impl is legal is a bit hazy)

xilem/src/view/button.rs
DJMcNab: Yeah - to be clear, I don't at all mind `button(label())` - just that it being mixed in this PR makes for an awkward review experience.

PR #1279, opened 2025-08-04
Add Padding property to Label

PoignardAzur:
Move LABEL_X_PADDING to theme module.

### DIFF
--- a/masonry/src/theme.rs
+++ b/masonry/src/theme.rs
@@ -29,9 +29,8 @@ pub const ACCENT_COLOR: Color = Color::from_rgb8(0x3b, 0x7e, 0xe4);
 pub const TEXT_COLOR: Color = Color::from_rgb8(0xf0, 0xf0, 0xea);
 pub const DISABLED_TEXT_COLOR: Color = Color::from_rgb8(0xa0, 0xa0, 0x9a);
 
-/// Added padding between each horizontal edge of the widget
-/// and the text in logical pixels.
-const LABEL_X_PADDING: f64 = 2.0;
+/// Default horizontal padding for [`Label`], in logical pixels.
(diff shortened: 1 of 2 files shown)


masonry/src/theme.rs
```diff
@@ -29,6 +29,10 @@ pub const ACCENT_COLOR: Color = Color::from_rgb8(0x3b, 0x7e, 0xe4);
 pub const TEXT_COLOR: Color = Color::from_rgb8(0xf0, 0xf0, 0xea);
 pub const DISABLED_TEXT_COLOR: Color = Color::from_rgb8(0xa0, 0xa0, 0x9a);
 
+/// Added padding between each horizontal edge of the widget
+/// and the text in logical pixels.
+const LABEL_X_PADDING: f64 = 2.0;
```
DJMcNab: Why is this constant private where everything else in this module is public?

We should either make it `pub` or document in the code why it is different.

masonry/src/theme.rs
```diff
@@ -29,6 +29,10 @@ pub const ACCENT_COLOR: Color = Color::from_rgb8(0x3b, 0x7e, 0xe4);
 pub const TEXT_COLOR: Color = Color::from_rgb8(0xf0, 0xf0, 0xea);
 pub const DISABLED_TEXT_COLOR: Color = Color::from_rgb8(0xa0, 0xa0, 0x9a);
 
+/// Added padding between each horizontal edge of the widget
+/// and the text in logical pixels.
```
DJMcNab: We should update the docs here to refer explicitly to `Label` rather than just "the widget".

DJMcNab (approved):
Looks good, thanks! Approving if and only if both review comments are actioned as described.

PR #1282, opened 2025-08-05
Add unit test for compose pass

PoignardAzur:
Adds test which checks that compose pass is runs when requested and sets the correct values.

Follow-up to https://github.com/linebender/xilem/pull/711.

### DIFF
--- a/masonry_testing/src/recorder_widget.rs
+++ b/masonry_testing/src/recorder_widget.rs
@@ -139,7 +139,7 @@ impl<W: Widget> Recorder<W> {
         &self.child
     }
 
-    /// Get a mutable reference to the child that's being recorded.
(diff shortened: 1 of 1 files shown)


masonry_testing/src/recorder_widget.rs
```diff
@@ -133,6 +133,16 @@ impl<W: Widget> Recorder<W> {
     pub fn recording(&self) -> &Recording {
         &self.recording
     }
+
+    /// Get a reference to the child that's being recorded.
+    pub fn inner(&self) -> &W {
+        &self.child
+    }
+
+    /// Get a mutable reference to the child that's being recorded.
```
DJMcNab:
```suggestion
    /// Get an exclusive reference to the child that's being recorded.
```

PR #1283, opened 2025-08-05
Add tests for pointer event pass

PoignardAzur:
Add tests for pointer capture.
Add test for focus-anchoring behavior.

### DIFF
--- a/masonry/src/tests/event.rs
+++ b/masonry/src/tests/event.rs
@@ -7,24 +7,51 @@ use assert_matches::assert_matches;
 use masonry_core::core::{NewWidget, TextEvent, Widget, WidgetTag};
 use masonry_testing::{ModularWidget, TestHarness, TestWidgetExt};
 use ui_events::keyboard::{Key, NamedKey};
-use ui_events::pointer::{PointerButton, PointerEvent};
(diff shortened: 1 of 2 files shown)


masonry/src/tests/event.rs
```diff
@@ -3,6 +3,154 @@
...
+
+    let target_id = harness.get_widget_with_tag(target_tag).id();
+    let other_id = harness.get_widget_with_tag(other_tag).id();
+
+    harness.mouse_move_to(target_id);
+    harness.mouse_button_press(PointerButton::Primary);
+
+    assert_eq!(harness.pointer_capture_target_id(), Some(target_id));
+
+    // As long as 'target' is captured, 'other' doesn't get pointer events, event when the cursor is on it.
```
DJMcNab:
```suggestion
    // As long as 'target' is captured, 'other' doesn't get pointer events, even when the cursor is on it.
```

masonry/src/tests/event.rs
```diff
@@ -3,6 +3,154 @@
...
+    harness.mouse_move_to(other_id);
+    assert_matches!(harness.get_records_of(other_tag)[..], []);
+
+    // 'other' is considered hovered either.
```
DJMcNab: is or isn't?

DJMcNab (approved):

PR #1284, opened 2025-08-05
Add tests for text event pass

PoignardAzur:
Test basic focus.
Test tab-focus.
Test text fallback.

Add `TestHarness::press_tab_key()` helper method.

### DIFF
--- a/masonry/src/tests/event.rs
+++ b/masonry/src/tests/event.rs
@@ -9,7 +9,7 @@ use ui_events::pointer::{PointerButton, PointerEvent, PointerInfo, PointerType};
 use vello::kurbo::Size;
 
 use crate::theme::default_property_set;
-use crate::widgets::{Button, Flex, SizedBox};
+use crate::widgets::{Button, Flex, SizedBox, TextArea};
 
 // POINTER EVENTS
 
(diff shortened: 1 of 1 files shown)


masonry/src/tests/event.rs
```diff
@@ -153,4 +153,100 @@ fn click_anchors_focus() {
...
+
+    // We focus on the widget, now it gets text events.
+    harness.focus_on(Some(target_id));
+    harness.keyboard_type_chars("A");
+    let records = harness.get_records_of(target_tag);
+    assert!(records.iter().any(|r| matches!(r, Record::TextEvent(_))));
+}
+
+#[test]
+fn text_event_fallback() {
```
DJMcNab, line 180: It might be worth noting that this is interim behaviour/that a better fallback mechanism is desired.

Is there an issue for that?

masonry/src/tests/event.rs
```diff
@@ -153,4 +153,100 @@ fn click_anchors_focus() {
...
+    });
+    harness.flush_records_of(target_tag);
+
+    // We've added another child, now nobody gets text events when no widget is focused.
+    harness.keyboard_type_chars("A");
+    assert_matches!(harness.get_records_of(target_tag)[..], []);
+}
+
+#[test]
+fn tab_focus() {
```
DJMcNab, line 206: As noted in #1226, we probably should at least think about our behaviour with nested focusable widgets.

DJMcNab (approved):

masonry/src/tests/event.rs
PoignardAzur: Well there's the #1269 PR.

masonry/src/tests/event.rs
PoignardAzur: I'll add those tests later. Most of them will be in the `update.rs` file.

(Also, it's kind of awkward to write tests for behavior I'm planning to change anyway.)

PR #1287, opened 2025-08-05
Add tests for mutate pass.

PoignardAzur:
Fix bug where a callback would panic if it was called after the associated widget was removed.

### DIFF
--- a/masonry/src/tests/mutate.rs
+++ b/masonry/src/tests/mutate.rs
@@ -1,4 +1,59 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-// TODO - Add unit tests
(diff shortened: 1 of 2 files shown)


PoignardAzur:
Quick note if anyone's wondering, the documentation says this:

> **The mutate pass**
> 
> The **mutate** pass runs a list of callbacks with mutable access to the widget tree.
> These callbacks can be queued with the `mutate_later()` method of various context types.
> 
> "Mutable access" means that those callbacks are given a [`WidgetMut`] to the widget that requested them, something that is otherwise only accessible from the owner of the global [`RenderRoot`] object.
> 
> If a callback is scheduled to run on a widget which is deleted before the callback is run, that callback is silently dropped.

So panicking when a widget was deleted was indeed a bug compared to what we promised.

Issue #607, opened 2024-09-23
Implement `VerticalList` widget


PoignardAzur:
Currently, people who want to create a new container widget look at the `Flex` widget for inspiration. That implementation is fairly complex, and not a great introduction for newcomers.

We should implement a container widget with a little code as possible. This widget would display its children vertically with a fixed amount of free space between them. It would have no customization parameters aside from that spacing constant. Its implementation would serve as a guide for people implementing custom widgets.

There should be non-doc comments in other widget files pointing newcomers to the VerticalList implementation.

To make sure the implementation doesn't bit-rot, we should:

- Include unit tests and screenshot tests.
- Use VerticalList in at least one example (probably to_do_list).
- Have a vertical list element is Xilem as well.

DJMcNab:
What is the use-case for this widget outside of this demonstration?

That is, could this instead be a masonry example? (examples can have tests, if they have `test = true` set on them)

PoignardAzur:
I guess it could.

I think having it in the main widget folder would be better for discoverability, though.

Philipp-M:
Yeah my first thought was also, that this may just be an example for masonry and possibly xilem as well, like a custom widget with children.

> I think having it in the main widget folder would be better for discoverability, though.

I think regarding this, we should probably just start with proper (high-level) documentation.

jaredoconnell:
This was partially my intention with the grid widget. I kept it as simple as possible so that it can be used as an example widget for layouts. It's literally a quarter the size of the Flex widget, and I added more documentation.

If you want to add a VerticalList widget, I think VerticalStack may be a more appropriate name. It helps differentiate it from the VirtualList, clarifying that you need to put it in a portal for scrolling, as it just stacks things vertically.

PR #1292, opened 2025-08-06
Add Stack widget

PoignardAzur:
Add Stack view.
Include test screenshots.

Stack is a non-flexible container widget, for cases where a Flex is not needed.

### DIFF
--- a/masonry/src/widgets/stack.rs
+++ b/masonry/src/widgets/stack.rs
@@ -17,7 +17,7 @@ use crate::core::{
 use crate::properties::types::{CrossAxisAlignment, Length, MainAxisAlignment};
 use crate::properties::{Background, BorderColor, BorderWidth, CornerRadius, Padding};
 use crate::theme::DEFAULT_GAP;
-use crate::util::{debug_panic, fill, include_screenshot, stroke};
+use crate::util::{fill, include_screenshot, stroke};
 use crate::widgets::flex::get_spacing;
 
 /// A container with either horizontal or vertical layout.


masonry/src/widgets/stack.rs
```diff
@@ -0,0 +1,646 @@
...
+            cross_alignment: CrossAxisAlignment::Center,
+            main_alignment: MainAxisAlignment::Start,
+            gap: DEFAULT_GAP,
+        }
+    }
+
+    /// Create a new horizontal stack.
+    ///
+    /// The child widgets are laid out horizontally, from left to right.
+    ///
```
waywardmonkeys: This line isn't needed.

masonry/src/widgets/flex.rs
```diff
@@ -19,7 +19,7 @@ use crate::properties::{Background, BorderColor, BorderWidth, CornerRadius, Padd
 use crate::theme::DEFAULT_GAP;
 use crate::util::{debug_panic, fill, include_screenshot, stroke};
 
-/// A container with either horizontal or vertical layout.
+/// A flexible container with either horizontal or vertical layout.
```
waywardmonkeys, line 23: Might be worth calling out here that if you don't need features blah blah blah that you can use `Stack` instead.

masonry/src/widgets/stack.rs
```diff
@@ -0,0 +1,646 @@
...
+    PropertiesMut, PropertiesRef, RegisterCtx, UpdateCtx, Widget, WidgetId, WidgetMut, WidgetPod,
+};
+use crate::properties::types::{CrossAxisAlignment, MainAxisAlignment};
+use crate::properties::{Background, BorderColor, BorderWidth, CornerRadius, Padding};
+use crate::theme::DEFAULT_GAP;
+use crate::util::{debug_panic, fill, include_screenshot, stroke};
+use crate::widgets::flex::get_spacing;
+
+/// A container with either horizontal or vertical layout.
+///
```
waywardmonkeys, line 24: Having a note here that references `Flex` for more complicated scenarios (and which those are) would be useful.

masonry/src/widgets/stack.rs
```diff
@@ -0,0 +1,635 @@
...
+use crate::properties::types::{CrossAxisAlignment, Length, MainAxisAlignment};
+use crate::properties::{Background, BorderColor, BorderWidth, CornerRadius, Padding};
+use crate::theme::DEFAULT_GAP;
+use crate::util::{fill, include_screenshot, stroke};
+use crate::widgets::flex::get_spacing;
+
+/// A container with either horizontal or vertical layout.
+///
+/// All the children of this widget will be laid out with the same constraints,
+/// then placed one after the other depending on the set alignment.
```
DJMcNab, line 26: What are these same constraints? It's really helpful for a mental model if this is explained, even if we expect it to change later.

masonry/src/widgets/stack.rs
```diff
@@ -0,0 +1,635 @@
...
+
+        harness.edit_root_widget(|mut stack| {
+            Stack::set_cross_axis_alignment(&mut stack, CrossAxisAlignment::Baseline);
+        });
+        assert_render_snapshot!(harness, "stack_row_cross_axis_baseline");
+
+        harness.edit_root_widget(|mut stack| {
+            Stack::set_cross_axis_alignment(&mut stack, CrossAxisAlignment::Fill);
+        });
+        assert_render_snapshot!(harness, "stack_row_cross_axis_fill");
```
DJMcNab, line 470: I don't think that this test meaningfully shows a fill (because e.g. the label is always left aligned).

Maybe (at least for the vertical ones) giving the labels internally centre alignment would work? OTOH, if we ever got a safety rail for that case, we'd run into issues...

DJMcNab (commented):
Does this fix https://github.com/linebender/xilem/issues/607?

It's not clear to me why this is only used in a single example - do so many of our examples and other code make use of "advanced" flex features?
If so, it's not all that clear what the purpose of this type actually is.

(That isn't entirely clear to me anyway, to be honest).

Code looks good other than my comments

masonry/src/widgets/stack.rs
PoignardAzur: It's `parent_box_constraints.loosen()`, which is essentially "same maximum sized at the parent, no minimum size".

masonry/src/widgets/stack.rs
PoignardAzur: Maybe I should give them a border.

DJMcNab:
> So my ideal plan for this would be to track down flex uses in examples which don't set must_fill_main_axis or have flex children: those are the ones we can replace with stack.
> After that, I'd want to straight up remove the must_fill_main_axis parameter, which is non-standard, unintuitive, and makes the flex code more complex. That way, Flex and Stack would actually fill distinct niches.

How this will work is a little bit unclear to me still. Since the Stack widget supports distributing its spacing, it seems like it also sometimes fills the main axis? I'm probably misunderstanding something fundamental here.

masonry/src/widgets/stack.rs
PoignardAzur: See [#masonry > Child constraints of Flex and Stack widgets](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Child.20constraints.20of.20Flex.20and.20Stack.20widgets).

PoignardAzur:
> Since the Stack widget supports distributing its spacing, it seems like it also sometimes fills the main axis? I'm probably misunderstanding something fundamental here.

Right, so, the Stack widget distributes its spacing when it needs to fill the *minimum* constraint.

Ideally, I'd like to remove min constraints for Masonry entirely, and remove extra space code from Stack. 

Though we could also remove the MainAxisAlignment param, assume that alignment is always at the start, and remove most of that code.

> None of this discussion blocks this PR landing.

I know I've asked you to expedite the process before, but I'm actually okay with delaying this PR a little. It's not critical for anything, and those questions are worth exploring.

PR #1293, opened 2025-08-06
Add tests for update pass.

PoignardAzur:
Move some tests from `widgets/testing`.


DJMcNab (approved):

jaredoconnell (approved):

DJMcNab:
(To be clear for future readers, my approval there is based on it only being test code, rather than being based on a careful reading of the code)

PR #1295, opened 2025-08-06
Merge `TextColor` and `SpinnerColor` into `ContentColor`.

PoignardAzur:


masonry/src/doc/widget_properties.md
```diff
@@ -63,7 +63,7 @@ Some examples:
 - [`BorderColor`]
 - [`BorderWidth`]
 - [`Padding`]
-- [`TextColor`]
```
waywardmonkeys, line 66: We’re these previously in alphabetical ordering? Makes sense to reorder now.

masonry/src/properties/content_color.rs
```diff
@@ -6,46 +6,49 @@ use std::any::TypeId;
...
 use crate::peniko::color::{AlphaColor, Srgb};
 
-/// The color of a widget's text and text decorations.
+/// The color of a widget's content, often text and text decorations.
 ///
-/// **IMPORTANT:** This property is only defined for [`Label`] and [`TextArea`], *not*
+/// **IMPORTANT:** This property is defined for [`Label`] and [`TextArea`], *not*
 /// for widgets embedding them such as [`Button`], [`Checkbox`], [`TextInput`], [`Prose`], etc.
 ///
+/// This widget is also defined for [`Spinner`].
```
waywardmonkeys: This property…

masonry/src/doc/widget_properties.md
PoignardAzur: Nope, I just included a few as examples, it really doesn't matter.

waywardmonkeys (approved):
Not wild about `fn color` being the replacement for `fn text_color` rather than using `fn content_color`. Also wish you'd just be willing to keep things sorted.

PR #1296, opened 2025-08-07
Fix flex code to correctly distribute extra space in the presence of …

PoignardAzur:
…spacers


jaredoconnell (approved):
The logic seems sound to me. Clearly we're missing a test, though, since no test snapshots needed to be regenerated for these changes.

PR #1297, opened 2025-08-07
Add unit tests for event bubbling.

PoignardAzur:


DJMcNab (approved):
Seems reasonable. We really do need an issue tracking the "handled status", which nothing currently makes use of.

PR #1298, opened 2025-08-07
Add skip_layout_when_cached test

PoignardAzur:
This test checks that layout stays cached and isn't recomputed when box constraints stay the same.

### DIFF
--- a/masonry/src/tests/layout.rs
+++ b/masonry/src/tests/layout.rs
@@ -1,14 +1,15 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use assert_matches::assert_matches;
 use masonry_core::core::{NewWidget, WidgetTag};
-use masonry_testing::assert_debug_panics;
+use masonry_testing::{TestWidgetExt, assert_debug_panics};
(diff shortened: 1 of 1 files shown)


masonry/src/tests/layout.rs
```diff
@@ -120,7 +121,33 @@ fn unstash_then_run_layout() {
...
+
+    harness.flush_records_of(button_tag);
+    harness.edit_widget_with_tag(sibling_tag, |mut sized_box| {
+        SizedBox::set_width(&mut sized_box, 30.0);
+        SizedBox::set_height(&mut sized_box, 30.0);
+    });
+
+    // The button did not request layout and its input constraints are the same:
+    // Nothing should happen to it.
+    assert_matches!(harness.get_records_of(button_tag)[..], []);
```
DJMcNab, line 149: To confirm, the sized box (and the zstack?) should have done a relayout?

DJMcNab (approved):
Seems reasonable enough

masonry/src/tests/layout.rs
PoignardAzur: Yes.

PR #1301, opened 2025-08-07
Make sure focus fallback doesn't send text events to disabled widget

PoignardAzur:


masonry_core/src/passes/event.rs
```diff
@@ -257,7 +257,13 @@ pub(crate) fn run_on_text_event_pass(root: &mut RenderRoot, event: &TextEvent) -
         // (since the root widget cannot be swapped).
         let root_widget_children = root.get_root_widget().children();
         if root_widget_children.len() == 1 {
-            Some(root_widget_children[0].id())
+            let only_child = root_widget_children[0].id();
+            if root.is_still_interactive(only_child) {
+                Some(only_child)
+            }
+            else {
```
DJMcNab: This else should be on the same line as the `}` above, surely?

DJMcNab (approved):
Looks good apart from the wrong formatting

masonry_core/src/passes/event.rs
PoignardAzur: Uh. That's so weird. For some reason rustfmt refuses to format this file.

PoignardAzur: Well, if you're curious: https://github.com/rust-lang/rustfmt/issues/6627

DJMcNab: Hmm, is that not https://github.com/rust-lang/rustfmt/issues/3863? We've previously worked around it e.g. https://github.com/linebender/vello/pull/499 by reformatting into a format which rustfmt will fix.

PoignardAzur: Probably, now that you mention it.

Either way, #1269 removes this piece of code, so I'm not worried about it.

PR #1304, opened 2025-08-07
Rewrite SizedBox setter methods

PoignardAzur:
Have `SizedBox::set_width()` and `SizedBox::set_height()` take a `Length`
Add `SizedBox::set_raw_width()` and `SizedBox::set_raw_height()` for use in Xilem

Add `SizedBox::size()`
Add `SizedBox::set_size()`

Follow-up to #1250.

### DIFF
--- a/masonry/src/widgets/sized_box.rs
+++ b/masonry/src/widgets/sized_box.rs
@@ -74,6 +74,13 @@ impl SizedBox {
         self
     }
 
+    /// Set container's width and height.
+    pub fn size(mut self, width: Length, height: Length) -> Self {
+        self.width = Some(width.value());
+        self.height = Some(height.value());
+        self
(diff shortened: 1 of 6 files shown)


masonry/src/widgets/sized_box.rs
```diff
@@ -166,6 +180,18 @@ impl SizedBox {
         this.ctx.request_layout();
     }
 
+    /// Set the width directly. Intended for toolkits abstracting over `SizedBox`
```
DJMcNab:
```suggestion
    /// Set the width directly. Intended for toolkits abstracting over `SizedBox`.
    ///
    /// If `Some`, the value should be non-negative and not NaN.
```
(And the same for `set_raw_height`)

PR #1305, opened 2025-08-07
Add status flags test and fix problems it uncovered

PoignardAzur:
When a widget has pointer capture, keep hovering it if the cursor is over a child widget.
Remove some old focus chain code which set some focus flags incorrectly.
Check the right flags in code that tracks changes in has_xxx flags.

Follow-up to #1293.


jaredoconnell (approved):
Good to see that the tests are uncovering problems.

PR #1306, opened 2025-08-07
Move safety rail tests for pointer capture

PoignardAzur:


masonry/src/tests/event.rs
```diff
@@ -137,6 +137,41 @@ fn pointer_capture_suppresses_neighbors() {
     assert!(harness.get_widget_with_tag(other_tag).ctx().is_hovered());
 }
 
+#[test]
+fn try_capture_pointer_on_pointer_up() {
```
DJMcNab:
```suggestion
fn try_capture_pointer_on_pointer_move() {
```

PR #1310, opened 2025-08-08
Replace `WindowState` with `is_suspended` boolean

### LINKED ISSUE #1308: `Window` not added back on `resumed`
I have not tested this on a mobile phone yet, but just noticed this while messing around with `masonry_winit`.

In `event_loop_runner.rs`, the `handle_resumed` function seems to reset the `self.windows` values via `std::mem::take` but did not put it back in if the window was in the state `WindowState::Suspended` or `WindowState::Rendering`.
```rs
    pub fn handle_resumed(&mut self, event_loop: &ActiveEventLoop, app_driver: &mut dyn AppDriver) {
        if !self.initial_windows.is_empty() {
            for new_window in std::mem::take(&mut self.initial_windows) {
                self.create_new_window(event_loop, new_window);
            }
        } else {
            for mut window in std::mem::take(&mut self.windows).into_values() {
                match std::mem::replace(
                    &mut window.state,
                    // TODO: Is there a better default value which could be used?
                    WindowState::Uninitialized(WindowAttributes::default()),
                ) {
                    WindowState::Uninitialized(attributes) => {
                        self.create_window_inner(event_loop, attributes, window);
                    }
                    WindowState::Suspended {
                        handle,
                        accesskit_adapter,
                    } => {
                        window.state = WindowState::Rendering {
                            handle,
                            accesskit_adapter,
                        };
                    }
                    _ => {
                        // We have received a redundant resumed event. That's allowed by winit
                    }
                }
            }
        }

        self.handle_signals(event_loop, app_driver);
    }
```

The `WindowState::Uninitialized` part was ok because it will be added back from the `create_window_inner` function.

### LINKED ISSUE #1309: Replace `WindowState` with `AppState`
The following will mostly be referencing `event_loop_runner.rs` in `masonry_winit`.

> TL;DR: I propose to remove `WindowState` and have the contents inside stored within the `Window` itself. The `Resumed` and `Suspended` state will be stored in a new `enum AppState` and stored directly within `MasonryState`.

In the current setup, each `Window` has a `WindowState` which consists of `Uninitialized`, `Rendering`, and `Suspended`.

If we dive into the code, we notice that `Unintialized` state is almost never the case at all, because looking at the function `fn create_window_inner`, which is the only way to insert a `Window` into the `self.windows` hashmap:

```rs
fn create_window_inner(..) {
    ..
    window.state = WindowState::Rendering {
        handle,
        accesskit_adapter: adapter,
    };
    window
        .render_root
        .handle_window_event(WindowEvent::Rescale(scale_factor));

    tracing::debug!(window_id = window.id.trace(), handle=?handle_id, "creating window");
    self.window_id_to_handle_id.insert(window.id, handle_id);
    self.windows.insert(handle_id, window);
    ..
}
```
We quickly notice that the `state` is immediately set to `Rendering` right before the insertion.

If we look at where the `Rendering` and `Suspended` state is being changed, these 2 functions arise: `handle_resumed` & `handle_suspended`:
```rs
pub fn handle_resumed(&mut self, event_loop: &ActiveEventLoop, app_driver: &mut dyn AppDriver) {
    if !self.initial_windows.is_empty() {
        for (id, attrs, widget) in std::mem::take(&mut self.initial_windows) {
            self.create_window(event_loop, id, attrs, widget);
        }
    } else {
        for mut window in std::mem::take(&mut self.windows).into_values() {
            match std::mem::replace(
                &mut window.state,
                // TODO: Is there a better default value which could be used?
                WindowState::Uninitialized(WindowAttributes::default()),
            ) {
                WindowState::Uninitialized(attributes) => {
                    self.create_window_inner(event_loop, attributes, window);
                }
                WindowState::Suspended {
                    handle,
                    accesskit_adapter,
                } => {
                    window.state = WindowState::Rendering {
                        handle,
                        accesskit_adapter,
                    };
                }
                _ => {
                    // We have received a redundant resumed event. That's allowed by winit
                }
            }
        }
    }

    self.handle_signals(event_loop, app_driver);
}
```
```rs
pub fn handle_suspended(&mut self, _event_loop: &ActiveEventLoop) {
    for window in self.windows.values_mut() {
        match std::mem::replace(
            &mut window.state,
            // TODO: Is there a better default value which could be used?
            WindowState::Uninitialized(WindowAttributes::default()),
        ) {
            WindowState::Rendering {
                handle,
                accesskit_adapter,
            } => {
                window.state = WindowState::Suspended {
                    handle,
                    accesskit_adapter,
                };
            }
            _ => {
                // We have received a redundant resumed event. That's allowed by winit
            }
        }
    }
    self.surfaces.clear();
}
```

Firstly, there's an issue with `handle_resumed` where we lost the window if it has been initialized (refer to #1308 )
Secondly, according to the logic above, here's the lifecycle of a `Window`:

1. `resumed` will get called first, which creates all the `initial_windows` with the state as `WindowState::Rendering` via `create_window() -> create_window_inner()`.
2. If:
    - `suspended` is called: all windows are set to `WindowState::Suspended`. (And all surfaces are discarded? `self.surfaces.clear()`)
    - `resumed` is called redundantly: all windows are cleared from the hashmap via `std::mem::take` and it's set to `WindowState::Rendering` despite it.
3. If `resumed` is called after `suspended` (point 1 above): all windows are set to `WindowState::Rendering` because they were all in the `WindowState::Suspended`.

2 major issues arise here: the state of the windows seems to always be the same, and `WindowState::Uninitialized` value is not used at all.

I also find this flow quite hard to follow, which I think should be refactored up to some point?

Feel free to let me know your thoughts below! And if I'm wrong in anyway here!

nixon-voxell:
Resolves #1309 
Resolves #1308

### DIFF
--- a/masonry_winit/src/app_driver.rs
+++ b/masonry_winit/src/app_driver.rs
@@ -13,7 +13,6 @@ use winit::event_loop::ActiveEventLoop;
 use winit::window::{Window as WindowHandle, WindowAttributes};
 
 use crate::app::MasonryState;
-use crate::event_loop_runner::WindowState;
 
 /// A unique and persistent identifier for a window.
 ///
@@ -98,35 +97,23 @@ impl DriverCtx<'_, '_> {
     ///
     /// # Panics
     ///
-    /// Panics if the window cannot be found or the window is not in the rendering state.
+    /// Panics if the window cannot be found.
     pub fn window_handle(&self, window_id: WindowId) -> &WindowHandle {
         let window = self.state.window(window_id);
-        let WindowState::Rendering { handle: window, .. } = &window.state else {
-            panic!(
-                "window with id {window_id:?} is in {:?} state, expected Rendering state",
-                window.state
-            );
-        };
-        window
+        &window.handle
     }
 
     /// Access the [`WindowHandle`] and [`RenderRoot`] of the given window.
     ///
     /// # Panics
     ///
-    /// Panics if the window cannot be found or the window is not in the rendering state.
+    /// Panics if the window cannot be found.
     pub fn window_handle_and_render_root(
         &mut self,
         window_id: WindowId,
     ) -> (&WindowHandle, &mut RenderRoot) {
         let window = self.state.window_mut(window_id);
-        let WindowState::Rendering { handle, .. } = &window.state else {
-            panic!(
-                "window with id {window_id:?} is in {:?} state, expected Rendering state",
-                window.state
-            );
-        };
-        (handle, &mut window.render_root)
+        (&window.handle, &mut window.render_root)
     }
 
     /// Creates a new window.
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -21,7 +21,7 @@ use masonry_core::util::Instant;
 use masonry_core::vello::util::{RenderContext, RenderSurface};
 use masonry_core::vello::wgpu;
 use masonry_core::vello::{AaConfig, AaSupport, RenderParams, Renderer, RendererOptions, Scene};
-use tracing::{debug, error, info, info_span};
+use tracing::{debug, info, info_span};
 use ui_events_winit::{WindowEventReducer, WindowEventTranslation};
 use winit::application::ApplicationHandler;
 use winit::error::EventLoopError;
@@ -45,37 +45,24 @@ impl From<accesskit_winit::Event> for MasonryUserEvent {
     }
 }
 
+/// The state of the current app.
 #[expect(unnameable_types, reason = "TODO")]
-#[allow(
-    clippy::large_enum_variant,
-    reason = "we don't have that many instances of it"
-)]
-pub enum WindowState {
-    Uninitialized(WindowAttributes),
-    Rendering {
-        handle: Arc<WindowHandle>,
-        accesskit_adapter: Adapter,
-    },
-    Suspended {
-        handle: Arc<WindowHandle>,
-        accesskit_adapter: Adapter,
-    },
-}
-
-impl Debug for WindowState {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            Self::Uninitialized(attrs) => f.debug_tuple("Uninitialized").field(attrs).finish(),
-            Self::Rendering { .. } => f.debug_struct("Rendering").finish_non_exhaustive(),
-            Self::Suspended { .. } => f.debug_struct("Suspended").finish_non_exhaustive(),
-        }
-    }
+#[derive(Default, Debug)]
+pub enum AppState {
+    /// The app is currently in suspended state,
+    /// usually set during [`ApplicationHandler::suspended()`].
+    #[default]
+    Suspended,
+    /// The app is currently in resumed state,
+    /// usually set during [`ApplicationHandler::resumed()`].
+    Resumed,
 }
 
 /// Per-Window state
 pub(crate) struct Window {
     id: WindowId,
-    pub(crate) state: WindowState,
+    pub(crate) handle: Arc<WindowHandle>,
+    pub(crate) accesskit_adapter: Adapter,
     event_reducer: WindowEventReducer,
     pub(crate) render_root: RenderRoot,
 }
@@ -83,8 +70,9 @@ pub(crate) struct Window {
 impl Window {
     pub(crate) fn new(
         window_id: WindowId,
+        handle: Arc<WindowHandle>,
+        accesskit_adapter: Adapter,
         root_widget: NewWidget<dyn Widget>,
-        attributes: WindowAttributes,
         signal_sender: Sender<(WindowId, RenderRootSignal)>,
         default_properties: Arc<DefaultProperties>,
     ) -> Self {
@@ -93,7 +81,8 @@ impl Window {
 
         Self {
             id: window_id,
-            state: WindowState::Uninitialized(attributes),
+            handle,
+            accesskit_adapter,
             event_reducer: WindowEventReducer::default(),
             render_root: RenderRoot::new(
                 root_widget,
@@ -116,6 +105,7 @@ impl Window {
 /// `MasonryState` via [`MasonryState::new`] and forward events to it via the appropriate method (e.g.,
 /// calling [`handle_window_event`](MasonryState::handle_window_event) in [`window_event`](ApplicationHandler::window_event)).
 pub struct MasonryState<'a> {
+    app_state: AppState,
     render_cx: RenderContext,
     renderer: Option<Renderer>,
     // TODO: Winit doesn't seem to let us create these proxies from within the loop
@@ -276,6 +266,7 @@ impl MasonryState<'_> {
             std::sync::mpsc::channel::<(WindowId, RenderRootSignal)>();
 
         MasonryState {
+            app_state: AppState::Suspended,
             render_cx,
             renderer: None,
             event_loop_proxy,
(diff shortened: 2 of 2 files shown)


masonry_winit/src/app_driver.rs
```diff
@@ -98,35 +97,23 @@ impl DriverCtx<'_, '_> {
...
     pub fn window_handle(&self, window_id: WindowId) -> &WindowHandle {
         let window = self.state.window(window_id);
-        let WindowState::Rendering { handle: window, .. } = &window.state else {
-            panic!(
-                "window with id {window_id:?} is in {:?} state, expected Rendering state",
-                window.state
-            );
-        };
-        window
+        &window.handle
```
nixonyh, line 103: I am not sure if this is the right choice, however, if we still want to panic if it's in suspended state, then we can add an if check here.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -45,46 +45,34 @@ impl From<accesskit_winit::Event> for MasonryUserEvent {
...
-            Self::Uninitialized(attrs) => f.debug_tuple("Uninitialized").field(attrs).finish(),
-            Self::Rendering { .. } => f.debug_struct("Rendering").finish_non_exhaustive(),
-            Self::Suspended { .. } => f.debug_struct("Suspended").finish_non_exhaustive(),
-        }
-    }
+#[derive(Default, Debug)]
+pub enum AppState {
+    /// The app is currently in suspended state,
+    /// usually set during [`ApplicationHandler::suspended()`].
+    #[default]
```
DJMcNab: It's not clear why this Default implementation exists. Should we ever expect users to create an instance of this type?

masonry_winit/src/event_loop_runner.rs
```diff
@@ -45,46 +45,34 @@ impl From<accesskit_winit::Event> for MasonryUserEvent {
...
-impl Debug for WindowState {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            Self::Uninitialized(attrs) => f.debug_tuple("Uninitialized").field(attrs).finish(),
-            Self::Rendering { .. } => f.debug_struct("Rendering").finish_non_exhaustive(),
-            Self::Suspended { .. } => f.debug_struct("Suspended").finish_non_exhaustive(),
-        }
-    }
+#[derive(Default, Debug)]
+pub enum AppState {
```
DJMcNab: Can we workshop a better name than `AppState` (it's extremely generic/undescriptive). I'd lean towards a documented `suspended: bool` field myself.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -300,66 +291,43 @@ impl MasonryState<'_> {
 
     // --- MARK: RESUMED
     pub fn handle_resumed(&mut self, event_loop: &ActiveEventLoop, app_driver: &mut dyn AppDriver) {
+        self.app_state = AppState::Resumed;
```
DJMcNab: I think we can short circuit return if we've already been resumed

masonry_winit/src/event_loop_runner.rs
```diff
@@ -300,66 +291,43 @@ impl MasonryState<'_> {
...
-                    }
-                }
-            }
+        }
+
+        //  Recreate surfaces for all windows.
+        for (handle_id, window) in self.windows.iter() {
+            let surface = create_surface(&mut self.render_cx, window.handle.clone());
+            self.surfaces.insert(*handle_id, surface);
         }
```
DJMcNab: I guess it isn't documented that creating a surface twice for the same window is a problem. It certainly feels fragile, though, and so I'd like to change it.

masonry_winit/src/app_driver.rs
DJMcNab: This seems reasonable enough, I think.

DJMcNab (commented):
This makes sense to me. Thanks for these cleanups.
I do have a few notes for before this lands, and I'd also like to see the commented out code removed.

Does this also fix #1038?

masonry_winit/src/event_loop_runner.rs
nixonyh: Not really actually, let me remove that (I was thinking it'll be more like a guide for users initially.)

masonry_winit/src/event_loop_runner.rs
nixonyh: Sounds like a good option really.

PoignardAzur (commented):
I'm not super familiar with this code, so I can't validate it, but it does look like this PR makes it straightforwardly simpler.

masonry_winit/src/event_loop_runner.rs
nixonyh: I see, I'm not quite familiar with the constraints, is there any pointers on how to do that? Thanks!

DJMcNab: I've pushed a fix. Edit: Undone, as it was bad

DJMcNab: Something like this:
```suggestion
        // Recreate surfaces for all windows.
        for (handle_id, window) in self.windows.iter() {
            let Entry::Vacant(entry) = self.surfaces.entry(*handle_id) else {
                continue;
            };
            let surface = create_surface(&mut self.render_cx, window.handle.clone());
            entry.insert(surface);
        }
```

Because, we cannot create a surface twice. Indeed, the app crashes on startup currently on my machine.

nixonyh: I see, but we did short-circuit the function isn't it to prevent the recreation of the surfaces?

masonry_winit/src/event_loop_runner.rs
```diff
@@ -116,6 +92,9 @@ impl Window {
 /// `MasonryState` via [`MasonryState::new`] and forward events to it via the appropriate method (e.g.,
 /// calling [`handle_window_event`](MasonryState::handle_window_event) in [`window_event`](ApplicationHandler::window_event)).
 pub struct MasonryState<'a> {
+    /// True if in suspended state (usually set during [`ApplicationHandler::suspended()`]).
+    /// Fales if in resumed state (usually set during [`ApplicationHandler::resumed()`]).
+    is_suspended: bool,
```
DJMcNab:
```suggestion
    /// The event loop is suspended when the app is e.g. in the background on Android.
    /// We aren't allowed to have any `Surface`s, and we also don't expect to receive any events.
    is_suspended: bool,
```

masonry_winit/src/event_loop_runner.rs
```diff
@@ -378,32 +347,30 @@ impl MasonryState<'_> {
...
-        ))
-        .unwrap();
+
         let scale_factor = handle.scale_factor();
         let handle_id = handle.id();
-        window.state = WindowState::Rendering {
+
+        // Surface.
+        let surface = create_surface(&mut self.render_cx, handle.clone());
+        self.surfaces.insert(handle_id, surface);
```
DJMcNab: This is invalid if `self.is_suspended`, and so we should defer it. I think just wrapping it with that condition should be valid

masonry_winit/src/event_loop_runner.rs
```diff
@@ -418,36 +385,11 @@ impl MasonryState<'_> {
         // HACK: When we exit, on some systems (known to happen with Wayland on KDE),
         // the IME state gets preserved until the app next opens. We work around this by force-deleting
         // the IME state just before exiting.
-        if let WindowState::Rendering { handle, .. } = window.state {
-            handle.set_ime_allowed(false);
+        if !self.is_suspended {
+            window.handle.set_ime_allowed(false);
         }
```
DJMcNab: I'm fairly sure that this needn't be gated like this. It was probably only gated previously for convenience

masonry_winit/src/event_loop_runner.rs
```diff
@@ -660,14 +601,14 @@ impl MasonryState<'_> {
...
-                    return;
-                };
-                accesskit_adapter.update_if_active(|| tree_update);
+                // let WindowState::Rendering {
+                //     accesskit_adapter, ..
+                // } = &mut window.state
+                // else {
+                //     error!("Suspended inside event");
+                //     return;
+                // };
```
DJMcNab: Does this need to be here still?

masonry_winit/src/event_loop_runner.rs
```diff
@@ -456,16 +398,16 @@ impl MasonryState<'_> {
...
     ) {
-        let WindowState::Rendering { handle, .. } = &mut window.state else {
-            tracing::warn!("Tried to render whilst suspended or before window created");
-            return;
-        };
-        let scale_factor = handle.scale_factor();
+        // let WindowState::Rendering { handle, .. } = &mut window.state else {
+        //     tracing::warn!("Tried to render whilst suspended or before window created");
+        //     return;
+        // };
```
DJMcNab: Same here

DJMcNab (commented):
A few details we still need to get right. In particular, this doesn't work at on Wayland!

If you want me to push this to the end, let me know and I'll do so.

masonry_winit/src/event_loop_runner.rs
nixonyh: I added an early return in the latest push, could you help me check if that's ok?

masonry_winit/src/event_loop_runner.rs
DJMcNab: This is for the windows created from `self.initial_windows`. Which you have just created with a surface, and so are now trying to invalidly give a second active surface to.

masonry_winit/src/event_loop_runner.rs
DJMcNab: That looks ok.

masonry_winit/src/event_loop_runner.rs
nixonyh: Ahh, now I get it!

nixonyh: I think my latest push fixes it. We will re-create all the surfaces before creating new windows to prevent creating the same surface twice.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -293,76 +274,72 @@ impl MasonryState<'_> {
...
+                "attempted to create a window with id {window_id:?} but a window with that id already exists",
             );
         }
 
+        if self.is_suspended {
+            // Wait until resumed before creating the windows.
+            self.new_windows.push((window_id, attributes, root_widget));
+
+            return;
+        }
```
DJMcNab, line 341: The reason not to do it this way is that the caller of `create_window` might reasonably assume that the window will actually be created now. But I think as a solution for now, this is viable

DJMcNab (approved):
I've pushed a last couple of changes, and otherwise will land this now.

Thanks a lot! It's good to see this cleaned up.

masonry_winit/src/event_loop_runner.rs
nixonyh: I see. Would that mean we should panic if it's suspended?

DJMcNab: I don't see why we can't make the window, just without creating the surface. But that was my proposal earlier, and you evaluated against it for some reason. I think still creating the window is somewhat reasonable.

It isn't even meaningful to have multiple windows support on Android or iOS anyway, and we don't support the web.

nixonyh: Ohh I see!

Issue #1307, opened 2025-08-07
Text overflow in TextInput


ricvelozo:
When I resize the window to be very small, the text overflows the input. I expected it to clip the text:

<img width="213" height="244" alt="Image" src="https://github.com/user-attachments/assets/cdcebbf2-5e9e-4aed-99f9-690df415f210" />

The code:

```rust
fn view(state: &ColorPreview) -> impl WidgetView<ColorPreview, Message> + use<> {
    let can_remove = state.colors.len() > 1;
    portal(
        flex_row(
            sized_box(
                flex((
                    state
                        .colors
                        .iter()
                        .enumerate()
                        .map(|(index, color_entry)| {
                            let id = color_entry.id();
                            flex_row((
                                text_input(color_entry.red.to_string(), move |_, input| {
                                    Message::RedInput(index, input)
                                })
                                .flex(1.),
                                text_input(color_entry.green.to_string(), move |_, input| {
                                    Message::GreenInput(index, input)
                                })
                                .flex(1.),
                                text_input(color_entry.blue.to_string(), move |_, input| {
                                    Message::BlueInput(index, input)
                                })
                                .flex(1.),
                                button("Remover", move |_| Message::Remove(id))
                                    .disabled(!can_remove),
                            ))
                            .gap(8.)
                        })
                        .collect::<Vec<_>>(),
                    button("Adicionar", |_| Message::Add),
                ))
                .gap(8.)
                .main_axis_alignment(MainAxisAlignment::Center),
            )
            .padding(8.)
            .width(360.)
            .expand_height(),
        )
        .main_axis_alignment(MainAxisAlignment::Center)
        .must_fill_major_axis(true),
    )
}
```

PR #1311, opened 2025-08-08
Use TextInput clipping in Xilem

DJMcNab:
Fixes https://github.com/linebender/xilem/issues/1307

Also adds a test for how clipping is rendered.


masonry/screenshots/text_input_clip.png
DJMcNab, line 1: The rendering on the right edge here isn't right; I've added some relevant TODO comments, but I'm calling this out-of-scope of this PR.

PoignardAzur (approved):
This touches a few places in the code where it's obvious we made unhappy compromises.

But aside from that, it's a straightforward upgrade (making a masonry feature available in xilem).

LGTM.

PR #1312, opened 2025-08-08
Masonry: Remove the generic on `VirtualScroll`

DJMcNab:
That generic provided little to no value, and made using `VirtualScroll` a bit harder.


PoignardAzur (approved):
Nice. I've wanted to do that with Portal for months.

PR #1318, opened 2025-08-11
Move `on_start` handling to the first resume

DJMcNab:
This is when the actual render roots exist.

This is still clear a hack; it's pretty shoddy design that each render root has its own font context, and I'm unsure why that was signed-off on.

This fixes the `variable_clock` example


DJMcNab:
This is blocked on #1310

PR #1320, opened 2025-08-12
Add `NewWindow` struct to encapsulate `(id, attrs, root_widget)`

nixon-voxell:
This is a proposal PR to switch over from tuples to a `NewWindow` struct.

This PR is actually quite similar to #1215

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -45,6 +45,12 @@ impl From<accesskit_winit::Event> for MasonryUserEvent {
     }
 }
 
+pub struct NewWindow {
+    pub id: WindowId,
+    pub attributes: WindowAttributes,
+    pub root_widget: NewWidget<dyn Widget>,
+}
+
 /// Per-Window state
 pub(crate) struct Window {
     id: WindowId,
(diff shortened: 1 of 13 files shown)


masonry_winit/src/event_loop_runner.rs
```diff
@@ -45,6 +45,12 @@ impl From<accesskit_winit::Event> for MasonryUserEvent {
     }
 }
 
+pub struct NewWindow {
+    pub id: WindowId,
+    pub attributes: WindowAttributes,
+    pub root_widget: NewWidget<dyn Widget>,
+}
```
DJMcNab: These need docs

DJMcNab (commented):
Other than the missing docs, this looks good!

masonry_winit/src/event_loop_runner.rs
nixonyh: Added!

DJMcNab: That I didn't make this clear - I'd also like some docs on the fields as well. For example:
- for `id`, we can talk about that the id is set by the app, and the window can be accessed using this id in `DriverContext::[name_of_method]`.
- For attributes, just a stub documentation is fine - "The attributes the window will be created with; can be created using `masonry_winit::winit::window::Window::default_attributes`"
- We can say that root widget is the widget which will take up the entire contents of the new window.

DJMcNab (commented):
Some more complete docs are still desirable here; I've given some hints

PR #1321, opened 2025-08-13
When a new font is loaded, relayout everything

DJMcNab:
This avoids issues where the initial render (say) doesn't have a given font, but the font is later loaded.

This is always the case with when Xilem accesses the font context, but I think it's also just generally the correct behaviour here.

This also fixes a minor issue with `request_render_all`, where it didn't take into account `post_paint`.

### DIFF
--- a/masonry_core/src/app/render_root.rs
+++ b/masonry_core/src/app/render_root.rs
@@ -453,10 +453,13 @@ impl RenderRoot {
     /// Returns a list of pairs each containing the family identifier and fonts
     /// added to that family.
     pub fn register_fonts(&mut self, data: Blob<u8>) -> Vec<(FamilyId, Vec<FontInfo>)> {
-        self.global_state
+        let ret = self
+            .global_state
             .font_context
             .collection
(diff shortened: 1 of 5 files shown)


masonry_core/src/app/render_root.rs
```diff
@@ -337,7 +337,14 @@ impl RenderRoot {
         };
 
         if let Some(test_font_data) = test_font {
-            let families = root.register_fonts(test_font_data);
+            // We don't use `register_fonts` here because that requests a global relayout.
+            // However, because we are not yet fully initialised (we are before the below call
+            // to `run_rewrite_passes`), `request_layout_all`
```
PoignardAzur: Incomplete doc. I assume the full doc would be "`request_layout_all` would panic".

PoignardAzur (dismissed):
LGTM

DJMcNab:
This hasn't actually fixed the issue...

PoignardAzur (approved):
This code is really making me wish we had a full reactive programming engine.

Other than that, LGTM.

PR #1328, opened 2025-08-13
Make remove_child method recursive

PoignardAzur:
This fixes a memory leak where widgets would never be removed from the scene cache if their grandparent was removed. Also makes focus_anchor behavior slightly more consistent.


DJMcNab (approved):
Seems reasonable to me! Thanks

PR #1329, opened 2025-08-13
Make non-primary click events focus buttons and checkboxes.

PoignardAzur:
Keeping this as a draft, because I think we already want non-primary clicks on buttons to send actions. This very low-priority for now.

### DIFF
--- a/masonry/src/tests/event.rs
+++ b/masonry/src/tests/event.rs
@@ -233,7 +233,10 @@ fn click_anchors_focus() {
         ))
         .with_child(NewWidget::new(Button::with_text("")))
         .with_child(NewWidget::new(Button::with_text("")))
-        .with_child(NewWidget::new_with_tag(Button::with_text(""), child_3))
+        .with_child(NewWidget::new_with_tag(
+            SizedBox::empty().size(10.px(), 10.px()),
+            child_3,
+        ))
         .with_child(NewWidget::new_with_tag(Button::with_text(""), child_4))
         .with_child(NewWidget::new(Button::with_text("")))
         .with_auto_id();
@@ -244,7 +247,7 @@ fn click_anchors_focus() {
     let child_4_id = harness.get_widget(child_4).id();
     let other_id = harness.get_widget(other).id();
 
-    // Clicking a button doesn't focus it.
+    // Clicking a sized box doesn't focus it.
     harness.mouse_click_on(child_3_id);
     assert_eq!(harness.focused_widget_id(), None);
 


waywardmonkeys:
Curious what the reasoning is behind this and whether anyone else does this?

DJMcNab:
The right click behaviour matches e.g. Firefox (but obviously, that also open a context menu, which we don't have).

PoignardAzur:
It's also the behavior for middle-click, and I assume other non-primary clicks? I haven't tested it in that much detail.

PoignardAzur:
On closer inspection, I realized that the web behavior is "Any click on an focusable widget focuses it". It's just that primary clicks on buttons and links will usually navigate away, so we don't notice them.

masonry/src/widgets/button.rs
```diff
@@ -112,15 +112,24 @@ impl Widget for Button {
...
-                // Changes in pointer capture impact appearance, but not accessibility node
-                ctx.request_paint_only();
-                trace!("Button {:?} pressed", ctx.widget_id());
+            PointerEvent::Down { button, .. } => {
+                if *button == Some(PointerButton::Primary) {
+                    ctx.capture_pointer();
+                    // Changes in pointer capture impact appearance, but not accessibility node
+                    ctx.request_paint_only();
+                    trace!("Button {:?} pressed", ctx.widget_id());
+                }
```
DJMcNab, line 121: This change means that buttons are no longer usable on touch screens (e.g. on Android).
We cannot land this change with that regression in place.

It also breaks `button_any_pointer` on the Xilem side, which is fine, but we should just remove it if we're doing that. But no longer supporting touch screens is not a good plan. `button_any_pointer` was added as a quick hack to unblock a user, but I'm not going to dig into who that was at the moment.

masonry/src/widgets/button.rs
```diff
@@ -363,6 +373,19 @@ mod tests {
...
+        let button = NewWidget::new(Button::with_text("Hello"));
+
+        let mut harness = TestHarness::create(default_property_set(), button);
+        let button_id = harness.root_id();
+
+        harness.mouse_move_to(button_id);
+        harness.mouse_button_press(PointerButton::Secondary);
+        assert_eq!(harness.focused_widget_id(), Some(button_id));
+        assert_matches!(harness.pop_action_erased(), None);
+    }
```
DJMcNab, line 387: We need a test for an emulated touch event here. That's a `Down` event with a button of `None`.

masonry/src/widgets/button.rs
```diff
@@ -112,15 +112,24 @@ impl Widget for Button {
...
-                trace!("Button {:?} pressed", ctx.widget_id());
+            PointerEvent::Down { button, .. } => {
+                if *button == Some(PointerButton::Primary) {
+                    ctx.capture_pointer();
+                    // Changes in pointer capture impact appearance, but not accessibility node
+                    ctx.request_paint_only();
+                    trace!("Button {:?} pressed", ctx.widget_id());
+                }
+                // Any click event should lead to this widget getting focused.
+                ctx.request_focus();
```
DJMcNab, line 123: Should this set the focus anchor, rather than actually setting the focus directly? I think either way would be viable. On the web, it seems to do the former, from some brief testing.

This doesn't block this PR.

masonry/src/widgets/button.rs
PoignardAzur: So what kind of filter could we use that accepts touchscreen presses but rejects right-clicks?

Is it `Some(PointerButton::Primary) | None` ?

DJMcNab: Yes. I should have said that explicitly

PR #1331, opened 2025-08-13
Rename `Harness::get_records_of` to `Harness::take_records_of`

PoignardAzur:
The name better conveys that the method drains the records from the given widget.


waywardmonkeys:
Or `drain`

DJMcNab (approved):
For test code, performance shouldn't be a *huge* concern, so the misleadingness of the name `take` (that it suggests that new records require allocation) doesn't matter, and semantically you are taking them. It's certainly an improvement over `get`.

I definitely prefer `take` over `pop`, but I don't have a strong preference between `drain` and `take`.

PR #1332, opened 2025-08-14
Add base color customization to `Windows` in `masonry_winit`

nixonyh:
Allow users to customize the base color of the window via
```rs
NewWindow::new(..).with_base_color(Color::WHITE)
```

### DIFF
--- a/masonry_winit/README.md
+++ b/masonry_winit/README.md
@@ -114,7 +114,7 @@ fn main() {
         .unwrap();
     masonry_winit::app::run_with(
         event_loop,
-        vec![NewWindow::new(
+        vec![NewWindow::new_with_id(
             driver.window_id,
             window_attributes,
             NewWidget::new(main_widget).erased(),
--- a/masonry_winit/src/lib.rs
+++ b/masonry_winit/src/lib.rs
@@ -92,7 +92,7 @@
 //!         .unwrap();
 //!     masonry_winit::app::run_with(
 //!         event_loop,
-//!         vec![NewWindow::new(
+//!         vec![NewWindow::new_with_id(
 //!             driver.window_id,
 //!             window_attributes,
 //!             NewWidget::new(main_widget).erased(),


masonry_winit/src/app_driver.rs
```diff
@@ -110,6 +111,15 @@ impl DriverCtx<'_, '_> {
...
 
+    /// Set the base color of the window.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the window cannot be found.
+    pub fn set_window_base_color(&mut self, window_id: WindowId, base_color: Color) {
+        self.state.window_mut(window_id).base_color = base_color;
+    }
+
```
nixonyh: This feels a little "bad" API wise?

nixonyh: We could also add a `get_window_base_color` API...?

waywardmonkeys: This wasn't configurable before. What if that isn't added here yet and we tackle that later?

nixonyh: Sure, we can do that.

DJMcNab (commented):
This needs to be hooked up to Xilem, and I'd like to see at least one of the Masonry examples use it.

The code otherwise looks good. I really wish we had https://github.com/rust-lang/rust/issues/132162 to use here.

nixon-voxell:
> The code otherwise looks good. I really wish we had [rust-lang/rust#132162](https://github.com/rust-lang/rust/issues/132162) to use here.

Thanks for linking the RFC, nice that they are working on this!

DJMcNab:
I'd lean towards reverting/removing 814b52fe7991c6ef97f27d8269edb699c2cfceb8 here. I think the commit beforehand was still fine. I can see the motivation, but it's an awful lot of noise for very little gain. `NewWindow` was effectively created to allowing adding new fields, so it's not clear that nesting further here has any advantage.

We can defer integrating this into Xilem - now that it's used in at least one example, that means that this PR is complete.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -63,6 +63,37 @@ pub struct NewWindow {
...
+
+    /// Create a new window with auto assigned [`WindowId::next()`].
+    pub fn new_with_auto_id(
+        attributes: WindowAttributes,
+        root_widget: NewWidget<dyn Widget + 'static>,
+    ) -> Self {
+        Self::new(WindowId::next(), attributes, root_widget)
+    }
+
+    /// Set the base color of the new window.
```
DJMcNab, line 90:
```suggestion
    /// Set the base color of the new window.
    ///
    /// The base color is the color of the background which all widgets in the window draw on top of.
    /// Masonry's current default theme assumes that this will be a very dark color for sufficient contrast.
    /// This is most useful for apps which want to for example support light mode.
    ///
    /// Please note that it is not currently supported to modify this once the app is running.
    /// This is not a fundamental limitation, and is only due to missing api design.
```

masonry_winit/src/event_loop_runner.rs
```diff
@@ -63,6 +63,37 @@ pub struct NewWindow {
     pub attributes: WindowAttributes,
     /// The widget which will take up the entire contents of the new window.
     pub root_widget: NewWidget<dyn Widget>,
+    /// The base color of the window.
+    pub base_color: Color,
+}
+
+impl NewWindow {
+    pub fn new(
```
DJMcNab: This method should be documented (we have allow `missing_docs` in this module for historical reasons, but that doesn't mean we should still have no docs for new fields.

It only needs extremely cursory docs, as the docs on the fields largely cover things.

```suggestion
    /// A default `NewWindow`.
    ///
    /// Further properties of the window which will be created using
    /// builder-style methods on this type.
    /// See the documentation on the fields of this type for details of the parameters.
    pub fn new(
```

nixon-voxell:
Let me know if there's anything else to add!

masonry_winit/src/event_loop_runner.rs
```diff
@@ -63,6 +63,37 @@ pub struct NewWindow {
...
+        }
+    }
+
+    /// Create a new window with auto assigned [`WindowId::next()`].
+    pub fn new_with_auto_id(
+        attributes: WindowAttributes,
+        root_widget: NewWidget<dyn Widget + 'static>,
+    ) -> Self {
+        Self::new(WindowId::next(), attributes, root_widget)
+    }
```
PoignardAzur: I'd recommend the other way around: have `new` be the short version and `new_with_id` the longer one.

nixonyh: Made the changes.

PR #1333, opened 2025-08-14
Add brief note about rare rust feature

nilsmartel:
The rust feature in question (e.g. use<>) is notoriously hard to google and very rarely used. Still, it's present throughout xilem.

This combination makes it likely that newcomers and even experienced rust devs will experience a lot of frustration with this codebase.

We can lift that frustration by actively linking the right documents.

Use case:
Me. I had a really hard time to find out, what was going on here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,19 @@ The [Rust code of conduct] applies.
 
 Unless you explicitly state otherwise, any contribution intentionally submitted for inclusion in the work by you, as defined in the Apache 2.0 license, shall be licensed as noted in the [License](#license) section, without any additional terms or conditions.
 
+## Recommended Reading
+
+Throughout Xilem you will find usage a rust syntax many people are not familiar with, the `+ use<>`.
+Here's a snippet from the xilem examples:
+
+```rust
+fn app_logic(data: &mut EmojiPagination) -> impl WidgetView<EmojiPagination> + use<> {
+   // ...
+}
+```
+
+You can read more about it in [this document](https://github.com/rust-lang/rfcs/blob/master/text/3617-precise-capturing.md).
+
 ## License
 
 Licensed under the Apache License, Version 2.0 ([LICENSE](LICENSE) or <http://www.apache.org/licenses/LICENSE-2.0>)


README.md
```diff
@@ -193,6 +193,19 @@ The [Rust code of conduct] applies.
...
+Throughout Xilem you will find usage a rust syntax many people are not familiar with, the `+ use<>`.
+Here's a snippet from the xilem examples:
+
+```rust
+fn app_logic(data: &mut EmojiPagination) -> impl WidgetView<EmojiPagination> + use<> {
+   // ...
+}
+```
+
+You can read more about it in [this document](https://github.com/rust-lang/rfcs/blob/master/text/3617-precise-capturing.md).
```
DJMcNab: Can we use a link to https://doc.rust-lang.org/stable/std/keyword.use.html#precise-capturing, instead.

We probably should call this whole section "Precise Capturing" or similar.

README.md
```diff
@@ -193,6 +193,19 @@ The [Rust code of conduct] applies.
 
 Unless you explicitly state otherwise, any contribution intentionally submitted for inclusion in the work by you, as defined in the Apache 2.0 license, shall be licensed as noted in the [License](#license) section, without any additional terms or conditions.
 
+## Recommended Reading
```
DJMcNab: This is in a bit of a strange position in the README. I'd expect it to be at least before the MSRV section (and probably also before Prerequisites)

README.md
DJMcNab, line 1: Whatever we end up with probably also need to be integrated into `xilem/README.md` (maybe also `xilem_core/src/lib.rs`, but that can be outside this PR)

README.md
```diff
@@ -193,6 +193,19 @@ The [Rust code of conduct] applies.
...
+
+Throughout Xilem you will find usage a rust syntax many people are not familiar with, the `+ use<>`.
+Here's a snippet from the xilem examples:
```
DJMcNab:
```suggestion
Here's a snippet from the Xilem examples:
```

README.md
```diff
@@ -193,6 +193,19 @@ The [Rust code of conduct] applies.
...
+
+```rust
+fn app_logic(data: &mut EmojiPagination) -> impl WidgetView<EmojiPagination> + use<> {
+   // ...
+}
+```
+
```
DJMcNab: Can we have a paragraph like:
> This is the precise capturing syntax, which indicates that the returned view does not make use of the lifetime of data.
> This is required because the view types in Xilem must be `'static`, but as of Rust's edition 2024, `impl Trait`
> assumes that parameter's lifetimes will be used in the return value.
> That is a good assumption for most Rust code, but this is mismatched with how Xilem works.

README.md
nilsmartel: I've taken some time to really find a place in the readme, and I think from a user perspective it belongs somewhat close to things like "memoization" and "components".

E.g. from a readers perspective these are interesting things that are just good to know about for programming with this. You will come into contact with it in a similar way.

Since it's not related to the overlying architecture, I have the section of "precise capturing" on it's own between "architecture" and "prerequisits".

Sitting right under technical things like Optional Type Erasure I feel good about this being a meaningful place.

image

README.md
nilsmartel: nice!

README.md
nilsmartel: The documentation is now in better shape thanks to your counseling.
I really enjoy the quality standard you maintainers are setting, this is a jewel of engineering in many regards thanks to that spirit!

If you think this is ready, I would maybe just copy and paste this section under "## View Elements" into xilem/Readme.md and then merge

nilsmartel: If you think it should be a separate PR you can just merge as it is and comment.
I think it's best to leave xilem_core alone.
xilem/xilem is curated better sand most people should look there anyway.
Since xilem_core/Readme looks more like a WIP it makes sense to actively shift focus to xilme/xilem instead

PR #1335, opened 2025-08-14
Add `assert_any`, `assert_all` and `assert_none` helper macros.

PoignardAzur:
The macros gives rich error messages for cases where we want apply a predicate to an entire collection.

### DIFF
--- a/masonry_testing/src/assert_any.rs
+++ b/masonry_testing/src/assert_any.rs
@@ -1,0 +1,103 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// Helper macro, checks that at least one item of the given `IntoIterator` matches the given predicate.
+#[macro_export]
+macro_rules! assert_any {
+    ($iter:expr, $pred:expr $(,)?) => {
+        $crate::assert_any_inner($iter, $pred)
+    };
+}
+
+/// Helper macro, checks that all items of the given `IntoIterator` match the given predicate.
+#[macro_export]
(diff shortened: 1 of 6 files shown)


masonry_testing/src/assert_any.rs
```diff
@@ -0,0 +1,103 @@
...
+    };
+}
+
+/// Helper macro, checks that no item of the given `IntoIterator` matches the given predicate.
+#[macro_export]
+macro_rules! assert_none {
+    ($iter:expr, $pred:expr $(,)?) => {
+        $crate::assert_none_inner($iter, $pred)
+    };
+}
```
DJMcNab, line 26: Given that these macros' bodies are literally just a function call with the given arguments, let's just give the functions these names and call them as functions.

DJMcNab (approved):
If you're going to be firm here, I guess I can approve and make an immediate PR removing the macros.

There definitely are benefits to it not being a macro. The interface of these functions (which, let's be clear, are *somewhat* complicated) will be documented clearly by the compiler/rustdoc, rather than the opaque functionality of the macro. Like, there is documentation that says it takes an `IntoIterator`, but nothing about what the predicate is (if it's a function, how many arguments, etc.)

Additionally, if we're talking about expectations, then I'm sure that it being a macro will users will expect there to be support for an assertion message, just like all of the standard library's assertion functionality. And indeed, that is why `assert_eq`/`assert` are macros, instead of functions.
I don't think it would be a good use of time to add that assertion message functionality, but without it, I don't think these being a macro is the right design.

PR #1337, opened 2025-08-14
Placehero: Use virtual scrolling for timelines

DJMcNab:
image

I've been slower to get to this than I would have liked, but the actual implementation wasn't especially difficult.

I haven't tested for edge cases, and currently the storage just grows in an unbounded fashion. But it does appear to be working! See the linked statuses from 2023, around the excellent Rustlab trip.

### DIFF
--- a/placehero/src/components/timeline.rs
+++ b/placehero/src/components/timeline.rs
@@ -100,7 +100,7 @@ impl Timeline {
                     } else {
                         // We would like to cache the status, but this is currently not supported due to `Avatars`
                         // (and everything environment-system) not being properly rebuild aware.
-                        // This is planned, but doesn't happen
+                        // This is planned, but the infra isn't in place yet.
                         // match timeline.cached_statuses.entry(idx) {
                         //     hash_map::Entry::Occupied(occupied_entry) => occupied_entry.get().clone(),
                         //     hash_map::Entry::Vacant(vacant_entry) => vacant_entry
--- a/placehero/src/components/timeline.rs
+++ b/placehero/src/components/timeline.rs
@@ -4,14 +4,18 @@
 use megalodon::Megalodon;
 use megalodon::entities::{Account, Status};
 use megalodon::megalodon::GetAccountStatusesInputOptions;
-use xilem::WidgetView;
 use xilem::core::fork;
+use xilem::core::one_of::{OneOf, OneOf3};
+use xilem::masonry::core::ArcStr;
+use xilem::masonry::properties::types::AsUnit;
 use xilem::palette::css;
 use xilem::style::Style;
 use xilem::tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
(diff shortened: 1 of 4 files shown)


placehero/src/components/timeline.rs
```diff
@@ -1,28 +1,146 @@
...
 /// and own their own boxes
-// TODO: Generalise to e.g. the explore page.
-pub(crate) fn timeline(statuses: &mut [Status]) -> impl WidgetView<Placehero, Navigation> + use<> {
-    portal(
-        flex(statuses.iter().map(timeline_status).collect::<Vec<_>>()).padding(Padding {
-            // Leave room for scrollbar
-            right: 20.,
-            ..Padding::all(5.0)
-        }),
+pub(crate) fn timeline(
```
jaredoconnell: Can you add more comments to the entire timeline function? I think that this will end up being an example for a more advanced use case of virtual school so I think the comments would be worth it.

DJMcNab:
>  I tested one edge case: loss of internet. It seemed to handle it well. Though a loading indicator would be preferred.

I'm surprised that we don't panic here. The error handling is all `todo!()`!

placehero/src/components/timeline.rs
```diff
@@ -46,105 +52,123 @@ impl Timeline {
...
+                                )
+                                .await;
+                            drop(proxy.message(result));
+                        }
                     }
-                    Err(e) => {
-                        todo!("handle {e}")
+                },
+                |timeline: &mut Self, sender| {
+                    // `worker` creates a channel pair for us; we need to keep out own track
```
jaredoconnell: Typo: out/our

placehero/src/components/timeline.rs
```diff
@@ -46,105 +52,123 @@ impl Timeline {
...
-            },
-            |timeline: &mut Timeline, resp| {
-                match resp {
-                    Ok(mut instance) => {
-                        // If we're at the oldest post, there's no use making the same request again
-                        if !instance.json.is_empty() {
-                            timeline.pending_id = false;
+                    if local_idx == timeline.statuses.len() {
+                        if timeline.at_end {
+                            OneOf3::A(prose("End of timeline.").text_alignment(TextAlign::Center))
```
jaredoconnell, line 96: Why are `OneOf` and `OneOf3` mixed?

jaredoconnell (approved):
Looks great! Just some minor comments.

placehero/src/components/timeline.rs
DJMcNab: I believe this to be the idiomatic form (although we've had a few questions about this before), and it's not really documented.

Essentially, we use `OneOf3` as the first variant to let Rust infer that `D to I` are all `Never`, but then use `OneOf` for the remainder. This allows us to add new variants whilst only needing to change the "first" one - not needing to update every item.

PR #1348, opened 2025-08-15
Add the `flex_row()` and `flex_col()` helper functions, and refactor the `flex()` function.

bowenxuuu:
`Horizontal` and `Vertical` layouts are used very frequently in GUI development, so it is necessary to add the `row()` and `column()` helper functions.

Related discussion: https://github.com/linebender/xilem/pull/1102

### DIFF
--- a/xilem/examples/flex.rs
+++ b/xilem/examples/flex.rs
@@ -20,7 +20,7 @@ fn big_button(
 }
 
 fn app_logic(data: &mut i32) -> impl WidgetView<i32> + use<> {
-    // This is the flex view, alternatives are `column` or `flex` which allows dynamically switching the axis
+    // This is the flex view, alternatives are `flex_col` or `flex` which allows dynamically switching the axis
     flex_row((
         FlexSpacer::Fixed(30.px()),
         big_button("-", |data| {


xilem/src/view/flex.rs
```diff
@@ -36,7 +36,7 @@ use crate::{AnyWidgetView, Pod, PropertyTuple as _, ViewCtx, WidgetView};
 /// }
 ///
 /// fn app_logic(data: &mut i32) -> impl WidgetView<i32> + use<> {
-///     flex((
+///     flex(Axis::Horizontal, (
```
Philipp-M, line 40: Might be worth mentioning the specialized components `column`,`row` as an alternative to using this, somewhere within this doc-comment.

xilem/examples/flex.rs
```diff
@@ -20,7 +20,7 @@ fn big_button(
 }
 
 fn app_logic(data: &mut i32) -> impl WidgetView<i32> + use<> {
-    flex_row((
+    row((
```
Philipp-M: As it's not immediately obvious anymore with that name, maybe a comment above:

```suggestion
    // This is the flex view, alternatives are `column` or `flex` which allows dynamically switching the axis
    row((
```

Philipp-M (approved):
I think the comments (which are seemingly lost in Githubs new buggy review section...) should be handled to make it more obvious that `row` and `column` are `Flex` views.

Approving based on the comment from Daniel.
I have slightly mixed feelings about this, on the one hand this certainly reads nicer, on the other hand, the names `column` and `row` are quite general, and they could e.g. be misunderstood, and be used in e.g. a grid (which is possible, but might have different effects than expected).

I'll let the chance of others giving it a review and merge this.

xb284524239:
After consideration, I think `h_flex()` and `v_flex()` may be more appropriate than `row()` and `column()`. The reasons are as follows:
 
The prefixes `h_` and `v_` can clearly indicate `horizontal` and `vertical` layouts.
The suffix `_flex` indicates that they are related to `flex()`.
As Philip said, `row()` and `column()` are quite general and may be used in `grids`, for example.
 
Please let me know if you agree to use `h_flex()` and `v_flex()`. I'd be happy to make the changes.

xilem/src/view/flex.rs
```diff
@@ -61,11 +60,12 @@ use crate::{AnyWidgetView, Pod, PropertyTuple as _, ViewCtx, WidgetView};
 /// }
 /// ```
 pub fn flex<State, Action, Seq: FlexSequence<State, Action>>(
+    axis: Axis,
```
T-256, line 64: Couldn't get rid of this function instead of adding a new parameter to it?

bowenxuuu: I suggest keeping `flex()` for now. The axis parameter will be useful in scenarios where dynamic switching between horizontal or vertical layouts is needed.

xilem/src/view/flex.rs
bowenxuuu: Ok.

DJMcNab (approved):
Thanks! I realise this PR is going to be a merge conflict nightmare, so I'm going to merge now.

PR #1349, opened 2025-08-15
Move examples from `masonry_winit` to `masonry`

PoignardAzur:
Rewrite their respective crate roots.

Supersedes https://github.com/linebender/xilem/pull/1158


masonry_winit/src/lib.rs
```diff
@@ -104,28 +62,11 @@
...
 //!
-//! [winit]: https://crates.io/crates/winit
-//! [Druid]: https://crates.io/crates/druid
-//! [Masonry]: https://docs.rs/masonry
-//! [Xilem]: https://crates.io/crates/xilem
-//! [tracing_tracy]: https://crates.io/crates/tracing-tracy
-// TODO: Add screenshot. This can't use include_screenshot as that doesn't work with cargo-rdme
-// See https://github.com/linebender/xilem/issues/851
+//! [Masonry's documentation]: https://docs.rs/masonry
+//! [Winit]: https://crates.io/crates/winit
```
DJMcNab: Technically, this should be an intra-doc link to the winit dependency of this crate (for better offline browsing). But it is my fault that we don't have good enough guidance for how you would actually implement that (essentially, you move this link reference definition to the top of README.md, outside the cargo-rdme block).

This is non-blocking

masonry_winit/src/lib.rs
```diff
@@ -137,14 +78,10 @@
...
-#![cfg_attr(
-    test,
-    expect(
-        unused_crate_dependencies,
-        reason = "False-positive with dev-dependencies only used in examples"
-    )
+#![expect(
+    clippy::needless_doctest_main,
+    reason = "Having a main function is a deliberate part of the root doc."
 )]
```
DJMcNab, line 146: This is a bit vestigal! Good to have it removed

masonry/README.md
```diff
@@ -19,15 +19,141 @@ To edit the following section, update it in lib.rs, then run:
...
+Masonry is built on top of:
+
+- [Vello] and [wgpu] for 2D graphics.
+- [Parley] for the text stack.
+- [AccessKit] for plugging into accessibility APIs.
+
+There are currently two backends for creating windows:
+
+- [masonry_winit] for most platforms.
+- `masonry_android_view` for Android.
```
DJMcNab, line 54: Maybe we should comment that this is extremely work in progress, and our most practical Android support currently is still using Masonry Winit?

masonry/README.md
```diff
@@ -19,15 +19,141 @@ To edit the following section, update it in lib.rs, then run:
...
+
+See [Xilem] as an example of a reactive UI library built on top of Masonry.
```
DJMcNab: I wonder if we should mention Xilem higher up - something like:

> The developers of Masonry are also developing Xilem, a reactive UI library built on top of Masonry. 
> Masonry's API is geared towards creating UI libraries; if you are creating an application, we recommend using Xilem.

OTOH, it's also fair to say that Masonry is more ready for prime-time than Xilem itself is (which is a vindication of the Masonry/Xilem split, of course!).

This is also non-blocking

masonry/README.md
```diff
@@ -19,15 +19,141 @@ To edit the following section, update it in lib.rs, then run:
...
+
+- `tracy`: Enables creating output for the [Tracy](https://github.com/wolfpld/tracy) profiler using [`tracing-tracy`][tracing_tracy].
+  This can be used by installing Tracy and connecting to a Masonry with this feature enabled.
+
+### Debugging features
+
+Masonry apps currently ship with two debugging features built in:
+- A rudimentary widget inspector - toggled by F11 key.
+- A debug mode painting widget layout rectangles - toggled by F12 key.
+
```
DJMcNab, line 161:
```suggestion
- Automatic tracing output, including to a file in debug mode.

```
Plus updating the count, or changing to "several, including"

masonry/src/lib.rs
```diff
@@ -1,10 +1,143 @@
...
+//! - A rudimentary widget inspector - toggled by F11 key.
+//! - A debug mode painting widget layout rectangles - toggled by F12 key.
+//!
+//! [masonry_winit]: https://crates.io/crates/masonry_winit
+//! [Vello]: https://crates.io/crates/vello
+//! [wgpu]: https://crates.io/crates/wgpu
+//! [Parley]: https://crates.io/crates/parley
+//! [AccessKit]: https://crates.io/crates/accesskit
+//! [Xilem]: https://crates.io/crates/xilem
+//! [tracing_tracy]: https://crates.io/crates/tracing-tracy
```
DJMcNab: Similarly, most of these should be intra-doc-links. Again, non-blocking

masonry/README.md
xStrom: Commenting about work-in-progress is fine. However, if our vision of the future is using Android View and not Winit, then I don't think we should direct anyone to the Winit version, even if it still works better. (*Similarly, Druid has features beyond Xilem, but we're not directing anyone there.*)

DJMcNab: It's not too important, I guess. The fact is that it's not currently practical to use Android View without just forking that repo, which is so far from the long term plan that I don't think pointing people towards using it is actually doing a service to those users.

xStrom: Forking is required? Ok that does sound extreme.

However, it's not really about doing users a service. It's about directing focus. If we want to focus on Android View, then we don't really want to deal with any bug reports or PRs related to the Winit Android implementation.

Maybe as a compromise then, a note about the Winit Android version that clearly states it's deprecated?

DJMcNab: If are interested in Masonry Android View, they can open an issue or ask on Zulip.

> Maybe as a compromise then, a note about the Winit Android version that clearly states it's deprecated?

We haven't had this causing any issues for us in the past, so there's probably no need to do this. We can revisit if it becomes a problem.
It is correct that the state on Android with Winit is *similar* to the state on the web, which is that we can't condone it, for user experience reasons.

I don't think any change needs to (or is going to) happen in this PR. I propose just resolving this thread.

xStrom: Sure, don't need to bikeshed this too much. 👍

DJMcNab:
I'm going to take this over - the longer this waits, the more it will bitrot.

masonry/src/lib.rs
```diff
@@ -1,10 +1,143 @@
...
 
-//! Traits and types of the Masonry toolkit.
-//! See [Masonry Winit's documentation] for more details, examples and resources.
+//! Masonry is a foundational framework for building GUI libraries in Rust.
 //!
-//! [Masonry Winit's documentation]: https://docs.rs/masonry_winit/latest/
+//! Masonry gives you a platform-independent manager, which owns and maintains a widget tree.
+//! It also gives you tools to inspect that widget tree at runtime, write unit tests on it, and generally have an easier time debugging and maintaining your app.
+//!
+//! The framework is not opinionated about what your user-facing abstraction will be: you can implement immediate-mode GUI, the Elm architecture, functional reactive GUI, etc, on top of Masonry.
```
tomcur, line 12: Nit: missing period after "etc."

```suggestion
//! The framework is not opinionated about what your user-facing abstraction will be: you can implement immediate-mode GUI, the Elm architecture, functional reactive GUI, etc., on top of Masonry.
```

masonry/src/lib.rs
```diff
@@ -1,10 +1,142 @@
...
+//!         driver,
+//!         default_property_set(),
+//!     )
+//!     .unwrap();
+//! }
+//! ```
+//!
+//! # Crate feature flags
+//!
+//! The following feature flags are available:
```
tomcur, line 121: Nit: as its just one for now, maybe just:

```suggestion
//! The following feature flag is available:
```

Feels a bit template-y otherwise.

masonry/README.md
```diff
@@ -16,18 +16,152 @@
...
+Masonry is built on top of:
+
+- [Vello][vello] and [wgpu][vello::wgpu] for 2D graphics.
+- [Parley][parley] for the text stack.
+- [AccessKit][accesskit] for plugging into accessibility APIs.
+
+There are currently two backends for creating windows:
+
+- [masonry_winit] for most platforms.
+- `masonry_android_view` for Android.
```
tomcur, line 54: Could we link to https://github.com/rust-mobile/android-view? I do understand with the other discussion thread (https://github.com/linebender/xilem/pull/1349/files#r2282377768) we want to be careful with the exact messaging of the current state.

masonry/src/lib.rs
```diff
@@ -1,10 +1,142 @@
...
+//!
+//! It *is* opinionated about its internals: things like text focus, pointer interactions and accessibility events are often handled in a centralized way.
+//!
+//! Masonry is built on top of:
+//!
+//! - [Vello][vello] and [wgpu][vello::wgpu] for 2D graphics.
+//! - [Parley][parley] for the text stack.
+//! - [AccessKit][accesskit] for plugging into accessibility APIs.
+//!
+//! There are currently two backends for creating windows:
```
tomcur, line 22: Could (/ should) we mention that Masonry is not opinionated about the backend, perhaps linking to `masonry_core`? (I'm not entirely up to date on how well Masonry is currently abstracted, though.)

DJMcNab:
Thanks for the feedback. I think I'll action it in an (immediate) follow-up, as this PR is a merge conflict magnet

PR #1352, opened 2025-08-18
Add Error Handling to virtual cats example

nilsmartel:
in case of failing to download the image.

I wanted to practice and get to know xilem a little.
When I found this TODO in the source code I thought this nice and silly little example might be a good fit, where I can cause no harm.

Here is the finished result which you may see running this without a connection

Bildschirmfoto 2025-08-19 um 16 56 33

### DIFF
--- a/xilem/examples/virtual_cats.rs
+++ b/xilem/examples/virtual_cats.rs
@@ -11,7 +11,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use masonry::core::ArcStr;
-use masonry::properties::types::{AsUnit, UnitPoint};
+use masonry::properties::types::{AsUnit, Length, UnitPoint};
 use masonry::properties::{LineBreaking, Padding};
 use vello::peniko::{Blob, Image};
 use winit::dpi::LogicalSize;
@@ -23,7 +23,7 @@ use xilem::view::{
     ObjectFit, ZStackExt, flex, image, label, prose, sized_box, spinner, virtual_scroll, zstack,
 };
 use xilem::{
-    Color, EventLoop, EventLoopBuilder, TextAlign, WidgetView, WindowOptions, Xilem, palette,
+    Color, EventLoop, EventLoopBuilder, FontWeight, TextAlign, WidgetView, WindowOptions, Xilem,
 };
 use xilem_core::one_of::OneOf3;
 
@@ -43,7 +43,6 @@ struct Status {
 /// The state of the download of an image from a URL
 enum ImageState {
     Pending,
-    // Error(?),
     Available(Image),
     Error(anyhow::Error),
 }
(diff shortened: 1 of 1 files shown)


xilem/examples/virtual_cats.rs
```diff
@@ -71,50 +76,74 @@ impl VirtualCats {
                         let result = image_from_url(&url).await;
                         match result {
                             // We choose not to handle the case where the event loop has ended
-                            Ok(image) => drop(proxy.message(image)),
-                            // TODO: Report in the frontend
```
nilsmartel, line 75: This line prompted the endeavor

xilem/examples/virtual_cats.rs
```diff
@@ -42,6 +45,7 @@ enum ImageState {
     Pending,
     // Error(?),
```
DJMcNab: This comment is now outdated

xilem/examples/virtual_cats.rs
```diff
@@ -71,50 +75,74 @@ impl VirtualCats {
...
                             // We choose not to handle the case where the event loop has ended
-                            Ok(image) => drop(proxy.message(image)),
-                            // TODO: Report in the frontend
+                            Ok(image) => drop(proxy.message(Ok(image))),
                             Err(err) => {
                                 tracing::warn!(
                                     "Loading image for HTTP status code {code} from {url} failed: {err:?}"
                                 );
+                                drop(proxy.message(Err(err)));
                             }
```
DJMcNab: I think this whole match could be replaced with `drop(proxy.message(result))`, now. We would lose the logging, but I suspect that's fine - we show the error on screen (admittedly with less context). 
We could log in an [`inspect_err`](https://doc.rust-lang.org/stable/core/result/enum.Result.html#method.inspect_err), but I don't think that's really worthwhile.

xilem/examples/virtual_cats.rs
```diff
@@ -71,50 +75,74 @@ impl VirtualCats {
...
                     }
                 },
-                move |state: &mut Self, image| {
-                    if let Some(status) = state.statuses.iter_mut().find(|it| it.code == code) {
-                        status.image = ImageState::Available(image);
-                    } else {
+                move |state: &mut Self, message| {
+                    let Some(status) = state.statuses.iter_mut().find(|it| it.code == code) else {
                         unreachable!("We never remove items from `statuses`")
+                    };
```
DJMcNab: Nice!

xilem/examples/virtual_cats.rs
```diff
@@ -71,50 +75,74 @@ impl VirtualCats {
...
+            ImageState::Error(err) => {
+                // the people deserve their cat.
+                // It is vital that the cat explains what went wrong.
+                let asciicat = label(
+                    "
+                  /\\_/\\  !
+                 ( o.o )
+                  > ^ <",
+                )
+                .color(BLACK);
```
DJMcNab: I'm of two minds here. This is of course cute, which is good.
However, I can see a few issues:
- Is there attribution for this art? We need to make sure we're following whatever license it's under. That adds complication.
- For accessibility purposes, I suspect that this is pretty hard to understand. We don't have good ways to exclude things from accessibility currently.

Maybe we can just use a 😿 instead?

xilem/examples/virtual_cats.rs
```diff
@@ -71,50 +75,74 @@ impl VirtualCats {
...
+                let errorstring = prose(err.to_string()).text_color(Color::from_rgb8(255, 0, 0));
```
DJMcNab: This text color/background combination possibly isn't sufficiently high contrast:
https://webaim.org/resources/contrastchecker/?fcolor=FF0000&bcolor=FFFFFF (I'm not certain this is the best website, but it gets the point across).
It's at least close, so it might be fine? Maybe we should make this text slightly larger or bold?

DJMcNab (commented):
This is looking really good. The only thing I'd really call a blocker is the ascii art (especially the license question, but also maybe just removing it entirely for accessibility).

xilem/examples/virtual_cats.rs
nilsmartel: 😿 it is

xilem/examples/virtual_cats.rs
nilsmartel: Completed a suggested. Good catch.
Found something similar, my rust skills are a little rusted it seems.

xilem/examples/virtual_cats.rs
nilsmartel: I checked for a color that is more suiting and increased the size and boldness of both the text and the cat emoji

nilsmartel:
@DJMcNab thank you for upholding such quality as to check the contrast ration, even for the smallest of aspects of this codebase. Xilem has a special place in my heart as one of the few things I feel are made in spirit of great quality.

DJMcNab (approved):
Thank you! I've pushed a couple of tiny stylistic changes

PR #1354, opened 2025-08-19
Allow loading a given user's statuses

DJMcNab:
image

This has run into difficulty with teardown on the timelines. We work around this here, but a more stable solution would be good.

### DIFF
--- a/placehero/src/components.rs
+++ b/placehero/src/components.rs
@@ -29,6 +29,7 @@ fn base_status<State: 'static>(
 ) -> impl FlexSequence<State, Navigation> + use<State> {
     // TODO: This really should be Arced or something.
     let status_clone: Status = status.clone();
+    let acct_clone = status.account.acct.clone();
     // TODO: In theory, it's possible to reblog a reblog; it's not clear what happens in this case.
     debug_assert!(status.reblog.is_none(), "`base_status` can't show reblogs.");
     // We return a child list.
@@ -42,7 +43,7 @@ fn base_status<State: 'static>(
                     .text_alignment(TextAlign::Start)
                     .text_size(20.)
                     .flex(CrossAxisAlignment::Start),
-                inline_prose(status.account.username.as_str())
+                inline_prose(status.account.acct.as_str())
                     .weight(FontWeight::SEMI_LIGHT)
                     .text_alignment(TextAlign::Start)
                     .flex(CrossAxisAlignment::Start),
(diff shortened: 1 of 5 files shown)


xilem_core/src/view.rs
```diff
@@ -96,6 +96,10 @@ pub trait View<State, Action, Context: ViewPathTracker>: ViewMarker + 'static {
         view_state: &mut Self::ViewState,
         ctx: &mut Context,
         element: Mut<'_, Self::Element>,
+        // TODO: If the app state no longer exists (e.g. if we're a view over some data, and
+        // that data has been removed) what do we do?
+        // There is potential value in the app state still being available, but maybe
+        // those use cases should migrate to the environment?
```
Philipp-M, line 102: I'm actually playing with the thought of completely removing `State` from `View<State>` to just `View`
but this is far from thought through, we first need solid ways to interact with the environment, and some experience with it.

placehero/src/lib.rs
```diff
@@ -113,20 +142,28 @@ impl Placehero {
...
+                OneOf6::A(thread(show_context, context))
             } else {
                 OneOf::B(prose("Loading thread"))
             }
+        } else if self.loading_timeline {
+            // Hack: Flex allows the sized box to not take up the full size.
+            OneOf::C(flex(sized_box(spinner()).width(50.px()).height(50.px())))
+        } else if let Some(acct) = self.not_found_acct.as_ref() {
+            OneOf::D(prose(format!(
+                "Could not find account @{acct} on this server.\
```
jaredoconnell: Needs a space or new line after the period.

placehero/src/components.rs
```diff
@@ -42,14 +43,19 @@ fn base_status<State: 'static>(
...
-                inline_prose(status.account.username.as_str())
+                inline_prose(status.account.acct.as_str())
                     .weight(FontWeight::SEMI_LIGHT)
                     .text_alignment(TextAlign::Start)
                     .flex(CrossAxisAlignment::Start),
             ))
             .main_axis_alignment(MainAxisAlignment::Start)
             .gap(1.px()),
             FlexSpacer::Flex(1.0),
+            button("Open Profile", move |_| {
```
jaredoconnell, line 54: This UI design shows how long-term we're going to want to make it so this action can be done from clicking on the profile picture or the username, since that's how users on web expect the UX.

jaredoconnell (approved):
Looks good. Just one small change needed. It's awesome to see this progressing.

placehero/src/components.rs
DJMcNab: Yeah, that definitely makes sense.

PR #1355, opened 2025-08-20
Better documentation for `NewWindow`

nixonyh:
A follow up PR for #1332

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -93,6 +93,13 @@ impl NewWindow {
     }
 
     /// Set the base color of the new window.
+    ///
+    /// The base color is the color of the background which all widgets in the window draw on top of.
+    /// Masonry's current default theme assumes that this will be a very dark color for sufficient contrast.
+    /// This is most useful for apps which want to for example support light mode.
+    ///
+    /// Please note that it is not currently supported to modify this once the app is running.
+    /// This is not a fundamental limitation, and is only due to missing api design.
     pub fn with_base_color(mut self, base_color: Color) -> Self {
         self.base_color = base_color;
         self


nixon-voxell:
I want to land this first before implementing a new feature. (I'm looking at adding a builder pattern to `NewWidget`)

masonry_winit/src/event_loop_runner.rs
```diff
@@ -68,12 +68,19 @@ pub struct NewWindow {
 }
 
 impl NewWindow {
-    /// Create a new window with auto assigned [`WindowId::next()`].
+    /// A default `NewWindow`.
+    ///
+    /// Further properties of the window which will be created using
+    /// builder-style methods on this type.
+    /// See the documentation on the fields of this type for details of the parameters.
```
DJMcNab: Looking at this again, my use of the word "default" here wasn't great. Additionally, this API has changed since I wrote these docs. Perhaps something as simple as:

```suggestion
    /// Create a new window with an automatically assigned [`WindowId`].
    ///
    /// See the documentation on the fields of this type for details of the parameters.
```
is best.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -68,12 +68,19 @@ pub struct NewWindow {
...
     pub fn new(attributes: WindowAttributes, root_widget: NewWidget<dyn Widget + 'static>) -> Self {
         Self::new_with_id(WindowId::next(), attributes, root_widget)
     }
 
-    /// Create a new window with a custom assigned [`WindowId`].
+    /// Create a `NewWindow` with a custom assigned [`WindowId`].
+    ///
+    /// Use this when you need to specify a unique ID for the window, for example,
+    /// for external tracking or state management.
```
DJMcNab, line 81: It isn't clear to me what use cases there are for creating a window in Masonry Winit, but not knowing its ID (because you always need the id to do any editing of the window).

And so I think:
```rust
let id = WindowId::next();
create_window(NewWindow::new(id, ..., ...));
```
is clearer than:
```
let new_window = NewWindow::new(..., ...);
let id = new_window.id;
create_window(new_window);
```

I think this splitting of methods probably should have been combined with making `create_window` return the id of the created window.

(This shouldn't be actioned in this PR - we probably should just split this into an issue, or do it in your follow-up)

DJMcNab (approved):
Why hasn't https://github.com/linebender/xilem/pull/1332#discussion_r2278775896 also been applied here? Of the two documentation items in that comment, I think that one is the more important.

masonry_winit/src/event_loop_runner.rs
nixonyh: Noted, and I like the suggestion of returning a new id when creating a new window.

nixon-voxell:
> Is there a specific reason that [#1332 (comment)](https://github.com/linebender/xilem/pull/1332#discussion_r2278775896) hasn't also been applied here? Of the two documentation items in that comment, I think that one is the more important.

Oh, I must have missed out on that!

PR #1357, opened 2025-08-20
Xilem: Fix handling of TextInput disabled property

DJMcNab:
In current Xilem, this means that if you have a TextInput which starts disabled, you can never re-enable it.

This has impacted Placehero since #1354


jaredoconnell (approved):
You just need to remove the unused import before merging.

PR #1363, opened 2025-08-25
fix: declare_property_tuple macro uses private scope

alexmoon:
The `declare_property_tuple` macro is currently unusable outside of the `xilem` crate because it references the `PropertyTuple` trait via its module path (`$crate::property_tuple::PropertyTuple`) and that module is private. This PR changes the macro to access the trait via its public re-export in the crate root.

### DIFF
--- a/xilem/src/property_tuple.rs
+++ b/xilem/src/property_tuple.rs
@@ -109,7 +109,7 @@ macro_rules! __declare_property_tuple_loop {
             }
         }
 
-        impl $crate::property_tuple::PropertyTuple for $Props
+        impl $crate::PropertyTuple for $Props
         {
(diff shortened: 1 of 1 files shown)


Philipp-M:
Be aware that #1278 entirely removes this for the sake of a dedicated view (type) `Prop<P>` that handles all properties in Xilem (at least initially).

alexmoon:
> Be aware that #1278 entirely removes this for the sake of a dedicated view (type) `Prop<P>` that handles all properties in Xilem (at least initially).

That sounds like a better long-term solution. This change lets my custom views work on the current main branch.

DJMcNab (approved):
Thanks!

PR #1364, opened 2025-08-25
Remove app_state parameter from teardown

alexmoon:
Per the discussion in https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Incoherent.20state.20issues, the consensus seems to be to remove the `app_state` parameter from the teardown functions since the state may no longer exist or be valid at that point.

### DIFF
--- a/xilem/src/driver.rs
+++ b/xilem/src/driver.rs
@@ -184,7 +184,6 @@ where
             &mut window.view_state,
             &mut window.view_ctx,
             ctx.window_handle_and_render_root(window_id),
-            &mut self.state,
         );
         self.windows.remove(&window_id);
         ctx.close_window(window_id);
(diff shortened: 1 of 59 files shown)


DJMcNab (approved):
Thanks! It's good to see this actioned.

I agree here for two reasons.
1) This brings us to a much more easily usable state directly.
2) The things which this was previously useful for are easily compatible with the environment system (not necessarily the one which is implemented, but future extensions of it).

Some procedural notes:
1) We'll need to follow-up in Placehero, which has hacks to work around this. This should not be in this PR.
2) We need to document this somewhere; even a stub comment on `View::teardown` or a `TODO` there would be good enough for now.
3) I've reached the end of my day, and haven't read this change carefully. I know it's all straightforward, but someone does need to skim over it to check for unwanted consequences. I can do this tomorrow if needed.
4) I'd *like* to get sign-off from @Philipp-M here, to check that this isn't missing something. I'm more than happy for him to press merge on this here. It's likely to be a merge conflict magnet, so I won't block on this review, and will try to land tomorrow. We can always revert if needed.

PR #1365, opened 2025-08-25
Associate a count with each view sequence

DJMcNab:
Part of [#xilem > Mad Science: Making &#96;ViewSequence&#96; just another &#96;View&#96;](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Mad.20Science.3A.20Making.20.60ViewSequence.60.20just.20another.20.60View.60/with/535140825)

My vague forward looking plan is as follows:
1) Have `View` still available, as the main trait implemented e.g. in Xilem. This is effectively shorthand for implementing `ViewSequence` for multiple element types (e.g. `Pod<DynWidget>`, `Pod<FlexWidget>`, etc.). This is the same as it is now.
2) Change all wrapper views to implement `ViewSequence` directly. Currently, they would only return `One` child.
3) Change each bound which accepts a `View` to accept a `ViewSequence`, with counts as relevant (e.g. `memoize` would accept any count; `sized_box` would only accept `One` child; etc.). The counts would be passed through the `ViewSequence` implementations.

Steps 2 and 3 might need to happen at the same time, but they should be able to happen somewhat piecemeal (so long as we start from the outermost parts and work inwards).
`AnyView` does become more complex in this world; I don't have all the answers there.

Note that there are reasonable arguments for changing the names of types; the main trait in this world becomes `ViewSequence`, and `View` is just a shortcut for a commonly needed implementation. `View` and `Views` is clearly unviable.


xilem_core/src/sequence.rs
```diff
@@ -64,6 +64,59 @@ impl<T> Default for AppendVec<T> {
...
+    /// This sequence is known to have exactly one element.
+    One,
+    // TODO: AtMostOne - useful for Option (and e.g. `SizedBox`?).
+    /// This sequence may have any number of elements.
+    Many,
+    /// The number of elements this sequence has is not statically known.
+    ///
+    /// This is used for [`AnyView`](crate::AnyView).
+    Unknown,
+}
```
Philipp-M: Maybe we can if known also specify the exact count (if possible), or something like `AtLeast(usize)` or `AtMost(usize)`, but I haven't carefully thought about this yet, just some food for thought.

xilem_core/src/sequence.rs
```diff
@@ -64,6 +64,59 @@ impl<T> Default for AppendVec<T> {
...
+
+impl Count {
+    /// Combine the counts of multiple children.
+    pub const fn combine<const N: usize>(vals: [Self; N]) -> Self {
```
Philipp-M, line 86: Pretty cool what we can already do with const generics

xilem_core/src/sequence.rs
```diff
@@ -104,6 +157,9 @@ where
     /// [`ViewState`]: View::ViewState
     type SeqState;
 
+    /// The number of elements this sequence will manage.
```
Philipp-M: Not sure if this is accurate, as we don't *really* specify the count here.

Philipp-M (approved):
Will be interesting to see where this leads to.

xilem_core/src/sequence.rs
DJMcNab: It's a good question. Essentially, I don't think that those counts are actually useful for anything, and so would just add a lot of bookkeeping for little gain.
The only use I can think of is "split", which wants exactly two children, and that's not an especially compelling reason...

xilem_core/src/sequence.rs
DJMcNab: I've clarified things here, talking more about classes/groupings.

xilem_core/src/sequence.rs
Philipp-M: Yeah that's what I fear too, I guess when there's a real use-case for that we can still come back to this.

PR #1314, opened 2025-08-09
Checkbox: Focused indicator around entire checkbox

tannal:
Fixing #1274

PR #1366, opened 2025-08-26
Masonry: Give driver control over the Checkbox state

DJMcNab:
This makes things easier for Xilem, as it controls the source of truth for being checked. This operates somewhat similarly to `VirtualScroll`, and also addresses the issues found in https://github.com/linebender/xilem/pull/265.
I think it better reflects Masonry's low-level nature.

Working on this has uncovered another issue, which is that the tests aren't properly updated since #1314.
That PR added a large indicator, even when hovered.
However, those indicators are inflated to draw outside of the widget's area, so aren't captured here.
I don't have the motivation to fix this at the moment, but @tannal if you are willing to that would be great. A similar "flex" approach as in the new test from #1314 would be fine; just applying that to both tests (this should likelyalso use `WidgetTag`).

I'd also like to see a test of checkbox with the Padding property, because I'm not sure the rendering is correct, since `paint` doesn't take it into account. (Either for the focused indicator, or for the "box" position.
I suppose we should open an issue for that?

### DIFF
--- a/masonry/src/widgets/checkbox.rs
+++ b/masonry/src/widgets/checkbox.rs
@@ -29,7 +29,13 @@ use crate::widgets::Label;
 ///
 #[doc = include_screenshot!("checkbox_hello_checked.png", "Checkbox with checked state.")]
 ///
-/// Emits [`CheckboxToggled`] when toggled.
+/// Emits [`CheckboxToggled`] when it should toggle.
+/// Note that the checked state does not automatically toggle, and so one of
+/// the responses to a `CheckboxToggled` is to call [`Checkbox::set_checked`]
+/// on the originating widget.
+///
+/// This allows higher-level components to choose how the checkbox responds,
+/// and ensure that its value is based on their correct source of truth.
 pub struct Checkbox {
     checked: bool,
     // FIXME - Remove label child, have this widget only be a box with a checkmark.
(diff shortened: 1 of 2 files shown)


masonry/src/widgets/checkbox.rs
```diff
@@ -95,18 +101,13 @@ impl Widget for Checkbox {
         match event {
             PointerEvent::Down { .. } => {
                 ctx.capture_pointer();
-                // Checked state impacts appearance and accessibility node
-                ctx.request_render();
```
Philipp-M, line 99: I have not checked this in detail, but is this safe to do? On a skim, `capture_pointer` results in `widget_state.is_active == true`, which results in a different paint state.

masonry/src/widgets/checkbox.rs
```diff
@@ -95,18 +101,13 @@ impl Widget for Checkbox {
...
             }
             PointerEvent::Up { .. } => {
                 if ctx.is_active() && ctx.is_hovered() {
-                    self.checked = !self.checked;
-                    ctx.submit_action::<Self::Action>(CheckboxToggled(self.checked));
+                    ctx.submit_action::<Self::Action>(CheckboxToggled(!self.checked));
                     trace!("Checkbox {:?} released", ctx.widget_id());
                 }
-                // Checked state impacts appearance and accessibility node
-                ctx.request_render();
```
Philipp-M, line 109: Same here.

masonry/src/widgets/checkbox.rs
DJMcNab: I believe so, because the "active" change is handled in `fn update`. And indeed, I can see that it works - holding down the button makes the background change as expected.

DJMcNab: You are right that this is more of a drive-by change, so thanks for making me check that.

masonry/src/widgets/checkbox.rs
DJMcNab: As above.

PR #1370, opened 2025-08-28
Show the blurhash of media attachments

DJMcNab:
This uses https://github.com/whisperfish/blurhash-rs The implementation uses the "stupid" approach of recompiling the blur every "app_logic" call, but the rebuilds are rare enough that it's fine in practise

image

This also avoids needing to think about retaining resources for images, so that's positive.


DJMcNab:
> Looks good to me. It's a little laggy in Debug mode, so maybe there is room for optimization later. But in release mode it's fine.

I've added a simple performance checker (log if it takes more than 5ms). My observation is that it really struggles on the large images (e.g. https://mastodon.online/@b_n@techhub.social/114499980929122894) - ~22ms for each of those images, and that reruns happen more regularly than I expected (because it actually does a rebuild for each frame that a new item is loaded, which makes sense but which I hadn't considered the implications of properly). In release mode, the time is about 1.5ms, which is sufficiently negligible for the moment.

I think having it be this way is still fine for now - but it suggests that the blurhash images should be retained more cleverly, probably in a similar way to how the actual images will be. I think that should be a follow-up.

Issue #788, opened 2024-12-14
Add unit test for window of size 0x0


PoignardAzur:
Masonry's widgets should be able to handle cases where the available space is null and proceed without crashing.

To check that this happens, we should create a test case with a large widget tree (ideally all of the built-in widgets) and check that the tree can be created with a size of zero without crashing.

zxsa89675:
Hello, we tried to solve the issue.

This is what we did:

Modified BoxConstraints to handle zero-size constraints and added a test case in TestHarness to create a widget tree with zero size.

You can review changes in this commit: https://github.com/zxsa89675/linebender-xilem-788/commit/b2df1a45ee5a01421fb571b2475b122048548051.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

PoignardAzur:
Well... that's new.

Anyone know how to block an account from an organization?

PR #1371, opened 2025-08-28
Optimize and fix window resizing.

### LINKED ISSUE #979: When the minimize button was clicked, the program crashed.

xStrom:
The Masonry `Window` struct was being created with a `0x0` size and a `1.0` scale factor. Initial layout and render passes ran with that. Then there would be an event for scale change (but still zero size) and another pass. Then there would be a size change event and useful work would finally take place.

This PR removes all that early useless work and creates the `Window` struct immediately with the correct info.

---

A bunch of places would ask Winit for the window size, including the `render` function. At least on Windows this would cause Winit to execute a syscall. That is needless overhead, especially in a hot path, because we already store that info in the `RenderRoot` struct.

This PR exposes that size info.

---

Masonry has been suffering from crashes on Windows 11 whenever a window would be minimized or otherwise resized to zero (#979).

This PR resolves that issue. Fixes #979.

First, I refactored the redraw code so that there is a single source of truth for the behavior. There was already divergence with the first frame not doing the `AnimFrame` event and more issues could easily follow with such a duplicated approach.

The redraw refactoring also introduced on-demand surface creation. This is useful to actually solve the Windows crash issue, as now we can just delete the old surface when a zero sized resize event comes in. Whenever a reasonable size redraw happens, the surface will be recreated.

---

I have tested this as working nicely on Windows 11. Trying out crazy resizings (especially to zero!), minimizing, and suspending on other platforms would be very welcome.

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -327,6 +327,7 @@ impl<W: Widget> TestHarness<W> {
                     default_properties: Arc::new(default_props),
                     use_system_fonts: false,
                     size_policy: WindowSizePolicy::User,
+                    size: window_size,
                     scale_factor: params.scale_factor,
                     test_font: Some(data),
                 },
@@ -349,7 +350,6 @@ impl<W: Widget> TestHarness<W> {
         };
 
         // Set up the initial state, and clear invalidation flags.
-        harness.process_window_event(WindowEvent::Resize(window_size));
         harness.process_window_event(WindowEvent::EnableAccessTree);
         harness.animate_ms(0);
 
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -24,6 +24,7 @@ use masonry_core::vello::{AaConfig, AaSupport, RenderParams, Renderer, RendererO
 use tracing::{debug, info, info_span};
 use ui_events_winit::{WindowEventReducer, WindowEventTranslation};
 use winit::application::ApplicationHandler;
+use winit::dpi::PhysicalSize;
 use winit::error::EventLoopError;
 use winit::event::{DeviceEvent as WinitDeviceEvent, DeviceId, WindowEvent as WinitWindowEvent};
 use winit::event_loop::ActiveEventLoop;
@@ -125,10 +126,9 @@ impl Window {
         signal_sender: Sender<(WindowId, RenderRootSignal)>,
         default_properties: Arc<DefaultProperties>,
         base_color: Color,
+        size: PhysicalSize<u32>,
+        scale_factor: f64,
     ) -> Self {
-        // TODO: We can't know this scale factor until later?
-        let scale_factor = 1.0;
-
         Self {
             id: window_id,
             handle,
@@ -143,6 +143,7 @@ impl Window {
                     default_properties,
                     use_system_fonts: true,
                     size_policy: WindowSizePolicy::User,
+                    size,
                     scale_factor,
                     test_font: None,
                 },
@@ -354,11 +355,7 @@ impl MasonryState<'_> {
 
         self.is_suspended = false;
 
-        //  Recreate surfaces for all existing windows.
-        for (handle_id, window) in self.windows.iter() {
-            let surface = create_surface(&mut self.render_cx, window.handle.clone());
-            self.surfaces.insert(*handle_id, surface);
-        }
+        // Surfaces for existing windows will be recreated on-demand in the redraw method
 
         // Create new windows.
         if !self.new_windows.is_empty() {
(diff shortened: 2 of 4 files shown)


jaredoconnell (commented):
This looks like a good improvement. I tested it on Mac OS and didn't notice any issues. I'll test it on Windows later.

masonry_testing/src/harness.rs
```diff
@@ -349,7 +350,6 @@ impl<W: Widget> TestHarness<W> {
         };
 
         // Set up the initial state, and clear invalidation flags.
-        harness.process_window_event(WindowEvent::Resize(window_size));
```
DJMcNab, line 352: As far as I understand, the initial size being zero was an intentional decision, as a sort of passive version of #788. For tests specifically, I think that defaulting to that behaviour is possibly actually a good thing.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -458,6 +457,58 @@ impl MasonryState<'_> {
...
+        // TODO: this calculation uses wall-clock time of the paint call, which
+        // potentially has jitter.
+        //
+        // See https://github.com/linebender/druid/issues/85 for discussion.
+        let last = self.last_anim.take();
+        let elapsed = last.map(|t| now.duration_since(t)).unwrap_or_default();
+
+        window
+            .render_root
+            .handle_window_event(WindowEvent::AnimFrame(elapsed));
```
DJMcNab, line 496: Fwiw, we probably need to handle signals after an animation (see e.g. #985). That being said, everything about animations will need to change in the future anyway, and this isn't a regression here.

DJMcNab (approved):
I'm not sure about the change in the test harness, because an initial layout/render at zero size does have value, in helping to encourage greater robustness.

Otherwise, this PR looks good and works well. For some reason it is forbidden in the Wayland protocol to have a zero sized *window* (even if there will be compositor provided decorations), so I haven't been able to test it on this machine (something in the stack sets a default minimum size, and trying to work past it gives a gnarly error).
But it hasn't regressed anything, and 
I've also tested on Android (which ran into #1377), and the interaction with suspension/resumption seems to be stable.

masonry_winit/src/event_loop_runner.rs
xStrom: Makes sense, but yeah it wasn't done before either, so that'll be future work.

masonry_testing/src/harness.rs
xStrom: The initial window size in testing is controlled by `TestHarnessParams::window_size`, so if starting with zero is useful, those specific tests can set it accordingly. Indeed, that is exactly what the zero size tests added in #1232 do.

The removal of this specific event is because it's no longer needed and no longer matches actual app behavior. It would also always send the same value that the window was initialized with. (*There's probably value in emulating the behavior of the startup Winit resize event, but that would be a bit later, not at this moment in the lifecycle.*)

xStrom:
I'm especially happy to hear that Android suspend still works!

I'll go ahead and merge this. If anyone finds issues, we'll deal with those as follow-up work.

masonry_testing/src/harness.rs
DJMcNab: The point is that this change reduces coverage of zero size handling for a large number of widgets.

Now maybe that's fine, and we probably should have such tests explicitly instead. But it is a regression in what's covered.

Either way, this isn't worth worrying about too much, but it probably is worth opening an issue for "explicit zero size testing of widgets individually"

PR #1374, opened 2025-08-29
masonry: Refactor local transform as `Property`

Philipp-M:
This got a little bit larger than I would've hoped, and I'm not entirely sure about it (more code without immediate benefit, apart from making the transform a property, which possibly allows other APIs e.g. in Xilem).

It introduces a new marker trait `GlobalProperty` with `impl<P: GlobalProperty, W: Widget> HasProperty<P> for W {}` as suggested in #1278 
And a core property `Transform` which is a new type wrapper around `kurbo::Affine`, it could reasonably also be a directly `Affine` (less boilerplate for sure), but I guess a new type probably is worth it non-the-less. 

This unfortunately disallows other things I had in mind, due to orphan rules (this would result in a similar behavior as xilem_webs `OneOf`):

```rust
impl<
    A: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    B: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    C: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    D: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    E: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    F: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    G: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    H: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    I: Widget + FromDynWidget + HasProperty<P> + ?Sized,
    P: Property,
> HasProperty<P> for OneOfWidget<A, B, C, D, E, F, G, H, I>
{
}
```

So another case why I'm not sure about merging this. But as it's basically finished, it's at the very least worth opening this.

### DIFF
--- a/masonry_core/src/core/properties/mod.rs
+++ b/masonry_core/src/core/properties/mod.rs
@@ -6,7 +6,10 @@ use std::collections::HashMap;
 use std::default::Default;
 
 use crate::core::Widget;
-use crate::util::AnyMap;
+use crate::util::{AnyMap, TypeSet};
+
+mod transform;
+pub use transform::Transform;
 
 /// A marker trait that indicates that a type is intended to be used as a widget's property.
 ///
@@ -32,6 +35,9 @@ pub trait Property: Default + Send + Sync + 'static {
     /// Ideally, when const generics are stable, we'll want to use `const Default` directly in the default impl.
     fn static_default() -> &'static Self;
 }
+
+/// A marker trait for properties that can be used on *any* [`Widget`]
+pub trait GlobalProperty: Property {}
 
 // TODO - Implement Debug.
 /// A collection of [properties](Property) that a widget can be created with.
@@ -76,6 +82,8 @@ pub struct DefaultProperties {
 /// This is not used directly by Masonry Core, but is instead provided for the convenience of external crates.
 pub trait HasProperty<P: Property> {}
 
+impl<P: GlobalProperty, W: Widget> HasProperty<P> for W {}
+
 impl Properties {
     /// Create an empty collection of properties.
     pub fn new() -> Self {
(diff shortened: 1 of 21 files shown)


masonry_core/src/passes/layout.rs
```diff
@@ -212,7 +219,7 @@ pub(crate) fn place_widget(child_state: &mut WidgetState, origin: Point) {
 
     // TODO - We may want to invalidate in other cases as well
     if origin != child_state.origin {
-        child_state.transform_changed = true;
+        changed_properties.mark_as_changed::<Transform>();
```
Philipp-M, line 222: Not sure whether it's the right thing to change this. Alternatively we could leave the `transform_changed` flag within `WidgetState` additionally, and check both in the compose pass (or set `transform_changed` as well in `WidgetMut::property_changed`).

masonry_core/src/core/properties/mod.rs
```diff
@@ -111,6 +119,34 @@ impl Properties {
     }
 }
 
+/// When properties of a widget were modified this should also be reflected by [`ChangedProperties::mark_as_changed`]
```
DJMcNab, line 122:
```suggestion
/// When properties of a widget were modified this should also be reflected by [`ChangedProperties::mark_as_changed`].
```

masonry_core/src/core/properties/mod.rs
```diff
@@ -111,6 +119,34 @@ impl Properties {
...
+#[derive(Default, Debug)]
+pub struct ChangedProperties {
+    set: TypeSet,
+}
+
+impl ChangedProperties {
+    /// Mark that the property `P` has changed
```
DJMcNab, line 129:
```suggestion
    /// Mark that the property `P` has changed.
```
etc.

masonry_core/src/core/properties/transform.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use vello::kurbo::Affine;
+
+use crate::core::{GlobalProperty, Property};
+
+/// The local transform of a widget
```
DJMcNab, line 8:
```suggestion
/// The local transform of a widget.
```

masonry_core/src/passes/layout.rs
DJMcNab: I think this alternative is the better way around. My understanding is that the semantics of `prop_has_changed` are very clear, which is that they mark whether the transform has changed during the lifetime of this `WidgetMut`, right?

DJMcNab (commented):
I'm finding this a bit hard to reason about.

PR #1376, opened 2025-08-29
Enable Label properties for VariableLabel too

RagibHasin:
On Xilem side ContentColor and other Label properties are not usable with VariableLabel. This intends to fix it.


DJMcNab:
The correct fix is probably something closer to how `any_button` works (although noting that we will still need a `Widget = Label` bound on the child.

PR #1380, opened 2025-09-01
Windows-specific window attrs

RagibHasin:
Introduces support for Windows-specific `WindowAttributes` extensions.

Multi-window support in Xilem introduced some basic support for `WindowAttributes`. This did not include any platform-specific support.

Support for other platforms are still to be added.

### DIFF
--- a/xilem/src/window_options.rs
+++ b/xilem/src/window_options.rs
@@ -26,6 +26,7 @@ pub(crate) struct ReactiveWindowAttrs {
     cursor: Cursor,
     min_inner_size: Option<Size>,
     max_inner_size: Option<Size>,
+    platform_specific: PlatformSpecificReactiveWindowAttrs,
 }
 
 /// These are attributes the user can change, so we cannot make them reactive.
@@ -35,6 +36,7 @@ pub(crate) struct InitialAttrs {
     position: Option<Position>,
     // TODO: move window_icon to ReactiveWindowAttrs once the winit type implements PartialEq
     window_icon: Option<Icon>,
+    platform_specific: PlatformSpecificInitialWindowAttrs,
 }
 
 pub(crate) struct WindowCallbacks<State> {
@@ -56,11 +58,13 @@ impl<State> WindowOptions<State> {
                 cursor: Cursor::default(),
                 min_inner_size: None,
                 max_inner_size: None,
+                platform_specific: PlatformSpecificReactiveWindowAttrs::default(),
             },
             initial: InitialAttrs {
                 inner_size: None,
                 position: None,
                 window_icon: None,
+                platform_specific: PlatformSpecificInitialWindowAttrs::default(),
             },
             callbacks: WindowCallbacks::default(),
         }
@@ -134,7 +138,9 @@ impl<State> WindowOptions<State> {
         if let Some(inner_size) = self.initial.inner_size {
             attrs = attrs.with_inner_size(inner_size);
         }
-        attrs
+        self.initial
+            .platform_specific
+            .build(self.reactive.platform_specific.build(attrs))
     }
 
     pub(crate) fn rebuild(&self, prev: &Self, window: &Window) {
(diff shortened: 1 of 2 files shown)


xilem/src/window_options.rs
```diff
@@ -183,5 +193,339 @@ impl<State> WindowOptions<State> {
...
+                no_redirection_bitmap: false,
+                drag_and_drop: true,
+                class_name: "Window Class".to_string(),
+                clip_children: true,
+            }
+        }
+    }
+
+    unsafe impl Send for PlatformSpecificReactiveWindowAttrs {}
+    unsafe impl Sync for PlatformSpecificReactiveWindowAttrs {}
```
DJMcNab: Why is this:
- Needed (all the fields appear to be `Send`, but I'm not on a windows machine to check)
- Sound (if there is a non-Send field, that will be for a good reason)

This needs a safety comment explaining both.

xilem/src/window_options.rs
```diff
@@ -183,5 +193,339 @@ impl<State> WindowOptions<State> {
...
+        }
+
+        #[inline]
+        fn with_corner_preference(mut self, corners: CornerPreference) -> Self {
+            self.reactive.platform_specific.corner_preference = Some(corners);
+            self
+        }
+    }
+
+    const NONE_COLOR: Color = unsafe { std::mem::transmute(0xfffffffe_u32) };
```
DJMcNab: This seems really strange; all the `Winit` apis use `Option::<Color>::None` to mean `Color(0xfffffffe_u32)`, but you seem to be specifying it as `Some(Color(0xfffffffe_u32))`. Can you clarify why just specifying `None` doesn't work?

DJMcNab (commented):
It seems reasonable to support this. Obviously it is not acceptable to remove support for non-Windows platforms.

The docs also need to be pointing to the right things (Xilem methods rather than Winit-specific).

The whole mechanism of how the platform specific stuff is wired up isn't great, but that's not your job to fix.

xilem/src/window_options.rs
RagibHasin: These are directly copy-pasted from Winit. I should verify their necessity though.

xilem/src/window_options.rs
RagibHasin: This is also copy-pasted. Ditto.

xilem/src/window_options.rs
RagibHasin: These can be removed. I think these can be removed from Winit too. These types are `Send + Sync` normally.

xilem/src/window_options.rs
RagibHasin: This can be removed too 😄. My bad.

RagibHasin:
I would like to expose other common window attributes from Xilem in a follow-up PR, if that is desirable.

xilem/src/window_options.rs
```diff
@@ -183,5 +193,360 @@ impl<State> WindowOptions<State> {
...
+                class_name: "Window Class".to_string(),
+                clip_children: true,
+            }
+        }
+    }
+
+    /// Extension setters for Windows-specific window options.
+    pub trait WindowOptionsExtWindows {
+        /// Set an owner to the window to be created. Can be used to create a dialog box, for example.
+        // / This only works when [`WindowAttributes::with_parent_window`] isn't called or set to `None`.
```
DJMcNab, line 354: Are these intentionally commented out? Maybe we should put them in a block comment instead if so?

DJMcNab (approved):
It's not great how much code there has to be here, but since this doesn't introduce any new unsafe code, this can land without massively careful review.

Thanks!

xilem/src/window_options.rs
RagibHasin: Yes, as I intend to follow up with [this](https://github.com/linebender/xilem/pull/1380#issuecomment-3242705139) and then we could uncomment and fix them.

Issue #1373, opened 2025-08-29
How to use virtual_scroll correctly?


oleid:
I've been toying with xilem recently and really like it a lot.
What I stumbled uppon, however, was the `virtual_scroll`. 
Having had  a look at placehero and the http_cats example, I thought using `virtual_scroll` would be simple enough. Yet, whenever I place something _below_ a `virtual_scroll`, apparently things are no longer visible.

This is an example of what I'm talking about:

https://github.com/linebender/xilem/compare/main...oleid:xilem:tests/lists-virtual-scroll

In this example, the button below `virtual_scroll` is no longer visible. I would have assumed, that `virtual_scroll` would just consume the space between both buttons. Is this a bug or am I holding it wrong?

jakobhellermann:
I think
```rs
flex(
  button,
  virtual_scroll.flex(1.0),
  button,
)
```
may work.

PR #1381, opened 2025-09-02
Add some basic docs on our flex behaviour

DJMcNab:
Fixes https://github.com/linebender/xilem/issues/1373

### DIFF
--- a/xilem/src/view/flex.rs
+++ b/xilem/src/view/flex.rs
@@ -39,6 +39,10 @@ use crate::{AnyWidgetView, Pod, ViewCtx, WidgetView};
 /// the proportion of remaining space (after all "non-flex" children are laid out) specified
 /// by its flex factor.
 ///
+/// When flexible children cannot expand to their allotted space, this space is allocated into the space
+/// between the children; the exact semantics are determined by the [`MainAxisAlignment`](Flex::main_axis_alignment).
+/// This is also discussed in the docs for the underlying [`Flex`](widgets::Flex) widget.
+///
 /// # Example
 /// ```rust,no_run
 /// use xilem::masonry::properties::types::{AsUnit, CrossAxisAlignment, MainAxisAlignment};


masonry/src/widgets/flex.rs
```diff
@@ -25,6 +25,23 @@ use crate::util::{debug_panic, fill, include_screenshot, stroke};
...
+/// In the web model, this is equivalent to the default `flex` being `none` (on the web, this is instead `auto`).
+/// This can lead to surprising results, including later siblings of the expanded child being pushed off-screen.
+/// A general rule of thumb is to set a flex factor on all "large" children in the flex axis, especially
+/// portals, sized boxes, and text inputs (in horizontal flex areas).
+/// That is, any item which needs to shrink to fit within the viewport should have a
+/// flex factor set.
+///
+/// There is also no support for flex grow or flex shrink; instead, each flexible child takes up
+/// the proportion of remaining space (after all "non-flex" children are laid out) specified
+/// by its flex factor.
```
jaredoconnell, line 43: We may want to also note that there is only a single pass for expansion, so if a widget cannot expand to the full width, the remaining space is distributed between widgets instead of to the remaining widgets themselves.

DJMcNab: Would you mind rewording this into a concrete suggestion I can apply?
I'm finding it quite hard to understand what exactly you're wanting to explain here, which suggests that it might need another draft! (In particular, which are the "remaining widgets")

jaredoconnell: Sure. It's a niche thing, but users will encounter it.

Here is something that can be added:
> In the web flex algorithm, if a widget cannot expand to its target flex size, that reaming space is distributed to the other flex widgets recursively. However, there is no support for that in this widget presently. Instead, if a widget cannot expand to its target size, that difference is distributed to the space between widgets according to `MainAxisAlignment`.

DJMcNab: I've added something based on this. As this PR is only aiming to add some cursory docs, I'm going to land this now.

PR #1386, opened 2025-09-12
Text area improvements

ricvelozo:
* Change the selection color when the window is unfocused
* Add properties to customize the colors of selection and caret
* Add animation to blink the caret

### DIFF
--- a/masonry/src/properties/selection.rs
+++ b/masonry/src/properties/selection.rs
@@ -8,7 +8,7 @@ use crate::peniko::color::{AlphaColor, Srgb};
 
 /// The color of a [`TextArea`]'s cursor.
 ///
-/// [`TextArea`]: crate::widgets::TextArea`
+/// [`TextArea`]: crate::widgets::TextArea
 #[expect(missing_docs, reason = "field names are self-descriptive")]
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct CaretColor {
@@ -26,7 +26,10 @@ impl Property for CaretColor {
 
 /// The background color of a [`TextArea`]'s selection.
 ///
-/// [`TextArea`]: crate::widgets::TextArea`
+/// Need to contrast the [`ContentColor`].
+///
+/// [`ContentColor`]: crate::properties::ContentColor
+/// [`TextArea`]: crate::widgets::TextArea
 #[expect(missing_docs, reason = "field names are self-descriptive")]
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct SelectionColor {
@@ -44,7 +47,10 @@ impl Property for SelectionColor {
 
 /// The background color of a [`TextArea`]'s selection when unfocused.
 ///
-/// [`TextArea`]: crate::widgets::TextArea`
+/// Need to contrast the [`ContentColor`].
+///
+/// [`ContentColor`]: crate::properties::ContentColor
+/// [`TextArea`]: crate::widgets::TextArea
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct UnfocusedSelectionColor(pub SelectionColor);
 
--- a/masonry/src/widgets/text_area.rs
+++ b/masonry/src/widgets/text_area.rs
@@ -778,8 +778,11 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
     fn register_children(&mut self, _ctx: &mut RegisterCtx<'_>) {}
 
     fn property_changed(&mut self, ctx: &mut UpdateCtx<'_>, property_type: TypeId) {
+        CaretColor::prop_changed(ctx, property_type);
         ContentColor::prop_changed(ctx, property_type);
         DisabledContentColor::prop_changed(ctx, property_type);
+        SelectionColor::prop_changed(ctx, property_type);
+        UnfocusedSelectionColor::prop_changed(ctx, property_type);
     }
 
     fn update(&mut self, ctx: &mut UpdateCtx<'_>, _props: &mut PropertiesMut<'_>, event: &Update) {
@@ -852,6 +855,7 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
             self.editor.try_layout().unwrap()
         };
         if ctx.is_focus_target() {
+            let caret_color = props.get::<CaretColor>().color;
             let selection_color = if ctx.is_window_focused() {
                 props.get::<SelectionColor>().color
             } else {
@@ -867,13 +871,7 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
                 );
             }
             if let Some(cursor) = self.editor.cursor_geometry(1.5) {
-                scene.fill(
-                    Fill::NonZero,
-                    Affine::IDENTITY,
-                    props.get::<CaretColor>().color,
-                    None,
-                    &cursor,
-                );
+                scene.fill(Fill::NonZero, Affine::IDENTITY, caret_color, None, &cursor);
             };
         }
 


ricvelozo:
I'm having a hard time figuring out how to change the text color in the selection range, so this just change the background color for now. 🇧🇷 

Captura de tela de 2025-09-13 17-20-19

It's ready for review. I can send another PR later when I figure out how the `render_text()` and `BrushIndex` works.

jaredoconnell (commented):
There are a lot of implementation details in how this is implemented, so I'll not approve or deny this PR at this time. I'll let other reviewers take a look.

I think that the addition of a cursor blink is very good, and improves the feel of the project.

The timing felt off. 600ms feels slow to me. My system defaults to 500ms.
At a minimum I would use a constant for the blink rate and have all the calculations use that as opposed to using hard-coded values.

Long-term we can use the system value to get the system-wide cursor blink rate.

ricvelozo:
> There are a lot of implementation details in how this is implemented, so I'll not approve or deny this PR at this time. I'll let other reviewers take a look.
> 
> I think that the addition of a cursor blink is very good, and improves the feel of the project.
> 
> The timing felt off. 600ms feels slow to me. My system defaults to 500ms. At a minimum I would use a constant for the blink rate and have all the calculations use that as opposed to using hard-coded values.
> 
> Long-term we can use the system value to get the system-wide cursor blink rate.

Yes, it will be cool when Xilem can subscribe to user preferences like accent color, color scheme, etc.

ricvelozo:
Something I forgot is the blink animation should stop after some time in some systems. In GNOME it stops after 10s, and in Android it doesn't at all.

Also, in GNOME the interval is 600ms, really. And in Firefox it is very slow and stops after 5 times.

masonry/src/properties/selection.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+/// The background color of a [`TextArea`]'s selection.
+///
+/// [`TextArea`]: crate::widgets::TextArea`
+#[expect(missing_docs, reason = "field names are self-descriptive")]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct SelectionColor {
+    pub color: AlphaColor<Srgb>,
```
DJMcNab, line 36: We should specify that this must contrast with the `ContentColor` used.

masonry/src/widgets/text_area.rs
```diff
@@ -418,6 +424,20 @@ pub enum TextAction {
...
+        _props: &mut PropertiesMut<'_>,
+        interval: u64,
+    ) {
+        self.t += (interval as f64) * 1e-9;
+        if self.t >= 1.2 {
+            self.t = self.t.rem_euclid(1.2);
+        }
+        ctx.request_anim_frame();
+        ctx.request_paint_only();
+    }
```
DJMcNab: There are a few things about this implementation I'm not sure about. In particular, I'm not especially pleased that it:

- Polls to see when the time changes (see #792, which would be a great PR to adopt if you're looking for something to work on next). This doesn't block getting cursors blink landed, to be clear.
- Runs even when the cursor isn't visible (i.e. the text area is not focused)
- Requests a repaint even when the cursor's "shown status" wouldn't have changed.
- Uses floats, but only ever as a whole number multiple of nanoseconds.

What do you think of splitting the cursor blink feature into its own PR, so we can review that separately. It's pretty orthogonal to the rest of this change, and is a lot subtler.

masonry/src/widgets/text_area.rs
```diff
@@ -848,23 +872,28 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+                    selection_color,
                     None,
                     &rect,
                 );
             }
-            if let Some(cursor) = self.editor.cursor_geometry(1.5) {
-                // TODO: Make configurable
+            if let Some(cursor) = self.editor.cursor_geometry(1.5)
+                && window_is_focused
+                && self.t <= 0.6
```
DJMcNab: As mentioned, we should have a comment here (the comment in https://github.com/linebender/parley/pull/152 is easy to copy). But as noted, I'd rather this were moved to its own PR; treat this as pre-review for that PR.

DJMcNab: I'd also slightly prefer us to use 0.5secs here, but that's only for consistency with https://github.com/linebender/parley/pull/152.

masonry/src/widgets/text_area.rs
```diff
@@ -848,23 +872,28 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
...
+            {
                 scene.fill(
                     Fill::NonZero,
                     Affine::IDENTITY,
-                    palette::css::WHITE,
+                    props.get::<CaretColor>().color,
```
DJMcNab: Our style is for all `props.get` calls to be on their own line (i.e. `let caret_color = props.get::<CatetColor>()`) . I believe this is to make it easier to validate that `properties_changed` implementation.
It isn't clear to me that this is entirely necessary, but it would be better to maintain that for now.

masonry/src/widgets/text_area.rs
DJMcNab, line 1: The calls to `prop_changed` for the new property types is missing, I believe.

xilem/src/view/text_input.rs
```diff
@@ -70,6 +73,30 @@ impl<State: 'static, Action: 'static> TextInput<State, Action> {
...
+    /// Set the selection's color when the window is unfocused.
+    ///
+    /// This overwrites the default `UnfocusedSelectionColor` property for the inner `TextArea` widget.
+    pub fn unfocused_selection_color(
+        self,
+        color: Color,
+    ) -> Prop<UnfocusedSelectionColor, Self, State, Action> {
+        self.prop(UnfocusedSelectionColor(SelectionColor { color }))
+    }
+
```
DJMcNab, line 99: I *think* these also should be made available on `Prose`. However, it's possible that I'm missing some detail which makes that worth skipping for now.

masonry/src/widgets/text_input.rs
DJMcNab, line 1: I think we also need to do the same thing for `Prose`. This also applies on the Xilem side. This can be deferred to after this PR.

masonry/src/theme.rs
```diff
@@ -87,17 +88,37 @@ pub fn default_property_set() -> DefaultProperties {
...
     properties.insert::<TextArea<true>, _>(ContentColor::new(TEXT_COLOR));
     properties
         .insert::<TextArea<true>, _>(DisabledContentColor(ContentColor::new(DISABLED_TEXT_COLOR)));
+    properties.insert::<TextArea<true>, _>(CaretColor { color: TEXT_COLOR });
+    properties.insert::<TextArea<true>, _>(SelectionColor {
+        color: ACCENT_COLOR,
+    });
+    properties.insert::<TextArea<true>, _>(UnfocusedSelectionColor(SelectionColor {
+        color: DISABLED_TEXT_COLOR,
+    }));
```
DJMcNab, line 121: Do these defaults not need to be on `TextInput`, not `TextArea`? (as they will get immediately overwritten to the static defaults by the `WidgetAdded` code)

masonry/src/widgets/text_area.rs
DJMcNab, line 1: Would you be willing to add a new screenshot test for the unfocused functionality? If the test harness doesn't support that yet, then we can skip it.

masonry/src/theme.rs
ricvelozo: I thought that `TextArea` could be used independently of `TextInput`, like a multiline version.

ricvelozo:
I'm uncertain about the properties of `TextArea` in `Prose` & `TextInput`.

* Are they supposed to be all added to both `Prose` & `TextInput`, including `ContentColor` (and have theme defaults)?
* Do they need a equivalent method in Xilem, or `.prop()` is enough?

Currently there is just a property for the placeholder label in `TextInput`.

masonry/src/theme.rs
DJMcNab: `TextArea` is "morally" a private implementation detail of both `TextInput` and `Prose` (hence the ugly generic, for example). This isn't well documented, and is something we might need to redesign at some point.

In this case, we could apply them to both `TextArea<true>` and `TextInput` (which I see you've now pushed). ~~The point I was making is that if you have a `TextInput`, these default which you're applying wouldn't actually work, which is not an ideal situation.~~

For clarity, `TextInput` does support multi-line text.

masonry/src/widgets/text_area.rs
DJMcNab: I've unresolved this comment as it still remains as a necessary follow-up.

DJMcNab (approved):
Thank you! Thanks for addressing the review comments as well.

> I'm uncertain about the properties of `TextArea` in `Prose` & `TextInput`.
> 
>     * Are they supposed to be all added to both `Prose` & `TextInput`, including `ContentColor` (and have theme defaults)?
>     * Do they need a equivalent method in Xilem, or `.prop()` is enough?
> 
> Currently there is just a property for the placeholder label in `TextInput`.

Right, this is my fault - I missed that `Prose` currently doesn't have the equivalent of the code in `TextInput` to propagate the properties down. So, for now, we can ignore `Prose` entirely (as this PR already does); in the long term, that's obviously not ideal, but it's not making things worse. Thanks!

masonry/src/widgets/text_area.rs
ricvelozo: Oops! I confused it with the other PR, sorry!

ricvelozo: Addressed in #1401.

PR #1388, opened 2025-09-15
Simplify multi-window code

RagibHasin:
Also adds window base-color in Xilem.

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -152,7 +152,7 @@ impl Window {
         }
     }
 
-    /// Access the [`WindowHandle`] of this window.
+    /// Access the the underlying [Winit window](WindowHandle) of this window.
     pub fn handle(&self) -> &WindowHandle {
         &self.handle
     }
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -53,7 +53,7 @@ pub struct NewWindow {
     /// The id is set by the App, and can be created using the [`WindowId::next()`] method.
     ///
     /// Once the window is created, it can be accessed using this `id` through the
-    /// [`DriverCtx::window_handle()`] method.
+    /// [`DriverCtx::window()`] method.
     pub id: WindowId,
     /// Window attributes for the winit's [`Window`].
     ///
@@ -108,7 +108,7 @@ impl NewWindow {
 }
 
 /// Per-Window state
-pub(crate) struct Window {
+pub struct Window {
     id: WindowId,
     pub(crate) handle: Arc<WindowHandle>,
     pub(crate) accesskit_adapter: Adapter,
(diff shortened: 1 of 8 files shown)


xilem/src/window_view.rs
```diff
@@ -2,33 +2,54 @@
...
+    pub(crate) id: WindowId,
     options: WindowOptions<State>,
     root_widget_view: Box<AnyWidgetView<State, ()>>,
+    /// The base color of the window.
+    base_color: Color,
+}
+
+/// A view representing a window.
+///
+/// `id` can be created using the [`WindowId::next()`] method.
```
DJMcNab: We need to specify that the id *must* be the same each frame for the same window (and so you should store it in your state somewhere)

masonry_winit/src/event_loop_runner.rs
```diff
@@ -151,6 +151,21 @@ impl Window {
             base_color,
         }
     }
+
+    /// Access the [`WindowHandle`] of this window.
```
DJMcNab: We can be more specific here (e.g. something like "access the underlying Winit window"). The docs previously assumed that you know what `WindowHandle` means, so now is an opportunity to do better.

This was despite the fact that `WindowHandle` isn't even a public re-export!

masonry_winit/src/event_loop_runner.rs
```diff
@@ -151,6 +151,21 @@ impl Window {
...
+    pub fn handle(&self) -> &WindowHandle {
```
DJMcNab: In fact, how about we just make this:

```suggestion
    pub fn handle(&self) -> &winit::window::Window {
```

(Whether we replace `WindowHandle` with that in the rest of this crate is up to you, as it would be an internal only change; I'd lean slightly towards it, but I haven't fully what impact that has on the entire readability)

xilem/src/window_view.rs
```diff
@@ -2,33 +2,54 @@
...
+pub fn window<V: WidgetView<State>, State: 'static>(
+    id: WindowId,
+    title: impl Into<String>,
+    root_view: V,
+) -> WindowView<State> {
+    WindowView {
+        id,
+        options: WindowOptions::new(title),
+        root_widget_view: root_view.boxed(),
+        base_color: Color::BLACK,
```
DJMcNab, line 36: I'm probably missing it, but I can't actually see where you overwrite this.
Is this being used in any of the examples?

DJMcNab (commented):
This broadly looks good; there are just a few details to iron out. Thanks!

xilem/src/window_view.rs
RagibHasin: You mean `base_color`? It is in [here](https://github.com/RagibHasin/xilem/blob/0ccb8ec6827e313389868a66337ad28b5c1d2d4b/xilem/src/window_view.rs#L90-L92).

RagibHasin: Oh, yes... 😬

masonry_winit/src/event_loop_runner.rs
RagibHasin: It would be `winit::window::Window` not `winit::Window`, which is far too long a name for such frequently used type. I say we should leave it as is for now.

DJMcNab: Sure; so long as we rename it here, that's fine. Something like `WinitWindow` might work better, but as I say as an internal change, it isn't that important.

RagibHasin: Rustdoc does not expand non-public re-exports? I think it should. Then we could keep using our concise name.

xilem/src/window_view.rs
DJMcNab: It would be nice to use this in an example (e.g. the one it was removed from in #992)

PR #1393, opened 2025-09-20
Fix blitting on a transparent window

RagibHasin:


DJMcNab (changes requested):
Discussion is happening in [#masonry > Transparency woes](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Transparency.20woes/with/540602611), but we shouldn't land this as-is (it's basically trading off being correct on some platforms with being correct on a different set of platforms).

We should aim to make sure this works properly, by handling this robustly (using https://docs.rs/wgpu/latest/wgpu/struct.SurfaceCapabilities.html#structfield.alpha_modes). To be clear, I really don't like that this is so hard to get right, but I believe that it's achievable.

masonry_winit/src/vello_util.rs
```diff
@@ -0,0 +1,269 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Simple helpers for managing wgpu state and surfaces.
```
DJMcNab, line 4: We should document that this is taken from Vello's utils module.

masonry_winit/src/vello_util.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        let device_handle = &self.devices[dev_id];
+        let capabilities = surface.get_capabilities(&device_handle.adapter);
+        let format = capabilities
+            .formats
+            .into_iter()
+            .find(|it| matches!(it, TextureFormat::Rgba8Unorm | TextureFormat::Bgra8Unorm))
+            .ok_or(Error::UnsupportedSurfaceFormat)?;
+
+        const PREMUL_BELND_STATE: BlendState = BlendState {
```
DJMcNab: Can you also submit `belnd` to https://github.com/crate-ci/typos?

masonry_winit/src/vello_util.rs
```diff
@@ -0,0 +1,269 @@
...
+            let texture_blitter =
+                if cfg!(windows) && device_handle.adapter.get_info().name.contains("AMD") {
+                    tracing::info!(
+                        "on Windows with AMD GPUs use premultiplied blitting even on opaque surface"
+                    );
+                    TextureBlitterBuilder::new(&device_handle.device, format)
+                        .blend_state(PREMUL_BELND_STATE)
+                        .build()
+                } else {
+                    TextureBlitter::new(&device_handle.device, format)
```
DJMcNab, line 123: We *probably* should differentiate between "opaque" and "inherit" here, but that isn't blocking (e.g. give an `info` for inherit that we're currently assuming postmultiplied, but have the warning I mention in the other comment for opaque).
Can we have a TODO comment for this?

DJMcNab, line 123: Can we also have a TODO comment for a warning if the user uses a transparent base colour their machine doesn't support a non-opaque alpha compositing mode (where of course we pretend it does for the AMD hack)?

masonry_winit/src/vello_util.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        const PREMUL_BELND_STATE: BlendState = BlendState {
+            alpha: BlendComponent::REPLACE,
+            color: BlendComponent {
+                src_factor: BlendFactor::SrcAlpha,
+                dst_factor: BlendFactor::Zero,
+                operation: wgpu::BlendOperation::Add,
+            },
+        };
+        let (alpha_mode, blitter) = if capabilities
```
DJMcNab, line 93: I think technically if the output is known to be opaque (i.e. if the `window.set_transparency` is false), we should prefer `CompositeAlphaMode::Opaque`; I think it could be better for power usage, for example.

I don't think this blocks this PR, though, but we probably want at least a TODO comment.

DJMcNab, line 93: I'd also like a warning for a transparent base colour without the `with_transparent(true)` window attribute set. This can be a TODO comment.

DJMcNab (approved):
I'd like to see at least one example using a translucent base color before landing this. I made to_do_list do this, and it works for me now.

I have several suggested TODO comments. It might be worth opening an issue for all of them as a group, e.g. "transparency tracking issue". The "use `Opaque` for power efficiency" and "warning if trying to use transparency without telling Winit" are probably the most important of those.

Approval conditional on comments being actioned, especially the typo fix.

masonry_winit/src/vello_util.rs
RagibHasin: It would require that base color setting code query for device capabilities. Though it won't hurt to add a todo for this I suppose.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -415,6 +415,14 @@ impl MasonryState<'_> {
             return;
         }
 
+        // TODO: move this check to modification of base_color once winit exposes window transparency state
+        if !new_window.attributes.transparent && new_window.base_color.components[3] != 0. {
```
DJMcNab: Isn't opaque `1.0`, not zero?

masonry_winit/src/event_loop_runner.rs
```diff
@@ -415,6 +415,14 @@ impl MasonryState<'_> {
...
+            tracing::warn!(
+                window_id = ?new_window.id,
+                "opaque window with non-opaque base color"
```
DJMcNab:
```suggestion
                "New window with non-opaque base color doesn't support transparency - \
                you should call `.with_transparent(true)` on the new window's `WindowAttributes`."
```

xilem/examples/transparent_window.rs
```diff
@@ -0,0 +1,59 @@
...
+use masonry::properties::types::AsUnit;
+use winit::error::EventLoopError;
+use xilem::{Color, EventLoop, WindowId, WindowView, Xilem, style::Style, view::*};
+
+#[derive(Debug, Clone)]
+struct AppState {
+    opacity_level: u8,
+
+    window_id: WindowId,
+    is_running: bool,
```
DJMcNab, line 15: We're missing the `on_close` handler to set `is_running` to false.

xilem/examples/transparent_window.rs
```diff
@@ -0,0 +1,59 @@
...
+
+//! Showcase a trasnsparent window.
+
+use masonry::properties::types::AsUnit;
+use winit::error::EventLoopError;
+use xilem::{Color, EventLoop, WindowId, WindowView, Xilem, style::Style, view::*};
+
+#[derive(Debug, Clone)]
+struct AppState {
+    opacity_level: u8,
```
DJMcNab: I'd *probably* do this as 0-255 scale, even if it's still with jumps of 64. I do realise that you hit a fencepost problem, in that it's possible to hit each of `{0, 63, 64, 127, 128, 191, 192, 255}` in that proposal, depending on which direction you go in.
Alternatively, doing it as floats, of `0.0` to `1.0` with jumps of `0.25`.

Either of these is a much more common mapping, which lowers the cognitive load of this example.

This isn't blocking, just a thought for consideration.

RagibHasin: I went with the float approach.

PR #1394, opened 2025-09-22
Fixup `include_screenshot` to fallback to the online image

### LINKED ISSUE #851: Integrate screenshots in documentation

DJMcNab:
Follow-up to https://github.com/linebender/xilem/pull/1391.

This uses the technique discussed in https://blog.sentry.io/fallbacks-for-http-404-images-in-html-and-javascript/#image-fallbacks-in-html, which seems to work well.

Fixes #851, in a way which is robust to all realistic scenarios (including local building of the docs!).

I've also added a whole pile of (internal) docs on this macro.

### DIFF
--- a/masonry_core/src/util.rs
+++ b/masonry_core/src/util.rs
@@ -129,15 +129,61 @@ pub fn get_debug_color(id: u64) -> Color {
 
 pub use crate::include_screenshot;
 
+/// Markdown input to display a screenshot from the current crate's `screenshots` directory.
+///
+/// This can be added to docs as follows:
+///
+/// ```rust,ignore
+/// /// Some docs here.
+/// ///
+/// #[doc = include_screenshot!("button_hello.png", "Button with text label.")]
+/// ```
+///
+/// The caption should have a full-stop at the end, as it's being used as alt-text.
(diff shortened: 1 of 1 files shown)


masonry_core/src/util.rs
```diff
@@ -129,15 +129,61 @@ pub fn get_debug_color(id: u64) -> Color {
...
+                // Ideally, the "base URL" would be customisable, so end-users could use this macro too.x
+                // The `v` is because of our tag name convention.
+                "(https://raw.githubusercontent.com/linebender/xilem/v", env!("CARGO_PKG_VERSION"), "/", env!("CARGO_PKG_NAME"), "/screenshots/", $path,
+                ")",
+                "\n</object>"
         )
+        // concat!(
+        //     " ![", $($caption,)? "]",
+        //     "(", env!("CARGO_MANIFEST_DIR"), "/screenshots/", $path, ")",
+        // )
```
PoignardAzur: Leftover comment.

PR #1397, opened 2025-09-23
Animate the insertion caret (blink)

ricvelozo:
Split-out from #1386.

I'll address the review comments later.


ricvelozo:
Ready for review, @DJMcNab.

masonry/src/widgets/text_area.rs
```diff
@@ -476,6 +529,12 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
         _props: &mut PropertiesMut<'_>,
         event: &TextEvent,
     ) {
+        // Reset the blink animation.
+        self.anim_prev_interval = 0;
+        self.anim_elapsed = 0;
+        ctx.request_anim_frame();
+        ctx.request_paint_only();
```
DJMcNab: This being here leads to some slightly weird behaviour - namely, pressing something like <kbd>ctrl</kbd> leads to the animation being reset.

I don't think that's necessarily a huge problem, though. Certainly in a first draft, that's fine to ignore.

masonry/src/widgets/text_area.rs
```diff
@@ -674,7 +733,11 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
                 }
             }
 
-            TextEvent::WindowFocusChange(_) => {
+            TextEvent::WindowFocusChange(focused) => {
+                if self.anim_cursor_visible != *focused {
+                    self.anim_cursor_visible = *focused;
+                    ctx.request_anim_frame();
+                }
```
DJMcNab: This block is a bit confusing, given the "reset the blink animation" section above.
To make this clearer, I've suggested moving the `false` case to `paint`. If we do that, I think that this block can safely be removed entirely.
Were you seeing a case where the animation frame happens after paint?

masonry/src/widgets/text_area.rs
```diff
@@ -870,7 +933,9 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
                     &rect,
                 );
             }
-            if let Some(cursor) = self.editor.cursor_geometry(1.5) {
+            if let Some(cursor) = self.editor.cursor_geometry(1.5)
+                && self.anim_cursor_visible
```
DJMcNab, line 940:
```suggestion
                && self.anim_cursor_visible
                && ctx.is_window_focused()
```

masonry/src/widgets/text_area.rs
```diff
@@ -420,6 +432,47 @@ pub enum TextAction {
...
     type Action = TextAction;
 
+    fn on_anim_frame(
+        &mut self,
+        ctx: &mut UpdateCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        interval: u64,
+    ) {
+        // TODO: for real world use, this should be reading from the system settings
+        const CURSOR_BLINK_TIME: u64 = 1000; // ms
```
DJMcNab: Maybe something like
```suggestion
        /// The time for a complete blink cycle, in milliseconds.
        /// For the first half (i.e. currently 0.5s) the cursor is shown, and for the second half, it is hidden.
        const CURSOR_BLINK_TIME: u64 = 1000; // ms
```
would prime readers of this code more clearly about what it's doing/what the parameter represents exactly.

I'd also propose moving the comment to after these items - and rewording it to something like:
```rust
        // TODO: These should be reading from the system settings, but we currently
        // aren't aware of a robust way to read that cross-platform.
```

masonry/src/widgets/text_area.rs
```diff
@@ -420,6 +432,47 @@ pub enum TextAction {
...
+        const CURSOR_BLINK_TIMEOUT: u64 = 10_000; // 10 seconds
```
DJMcNab, line 448:
```suggestion
        /// The timeout, in milliseconds, after which the cursor will stop blinking (i.e. stay solid).
        const CURSOR_BLINK_TIMEOUT: u64 = 10_000; // 10 seconds
```

DJMcNab (commented):
We also need to handle the animation starting/stopping when moving the cursor due to the mouse. (E.g. play around in the mason example with a mouse to see the current jank).

Otherwise, this is looking really good. There are a few small things I'd like to change, but this is very close to landing. Thanks so much for your patience; review has been slow as I was ill at the end of last week.

masonry/src/widgets/text_area.rs
ricvelozo: I copied the GTK 4 behavior here as well; I don't know how it works on Windows.

DJMcNab: I don't know; it's different behaviour to Firefox on my KDE system, which is why I called it out.

It's definitely a corner case, so as I say, not worth worrying about.

ricvelozo:
> We also need to handle the animation starting/stopping when moving the cursor due to the mouse. (E.g. play around in the mason example with a mouse to see the current jank).
> 
> Otherwise, this is looking really good. There are a few small things I'd like to change, but this is very close to landing. Thanks so much for your patience; review has been slow as I was ill at the end of last week.

I couldn't see the jank  (or I didn't understand haha):

https://github.com/user-attachments/assets/831be2e6-7ea9-4402-a5cb-c364f9f22e85

masonry/src/widgets/text_area.rs
ricvelozo: I think this became unnecessary in the rebase, I don't remember.

DJMcNab (approved):
Other than the blink not being handled after a mouse move for `prose`, this is looking good. The behaviour of the cursor blink during a mouse move is a fair bit more complicated still, so I think that leaving it for a follow-up is actually better.

To clarify what I mean, you can use:
```sh
cargo run --example mason
```
If you then click into the big block of lorem ipsum, then see that the cursor sometimes disappears.

Let me know when you're happy for us to merge this :)

The docs job is tracked in [#masonry > CI: Docs failing because of macros](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/CI.3A.20Docs.20failing.20because.20of.20macros/with/542427623), but isn't a blocker.

ricvelozo:
> I don't think that is the `mason` example; I was pointing you to an example with a `prose`; if you click into that with a mouse, the animation just doesn't work right at all (e.g. the blink not updating)
> 
> The reason it's not an issue for text input is that we send IME events whenever the cursor moves, which re-syncs this.

I see now. If any `TextInput` get focus first, the `Prose` will work too. Changing `TextArea::accepts_text_input()` to unconditionally return `true` fix this. Also, `Prose` don't accept focus by tabbing, but will accept by clicking even if `TextArea::accepts_focus()` returns false.

As a reference, on GTK the caret will not blink if the "text area" is immutable to convey that to the user, but will accept focus (tab and arrows).

ricvelozo:
GNOME Text Editor:

https://github.com/user-attachments/assets/22db8c12-5aa6-4bf9-9ad2-9c816558d36a

ricvelozo:
TLDR: I think this PR can be merged as is, and we discuss changing the behavior of `TextArea<false>` / IME in another PR.

DJMcNab:
> TLDR: I think this PR can be merged as is, and we discuss changing the behavior of `TextArea<false>` / IME in another PR.

:+1: I'll rebase this PR now.

FWIW, I believe that your analysis of the correct way to fix this is wrong, and I stand by my prior analysis. But I do agree that we should land this change anyway, and I also agree that non-editable text does want different behaviour here.

PR #1398, opened 2025-09-24
Follow-up on Slider widget

RagibHasin:
* Use properties system to style a `Slider`
* Rearrange code to adhere to Masonry customs

### DIFF
--- a/xilem/examples/slider_demo.rs
+++ b/xilem/examples/slider_demo.rs
@@ -10,7 +10,7 @@
 
 use masonry::peniko::Color;
 use masonry::properties::{
-    Background,
+    Background, BarColor, ThumbColor, ThumbRadius,
     types::{AsUnit, CrossAxisAlignment},
 };
 use winit::dpi::LogicalSize;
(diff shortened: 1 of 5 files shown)


xilem/examples/slider_demo.rs
```diff
@@ -117,9 +117,9 @@ fn app_logic(state: &mut AppState) -> impl WidgetView<AppState> + use<> {
                 })
                 .step(5.0)
                 .disabled(!state.use_transparency)
-                .active_track_color(Color::from_rgb8(0x78, 0x71, 0x6c))
-                .thumb_color(Color::WHITE)
-                .thumb_radius(10.0),
+                .prop(BarColor(Color::from_rgb8(0x78, 0x71, 0x6c)))
+                .prop(ThumbColor(Color::WHITE))
+                .prop(ThumbRadius(10.0)),
```
Philipp-M, line 122: As we currently don't use `.prop` anywhere else in the examples, I wonder whether this should be abstracted in `Style`. Though either is fine for me.

Philipp-M (approved):
Thanks, nice improvement!

xilem/examples/slider_demo.rs
RagibHasin: OTOH, I think the `.prop` method should have a demo in example 😄.

Anyway, I don't like the currently available customization options very much. I think the whole thumb should be customizable, shape and all. As such I am against adding these style abstractions to `Style`.

PR #1400, opened 2025-09-24
Xilem: Add an example for a scrollable textbox.

Cyncrovee:
This file is meant to be an example of how to create a `textbox` with a buffer handled by the app state, update that `textbox` with user input, as well as a simple example of using a `virtual_scroll`.


Cyncrovee:
You may want to add license information at the top like the other examples, I wasn't sure if I'm meant to do that in the pull request. But for the sake of clarity I do allow this code to be used for the Xilem crate and repository.

Here is what I'm talking about if you want to copy/paste:
```

//! A simple example of using virtual_scroll and a textbox
```

Cyncrovee:
I have found an issue with the scrolling, in which it will always try to scroll to the bottom when inputting text into the `textbox`, I'm unsure if this is an issue with the `virtual_scroll` itself or how I've implemented it.

xilem/examples/scrollable_textbox.rs
```diff
@@ -0,0 +1,39 @@
...
+impl Default for State {
+    fn default() -> Self {
+        State {
+            buffer: "".to_string(),
+        }
+    }
+}
+
+fn logic(_state: &mut State) -> impl WidgetView<State> + use<> {
+    virtual_scroll(0..1, |state: &mut State, _| {
```
DJMcNab, line 20: Virtual Scroll is not the right component for this use case; you should instead use `portal`.

Virtual Scroll is intended for virtualised scrolling, where elements outside of the view will be unloaded, to improve rendering and layout performance. If there's only one item, it's strictly worse than portal (as it doesn't support scrollbars - and on a technical level, it can't support them as well as portal can, because portal has more information).

DJMcNab (changes requested):
Welcome!

I do agree that we should have better docs for using `text_input`, but I think a better place to have that is probably as documentation on the `text_input` component function - i.e. 
https://github.com/linebender/xilem/blob/bbcf9a62c6fea1929d28580c3f594f7359d057fc/xilem/src/view/text_input.rs#L21-L26

Would you be willing to instead write some docs there (e.g. a similar example to what you have here as a doctest, but with only the `textbox`/`text_input` view inside `app_logic`). I think once we have that, this example probably doesn't add much to the Xilem project, and so I'd lean towards removing it.
That being said, if the issue you describe with scrolling still reproduces with `portal`, then having this example in a compiling state would be helpful to reproduce that. Perhaps then improving the docs should be a new PR?

This example also does not compile for several reasons, and you should address those. The code looks correct for Xilem as of maybe a month ago; the code needs to be updated to take the changes since then into account. (I'm slightly confused about how you reached this point, as the merge-base is a very recent commit)

Please also do add the copyright header. It's clearer for us if you add that yourself, although the comment that you are happy with us adding it is also helpful.

Cyncrovee:
I'm having problem with using `portal`, I'm not sure it works well with `text_input`. I might drop the example from the repository and instead focus on the `text_input` documentation.

DJMcNab:
> I'm having problem with using portal, I'm not sure it works well with text_input. 

This isn't hugely surprising to me; we haven't tested very large text areas yet; we need to work out a design for scrolling, etc. with them.

Issue #1344, opened 2025-08-15
Tracking issue for Portal/Scroll widget


PoignardAzur:
Portals are widgets which handle overflow by providing a scrollable area.

Their name is basically an historical artifact of how they used to work (there was a parent Scroll widget and Portal was the small widget -> big widget interface) and also of me trying too hard to be clever.

The Portal widget is some of the oldest and most cursed code in Masonry.

Here's some things we could do to improve it:

## Short terms improvements

- Rename to `Scroll` (or ScrollArea or whatever).
- Remove type parameter (we've removed them everywhere else).
- Handle horizontal scrolling in `on_pointer_event` method.
- Set `content_must_fill` to true by default, or even remove it.
- Handle Home/End/Arrows/PageUp/PageDown keys.
- Handle AccessKit events.

## Refactor reactive logic

Overall, Portal tries to do two-way reactivity relation between "where the scrollbar is" and "how far the children is scrolled", but that relation is managed in a bespoke way.

That relation is managed in:
- The `PointerEvent::Scroll` event code.
- The `Update::RequestPanToChild` event code.
- Layout code.

Also, part of the complexity is that some of the values which act as sources of truth are stored in the child `ScrollBar` widgets.

Having those be separate widgets is useful for giving scrollbars pointer capture, but it decentralizes the logic in an inconvenient way.

We could use event bubbling instead to handle pointer events to the scrollbars directly in the Portal widget.

eliheuer:
How should styling the scrollbar work in a Xilem/Masonry application? Should it just get styling methods from the view like buttons?

I'm working a Xilem/Masonry application and would like to fade out scrollbar viability when scrolling is not happening and make it a color that matches my buttons for example.

ricvelozo:
> How should styling the scrollbar work in a Xilem/Masonry application? It looks like scrollbar styles are currently hardcoded without styling methods I can find. Should it just get styling methods from Style/HasProperty like buttons?
> 
> I'm working a Xilem/Masonry application and would like to fade out scrollbar viability when scrolling is not happening and make the scrollbar a color that matches my buttons for example.
> 
> Is this how this kind of thing should work?
> 
> ```
> portal(content).scrollbar_visible(false)
> ```

It need to be implemented. The fade animation can be tricky because Masonry don't has animation abstractions yet, just a frame callback. But the color stuff can be easily done. Take this as a reference: #1386.

For the animations, #792 is the first step to it. After that, I have some ideas for animation primitives.

[...]

PR #1402, opened 2025-09-25
masonry: Fix horizontal scrolling in Portal component

yu2344:


masonry/src/widgets/portal.rs
```diff
@@ -286,11 +291,19 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
...
+                    scrollbar.cursor_progress =
+                        self.viewport_pos.x / (content_size - portal_size).width;
+                    scrollbar_ctx.request_render();
+                } else {
+                    let (scrollbar, mut scrollbar_ctx) =
+                        ctx.get_raw_mut(&mut self.scrollbar_vertical);
+                    scrollbar.cursor_progress =
+                        self.viewport_pos.y / (content_size - portal_size).height;
+                    scrollbar_ctx.request_render();
+                }
```
DJMcNab: I believe that this code is incorrect. I presume that what you're trying to do here is add the "if pressing shift and scroll down with a scroll wheel, scroll to the right"?

What this code actually does is it updates either the vertical or horizontal scrollbar's position correctly, depending on whether shift was pressed.

What I think need instead is:
1) Both of these sections to be here in a row (i.e. just remove the `if modifiers.shift() {` `} else {` and `}`, but keep the contents); that is, the PR would then only be added the equivalent line for "horizontal" as it does for vertical.
2) In the above `match delta` block, if the scroll device is a mouse and shift is pressed, swap the x and y deltas (cc @xorgy, is this the right way to distinguish between touchpad swipe gestures and scroll wheels? Do you happen to know off  the top of your head the right affordance for mice with horizontal scrollwheel support if you press shift?)

I'd approve this PR if it only did the former, and we could then split the "shift" support into its own PR, if that works for you.

masonry/src/widgets/portal.rs
```diff
@@ -457,7 +470,10 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
     }
 
     fn compose(&mut self, ctx: &mut ComposeCtx<'_>) {
-        ctx.set_child_scroll_translation(&mut self.child, Vec2::new(0.0, -self.viewport_pos.y));
+        ctx.set_child_scroll_translation(
+            &mut self.child,
+            Vec2::new(-self.viewport_pos.x, -self.viewport_pos.y),
+        );
```
DJMcNab, line 472: 😆

xilem/examples/scroll.rs
```diff
@@ -0,0 +1,71 @@
...
+                        .width(50.px())
+                        .height(50.px())
+                        .background_color(palette::css::WHITE.map_lightness(|l| {
+                            l * (col_idx as f32) / (state.horizontal_count as f32)
+                        }))
+                })
+                .collect::<Vec<_>>();
+            flex_row(horizontal_blocks)
+        })
+        .collect::<Vec<_>>();
```
DJMcNab: I *think* I'd probably use a `grid` here, rather than nested flex rows (especially as each item is the same size). But this version is fine.

DJMcNab (changes requested):
Welcome! Linking to https://github.com/linebender/xilem/issues/1344 as this is related. As noted in that issue, this portal code needs some serious refactoring, but it's good to fix the bugs in this current version as well.

The fix for horizontal scrolling never being applied looks good, despite the embarrassing prior state. The other addition of <kbd>shift</kbd> handling for scrollbars unfortunately doesn't work for me. I've proposed an alternative implementation which should work better, and I'd like to see that moved into a new PR if you're willing to do so.

Just as a note from a reviewer's perspective - it can help immensely to have some introductory context, such as in the PR description. In this case, I had to work out the two things which you're trying to fix/add (in addition to the example).

xilem/examples/scroll.rs
yu2344: Yes, using a grid is definitely more suitable. However, in my tests I noticed that the grid seems to have a layout issue (the `GridItem` don’t properly adapt to the size of the `SizedBox`). I’ve implemented both approaches for now, so that once the grid layout is fixed, it will be easy to test and switch over.

DJMcNab: Oh, of course; the grid always uses the available space. That makes sense.

PR #1405, opened 2025-09-27
DRAFT: Add layer system to Masonry (contd.)

RagibHasin:


PoignardAzur:
(Note to self: This PR has been in limbo for a while. Definitely need to review it soon.)

masonry_core/src/app/render_root.rs
```diff
@@ -268,6 +269,57 @@ pub enum RenderRootSignal {
...
+            Self::ShowWindowMenu(pos) => f.debug_tuple("ShowWindowMenu").field(pos).finish(),
+            Self::WidgetSelectedInInspector(widget_id) => f
+                .debug_tuple("WidgetSelectedInInspector")
+                .field(widget_id)
+                .finish(),
+            Self::NewLayer(root_widget, pos) => f
+                .debug_tuple("NewLayer")
+                .field(&root_widget.widget.short_type_name())
+                .field(&pos)
+                .finish(),
```
PoignardAzur: (Non-blocking) I'd rather add a Debug impl to NewWidget than remove the Debug impl from RenderRootSignal.

masonry_core/src/passes/update.rs
```diff
@@ -219,17 +219,18 @@ fn update_widget_tree(
 pub(crate) fn run_update_widget_tree_pass(root: &mut RenderRoot) {
     let _span = info_span!("update_new_widgets").entered();
 
-    if root.root.incomplete() {
+    // REROOT
+    if root.layer_stack.incomplete() {
```
PoignardAzur, line 223: If I remember what this block does, I don't think "REROOT" is a good descriptor for it. It's only called once when the widget root is initialized.

Something like "INIT TREE" would fit better.

masonry/examples/layers.rs
```diff
@@ -0,0 +1,165 @@
...
+impl AppDriver for Driver {
+    fn on_action(
+        &mut self,
+        _window_id: WindowId,
+        _ctx: &mut DriverCtx<'_, '_>,
+        _widget_id: WidgetId,
+        _action: ErasedAction,
+    ) {
+    }
+}
```
PoignardAzur, line 35: Makes me wonder if we should add a `NullDriver` type to Masonry in a future PR.

masonry_core/src/app/layer_stack.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Panics if the the intended layer the base layer in debug mode.
+    pub(crate) fn remove_layer(this: &mut WidgetMut<'_, Self>, root_id: WidgetId) {
+        match this
+            .widget
+            .layers
+            .iter()
+            .position(|layer| layer.widget.id() == root_id)
+        {
+            Some(0) => debug_panic!("Cannot remove initial layer"),
+            None => tracing::warn!("layer with root widget {root_id:?} not found"),
```
PoignardAzur: I can't imagine a case where this happens that isn't straightforwardly the framework writer's fault. This should definitely be a debug_panic.

masonry_core/src/app/layer_stack.rs
```diff
@@ -0,0 +1,200 @@
...
+        new_origin: Point,
+    ) {
+        match this
+            .widget
+            .layers
+            .iter()
+            .position(|layer| layer.widget.id() == root_id)
+        {
+            Some(0) => debug_panic!("Cannot reposition initial layer"),
+            None => tracing::warn!("layer with root widget {root_id:?} not found"),
```
PoignardAzur: Same.

masonry_core/src/app/render_root.rs
```diff
@@ -619,6 +711,49 @@ impl RenderRoot {
...
+    ///
+    /// Panics if the intended layer is the base layer in debug mode.
+    pub fn remove_layer(&mut self, root_id: WidgetId) {
+        mutate_widget(self, self.root_id(), |mut layer_stack| {
+            let mut layer_stack = layer_stack.downcast::<LayerStack>();
+            LayerStack::remove_layer(&mut layer_stack, root_id);
+        });
+
+        self.run_rewrite_passes();
+    }
```
PoignardAzur, line 696: We should document what happens when the id isn't a layer root, whether we make it a panic or keep it a warning.

masonry_core/src/app/render_root.rs
```diff
@@ -619,6 +711,49 @@ impl RenderRoot {
...
+    ///
+    /// Panics if the intended layer is the base layer in debug mode.
+    pub fn reposition_layer(&mut self, root_id: WidgetId, new_origin: Point) {
+        mutate_widget(self, self.root_id(), |mut layer_stack| {
+            let mut layer_stack = layer_stack.downcast::<LayerStack>();
+            LayerStack::reposition_layer(&mut layer_stack, root_id, new_origin);
+        });
+
+        self.run_rewrite_passes();
+    }
```
PoignardAzur, line 713: Same.

PoignardAzur (commented):
Very high-quality PR!

I especially like using root widgets' ids to identify layers in the API. This makes this change feels much less disruptive than if we had to add a LayerId type.

(I still need to run the example and re-read the example code.)

Aside from the nits I mention and the CI errors, this feels ready to merge.

That said, I think this feature needs more documentation. The whole "RenderRoot emits a signal to ask for a new layer, and then the driver calls RenderRoot methods to add the layer" design is a bit counter-intuitive. People will obviously wonder "Why can't RenderRoot add the layer it wants directly?"; so we need to document why it's done this way. It can be done in this PR or in a follow-up (in which case a TODO would be nice).

Thanks again for the PR! Sorry for not reviewing earlier.

masonry/examples/layers.rs
```diff
@@ -0,0 +1,165 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! An example of layers.
```
PoignardAzur:
```suggestion
//! An example showcasing how layers can be used to implement rudimentary tooltips.
```

masonry_core/src/app/render_root.rs
RagibHasin: I initially went that way but could not decide how the debug impl for `NewWidget` should look like.

What should the impl show, you think?

PoignardAzur: I think using `short_type_name()` the way you're doing is fine.

PR #1412, opened 2025-10-03
Upgrade to parley 0.6.0-dev

nicoburns:
Upgrades Parley to the [0.6 release PR](https://github.com/linebender/parley/pull/426)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,6 +818,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "core_maths"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77745e017f5edba1a9c1d854f6f3a52dac8a12dd5af5d2f54aecf61e43d80d30"
+dependencies = [
+ "libm",
+]
+
+[[package]]
 name = "counter"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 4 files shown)


masonry/src/widgets/text_area.rs
```diff
@@ -26,6 +26,11 @@ use crate::theme::default_text_styles;
 use crate::util::debug_panic;
 use crate::{TextAlign, theme};
 
+/// Convert `parley::BoundingBox` into `vello::kurbo::Rect`
```
DJMcNab:
```suggestion
/// Convert `parley::BoundingBox` into `vello::kurbo::Rect`.
```

masonry/src/widgets/text_area.rs
```diff
@@ -26,6 +26,11 @@ use crate::theme::default_text_styles;
...
+fn convert_rect(rect: parley::BoundingBox) -> vello::kurbo::Rect {
+    vello::kurbo::Rect::new(rect.x0, rect.y0, rect.x1, rect.y1)
+}
```
DJMcNab: What do you think about copying this function into the Parley changelog as well?

DJMcNab (approved):
Looks good, thanks! It's reassuring that this was so simple.

masonry/src/widgets/text_area.rs
DJMcNab: Note that this method is about to land in a different place in #1408.

nicoburns: > What do you think about copying this function into the Parley changelog as well?

done

masonry/src/widgets/text_area.rs
nicoburns: This function has disappeared after rebasing on top of the vello bump.

PR #1423, opened 2025-10-21
Refactor how screenshots are included in doc

PoignardAzur:
Remove screenshot folder from package uploaded to crates.io. Rework `include_screenshot!` macro. Add `include_screenshot_reference!` macro. Add screenshot to Masonry's documentation root.

See [#masonry > Screenshots in doc @ 💬](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Screenshots.20in.20doc/near/546193142) for rationale.

### DIFF
--- a/masonry/src/lib.rs
+++ b/masonry/src/lib.rs
@@ -168,6 +168,7 @@
     not(docsrs),
     doc = "**Warning: This documentation is meant to be read on docs.rs. Screenshots may fail to load otherwise.**\n\n"
 )]
+// Screenshot generated in the unit test for the `to_do_list` example.
 #![doc = util::include_screenshot_reference!("to-do-screenshot", "example_to_do_list_initial.png")]
 //!
 


masonry/README.md
```diff
@@ -170,8 +191,11 @@ If you want to use your own subscriber, simply set it before starting masonry -
 [Xilem]: https://github.com/linebender/xilem/tree/main/xilem
 [tracing_tracy]: https://crates.io/crates/tracing-tracy
 
+
 <!-- cargo-rdme end -->
 
+[to-do-screenshot]: ./screenshots/example_to_do_list_initial.png
```
DJMcNab: This should *probably* be in the big block of links at the top.

masonry/src/lib.rs
```diff
@@ -141,9 +163,13 @@
...
 //! [tracing_tracy]: https://crates.io/crates/tracing-tracy
-
-// TODO: Add screenshot. This can't use include_screenshot as that doesn't work with cargo-rdme
-// See https://github.com/linebender/xilem/issues/851
+//!
+#![cfg_attr(
+    not(docsrs),
+    doc = "**Warning: This documentation is meant to be read on docs.rs. Screenshots may fail to load otherwise.**\n\n"
+)]
+#![doc = util::include_screenshot_reference!("to-do-screenshot", "example_to_do_list_initial.png")]
```
DJMcNab, line 172: We should have a comment here noting that this isn't a screenshot generated by the doc test.

Of course, ideally, we would just generate it in the doc test, but that's not for this PR.

masonry_core/src/util.rs
```diff
@@ -133,9 +133,24 @@ pub fn get_debug_color(id: u64) -> Color {
...
 // ---
 
+// Macros are exported from crate root by default. Re-export them from here.
 pub use crate::include_screenshot;
+pub use crate::include_screenshot_reference;
 
-/// Markdown input to display a screenshot from the current crate's `screenshots` directory.
+// If we made this into proc macros, we would gain the following features:
+// 1) Automatic detection of the file existing - see https://github.com/linebender/xilem/issues/1080
+// 2) Using the image's dimensions to set the `width` and `height` attributes of the object and `img` tag
```
DJMcNab: I don't think this item is true now.

masonry_core/src/util.rs
```diff
@@ -133,9 +133,24 @@ pub fn get_debug_color(id: u64) -> Color {
...
+// 3) Extract the "repository" from CARGO_PKG_REPOSITORY and auto-generate the online URL version.
+
+// We want to show the "local" image if it's present (e.g. from a git dependency or in the local repository).
+// However, if we're on docs.rs (or building from the crates.io registry),
+// the screenshots aren't available (as they shouldn't be in the published package for space reasons).
+// For those environments, we want to instead show the image from `raw.githubusercontent.com`.
+// This means that screenshots may fail to display in some cases, e.g. the user is pulling a crate as a
+// crates.io dependency and then generating its doc locally.
+// Masonry's documentation has a few warnings for these cases.
```
DJMcNab, line 152: I've proposed a minor clean up for this block, to better clarify the limitations and the reasons for them.

```suggestion
// We want to show the "local" image if it's present (e.g. from a git dependency or in the local repository).
// The image won't be available locally if our docs are being built on docs.rs or from a crates.io dependency, 
// as we don't include the screenshots in the published package (for space/bandwidth reasons).
// This fall back uses `raw.githubusercontent.com`, which allows it to access the correct version of the screenshot for the crate's version.
// Unfortunately, it isn't currently possible to detect that this fallback is needed (without a procedural macro or build script);
// as such, we currently use `cfg(docsrs)` as a proxy for whether to use a fallback.
// This does mean that screenshots may fail to display in some cases, e.g. the user is pulling a crate as a
// crates.io dependency and then generating its doc locally.
// Masonry's documentation has a few warnings for these cases.
```

masonry_core/src/util.rs
```diff
@@ -146,46 +161,81 @@ pub use crate::include_screenshot;
 /// ```
 ///
 /// The caption should have a full-stop at the end, as it's being used as alt-text.
-/// This macro will only function correctly for packages in the Xilem repository,
-/// as it hardcodes the supported GitHub repository.
+///
+/// **Warning: This macro will only function correctly for packages in the Xilem repository,
+/// as it hardcodes the supported GitHub repository.**
+#[cfg(docsrs)]
```
DJMcNab: I think we want the documentation to be on the "other" macro, because if we use `--document-private-items` or go-to-definition, it will take us to the `not(docsrs)` item.

Practically, I think swapping them around is the key.

masonry_core/src/util.rs
```diff
@@ -146,46 +161,81 @@ pub use crate::include_screenshot;
...
+#[doc(hidden)]
+#[macro_export]
+macro_rules! include_screenshot {
+    ($path:literal $(, $caption:literal)? $(,)?) => {
+        concat!(
+            // This space at the start avoids triggering https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_doc_comments
+            // when using this macro in a `doc` attribute
+            " ![", $($caption,)? "]",
```
DJMcNab: I'd like to at least try:
```suggestion
            "![", $($caption,)? "]",
```
again.

https://github.com/rust-lang/rust-clippy/issues/14265 is supposedly fixed, and we only run stable clippy.

masonry/src/lib.rs
PoignardAzur: > Of course, ideally, we would just generate it in the doc test, but that's not for this PR.

I think the doc example is already too big for something that shows up in the main README.

PoignardAzur:
> I hope that we'll add screenshots to all the markdown docs in a follow-up?

Yeah, I don't have the motivation to do that right now.

Part of the problem is that most of the screenshots are "single color rectangle with no context". It's a problem with other examples as well (e.g. the button screenshot barely looks like a button). I think the solution will be to add a "root padding" parameter to the test harness so the widgets get more "breathing space".

masonry/src/lib.rs
DJMcNab: Fwiw, this would probably still be `#`ed out, i.e. not copied by cargo rdme or shown by rustdoc. The example as it currently stands is likely too big, but having it automatically tested shouldn't add to that.

PR #1424, opened 2025-10-22
Unify our `cargo rdme` handling

DJMcNab:
Also adds the command to run to `lib.rs`, as I often find myself needing it from there.

This adds cargo-rdme to masonry_testing, and otherwise makes the heading base levels always `=1` (the default)

### DIFF
--- a/masonry/src/lib.rs
+++ b/masonry/src/lib.rs
@@ -1,8 +1,8 @@
 // Copyright 2018 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
+// After you edit the crate's doc comment, run this command, then check README.md for any missing links
 // cargo rdme --workspace-project=masonry
-// After editing the below, then check links in README.md
 
 //! Masonry is a foundational framework for building GUI libraries in Rust.
 //!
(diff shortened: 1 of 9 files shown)


masonry/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2018 the Xilem Authors and the Druid Authors
 // SPDX-License-Identifier: Apache-2.0
 
+// cargo rdme --workspace-project=masonry
+// After editing the below, then check links in README.md
```
PoignardAzur: This comment is a bit unclear, and there's a typo in the second line. Same for all the other files with the same comment.

PoignardAzur (changes requested):
LGTM, aside from the typo.

PoignardAzur (approved):

masonry/src/lib.rs
DJMcNab: I wanted to get the command on its own line, for easier copying and pasting, and I didn't want this to spill to three lines.

I have now written what should be a clearer version of this. Hope that helps

(Even knowing there's a typo, I can't find it, so I don't know if I've fixed it.)

PoignardAzur: "After editing the below" should probably have been "After editing the doc below".

In any case, current text is fine.

PR #1426, opened 2025-10-23
Update Masonry Core's documentation

DJMcNab:
As we're looking at getting a crate release out soon, I thought it would be worth making the crate structure easier to understand. As part of this, I've written some docs for Masonry Core, most importantly the README.

### DIFF
--- a/masonry_core/README.md
+++ b/masonry_core/README.md
@@ -34,7 +34,8 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 Masonry Core provides the base GUI engine for Masonry.
 
 Masonry's widgets are implemented in the Masonry crate, which re-exports this crate as `masonry::core`.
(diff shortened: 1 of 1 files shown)


masonry_core/src/lib.rs
```diff
@@ -1,10 +1,42 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//! Traits and types of the Masonry toolkit.
-//! See [Masonry's documentation] for more details, examples and resources.
+//! Masonry Core provides the base GUI engine for Masonry.
+//!
+//! Masonry's widgets are implemented in the Masonry crate, which re-exports this crate as `masonry::core`.
+//! Most users who wish to use Masonry for creating applications (and UI libraries) should prefer that crate.
```
jaredoconnell: There is a level of ambiguity by saying "that crate", since the prior sentence talks about two crates. It'd probably be better to be explicit with the crate there.

PR #1427, opened 2025-10-23
Improve the setup of `masonry_core`'s default tracing backend

StT191:
This PR aims to improve the automatic setup of `masonry_core`'s default tracing subscriber.

The Problem it tries to fix is that before the default tracing subscriber would be setup regardless of if a user had already set a tracing subscriber. This would introduce side effects like creating a temp-file in "/tmp/" and logging a message to the user about it even though the temp-file would never be used.

The new version introduces a call to a `verify_subscriber_has_not_been_set()` check function before the default tracing subscriber setup is attempted. It uses an undocumented tracing function `tracing_core::dispatcher::has_been_set()` internally. The only caveat with this approach is that this function returns true if any tracing subscriber has been set before on the current thread regardless of if it was the global default or if it is even active at the moment.

The assumption is that users who had set a tracing subscriber manually will not rely on a default being set by the library. If they still wish to use `masonry_core`'s default subscriber they get access to it through the newly exposed `default_tracing_subscriber` function.

### DIFF
--- a/masonry_core/src/app/tracing_backend.rs
+++ b/masonry_core/src/app/tracing_backend.rs
@@ -31,8 +31,13 @@ fn default_tracing_subscriber_native(default_level: LevelFilter) -> impl Subscri
     let env_filter_builder = EnvFilter::builder()
         .with_default_directive(default_level.into())
         .with_env_var("RUST_LOG");
-    // We store the error until our env is set, *then* we display it
-    let env_var_error = env_filter_builder.from_env().err();
+    if let Some(err) = env_filter_builder.from_env().err() {
+        #[allow(clippy::print_stderr, reason = "Can only use stderr")]
+        {
+            // We print this message to stderr because tracing hasn't been set up yet
+            eprintln!("Failed to parse RUST_LOG environment variable: {:#}", err);
+        }
+    }
     let env_filter = env_filter_builder.from_env_lossy();
 
     // This format is more concise than even the 'Compact' default:
@@ -95,13 +100,6 @@ fn default_tracing_subscriber_native(default_level: LevelFilter) -> impl Subscri
     // After the above line because of https://github.com/linebender/android_trace/pull/17
     #[cfg(feature = "tracy")]
     let registry = registry.with(tracing_tracy::TracyLayer::default());
-
-    if let Some(err) = env_var_error {
-        tracing::error!(
-            err = &err as &dyn std::error::Error,
-            "Failed to parse RUST_LOG environment variable"
-        );
-    }
 
     registry
 }


masonry_core/src/app/tracing_backend.rs
```diff
@@ -119,29 +119,63 @@ fn try_init_wasm_tracing(max_level: LevelFilter) -> Result<(), SetGlobalDefaultE
...
+}
+
+impl std::error::Error for TracingSubscriberHasBeenSetError {}
+
+// Verify that a tracing subscriber has not been set already.
+fn verify_subscriber_has_not_been_set() -> Result<(), TracingSubscriberHasBeenSetError> {
+    // Return a TracingSubscriberHasBeenSetError if a tracing subscriber/dispatcher was already set:
+    // - We use the undocumented tracing::dispatcher::has_been_set() function
+    //   to test if a tracing dispatcher has already been set.
+    if tracing::dispatcher::has_been_set() {
```
DJMcNab: We should use the same function from `tracing_core`, which is still `doc(hidden)`, but importantly is *not* documented in the doc comment to not be public API. Compare <https://docs.rs/tracing/0.1.41/src/tracing/dispatcher.rs.html#139-145> vs <https://docs.rs/tracing-core/0.1.34/src/tracing_core/dispatcher.rs.html#336-344>.

masonry_core/src/app/tracing_backend.rs
```diff
@@ -119,29 +119,63 @@ fn try_init_wasm_tracing(max_level: LevelFilter) -> Result<(), SetGlobalDefaultE
...
+    }
+}
+
+impl std::error::Error for TracingSubscriberHasBeenSetError {}
+
+// Verify that a tracing subscriber has not been set already.
+fn verify_subscriber_has_not_been_set() -> Result<(), TracingSubscriberHasBeenSetError> {
+    // Return a TracingSubscriberHasBeenSetError if a tracing subscriber/dispatcher was already set:
+    // - We use the undocumented tracing::dispatcher::has_been_set() function
+    //   to test if a tracing dispatcher has already been set.
```
DJMcNab: At the moment, this is a "what" comment, and so it provides no value (or even serves to make this code more confusing to read!).
We should make it a "why" comment, saying something like "`tracing_core::dispatcher::has_been_set` is `doc(hidden)`. However, it is guaranteed to remain for the entire `tracing_core` 1.0 series, as `tracing` depends on it, and it isn't documented as unsupported."

masonry_core/src/app/tracing_backend.rs
StT191: I agree about the joy-factor. 

Anyway I switched to the `tracing_core` version of the `has_been_set` function as you suggested and added a simple test case.

masonry_core/src/app/tracing_backend.rs
StT191: I just replaced the whole section with your suggestion and added a doc comment to the function.

masonry_core/src/app/tracing_backend.rs
```diff
@@ -95,20 +101,12 @@ fn try_init_layered_tracing(default_level: LevelFilter) -> Result<(), SetGlobalD
     #[cfg(feature = "tracy")]
     let registry = registry.with(tracing_tracy::TracyLayer::default());
 
-    tracing::dispatcher::set_global_default(registry.into())?;
-
-    if let Some(err) = env_var_error {
-        tracing::error!(
-            err = &err as &dyn std::error::Error,
-            "Failed to parse RUST_LOG environment variable"
-        );
```
StT191, line 104: I converted the tracing logging of `env_var_error` to logging directly to stdout above as tracing won't be initialized during this setup function.

DJMcNab: Hmm. I think I'd still rather log it, even if that means we need to make an ugly function signature (either returning a tuple or passing in a `&mut Option<...>`).

DJMcNab:
> Is it really desirable for a library crate to attempt to set a tracing subscriber without being explicitly asked to do so by the user? As a first time user who just wanted to tinker with the library I found this behavior rather impolite.

It's a reasonable question, and partly why I say that this doesn't spark joy. My proposal for how I'd like to see this happen.

1) Export the functionality to set the default tracing publicly, as a single function call.
2) Maybe move it to the "higher-level" Masonry crate
3) When you initialise a render root (or at whatever level), we check if *some* tracing subscriber has been set, and output a really big warning if not (I'd argue for debug panicking, although we couldn't use `debug_panic!`, because we would want this to print to stderr rather than logging an error, for obvious reasons).
4) So at the start of main in each Xilem and Masonry application, you need to call `masonry::start_log_handler();`.

This would also fix some issues we have had in Xilem, where the first reconciliation run doesn't get logging.
Either way, that shouldn't be this PR.

PR #1428, opened 2025-10-23
Add RadioButton widget

ricvelozo:
Add RadioButton widget and respective `View`. It's already working, but I'll explore sending `Action` to communicate with a `RadioGroup` and centralize the state there.

Idea:

```rust
radio_group(
    state.filter,
    move |state: &mut TaskList, option| state.filter = option,
    flex_row((
        radio_button(label, Filter::All),
        radio_button(label, Filter::Active),
        radio_button(label, Filter::Completed),
    ))
)
```

I don't yet know if it's possible (or desirable) to change the state of a child widget.


masonry/src/widgets/radio_group.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+// --- MARK: WIDGETMUT
+impl RadioGroup {
+    /// Get mutable reference to the child widget, if any.
+    pub fn child_mut<'t>(this: &'t mut WidgetMut<'_, Self>) -> Option<WidgetMut<'t, dyn Widget>> {
+        let child = &mut this.widget.child;
+        Some(this.ctx.get_mut(child))
+    }
```
DJMcNab, line 36:
```suggestion
    /// Get mutable reference to the child widget.
    pub fn child_mut<'t>(this: &'t mut WidgetMut<'_, Self>) -> WidgetMut<'t, dyn Widget> {
        let child = &mut this.widget.child;
        this.ctx.get_mut(child)
    }
```

masonry/src/widgets/radio_group.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// --- MARK: IMPL WIDGET
+impl Widget for RadioGroup {
+    type Action = NoAction;
+
+    // TODO: navigation shortcuts.
```
DJMcNab, line 43: We should link to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role#keyboard_interactions

masonry/src/widgets/radio_group.rs
```diff
@@ -0,0 +1,83 @@
...
+        Role::RadioGroup
+    }
+
+    fn accessibility(
+        &mut self,
+        _ctx: &mut AccessCtx<'_>,
+        _props: &PropertiesRef<'_>,
+        _node: &mut Node,
+    ) {
+    }
```
DJMcNab, line 72: According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role, a radiogroup "must have an accessible name" set as a label.

I'm not sure exactly how we achieve that with this here. That can be a TODO.

masonry/src/widgets/radio_group.rs
```diff
@@ -0,0 +1,83 @@
...
+use tracing::{Span, trace_span};
+use vello::Scene;
+use vello::kurbo::{Point, Size};
+
+use crate::core::{
+    AccessCtx, BoxConstraints, ChildrenIds, LayoutCtx, NewWidget, NoAction, PaintCtx,
+    PropertiesMut, PropertiesRef, RegisterCtx, Widget, WidgetId, WidgetMut, WidgetPod,
+};
+
+/// A radio group container that holds radio buttons.
```
DJMcNab, line 16: This needs better documentation. Essentially, we should say:
"This widget currently only exists for creating a node with the `RadioGroup` accessibility role.
Ensuring that only a single radio button in this group is checked at a time is the responsibility of the user"

masonry/src/widgets/radio_button.rs
```diff
@@ -0,0 +1,437 @@
...
+impl HasProperty<BorderColor> for RadioButton {}
+impl HasProperty<BorderWidth> for RadioButton {}
+impl HasProperty<CornerRadius> for RadioButton {}
+impl HasProperty<Padding> for RadioButton {}
+impl HasProperty<DisabledCheckmarkColor> for RadioButton {}
+impl HasProperty<CheckmarkColor> for RadioButton {}
+
+/// The action type emitted by [`RadioButton`] when it is activated.
+///
+/// The field is the target toggle state (i.e. true is "this radio button would like to become checked").
```
DJMcNab, line 98: There is no field. I think this documentation wants to say:

```suggestion
/// All other radio buttons in the same radio group should be [unchecked](`set_checked`)
/// and this button checked in response to this action.
```

masonry/src/widgets/radio_button.rs
```diff
@@ -0,0 +1,437 @@
...
+                    ctx.submit_action::<Self::Action>(RadioButtonToggled);
+                }
+            }
+            _ => (),
+        }
+    }
+
+    fn accepts_focus(&self) -> bool {
+        // RadioButton can be tab-focused...
+        true
```
DJMcNab, line 145: Well... Really, only the currently checked radio button can be tab focused, see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radiogroup_role#tab.

This can be a TODO, as I believe that `accepts_focus` can't be toggled dynamically.

masonry/src/widgets/radio_button.rs
```diff
@@ -0,0 +1,437 @@
...
+        let bg = if ctx.is_disabled() {
+            &props.get::<DisabledBackground>().0
+        } else if is_pressed {
+            &props.get::<ActiveBackground>().0
+        } else {
+            props.get::<Background>()
+        };
+
+        let bg_rect = border_width.bg_rect(size, border_radius);
+        let border_rect = border_width.border_rect(size, border_radius);
```
DJMcNab, line 242: I'm not a massive fan of using rounded rectangles to create a circle. At the very least, I think this needs a comment.

cc @PoignardAzur, what would you recommend here?

xilem/src/view/radio_button.rs
```diff
@@ -0,0 +1,140 @@
...
+///     Lime,
+/// }
+///
+/// struct State {
+///     fruit: Fruit,
+/// }
+///
+/// // ...
+///
+/// flex_row((
```
DJMcNab, line 29: Should this not have a `radio_group`?

xilem/src/view/radio_button.rs
```diff
@@ -0,0 +1,140 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::core::{MessageContext, Mut, ViewMarker};
+use crate::{MessageResult, Pod, View, ViewCtx};
+
+use masonry::core::ArcStr;
+use masonry::widgets::{self, RadioButtonToggled};
+
+/// An element which can be in checked and unchecked state.
```
DJMcNab, line 10:
```suggestion
/// A single "radio" input, i.e. one option in a radio group.
///
/// The provided `callback` will be called when this radio button is pressed/selected.
/// The changes made in that callback should lead to only this radio button in the
/// group being checked.
```

masonry/src/widgets/radio_button.rs
PoignardAzur: It does seem surprising.

PR #1436, opened 2025-10-28
Rewrite root documentation

PoignardAzur:


tree_arena/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //! ## Safe Tree
 //!
 //! The safe tree contains a root `TreeArena` which owns the root nodes as `Vec<TreeNode<T>>`, and a`parents_map` tracking the parent of every node.
-//! Each `TreeNode` subsequently owns its own children as `Vec<TreeNode<T>>`. This model of owneship is thus checked by the rust compiler,
+//! Each `TreeNode` subsequently owns its own children as `Vec<TreeNode<T>>`. This model of ownership is thus checked by the rust compiler,
```
DJMcNab, line 23: We should probably add this to https://github.com/crate-ci/typos/issues/1383#issuecomment-3452034090

tree_arena/src/tree_arena_safe.rs
```diff
@@ -1,15 +1,10 @@
...
-//! In the meantime, we intentionally don't make the types in this module part of
-//! our public API, but still implement methods that a standalone crate would have.
-//!
 //! The types defined in this module don't *actually* implement an arena. They use
-//! 100% safe code, which has a significant performance overhead. The final version
-//! will use an arena and unsafe code, but should have the exact same exported API as
-//! this module.
+//! 100% safe code, which has a significant performance overhead.
+//! The version in `tree_arena_unsafe.rs` uses unsafe code, but should have the
+//! exact same exported API as this module.
```
DJMcNab, line 7: I wonder if we can `#![forbid(unsafe_code)]` in this module. Needn't be in this PR

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
 # ARCHITECTURE
-Xilem is a framework that aims to provide a performant and productive option for Rust GUI.
 
-This document describes the high-level architecture of xilem. If you want to familiarize yourself with the code base, you are just in the right place!
+Xilem is a family of high-level GUI frameworks.
+Xilem apps are written with idiomatic Rust code, with little to no reliance on macros and DSLs.
```
DJMcNab, line 4: Fwiw, I think that Xilem view functions are pretty DSL-like, and I'm also not sure the Rust code you write is entirely idiomatic.

But idk, this is quite nitpicky. I don't really have a better suggestion.

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
...
+- After every change, user-provided functions are called to generate a **view tree**, a lightweight representation of the app's UI.
+- The new view tree is compared against the previous view tree.
+- Based on the differences, the back-end creates an updates a retained **element tree**. elements are added, removed or mutated to match the current view tree.
+    - In `xilem`, the element tree is the Masonry widget tree.
+    - in `xilem_web`, the element tree is the DOM.
+
+This architecture is strongly inspired by React, Elm and SwiftUI, which work on similar principles.
+
+A Xilem app's code will usually be made of functions called **components** that create and return subsets of the view tree.
+A component function might look like this pseudo-code:
```
DJMcNab, line 21: Nice to have this terms I've been using actually defined in our docs!

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
...
+Xilem apps are written with idiomatic Rust code, with little to no reliance on macros and DSLs.
+
+
+## General architecture
+
+The most important thing about Xilem is that it is a *reactive* architecture:
+
+- After every change, user-provided functions are called to generate a **view tree**, a lightweight representation of the app's UI.
+- The new view tree is compared against the previous view tree.
+- Based on the differences, the back-end creates an updates a retained **element tree**. elements are added, removed or mutated to match the current view tree.
```
DJMcNab:
```suggestion
- Based on the differences, the back-end creates an updates a retained **element tree**.
  Elements are added, removed or mutated to match the current view tree.
```
Both one sentence per line, and fixing the capitalisation at the start of the second sentence.

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
...
+    )
+}
+```
+
 
 ## High-level Goals
-- **High Performance.** Lightweight state management, retained widget layer, massively parallel rendering backend (Vello), fast startup time, small binary size
-- **Productive.** Offer an ergonomic API, code is concise (low Rust tax), idiomatic and UI components compose well
-- **Rich 2D graphics model.**
-- **Wide platform support.** Windows, macOS, Linux (Wayland/X11), Android (Google funded), iOS
```
DJMcNab, line 10: I feel like I should have known about this saying "Android (Google funded)". I wonder how much of the perception that Xilem is Android specific has come from this?

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
...
-- **Batteries-included.** Advanced text layout, IME, Accessibility, Animation, Styling
 
-## Roadmap
-<!-- TODO -->
+- **Good Performance:** Xilem aims for the "performant by default" class of Rust software. Everything isn't always optimized, but it uses sober data structures and high-performance dependencies (Vello, Parley, wgpu). Static typing lets us get very efficient diffing of view trees by default.
```
DJMcNab: This block has a few sections which aren't one sentence per line, which we should fix.

ARCHITECTURE.md
```diff
@@ -0,0 +1,17 @@
+# ARCHITECTURE
+
+This repository holds the source code for the Xilem project and the Masonry project, including their sub-crates.
+
+- Xilem is a family of high-level GUI frameworks. Xilem apps are written with idiomatic Rust code, with little to no reliance on macros and DSLs.
+    - **`xilem_core`** includes the traits that define Xilem.
+    - **`xilem`** is the native framework, built on Masonry.
```
DJMcNab: Can we avoid the word "native" please - I generally prefer "cross-platform". This might occur in other places in this diff; I've not done another scan to check

ARCHITECTURE.md
```diff
@@ -0,0 +1,17 @@
...
+    - **`xilem_web`** is the web framework, built on the DOM.
+- Masonry is a foundational framework for building high-level Rust GUI libraries.
+    - **`masonry_core`** includes the core traits, types and abstractions.
```
DJMcNab: I never really liked this "core traits, types and abstractions" terminology; it's a very vague description. What do you think about the "base GUI engine" from https://github.com/linebender/xilem/pull/1426?

This isn't blocking.

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
...
+- **Productive:** Xilem offers a high-level API, designed to bypass the borrowing and state management issues that retained GUIs suffer from in Rust.
+- **Idiomatic:** Xilem apps are low on macros or DSLs. Writing a component is about creating and composing view objects, no special syntax needed.
+- **Batteries-included:** Through Masonry, Xilme includes advanced text layout, IME, accessibility trees and styling.
+- **Wide platform support:** Xilem supports the web through `xilem_web` and native platforms such as Windows, macOS, Linux (Wayland/X11), Android and iOS through `masonry` and `winit`. (Though iOS support is low-priority.)
```
DJMcNab: We've spent some time talking very carefully about how we talk about iOS support (although I don't see this written down anywhere), however, I don't think this lines up with that.
I think we've previously preferred to say "unofficial"  - maybe something like "Though iOS support is currently unofficial due to the difficulty of testing for that platform."

Maybe @waywardmonkeys is able to articulate better what we want to say here.

Alternatively, I'd also be in favour of not adding this aside at all. I get where it's coming from, and it isn't wrong, but it's maybe just not helpful?

xilem/ARCHITECTURE.md
```diff
@@ -1,27 +1,56 @@
...
+A Xilem app's code will usually be made of functions called **components** that create and return subsets of the view tree.
+A component function might look like this pseudo-code:
+
+```rust
+fn list_item(item: &ItemData) -> impl View<...> {
+    row(
+        image(item.image_url),
+        text(item.text),
+        button("Do thing").on_click(|...| do_thing(item.id)),
+    )
```
DJMcNab:
```suggestion
    row((
        image(item.image_url),
        text(item.text),
        button("Do thing").on_click(|...| do_thing(item.id)),
    ))
```
Although it remarkable how much nicer the non-tuple version looks...

tree_arena/src/tree_arena_safe.rs
PoignardAzur: Probably, yeah.

xilem/ARCHITECTURE.md
PoignardAzur: In this case, I think looking nicer is more important than being accurate. This is why I explicitly call this pseudo-code, and I don't use actual element names.

xilem/ARCHITECTURE.md
PoignardAzur: I think I'll just say "mobile" to sidestep this debate. ARCHITECTURE.md isn't supposed to be the place where you give an exhaustive feature list anyway.

ARCHITECTURE.md
PoignardAzur: Yeah, I think "base GUI engine" is better.

ARCHITECTURE.md
PoignardAzur: "Native" does point at something real that "cross-platform" doesn't: we're not using a web engine, we're using our own graphics primitives and stuff. I'm not sure what the best word would be.

xilem/ARCHITECTURE.md
DJMcNab: I don't think it's wrong to want this to look nice, but this document goes out of its way to say that the syntax is non-magical, then this example casually drops an impossible variadic function.

PoignardAzur: True.

ARCHITECTURE.md
DJMcNab: I don't disagree - I think however that the way in which "native" is misleading is worse than the way cross-platform is misleading.
I can't find where we've had these conversations before, unfortunately. Zulip's discovery of past threads isn't great.
I've made a thread - [#rust ui > Terminology: "Native" vs "Cross-Platform" vs ...](https://xi.zulipchat.com/#narrow/channel/422907-rust-ui/topic/Terminology.3A.20.22Native.22.20vs.20.22Cross-Platform.22.20vs.20.2E.2E.2E)

PR #1437, opened 2025-10-28
Split off part of README into `docs/xilem_tutorial.md`

PoignardAzur:
(This is a move-only change, contents weren't rewritten.)


README.md
```diff
@@ -32,100 +32,6 @@ Then Masonry is built on top of:
 - **Parley** for [the text stack](https://github.com/linebender/parley#the-Parley-text-stack).
 - **AccessKit** for plugging into accessibility APIs.
 
-## Architecture
-
-See [ARCHITECTURE.md](./ARCHITECTURE.md) for details.
```
DJMcNab, line 37: I note that this isn't retained in the new file. Is this removal intended?

PoignardAzur: It is, yes. I don't think the README needs to point to ARCHITECTURE.md, and it's already pretty bloated.

PR #1438, opened 2025-10-28
Rewrite root README

PoignardAzur:


README.md
```diff
@@ -12,120 +12,25 @@
...
-Like SwiftUI, it is strongly typed.
+Xilem and Masonry provide an experimental high-level architecture for writing GUI apps in Rust.
 
-## Project structure
+**Xilem** a high-level reactive framework inspired by React, SwiftUI and Elm.
+It lets users create a lightweight view tree, and changes the rendered app based on changes to the tree.
 
-This diagram gives an idea what the Xilem project is built on:
+**Masonry** is a foundational crate for building native GUIs in Rust.
+It provides user with a retained widget tree and runs event handling and update passes on it.
```
DJMcNab: Perhaps the following would be clearer:

```suggestion
It takes a user provided retained widget tree and runs event handling and update passes on it.
```
The current sentence doesn't seem well formed.

README.md
DJMcNab, line 1: I think it's worth saying explicitly that Xilem builds on Masonry, or provides a reactivity layer on top of Masonry, or something like that.

Something like "Masonry is the cross-platform element tree for Xilem".
At the moment, the high-level overview doesn't explain how Xilem and Masonry are linked, and makes it seem like they're unrelated things which happen to live in the same repository.

README.md
```diff
@@ -12,24 +12,23 @@
...
-Like all of these, it uses lightweight view objects, diffing them to provide minimal updates to a retained UI.
-Like SwiftUI, it is strongly typed.
+Xilem and Masonry provide an experimental high-level architecture for writing GUI apps in Rust.
 
-## Project structure
+**Xilem** a high-level reactive framework inspired by React, SwiftUI and Elm.
+It lets users create a lightweight view tree, and changes the rendered app based on changes to the tree.
 
-This diagram gives an idea what the Xilem project is built on:
+**Masonry** is a foundational crate for building native GUIs in Rust.
```
DJMcNab: As in the other PR, I'd prefer to use cross-platform over native (as native implies we're using win32 or w/e for some consumers)

DJMcNab (approved):
Lovely, thanks. Good to see [#rust ui > Terminology: "Native" vs "Cross-Platform" vs ...](https://xi.zulipchat.com/#narrow/channel/422907-rust-ui/topic/Terminology.3A.20.22Native.22.20vs.20.22Cross-Platform.22.20vs.20.2E.2E.2E/with/547646796) put into action!

PR #1440, opened 2025-10-29
Resize observer, for getting an update when the size of a widget changes

### LINKED ISSUE #1166: Access to Window size in Xilem
I just noticed that custom foreground and background colors are now supported for labels and buttons, and fixed custom text sizes as well. So I might finally start with my chess GUI. Pieces would be drawn using Unicode Glyphs, similar as in my tiny-chess EGUI app. It would be nice if I could scale font size when window is resized as the EGUI app does. So I would need read access to the window size, or to current widget sizes, and perhaps a winit resize event. I can not find such functionality. When it is possible, adding that to calc.rs would be a good idea, even when for most user interfaces dynamic font scaling is not required.

DJMcNab:
Fixes #1166.

Inspired by: https://github.com/linebender/xilem/blob/xilem_classic_archive/src/widget/layout_observer.rs

TODO:

- [x] Tests
- [x] Xilem View
- [x] Use in an example? - Just stuck into Mason for now

### DIFF
--- a/xilem/src/view/resize_observer.rs
+++ b/xilem/src/view/resize_observer.rs
@@ -4,11 +4,11 @@
 use std::any::type_name;
 use std::marker::PhantomData;
 use vello::kurbo::Size;
-use xilem_core::{Arg, MessageResult, ViewArgument, ViewId, ViewPathTracker};
+use xilem_core::{Arg, MessageCtx, MessageResult, ViewArgument, ViewId, ViewPathTracker};
 
 use masonry::widgets::{self, LayoutChanged};
 
-use crate::core::{MessageContext, Mut, View, ViewMarker};
+use crate::core::{Mut, View, ViewMarker};
 use crate::{Pod, ViewCtx, WidgetView};
 
 /// A view which calls the `on_resize` callback whenever the size of its child changes.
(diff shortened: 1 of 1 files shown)


xilem/src/view/resize_observer.rs
```diff
@@ -0,0 +1,148 @@
...
+/// This can be a useful primitive for making size-adaptive designs, such as
+/// scaling up a game board in response more space being available, or switching
+/// to use fewer columns when there is not space to fit multiple columns.
+/// This can be safely used to dynamically access the size of a window
+/// or tab in a [`split`](crate::view::split::split).
+/// You must make sure that the child takes up all the available space.
+/// This can be most easily achieved by making the child be
+/// an [expanded](crate::view::SizedBox::expand) `sized_box`.
+///
+/// See the documentation on the underlying [`ResizeObserver`](widgets::ResizeObserver) for more information.
```
Philipp-M, line 27: Might be nice to have an example here (also to serve as very basic test for Xilems API).

xilem/src/view/resize_observer.rs
```diff
@@ -0,0 +1,148 @@
...
+///
+/// See `resize_observer` documentation for more context.
+#[must_use = "View values do nothing unless provided to Xilem."]
+pub struct ResizeObserver<V, F, State, Action = ()> {
+    inner: V,
+    on_resize: F,
+    phantom: PhantomData<fn() -> (State, Action)>,
+}
+
+const RESIZE_OBSERVER_CONTENT_VIEW_ID: ViewId = ViewId::new(0);
```
Philipp-M: As noted in some historical PR (from you IIRC), maybe a random number here to distinct this of other view ids?

Philipp-M (approved):
Thanks, nice job on the doc-comments and tests

PR #1441, opened 2025-10-29
masonry: Use `ui-events` from `masonry_core`

waywardmonkeys:
This is preferable to both `masonry_core` and `masonry` having a dependency on `ui-events` when `masonry_core` also re-exports it and `masonry` depends on `masonry_core`.


xStrom (approved):
Makes sense, thanks!

PR #1443, opened 2025-10-31
Add Xilem controls to edit text views' size.

PoignardAzur:
Modify the to_do_mvc to have some larger controls, which is slightly closer to how the canonical to_do_mvc looks.


PoignardAzur:
Mostly it's about adding the capability, and using it in at least one example.

DJMcNab:
> If this is the direction we want to go, this works. 

I'm not convinced that this is the direction we want to go. I'd be more inclined for `Checkbox` in Xilem to be generic over any View which yields a Label.
However, I think as a near term hack, this is fine to have landed.

PR #1444, opened 2025-10-31
Allow non-contiguous app state in Xilem

DJMcNab:
See the proposal in [#xilem > A mirage of a future Xilem](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/A.20mirage.20of.20a.20future.20Xilem/with/548204242).

This is a reasonably small addition to Xilem's model, but with the potential to radically improve Xilem's usability. In it, we allow multiple references

The biggest downside identified so far is the required new `()` parameter to `lens`. This is due to https://doc.rust-lang.org/error_codes/E0502.html. This is obviously pretty bad, but I don't think it's a showstopper.
At the very least, I'd argue that Xilem needs an experiment like this to happen; our development velocity isn't high enough that post-facto reverting this would be unviable.

Status, of what needs to be ported:

- [X] Xilem Core
- [X] Xilem Core Examples/Tests
- [x] Xilem
- [x] Xilem Examples
- [x] Xilem Web

I'm proposing that these are left as follow-ups:

- [ ] Tests which use the new capabilities this provides
- [ ] New views for local variables
- [ ] Porting whichever examples this would be suitable for to use this
- [ ] The extension of the environment system into shared references.

Guidance for porting:

- Porting components is pretty trivial. Basically, any `View<State>` (or `WidgetView<State>`) needs to be changed to `View<Edit<State>>` (or `WidgetView<Edit<State>>`).
- Porting View implementations is also reasonably simple. We need:
  - new bounds that the `State` implements `ViewArgument`.
  - Parameters to `View` functions of the form `&mut State` need to be changed to `Arg<'_, State>`.
  - Uses of `app_state` need to replaced with `State::reborrow_mut(&mut app_state)`; in some cases, this can be skipped.

### DIFF
--- a/xilem_web/README.md
+++ b/xilem_web/README.md
@@ -44,9 +44,10 @@ use xilem_web::{
     elements::html::{button, div, p},
     interfaces::{Element as _, HtmlDivElement},
     App,
+    core::Edit,
 };
 
-fn app_logic(clicks: &mut u32) -> impl HtmlDivElement<u32> + use<> {
+fn app_logic(clicks: &mut u32) -> impl HtmlDivElement<Edit<u32>> + use<> {
     div((
         button(format!("clicked {clicks} times")).on_click(|clicks: &mut u32, _event| *clicks += 1),
         (*clicks >= 5).then_some(p("Huzzah, clicked at least 5 times")),


xilem_core/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+/// # use xilem_core::docs::{DocsView as WidgetView, stateless_component};
+/// # use xilem_core::map_state;
+/// use core::ops::Range;
+/// use xilem_core::{Read, Edit};
+/// fn slider() -> impl WidgetView<(Edit<f64>, Read<Range<f64>>)> {
+///     // ...
+/// #   map_state(stateless_component(), |(_result, _range), ()| ())
+/// }
+/// ```
+// TODO: Maybe the name `Mut` here would make sense? I don't hate edit, though.
```
jaredoconnell, line 113: I think `Mut` does make more sense, especially considering how it's literally a static reference to a mutable generic type. I think `Edit` implies more than is being done here.

xilem_core/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+/// # use xilem_core::docs::{DocsView as WidgetView, stateless_component};
+/// # use xilem_core::map_state;
+/// use core::ops::Range;
+/// use xilem_core::{Read, Edit};
+/// fn slider() -> impl WidgetView<(Edit<f64>, Read<Range<f64>>)> {
+///     // ...
+/// #   map_state(stateless_component(), |(_result, _range), ()| ())
+/// }
+/// ```
+pub type Read<T> = &'static T;
```
jaredoconnell, line 155: I like that this can be used instead of Edit/Mut where applicable. Are there no locations at which this can be used in the current examples?

DJMcNab: I don't know. I've not looked carefully.

A lot of places ideally would be using `()`, but that's also not super ergonomic at the moment.

jaredoconnell: Ah. I think giving it another look would be worth it, since if you can use it then that feature is tested.

xilem_core/src/state.rs
DJMcNab: I see a few problems with `Mut`:
1) The name overlaps with the existing `xilem_core::Mut`, making this messier.
2) `Read` doesn't really fit the naming theme of `Mut`; the equivalent name by analogy with `&mut` would be ` ` (i.e. the empty name), which isn't possible. I don't particularly like the name `Ref`, because `Mut` is also a "reference", just an exclusive one.

On the other hand, Ref and Mut are a pretty standard pair in Rust (e.g. in Bevy ECS, `WidgetMut`/`WidgetRef`).

I'd lean slightly towards leaving it as is for this PR - we can always rename this as a follow-up if we want to.

xilem_core/src/state.rs
DJMcNab: The other basic problem is that there's no automatic downgrading, so turning from writing to reading is an annoying operation.

The accesses to the state generics essentially all happen in event handlers currently, so it doesn't really make much sense to access the state in a read only way there. Any function which doesn't need to know about the state is already generic

I do have some ideas about how I'd change Xilem even more radically, building from this change, but the current examples aren't architected to use `Read`

xilem_core/src/state.rs
jaredoconnell: Looking up ideas, one option would be `Mut` and `Imm`.

Yeah, that name collision wouldn't be ideal.

I'm okay with leaving it as-is if you don't love this idea.

xilem/examples/external_event_loop.rs
```diff
@@ -16,18 +16,20 @@ use winit::event::ElementState;
...
-fn big_button(
+fn big_button<F: Fn(&mut i32) + Send + Sync + 'static>(
     label: impl Into<Label>,
-    callback: impl Fn(&mut i32) + Send + Sync + 'static,
-) -> impl WidgetView<i32> {
-    sized_box(button(label.into(), callback))
+    callback: F,
+) -> impl WidgetView<Edit<i32>> {
+    // This being fully specified is "a known limitation of the trait solver"
+    sized_box(button::<Edit<i32>, _, _, F>(label.into(), callback))
```
Philipp-M, line 27: Hmm having to explicitly type all of this is really annoying... I hope that this will be solved in the trait-solver, but it might also be another argument for [#xilem > Removing App State](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Removing.20App.20State/with/545554870)

xilem_core/src/state.rs
Philipp-M: I agree with leaving as is for now, I can also see this changing towards something semantically different (i.e. the more radical approach as proposed in [#xilem > Removing App State](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/Removing.20App.20State/with/545554870)). So I think not introducing more breaking changes than necessary is good for now (even renaming of types), additionally it further bloats the diff.

xilem_core/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// An alias to the "reference" form of a [`ViewArgument`].
+///
+/// This is used in `View` functions (instead of the expanded form) to decrease
+/// the amount of noise produced.
+// TODO: The name of this definitely needs rethinking.
+pub type Arg<'a, T> = <T as ViewArgument>::Params<'a>;
+
+// TODO: Do these need to be 'static?
```
Philipp-M, line 164: I think so? Otherwise `impl View` is not static is it (I think we had a discussion about this a while back)?

Neat trick by "abusing" a GAT with a `&'static` to avoid new type wrappers.

xilem_core/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+    type Params<'a> = (T0::Params<'a>, T1::Params<'a>);
+    fn reborrow_mut<'input, 'a: 'input>(
+        (t0, t1): &'input mut Self::Params<'a>,
+    ) -> Self::Params<'input> {
+        (T0::reborrow_mut(t0), T1::reborrow_mut(t1))
+    }
+}
+
+// TODO: Generate 2+ with a macro; maybe 1+, but then again for understandability purposes, having at least one
+// outside of the macro is appealing..
```
Philipp-M, line 207: I agree, the impl is short enough, that we can keep it for `(T0,)` and `(T0,T1)`. Should this happen in this PR?

Philipp-M (approved):
Thanks for investigating this, I think this is a reasonable way forward, albeit I think it should be an intermediate solution, towards something that ideally has less type-complexity, which I'm a little bit concerned about (as we already have a lot of type-fu).
I fear that this can repel new users without a deep knowledge of Rusts type-system.
This is visible in the diff of LOC, almost 1000 new lines, but the actual juice of this PR is 220 lines (state.rs), the rest is mostly explicit typing for Rust.

I think we definitely need an example what this PR enables, as this is not immediately clear to someone that doesn't know the internals of Xilem.

I could imagine this being extended in the components example, or it could be in its own, it could be a follow-up PR, as well as added here.

xilem_core/src/state.rs
DJMcNab: I'll do it as a follow-up so that we're not blocking merging this on more review, since it's a merge conflict magnet.

xilem/examples/external_event_loop.rs
DJMcNab: Yeah, it's frustrating.

xilem_core/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+///
+/// This is implemented for [`Edit<T>`], [`Read<T>`], and tuples thereof.
+/// It can also be implemented manually, which allows names to be given to fields.
+///
+/// # Examples
+///
+/// In these examples, the Action and Context parameters to `View` are elided.
+///
+/// - `View<Edit<f32>>` will read and write one `f32` value.
+/// - `View<(Edit<f32>, Range<f32>)>` will read and write one `f32` value, and read an `f32` range.
```
axelkar, line 21:
```suggestion
/// - `View<(Edit<f32>, Read<Range<f32>>)>` will read and write one `f32` value, and read an `f32` range.
```

Issue #1350, opened 2025-08-15
Tracking issue for top-level documentation


PoignardAzur:
The top-level documentation of Xilem, i.e. the first thing people see when they find the project, is in a poor state.

We should do the following:

- [x] Rewrite and/or split the top-level README, which mostly describes Xilem and ignores the other crates.
- [x] Rewrite the top-level ARCHITECTURE.md file.
- [x] Add some pretty screenshots to the top-level README.
- [x] Remove references to the "Xilem: an architecture for UI in Rust" article. The article is from 2022 article, and while the general concepts still apply, many of the items it refers to no longer exist.
- [x] Add screenshots to `masonry/README.md`.
- [x] Fill out `masonry_core/README.md`, and include it among the files updated by `cargo rdme`.
- [x] Fill out `masonry_testing/README.md`, and include it among the files updated by `cargo rdme`.
- [x] Move the examples in masonry_winit to masonry.

PR #1445, opened 2025-10-31
Add Canvas widget - cont.

philocalyst:


PoignardAzur:
To be clear: this is based on https://github.com/linebender/xilem/issues/875.

masonry/src/testing/harness.rs
PoignardAzur, line 1: This is a merge error. You should remove this file.

masonry/src/widgets/screenshots/masonry__widgets__canvas__tests__hello.png
PoignardAzur, line 1: You should remove this file.

masonry/src/widgets/snapshots/masonry__widgets__canvas__tests__simple_canvas.snap
PoignardAzur, line 1: We almost never do these kinds of snapshot tests anymore. You should remove this file.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,219 @@
...
+                &path,
+            );
+        });
+
+        let mut harness = TestHarness::create(
+            DefaultProperties::default(),
+            canvas.with_props(Properties::default()),
+        );
+
+        assert_debug_snapshot!(harness.root_widget());
```
PoignardAzur:
```suggestion
```

PoignardAzur (changes requested):
Thanks for picking up this work!

philocalyst:
Fixed both :)

▰▰▰▰▰
Miles Wirht 🙃

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,217 @@
...
+                &path,
+            );
+        });
+
+        let mut harness = TestHarness::create(
+            DefaultProperties::default(),
+            canvas.with_props(Properties::default()),
+        );
+
+        assert_render_snapshot!(harness, "hello_new");
```
ricvelozo: Change the snapshot name to "canvas_simple", to match the test name.

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,94 @@
...
+use masonry::core::WidgetMut;
+use masonry::widgets;
+use vello::Scene;
+use vello::kurbo::Size;
+use xilem_core::MessageContext;
+
+use crate::core::{Mut, ViewMarker};
+use crate::{MessageResult, Pod, View, ViewCtx};
+
+/// A non-interactive text element.
```
ricvelozo: Update the documentation, with an simple example.

ricvelozo: That example could be added to the examples directory too.

philocalyst: I added an example as you said, but I'm noticing that's not the norm (Just checked the prose element) Is there an active issue that's tracking this?

ricvelozo: #392 & #1350

masonry/src/widgets/screenshots/masonry__widgets__canvas__tests__hello.png
ricvelozo: You forgot this file, @philocalyst.

philocalyst: Oops, my fault @ricvelozo

ricvelozo (commented):
@PoignardAzur originally reviewed the first PR, and may want to request further changes. But it looks good to me.

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,96 @@
...
+        prev: &Self,
+        (): &mut Self::ViewState,
+        _ctx: &mut ViewCtx,
+        element: Mut<'_, Self::Element>,
+        _state: &mut State,
+    ) {
+        if !Arc::ptr_eq(&self.draw, &prev.draw) {
+            widgets::Canvas::set_painter_arc(element, self.draw.clone());
+        }
+    }
```
PoignardAzur: Alt-text handling is missing.

PoignardAzur (changes requested):

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,96 @@
...
+/// # Example
+///
+/// ```ignore
+/// use xilem::view::canvas;
+/// use masonry::{Scene, Size};
+///
+/// let my_canvas = canvas(|scene: &mut Scene, size: Size| {
+///     // Drawing a simple rectangle that fills the canvas.
+///     scene.fill_rect((0.0, 0.0, size.width, size.height), [0.2, 0.4, 0.8, 1.0]);
+/// });
```
Philipp-M: I think we should strive towards not ignoring the examples in doc-comments.
This has the tendency to quickly be out of sync, I think you can already see that in the diff.

```suggestion
/// ```
/// use xilem::masonry::{kurbo::{Rect, Size}, peniko::Fill, vello::Scene};
/// use xilem::{Affine, view::canvas};
/// # use xilem::WidgetView;
///
/// # fn fill_canvas() -> impl WidgetView<()> + use<> {
/// let my_canvas = canvas(|scene: &mut Scene, size: Size| {
///     // Drawing a simple rectangle that fills the canvas.
///     scene.fill(
///         Fill::NonZero,
///         Affine::IDENTITY,
///         xilem::palette::css::AQUA,
///         None,
///         &Rect::new(0.0, 0.0, size.width, size.height),
///     );
/// });
/// # my_canvas
/// # }
```

xilem/src/view/canvas.rs
philocalyst: I added it to build, but am confused what I should do for rebuild? As there may be a case where both the drawing of the element and the alt text has changed, and set_painter_arc consumes element, so I can't apply both changes..

DJMcNab: There are two answers here.

Firstly, the `WidgetMut` methods on the `Canvas` widget should be accepting `&mut WidgetMut<Self>`, rather than just plain `WidgetMut`. For example, you can look in:
https://github.com/linebender/xilem/blob/d408dd6abf5b963cbf283eb010e439dc62eaea15/masonry/src/widgets/button.rs#L77-L83
That would solve this issue.

However, there is also a trick for when you do need to fork a `WidgetMut`. That's the [`reborrow_mut`](https://docs.rs/masonry/latest/masonry/core/struct.WidgetMut.html#method.reborrow_mut) method.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    /// Set the text that will be used to communicate the meaning of the canvas to
+    /// those using screen readers.
+    ///
+    /// Users are encouraged to set alt text for the canvas.
+    /// If possible, the alt-text should succinctly describe what the canvas represents.
+    ///
+    /// If the canvas is decorative or too hard to describe through text, users should set alt text to `""`.
```
DJMcNab, line 49: Again, the same comment applies as in #875, namely:

```suggestion
    /// If the canvas is decorative users should set alt text to `""`.
    /// If it's too hard to describe through text, the alt text should be left unset.
    /// This allows accessibility clients to know that there is no accessible description of the canvas content.
```

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn with_alt_text(mut self, alt_text: impl Into<String>) -> Self {
+        self.alt_text = Some(alt_text.into());
+        self
+    }
+}
+
+// --- MARK: WIDGETMUT ---
+impl Canvas {
+    /// Update the draw function
```
DJMcNab, line 58:
```suggestion
    /// Update the draw function.
```

And the same for the other items. We want all doc comments to end with a full stop, for consistency.
Unfortunately, this currently isn't linted for.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,217 @@
...
+        this.ctx.request_render();
+    }
+
+    /// Set the alternative text for this widget
+    pub fn set_alt_text(mut this: WidgetMut<'_, Self>, alt_text: String) {
+        this.widget.alt_text = Some(alt_text);
+        this.ctx.request_accessibility_update();
+    }
+
+    /// Remove the existing alternative text on this widget (If there is any)
```
DJMcNab:
```suggestion
    /// Remove the existing alternative text on this widget.
```

The parenthetical suggests that this widget does something else in the case where there wasn't alt text (such as panicking), which isn't the case.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn on_pointer_event(
+        &mut self,
+        _ctx: &mut EventCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        _event: &PointerEvent,
+    ) {
+    }
+
+    fn accepts_pointer_interaction(&self) -> bool {
+        true
```
DJMcNab, line 101: Previous discussion at https://github.com/linebender/xilem/pull/875#discussion_r1976398994.

The only case this really impacts is using a canvas within a button; see for example #1429. I think that in most cases, we would want this to be `false`, as was my position in that original thread. But fixing this can be a follow-up; we certainly don't want to block on this.

That is concretely, please leave this as-is!

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    fn get_debug_text(&self) -> Option<String> {
+        self.alt_text.clone()
+    }
+
+    type Action
+        = NoAction
+    where
+        Self: Sized;
```
DJMcNab:
```suggestion
    type Action = NoAction;
```

For consistency with the rest of the widgets, I'd probably also move this to the start of this implementation. But that isn't blocking.

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,112 @@
...
+///     scene.fill(
+///         Fill::NonZero,
+///         Affine::IDENTITY,
+///         &Color::from_rgb8(51, 102, 204),
+///         None,
+///         &rect,
+///     );
+/// });
+/// ```
+pub fn canvas(draw: impl Fn(&mut Scene, Size) + Send + Sync + 'static) -> Canvas {
```
DJMcNab: The same comment applies as in #875, that this signature is wrong.
This signature can only be:

```suggestion
pub fn canvas(draw: Arc<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>) -> Canvas
// or
pub fn canvas(draw: &Arc<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>) -> Canvas
```

Because we want this to allow not resetting the draw function every rebuild. Currently the `!Arc::ptr_eq` line will *always* return true (i.e. that the items are not equal) because a new Arc is allocated in each call to `canvas`.

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,112 @@
...
+/// Creates a non-interactive drawing surface.
+///
+/// The `canvas` function provides a way to render custom graphics using a
+/// user-supplied drawing callback.
+///
+/// # Example
+///
+/// ```
+/// use xilem::view::canvas;
+/// use vello::{
```
Philipp-M: More of a nit, and I see reasons for using `vello` directly, my suggestion of using `vello` via xilem was intentional, as it's for one not necessary to have `vello` as direct dependency, and it avoids possible version mismatch.

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,112 @@
...
+///     kurbo::{Affine, Rect},
+///     peniko::{Color, Fill},
+///     Scene,
+/// };
+///
+/// let my_canvas = canvas(|scene: &mut Scene, size| {
```
Philipp-M: Same here, wrapping this in a function (even if not visible in the doc for the user) returning `impl WidgetView` has the advantage of checking whether this is *actually* a view (i.e. test on the type-checking-level).

Philipp-M:
These changes came from https://github.com/linebender/xilem/pull/1444

xilem/src/view/canvas.rs
philocalyst: I don't see it re-exported anywhere though... Should I do that?

Philipp-M: Check out my original suggestion (which I actually tested locally, and includes the suggestion below as well)

philocalyst: Now included :) Apologies, I should have integrated the suggestion at the time as I quickly forgot...

PR #1446, opened 2025-11-02
Add `ContentHost` widget; use it for Xilem `AnyView`

waywardmonkeys:
Introduce `masonry::widgets::ContentHost`, a pass‑through single‑child container that hosts exactly one child and supports dynamic replacement.

In Xilem, we replace the bespoke `DynWidget` with `ContentHost`. `AnyWidgetView` now uses `Pod<ContentHost>`.


xilem/src/driver.rs
```diff
@@ -312,7 +311,10 @@ where
         let fonts = std::mem::take(&mut self.fonts);
 
         for root in state.roots() {
-            if let Some(root_widget) = root.get_layer_root(0).downcast::<DynWidget>() {
+            if let Some(root_widget) = root
+                .get_layer_root(0)
+                .downcast::<masonry::widgets::ContentHost>()
+            {
```
waywardmonkeys, line 317: I'm working on a different PR that changes this ... but this preserves the current behavior.

masonry/src/doc/implementing_container_widget.md
```diff
@@ -20,6 +20,10 @@ Our widget was overall pretty simple, the tutorial skipped over a few methods wi
 However, in some cases you want to write a widget which "contains" other widgets.
 You want these child widgets to receive events and be painted as well, as part of the widget hierarchy.
 
+Note: If you only need a simple pass-through container that hosts exactly one child and you want to be able to dynamically swap that child at runtime, consider using [`ContentHost`](crate::widgets::ContentHost).
+It forwards layout/paint/accessibility to its content and exposes helpers to replace or edit the hosted child.
+For richer behavior (custom layout, chrome, etc.), continue with a dedicated container as described below.
```
PoignardAzur, line 25: I'm not sure this bit is warranted. This tutorial is specifically about creating a new container, not existing ones.

On the other hand, maybe there should be a broader section about container widgets?

This is non-blocking.

masonry/src/widgets/content_host.rs
PoignardAzur, line 1: I wonder if maybe this should go in `masonry_core` directly.

PoignardAzur (approved):
`ContentHost` seems like a confusing name. How about `WrapperWidget`?

Nits aside, this moves in the right direction.

masonry/src/doc/implementing_container_widget.md
DJMcNab: Yeah, I'd probably agree that this doesn't belong here/isn't needed

masonry/src/widgets/content_host.rs
```diff
@@ -0,0 +1,159 @@
...
+    PropertiesMut, PropertiesRef, RegisterCtx, Widget, WidgetId, WidgetMut, WidgetPod,
+};
+
+/// A pass-through container that hosts exactly one child, which may be replaced dynamically.
+///
+/// Useful when you need an insertion point for one widget that can be
+/// replaced at runtime, without adding layout or chrome of its own.
+///
+/// # Examples
+/// Create a host and later replace its content inside a mutate/edit callback:
```
DJMcNab, line 22:
```suggestion
/// # Examples
///
/// Create a host and later replace its content inside a mutate/edit callback:
```

PoignardAzur:
> What about `Passthrough`?

I like it!

> Some other naming options

I think most of those evoke a more complex behavior than what this does.

PR #1447, opened 2025-11-02
Introduce an opt‑in focus fallback policy to Masonry

waywardmonkeys:
This adds `FocusFallbackPolicy` (`FirstTextInput`, `FirstFocusable`, `AnchorSubtreeFirst`, `Tagged`) and two helpers on `RenderRoot`: `compute_focus_fallback` and `set_focus_fallback_policy`. The traversal is implemented once via a small depth‑first helper, so policies remain clean and extensible. The enum is non‑exhaustive, and matching inside the crate is explicit to keep future additions intentional.

Xilem now calls the policy at startup and after root rebuilds, giving apps a sensible default target for text events when nothing is focused. Masonry remains free of a default behavior; nothing changes unless a framework opts in.

### DIFF
--- a/masonry/src/tests/event.rs
+++ b/masonry/src/tests/event.rs
@@ -13,7 +13,8 @@ use vello::kurbo::Size;
 
 use crate::properties::types::AsUnit;
 use crate::theme::default_property_set;
-use crate::widgets::{Button, Flex, SizedBox, TextArea};
+use crate::widgets::{Button, Flex, Label, SizedBox, TextArea};
+use masonry_core::app::FocusFallbackPolicy;
 
 // POINTER EVENTS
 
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
This builds on top of #1446 to remove the special case behavior there.

masonry/src/tests/event.rs
```diff
@@ -345,6 +490,12 @@ fn text_event_fallback() {
     assert_matches!(harness.take_records_of(target_tag)[..], []);
 }
 
+// default fallback is not set automatically
+
+// default fallback is not updated automatically
+
+// default fallback is not set automatically
+
```
DJMcNab, line 498: Which of these cases are these comments describing:

- These are tests we want to have but haven't written yet
- These are things we want to have but aren't implemented yet so we can't test them
- These are things which are currently true, but which don't need a test because they're so trivial

I'm sure I'll be able to infer this from the rest of the code, but they're locally non-obvious.

masonry_core/src/app/render_root.rs
```diff
@@ -37,6 +37,25 @@ use crate::passes::update::{
...
 
+/// A policy used to compute a reasonable focus fallback.
+///
+/// Use [`RenderRoot::set_focus_fallback_policy`] to compute and apply a fallback according to this policy.
+///
+/// This enum does not change Masonry's default behavior.
+/// Frameworks compute and set a fallback explicitly when desired.
+#[non_exhaustive]
+#[derive(Clone, Debug)]
+pub enum FocusFallbackPolicy {
```
DJMcNab, line 48: Maybe "Strategy" is a better name than "Policy"; policy to me implies continual application, which you're having to argue against in the docs.

masonry_core/src/app/render_root.rs
```diff
@@ -287,6 +306,80 @@ pub(crate) struct InspectorState {
...
+            }),
+            FocusFallbackPolicy::Tagged(name) => {
+                let id = self.global_state.widget_tags.get(name).copied()?;
+                if self.is_still_interactive(id) {
+                    Some(id)
+                } else {
+                    None
+                }
+            }
+            FocusFallbackPolicy::AnchorSubtreeFirst(_) => None,
```
DJMcNab, line 368: Should this `debug_panic`?

masonry_core/src/app/render_root.rs
```diff
@@ -37,6 +37,25 @@ use crate::passes::update::{
...
+/// Frameworks compute and set a fallback explicitly when desired.
+#[non_exhaustive]
+#[derive(Clone, Debug)]
+pub enum FocusFallbackPolicy {
+    /// Pick the first interactive widget that accepts text input.
+    FirstTextInput,
+    /// Pick the first interactive widget that accepts focus.
+    FirstFocusable,
+    /// Prefer the subtree under the current focus anchor, then apply the inner policy globally.
+    AnchorSubtreeFirst(Box<FocusFallbackPolicy>),
```
DJMcNab, line 54: It seems to be that this is semantically a one-time boolean, rather than part of this enum. It's  meaningless recursively, for one thing.  It's also meaningless for "Tagged", which suggests it could be a field on `FirstTextInput` and `FirstFocusable`.

masonry_core/src/app/render_root.rs
```diff
@@ -287,6 +306,80 @@ pub(crate) struct InspectorState {
...
+    pub fn compute_focus_fallback(&self, policy: FocusFallbackPolicy) -> Option<WidgetId> {
+        let base = self.layer_root_id(0);
+        match &policy {
+            FocusFallbackPolicy::AnchorSubtreeFirst(inner) => {
+                if let Some(anchor) = self.global_state.focus_anchor
+                    && self.is_still_interactive(anchor)
+                    && let Some(id) = self.match_policy_at(inner, anchor)
+                {
+                    return Some(id);
+                }
```
DJMcNab, line 321: I wonder if this should also search through the entire layer the anchor is part of.
There are also arguments for having it also do a "nearest parent text-input" (or focusable).

Neither of these are blocking for this PR, though.

masonry_core/src/app/render_root.rs
```diff
@@ -287,6 +306,80 @@ pub(crate) struct InspectorState {
...
+        }
+    }
+
+    /// Depth-first pre-order search from `start` for first widget whose `predicate` returns true.
+    fn find_first_matching(
+        &self,
+        start: WidgetId,
+        predicate: impl Fn(&WidgetRef<'_, dyn Widget>) -> bool,
+    ) -> Option<WidgetId> {
+        let mut stack = vec![self.get_widget(start)?];
```
DJMcNab, line 336: I guess the reason we've not done a recursive search here is because the function can't be passed to the children? In other similar functions (e.g. `find_widget_by_id`), we've done a recursive search rather than allocating.

I believe that you can either accept the predicate by reference, or ensure that the predicate is `Copy` (which incidentally is equivalent but more ergonomic, as I believe that a non-copy predicate can just be borrowed to make it Copy)

masonry_core/src/app/render_root.rs
```diff
@@ -287,6 +306,80 @@ pub(crate) struct InspectorState {
...
+                }
+            }
+            FocusFallbackPolicy::AnchorSubtreeFirst(_) => None,
+        }
+    }
+
+    /// Compute and set a focus fallback according to the provided `policy`.
+    ///
+    /// Returns `true` if a fallback was found and set.
+    pub fn set_focus_fallback_policy(&mut self, policy: FocusFallbackPolicy) -> bool {
```
DJMcNab, line 375: I think this function name probably needs to change, as again, "policy" to me strongly indicates that this is continuous. I'm not sure what the oneshot name for this would be, though.

masonry_core/src/app/render_root.rs
```diff
@@ -426,6 +519,8 @@ impl RenderRoot {
             .state
     }
 
+    // (helper removed)
+
```
DJMcNab, line 523: What does this mean?

DJMcNab (commented):
The new tests are great!

From an accessibility perspective, it's a bit odd that the focus fallback is not truly focused.
For example, running the `mason` example, I can type in the textbox, but the caret isn't visible, nor even is my current selection. I also expect that IME probably doesn't work, but I've not checked that. Additionally, pressing <kbd>Tab</kbd> has behaviour which I believe would be unintuitive.

The original focus fallback mechanism was designed to handle something like a (potentially invisible) global keyboard shortcut matcher. For that use case, I agree that you don't want the fallback to change the real focus. The new "tagged" policy would be excellent for this use case, as well. (Incidentally, it would probably make sense if the focus fallback gets all *unhandled* text events).

However, using this mechanism for "default" focus doesn't seem right to me, for the reasons I've mentioned about.

This concern doesn't block this PR from my perspective. However, we will need a follow-up issue to work out a way forward which addresses these needs.

masonry_core/src/app/render_root.rs
DJMcNab: (Alternatively, no suffix at all might be even nicer)

PR #1448, opened 2025-11-03
Change default background and text colors

PoignardAzur:
The to_do_list screenshot is a good example of the new look.

Update test_property_set to avoid having to update all the screenshots.


PoignardAzur:
Never mind, the screenshot tests do fail in CI. That's inconvenient =/

jaredoconnell:
The darker value is better. Thanks for changing that.

PR #1449, opened 2025-11-04
Add root padding parameter to TestHarness

PoignardAzur:
This solves a problem where a lot of screenshots are kind of hard to parse because they're really tight (especially annoying in doc screenshots).

Add some padding to one of the button screenshots.

I expect to add padding to a lot of screenshots, but I'd rather do it in a follow-up PR to avoid committing to many new images to the repository.


jaredoconnell (approved):
That example really shows the value of this change.

masonry_testing/src/harness.rs
```diff
@@ -161,6 +163,10 @@ pub struct TestHarnessParams {
     /// The background color of the virtual window.
     /// Defaults to [`Self::DEFAULT_BACKGROUND_COLOR`].
     pub background_color: Color,
+    /// The padding between the virtual window and the root widget.
+    ///
+    /// Defaults to zero, but [`Self::ROOT_PADDING`] is recommended.
```
DJMcNab, line 168: Is the expected workflow that each test is rewritten into the much more verbose form to be able to set this?

I'd probably lean towards changing the default (not in this PR, for the reasons you've noted); if the plan is to do that, we should say something more like:

```rust
/// Currently defaults to zero, however we expect this to change in the future.
/// As such, if you're writing a test which requires there to be zero padding, you must set this explicitly.
```

masonry_testing/src/harness.rs
```diff
@@ -161,6 +163,10 @@ pub struct TestHarnessParams {
     /// The background color of the virtual window.
     /// Defaults to [`Self::DEFAULT_BACKGROUND_COLOR`].
     pub background_color: Color,
+    /// The padding between the virtual window and the root widget.
```
DJMcNab, line 166: This is on each side, right?

PoignardAzur: Yes.

masonry_testing/src/harness.rs
PoignardAzur: I've thought about having a default padding value, and I'd rather avoid it.

Basically, my reasoning is that having no padding is never surprising, but if a user is wondering "why is this widget not at the position I expect?", the default extra padding is *really hard* to make discoverable.

I think extra test verbosity isn't a big problem, as long as each line has a clear purpose. Nobody is going to read `params.root_padding = 10.;` and wonder what it could possibly mean.

(I should probably document all this.)

DJMcNab: I don't really like the sound of this. The only way I think your position makes any sense is if you're drawing a distinction between screenshot tests and tests which don't use a screenshot. E.g. for a test without a screenshot which manually moves the mouse, having default padding is reasonably confusing.

That being said, I would expect tests which manually move the mouse to an absolute position to be rare. There are a couple, including in the (disabled) `masonry/src/tests/transforms.rs` module. (Incidentally, it isn't clear why that module isn't in the module tree, but its sibling `widget_tree` is also not included, so I have to assume it's intentional). It's also used for testing scrollbars and sliders (although I'd claim that a relative mouse move would probably be clearer for those).

> I think extra test verbosity isn't a big problem

I'm not happy with adding at least two lines of boilerplate to (almost) every screenshot test. It means that:

- There's a significant lift required to make a test "screenshot ready", if a test was already existing, you have to review all the code to make sure the new shrunken size doesn't change any of your maths.
- If a screenshot test doesn't have this padding, it isn't clear if that's because the test needs it not to exist (I don't even know what kind of test that would be?), or if that's because the test writer just forgot to add it when making it use a screenshot.

> Nobody is going to read `params.root_padding = 10.;` and wonder what it could possibly mean.

As noted above, the ambiguity in the lack of this would be what's confusing.

--- 

In my mind, the question becomes can we get the best of both worlds, i.e. make the padding be a default part of taking a screenshot test, but not have it impact the actual layout. That's pretty trivial because of how Vello's Scene's work; we only need to add twice the padding to each dimension, then transform the output scene by translating it by the padding.

If we want to go that way, obviously the implementation strategy for this PR changes a bit. I think having it be configurable does still make sense, although I can't really imagine any reasonable case where you'd choose to configure it.

PoignardAzur: The problems I'm anticipating aren't about manual mouse position or anything. I'm more thinking of a situation where someone is trying to adjust the size of a widget or its visual "breathing room" or something, and is wondering why the values they write don't match the screenshots they see.

> In my mind, the question becomes can we get the best of both worlds,

Not what you're thinking of, but we could add it as a second argument to `TestHarness::create_with_size`. That makes the diff for this PR much bigger, but I think it's a reasonable API compromise.

PoignardAzur:
I'm merging this, and leaving the API concerns for later.

PR #1450, opened 2025-11-04
masonry_core: fix wasm build

0x241F31:
I decided not to share example deps and relax examples' common deps versions instead.
I believe decoupling will make it easier to bootstrap new project via example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ peniko = "0.5.0"
 winit = "0.30.12"
 tracing = { version = "0.1.41", default-features = false }
 tracing-wasm = "0.2.1"
-console_error_panic_hook = "0.1.2"
+console_error_panic_hook = "0.1.7"
 ui-events = { version = "0.2.0", default-features = false, features = ["kurbo"] }
 ui-events-winit = { version = "0.2.0", default-features = false }
 smallvec = "1.15.1"
--- a/masonry_core/Cargo.toml
+++ b/masonry_core/Cargo.toml
@@ -39,9 +39,9 @@ ui-events.workspace = true
 vello.workspace = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-web-time.workspace = true
 console_error_panic_hook.workspace = true
 tracing-wasm.workspace = true
+web-time.workspace = true
 
 [dev-dependencies]
 assert_matches = "1.5.0"
--- a/xilem_web/web_examples/todomvc/Cargo.toml
+++ b/xilem_web/web_examples/todomvc/Cargo.toml
@@ -17,11 +17,11 @@ bench = false
 workspace = true
 
 [dependencies]
+console_error_panic_hook = "0.1"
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -55,6 +55,8 @@ parley = { version = "0.6.0", features = ["accesskit"] }
 peniko = "0.5.0"
 winit = "0.30.12"
 tracing = { version = "0.1.41", default-features = false }
+tracing-wasm = "0.2.1"
+console_error_panic_hook = "0.1.2"
```
DJMcNab: Why this version? It's not what's in Cargo.lock, which we generally want to synchronise our direct dependencies to.

xilem_web/web_examples/todomvc/Cargo.toml
```diff
@@ -17,11 +17,11 @@ bench = false
 workspace = true
 
 [dependencies]
-console_error_panic_hook = "0.1.7"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.145"
 tracing = "0.1.41"
-tracing-wasm = "0.2.1"
+tracing-wasm = "0.2"
+console_error_panic_hook = "0.1"
```
DJMcNab: It isn't clear to me why you've weakened these specifications either. Can you clarify that?

masonry_core/Cargo.toml
```diff
@@ -40,6 +40,8 @@ vello.workspace = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-time.workspace = true
+console_error_panic_hook.workspace = true
+tracing-wasm.workspace = true
```
waywardmonkeys, line 43: Alphabetize these ...

xilem_web/web_examples/todomvc/Cargo.toml
waywardmonkeys: Should just make them workspace deps and also keep these alphabetized.

waywardmonkeys: Well, maybe not for workspace deps since for some reason, none of these are workspace deps ...

masonry_core/Cargo.toml
0x241F31: done

xilem_web/web_examples/todomvc/Cargo.toml
0x241F31: alphabetized

Cargo.toml
0x241F31: my bad, set correct version: 0.1.7

xilem_web/web_examples/todomvc/Cargo.toml
0x241F31: > It isn't clear to me why you've weakened these specifications either. Can you clarify that?

Idea is to relax examples deps and sync major versions only, so updating will be less frequent and it will be easier for users to get started with examples

PR #1453, opened 2025-11-05
Implement `ViewArgument` for more tuples, and fixup docs

DJMcNab:
Thanks to Axel Karjalainen for catching the typo

Follow-up to #1444

### DIFF
--- a/xilem_core/src/state.rs
+++ b/xilem_core/src/state.rs
@@ -207,12 +207,21 @@ impl<T0: ViewArgument, T1: ViewArgument> ViewArgument for (T0, T1) {
 }
 
 // We use manual impls for 0, 1 and 2 to show the pattern
+
 viewargument_tuple!(T0, T1, T2);
 viewargument_tuple!(T0, T1, T2, T3);
 viewargument_tuple!(T0, T1, T2, T3, T4);
 viewargument_tuple!(T0, T1, T2, T3, T4, T5);
 viewargument_tuple!(T0, T1, T2, T3, T4, T5, T6);
 viewargument_tuple!(T0, T1, T2, T3, T4, T5, T6, T7);
+// 8 items (above) is likely the absolute maximum that anyone should be reasonably using.
+// Instead, they should make a custom struct to encapsulate these arguments.
+// However, we still support more, to avoid cliff edges.
+// Certainly if we had variadic generics, we wouldn't artificially limit it here.
+viewargument_tuple!(T0, T1, T2, T3, T4, T5, T6, T7, T8);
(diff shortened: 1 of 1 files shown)


xilem_core/src/state.rs
```diff
@@ -203,17 +206,32 @@ impl<T0: ViewArgument, T1: ViewArgument> ViewArgument for (T0, T1) {
...
-            T2::reborrow_mut(t2),
-        )
-    }
+// We use manual impls for 0, 1 and 2 to show the pattern
+viewargument_tuple!(T0, T1, T2);
+viewargument_tuple!(T0, T1, T2, T3);
+viewargument_tuple!(T0, T1, T2, T3, T4);
+viewargument_tuple!(T0, T1, T2, T3, T4, T5);
+viewargument_tuple!(T0, T1, T2, T3, T4, T5, T6);
+viewargument_tuple!(T0, T1, T2, T3, T4, T5, T6, T7);
```
Philipp-M, line 216: Is there a reason why we "only" support 8 arguments? (I think 12 would be a reasonable value to be in sync with std trait impls like `Default` or `Debug` ?)

DJMcNab: It's a fair question. My reasoning was that these values are:

1) Semantically meaningful (that is, it's not like Bevy's `Bundle` where the order doesn't really matater).
2) Inherently unnamed (that is, it's not like Bevy's `SystemParams`, where they're function arguments with a name)

As such, having even as many as 8 in a function is an anti-pattern, and you should be falling back to a custom struct.
That being said, it can still be an anti-pattern whilst being supported. These trait impls are also simple, so I don't think they'd cause any actual issues.

12 is the ecosystem standard, so I'll move it to 12.

xilem_core/src/state.rs
```diff
@@ -10,15 +10,18 @@
 ///
 /// This trait is used to talk about "reference" versions of states.
 ///
-/// This is implemented for [`Edit<T>`], [`Read<T>`], and tuples thereof.
+/// This is implemented for [`Edit<T>`], [`Read<T>`], and tuples of other implementations (of up to length 8).
```
Philipp-M, line 13: Not sure if intentional, but I think this should then also be 12?

xilem_core/src/state.rs
```diff
@@ -10,15 +10,18 @@
...
 /// It can also be implemented manually, which allows names to be given to fields.
+/// Note that if you need more than 8 items, you can either use a manual implementation, or nest tuples.
```
Philipp-M, line 15: Here as well?

PR #1454, opened 2025-11-05
Add test_property_set

PoignardAzur:
This lets us change default properties without having to reset all screenshots every time.

### DIFF
--- a/masonry/src/widgets/checkbox.rs
+++ b/masonry/src/widgets/checkbox.rs
@@ -361,8 +361,7 @@ mod tests {
         let widget = NewWidget::new(Checkbox::new(false, "Hello"));
 
         let window_size = Size::new(100.0, 40.0);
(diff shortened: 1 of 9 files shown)


masonry/src/theme.rs
```diff
@@ -146,3 +146,12 @@ pub fn default_text_styles(styles: &mut StyleSet) {
...
 }
+
+/// Set of default properties used in unit tests.
+///
+/// This lets us change default properties without having to reset all screenshots every time.
+/// This should still be kept relatively close to `default_property_set()` so that screenshots look like end user apps.
+#[cfg(test)]
+pub(crate) fn test_property_set() -> DefaultProperties {
+    default_property_set()
+}
```
DJMcNab, line 157: The only question is whether this should have a different visibility, but that can always be a follow-up.

PR #1457, opened 2025-11-06
Add screenshot padding in `assert_render_snapshot` instead of layout

DJMcNab:
Alternative to #1449.

This feels much nicer to me than the mechanism currently. It's much easier to set this as the default, as it won't impact any non-screenshot tests. Additionally, it also can't break the rendering of any existing tests, because the rendered scene is identical, just the scene content is translated. The added padding color also matches our tests in Parley.

For "display" screenshots, the padding having a background color isn't great; we could consider making that optional or configurable.

We could also use a more "vibrant" padding color. E.g. Parley uses `#a6c8ff`.

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -163,7 +163,7 @@ pub struct TestHarnessParams {
     /// The background color of the virtual window.
     /// Defaults to [`TestHarnessParams::DEFAULT_BACKGROUND_COLOR`].
     pub background_color: Color,
-    /// Extra padding added to screenshots in [`assert_render_snapshot!`].
+    /// Extra padding added to screenshots in [`assert_render_snapshot`].
     ///
     /// Currently defaults to zero, but we plan to change this
     /// to [`TestHarnessParams::ROOT_PADDING`] soon.
@@ -173,6 +173,8 @@ pub struct TestHarnessParams {
     ///    layout box (e.g. drop shadows, focus indicators).
     /// 2) For full apps, as it allows (manual) validation that none of the app content is cut off by
     ///    the window border.
+    ///
+    /// [`assert_render_snapshot`]: crate::assert_render_snapshot
     pub root_padding: u32,
     /// The maximum difference between two pixel channels before the harness will fail a screenshot test.
     /// Defaults to [`TestHarnessParams::DEFAULT_SCREENSHOT_TOLERANCE`].


masonry_testing/src/harness.rs
```diff
@@ -158,17 +158,24 @@ pub struct TestHarness<W: Widget> {
 #[non_exhaustive]
 pub struct TestHarnessParams {
     /// The size of the virtual window the harness renders into for snapshot testing.
-    /// Defaults to [`Self::DEFAULT_SIZE`].
+    /// Defaults to [`TestHarnessParams::DEFAULT_SIZE`].
```
DJMcNab, line 161: We already used the `TestHarnessParams` form for `KIBIBYTE` in `max_screenshot_size`, and this is what users would actually type.

masonry_testing/src/harness.rs
```diff
@@ -505,12 +528,11 @@ impl<W: Widget> TestHarness<W> {
             .expect("Got non-Send/Sync error from creating renderer")
         });
 
-        // TODO - fix window_size
```
DJMcNab, line 508: I can re-add this comment if I'm given an explanation of what it means (cc @PoignardAzur)

masonry_testing/src/screenshots.rs
```diff
@@ -24,12 +24,8 @@ pub(crate) fn get_image_diff(
...
     max_screenshot_tolerance: u32,
 ) -> Option<RgbImage> {
-    // TODO - Handle this case more gracefully.
-    assert_eq!(
-        (ref_image.width(), ref_image.height()),
-        (new_image.width(), new_image.height()),
-        "New image (right) has different size from old image (left)."
-    );
+    let ref_size = (ref_image.width(), ref_image.height());
+    let new_size = (new_image.width(), new_image.height());
```
DJMcNab, line 28: This was an easy lift, as the diff image computation already supported differently-sized images.

masonry_testing/src/harness.rs
PoignardAzur: I think this was about `window_size` not being the same type as the one we end up giving to wgpu. Not sure what the ideal behavior would be.

masonry_testing/src/screenshots.rs
PoignardAzur: This is definitely useful independently of the rest.

masonry_testing/src/harness.rs
```diff
@@ -465,13 +472,29 @@ impl<W: Widget> TestHarness<W> {
...
+    /// The returned image contains a bitmap (an array of pixels) as an 8-bits-per-channel RGB image.
+    ///
+    /// The returned image also does not apply [`TestHarnessParams::root_padding`].
+    /// To get an image which uses that, use [`render_with_padding`](Self::render_with_padding) with the required padding.
+    // TODO: There are some users of this function which just use it assert that `paint`/`compose` doesn't crash.
+    // Those could avoid actually performing a real render.
+    pub fn render(&mut self) -> RgbaImage {
+        // No padding
+        self.render_with_padding(0)
+    }
```
PoignardAzur: Independently of my other doubts, I find this method a little confusing. As a user of this API, I would expect `render()` to default to the padding set in `TestHarnessParams`.

(And also, I'm not really sold on having two different render methods. I don't think there's many realistic situations where a user would want the harness to both produce images with and without padding. I think `TestHarnessParams` is the right customization point for that.)

DJMcNab: The objective was so that toggling this to be default doesn't break any existing code, in case that made the assumption that there's no padding. I'll re-combine these.

DJMcNab:
> As I've said before, I don't like the idea of adding padding by default, because I think these kinds of "make the image nicer look by default" tricks are often badly discoverable, and make testing/tweaking harder. For example when I was working on layout code for text inputs, I had a few false starts because the TextInput widget had extra padding around it, also to look nice by default. For screenshot tests in particular, which are about correctness first and aesthetic second, I would want any padding to be explicitly "requested" by the user.

Would it be better for this to make the padding background colour a much more obviously "system" colour, e.g. the `#a6c8ff` used by Parley. That would make sure that this padding is obviously not something added by the user.
For doc screenshots, we'd obviously need to use a different colour for the padding in that case (i.e. probably just make it colourless). For that, we'd potentially also then document that any content which is overlapping the padding could become non-visible in a real app (e.g. the button focus indicator thing)

> The main problem I see with the Scene trick you use is that, if we later add some viewport cullling to Masonry, the screenshots will not reflect what someone would see of the widget "in the wild". I'd rather use some actual widget code, even if it ends up being a little verbose.

Can you speak more to this; I'm unsure how this would play out in practise.

I read your comment as saying that in the future we might not render widgets which are fully outside the viewport. In that case, for a screenshot test using this method, that any children of the test harness which are outside of the viewport wouldn't render. This would be as opposed to in a real app you would see them, as they would not be taking up the full viewport. A way I could imagine this coming into play would be if the widget had an offset in compose, but then undid that offset in paint?

However, my reading must be wrong, because it seems to me that this solution would be an improvement for that case. This is because a widget which relied upon its children being visible in this way would be wrong (e.g. the rendering would break if the widget were near the edge of the viewport in the wrong way). As such, the widget would need to be using whatever escape hatch we had (presumably the paint rect?), which your proposal does not enforce. That is, this solution seems to improve correctness if we gained that feature, rather than decreasing it.

masonry_testing/src/harness.rs
DJMcNab: I'll leave it gone then; the split out variables are easier to work with anyway.

PoignardAzur:
> Would it be better for this to make the padding background colour a much more obviously "system" colour, e.g. the #a6c8ff used by Parley.

Maybe. I feel like that'd just remove the "breathing room" effect that padding currently has visually.

> A way I could imagine this coming into play would be if the widget had an offset in compose, but then undid that offset in paint?

No, if it doesn't render in that case, that is indeed a bug.

What I have in mind is something like "You have a list of 10 rects, Rect n°6 is just barely offscreen, the screenshot doesn't show Rect n°6 because it's not included in the Scene, even though it's "logically" there".

DJMcNab:
> Maybe. I feel like that'd just remove the "breathing room" effect that padding currently has visually.

What I think we're reaching is that the padding has several distinct functionalities:

1) Showing that widgets don't draw outside of their bounds unexpectedly.
2) Allowing there to be extra room outside of widgets to make screenshots prettier.
3) Allowing the content that widgets draw outside of their bounds to be tested.

For case 1, I think we do want a highlighted different background, and I would probably want this to be on by default. For cases 2 and 3, we want something slightly different. For case 2, we probably want there to be a colourless padding highlight, so the background "extends". For case 3, this probably needs to be case-by-case. E.g. only used for tests with box shadows, and in that case you probably want a colorless background. But most of the tests of the same widget would still have the highlighted background. You might also want the same test twice, once with a colored background and once without, to allow showing that shadows interact properly with the actual layout box.

This strongly suggests that we should look at redesigning this API to allow setting the background color before turning it on as a default.

> What I have in mind is something like "You have a list of 10 rects, Rect n°6 is just barely offscreen, the screenshot doesn't show Rect n°6 because it's not included in the Scene, even though it's "logically" there".

To clarify, in that case you also think that the bottom padding shouldn't be there? Because I don't think that the treatment of this case in the existing implementation makes sense.

I suspect that the only cases where you'd have such a test might be when you're testing the viewport culling itself. In which case you actually want rectangle No 6 to be visible...

For screenshots of full pages, both for showing in docs and otherwise, I think the padding should probably be none (or maybe a couple of pixels?). That's because you want to be testing the content which is actually on-screen. Full pages are the only other case I could imagine that you'd get overflowing content.

I'd also claim that we should cross that bridge when we come to it, rather than thinking about a future hypothetical.

PR #1461, opened 2025-11-07
Refactor xilem_core module structure

PoignardAzur:
The overall goal of this PR is to ensure that major types/traits are declared in modules of the same name, and that module names reflect their most important item. This also avoids "kitchen sink" modules which are hard to nagivate.

- Split View impls into `views::impl_xxx` modules.
- Split ViewSequence impls into `view_sequences::impl_xxx` modules.
- Make any_view a submodule of `views` .
- Split off `WithoutElements` into its own module.
- Split off `ElementSplice` into its own module.
- Split off `ViewPathTracker` into its own module.
- Rename `context` to `message_context`.
- Rename `deferred` to `message_proxy`.
- Rename `state` to `view_argument`.
- Rename `sequence` to `view_sequence`.

This is a move-only PR with no logic changes.


xilem_core/src/views/impl_arc.rs
DJMcNab, line 1: I feel that this split goes too far the other way. I'd like to at least combine `Arc` and `Rc`, given how similar they are (that the module for `Arc` imports the state from the module for `Rc` really gives me pause).

I'd probably want to name that module something like `alloc`, and also include the `Box` impl there, although if you really want to split the box and reference counted ones, I guess that's fine. The module name isn't user-facing, at least.

xilem_core/src/element.rs
```diff
@@ -3,6 +3,8 @@
 
 //! The types which can be used as elements in a [`View`](crate::View)
 
+// TODO - Rename back to `Element`
```
DJMcNab: Why? This trait isn't user facing, except to people implementing peers of Xilem and Xilem Web. As such, it's something I actively don't want people to see and think it's important, and so it having an out-of-the-way name isn't a bad thing.

xilem_core/src/view_sequences/mod.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+mod impl_array;
+mod impl_option;
+mod impl_tuples;
+mod impl_vec;
+mod without_elements;
```
DJMcNab, line 8: I'm not a huge fan of these modules being in a completely random part of the module tree from the trait that they're implementing.

Why is this not a set of modules beneath `view_sequence.rs`?

DJMcNab (changes requested):
If the implementations for `ViewSequence` were moved around, I'd be happy landing this.
The original layout has come together in a pretty piecemeal way, so a little more thought is good.

I'm not enamoured by the `impl_`  module names, but I can't be bothered to argue against it.

xilem_core/src/element.rs
PoignardAzur: Fair enough, I don't feel strongly about that one. I might rename the module, then.

xilem_core/src/view_sequences/mod.rs
PoignardAzur: Overall I think it matches a pattern we've been applying elsewhere: declare a trait at the root (or for masonry, in a separate crate) and then have a folder with nothing but implementations of that trait.

DJMcNab: > with nothing but implementations of that trait.

Well the other cases are "nothing but implementers", not "nothing but implementations"  of that trait. I guess this was would be consistent if we gained manual implementations of ViewSequence, so we can probably leave it...

xilem_core/src/views/impl_rc.rs
```diff
@@ -0,0 +1,156 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use alloc::rc::Rc;
+use alloc::sync::Arc;
+use core::ops::Deref;
+
+use crate::message::MessageResult;
+use crate::{Arg, MessageContext, Mut, View, ViewArgument, ViewMarker, ViewPathTracker};
+
+#[allow(unnameable_types)] // reason: Implementation detail, public because of trait visibility rules
```
DJMcNab, line 11: This is a reminder that we need to change these to `expect` (not in this PR)

DJMcNab (approved):

xilem_core/src/views/impl_rc.rs
PoignardAzur: It's on my todo list. (Literally, I wrote it down.)

Issue #1395, opened 2025-09-22
Allowing users of Xilem and Masonry to opt out of wgpu's default features


StT191:
Vello just recently added the ability for crate users to opt out of `wgpu`'s default features (via [PR #1229](https://github.com/linebender/vello/pull/1229)).

In order for users of `xilem` or `masonry` to benefit from this they would need to introduce a possibly similar mechanism to opt out of `vello`'s default features, which default to enabling `wgpu`'s defaults.

I'm wondering what the most ergonomic design for users and crate maintainability would be. I imagine three possible scenarios:

1. `xilem` and `masonry` both introduce a new feature "wgpu_default" or "vello_default" that will be enabled by default. I'm not however convinced transitively adding a "wgpu_default" feature to all the crates that depend on `vello`/`wgpu` is desireble.
2. Just add "vello/wgpu_default" as a default feature in `xilem` and `masonry`. Semantically very similar to (1.) but there is no need to introduce new feature flags in `xilem` and `masonry`.
3. Introduce a hard breaking change: Don't enable `wgpu`'s defaults at all and document for the users very explicitly that they are responsible for enabling the right features for `wgpu` via a `wgpu` dependency in Cargo.toml, namely what backends wgpu compiles with and whether to enable the recommended default feature "parking_lot". Going with such a default would minimize the feature flags that crates depending on `wgpu` would have to introduce but puts the whole responsibility on the users. I think for library crates depending on `wgpu` this should be the default way to go when well documented.

I guess this needs further consideration.

DJMcNab:
I don't know that this is tracking anything meaningful anymore, so I'm going to close it.

@StT191 let us know if we should re-open it. I notice that you avoided using one of the trigger phrases in #1467, so I think it's possible there's a reason to have left this open.

PR #1467, opened 2025-11-09
Put Vello's default features behind the default feature flag in transitive dependents

StT191:
This enables users of `xilem` and the `masonry` crates to opt out of `vello`'s default features. As described in [#1395](https://github.com/linebender/xilem/issues/1395) this is particularly useful to opt out of `wgpu`'s default backends.

As `vello` is a transitive dependency to `xilem` and the `masonry` crates it is necessary to change the default features in all of them. It also enables use cases where a user would want to opt out of default features of one of the crates in this repository if they were to introduce default features in the future.

### DIFF
--- a/masonry_testing/Cargo.toml
+++ b/masonry_testing/Cargo.toml
@@ -16,7 +16,7 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = []
 
 [features]
-default = ["masonry_core/default"]
+default = []
 
 [dependencies]
 accesskit_consumer.workspace = true
--- a/masonry_testing/src/lib.rs
+++ b/masonry_testing/src/lib.rs
@@ -46,14 +46,6 @@
 //!
 //! - To test applications: see the tests in Masonry's examples.
 //! - To test widgets: see the `tests` module in each widget in Masonry.
-//!
-//! # Feature flags
-//!
-//! The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
-//!
-//! - `default`: Enables the default features of [`masonry_core`][masonry_core].
-//!
-//! [masonry_core]: https://crates.io/crates/masonry_core
 
 // TODO: Remove any items listed as "Deferred"
 #![expect(missing_debug_implementations, reason = "Deferred: Noisy")]
--- a/masonry_winit/Cargo.toml
+++ b/masonry_winit/Cargo.toml
@@ -18,7 +18,7 @@ targets = []
 cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
 
 [features]
-default = ["masonry_core/default"]
+default = []
 # Enables tracing using tracy if the default Masonry tracing is used.
 # https://github.com/wolfpld/tracy can be connected to when this feature is enabled.
 tracy = ["dep:tracing-tracy", "dep:wgpu-profiler", "wgpu-profiler/tracy", "masonry_core/tracy"]
--- a/masonry_winit/src/lib.rs
+++ b/masonry_winit/src/lib.rs
@@ -67,18 +67,8 @@
 //!
 //! (See the Masonry documentation for more detailed examples.)
 //!
-//! # Feature flags
-//!
-//! The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
-//!
-//! - `default`: Enables the default features of [`masonry_core`][masonry_core].
-//! - `tracy`: Enables creating output for the [Tracy](https://github.com/wolfpld/tracy) profiler using [`tracing-tracy`][tracing_tracy].
-//!   This can be used by installing Tracy and connecting to a Masonry with this feature enabled.
-//!
 //! [Masonry's documentation]: https://docs.rs/masonry
 //! [Masonry]: https://crates.io/crates/masonry
-//! [masonry_core]: https://crates.io/crates/masonry_core
-//! [tracing_tracy]: https://crates.io/crates/tracing-tracy
 
 // LINEBENDER LINT SET - lib.rs - v3
 // See https://linebender.org/wiki/canonical-lints/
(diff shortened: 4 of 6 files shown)


masonry/Cargo.toml
```diff
@@ -35,8 +35,8 @@ assert_matches = "1.5.0"
 float-cmp = { version = "0.10.0", features = ["std"], default-features = false }
 image = { workspace = true, features = ["png"] }
 insta = { version = "1.43.2" }
-masonry_testing = { workspace = true }
-masonry_winit.workspace = true
+masonry_testing = { workspace = true, features = ["default"] }
+masonry_winit = { workspace = true, features = ["default"] }
```
PoignardAzur, line 39: Wait, shouldn't the `"default"` feature in these two cases be gated on the `default` feature of Masonry above?

StT191: I'm not sure. Since those are dev-dependencies I assumed it would be better to leave them with the default features. But I'd be happy to change it if you think that would be desirable.

PoignardAzur: Oh, that makes sense, yeah.

masonry/src/lib.rs
```diff
@@ -163,6 +165,8 @@
 //!
 //! If you want to use your own subscriber, simply set it before starting masonry - in this case masonry will not set a subscriber.
 //!
+//! [masonry_core]: https://crates.io/crates/masonry_core
+//! [masonry_testing]: https://crates.io/crates/masonry_testing
```
DJMcNab, line 169: This is wrong. We want to link to the local docs of Masonry Core/Masonry Testing, if you have them.
```suggestion
```

We expect to be documented with all features enabled, and are fine having broken intra-doc-links if that isn't the case.

DJMcNab: It is correct that Masonry Testing isn't linked to correctly on main, but the correct way to fix that is to add the links to the block of links at the top of README.md. I can do that as a follow-up if you'd prefer

masonry_core/src/lib.rs
```diff
@@ -39,8 +39,17 @@
...
+//!
+//! The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
+//!
+//! - `default`: Enables the default features of [`vello`].
+//! - `tracy`: Enables creating output for the [Tracy](https://github.com/wolfpld/tracy) profiler using [`tracing-tracy`][tracing_tracy].
+//!   This can be used by installing Tracy and connecting to a Masonry with this feature enabled.
+//!
 //! [Masonry's documentation]: https://docs.rs/masonry/latest/
 //! [Masonry Winit]: https://docs.rs/masonry_winit/latest/
+//! [tracing_tracy]: https://crates.io/crates/tracing-tracy
```
DJMcNab, line 52: ❤️

DJMcNab, line 52: (The reason the same thing doesn't apply to `tracing_tracy` here is not because we shouldn't handle it the same way; instead it's much less important, as it's very unlikely that anyone would want to explore the docs of that library from this crate, so the homepage is as appropriate)

masonry_testing/src/lib.rs
```diff
@@ -46,6 +46,14 @@
...
 //! - To test applications: see the tests in Masonry's examples.
 //! - To test widgets: see the `tests` module in each widget in Masonry.
+//!
+//! # Feature flags
+//!
+//! The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
+//!
+//! - `default`: Enables the default features of [`masonry_core`][masonry_core].
+//!
+//! [masonry_core]: https://crates.io/crates/masonry_core
```
DJMcNab: Same applies here. Again, if needed I can make it be a follow-up.
```suggestion
```

masonry_winit/src/lib.rs
```diff
@@ -67,8 +67,18 @@
...
+//!
+//! The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
+//!
+//! - `default`: Enables the default features of [`masonry_core`][masonry_core].
+//! - `tracy`: Enables creating output for the [Tracy](https://github.com/wolfpld/tracy) profiler using [`tracing-tracy`][tracing_tracy].
+//!   This can be used by installing Tracy and connecting to a Masonry with this feature enabled.
+//!
 //! [Masonry's documentation]: https://docs.rs/masonry
 //! [Masonry]: https://crates.io/crates/masonry
+//! [masonry_core]: https://crates.io/crates/masonry_core
```
DJMcNab:
```suggestion
```

masonry_testing/Cargo.toml
```diff
@@ -16,7 +16,7 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = []
 
 [features]
-default = []
+default = ["masonry_core/default"]
```
DJMcNab: I don't think this feature edge is necessary, and I'd lean slightly towards not having it. Masonry Testing pretty intentionally doesn't re-export Masonry Core, so I don't think it should be in the business of enabling this feature.

That is, if you have a dependency on Masonry Testing (or Masonry Winit), to make it useful you by definition also have a dependency on Masonry or Masonry Core. That peer dependency should be the one enabling this kind of feature.

This is the only concern I'd treat as blocking (and to be clear, a reason that my understanding is wrong would also be sufficient to "unblock")

masonry_winit/Cargo.toml
```diff
@@ -18,7 +18,7 @@ targets = []
 cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
 
 [features]
-default = []
+default = ["masonry_core/default"]
```
DJMcNab: Similarly here.

masonry_testing/Cargo.toml
StT191: I agree. I'll remove it ... coming up.

masonry/src/lib.rs
StT191: Yeah, it is a bit confusing to me. I'd be happy if you added a commit to fix the issues.

PR #1468, opened 2025-11-09
tree_arena: Support reparenting of root nodes

no-materials:
Enables moving root nodes under new parents, removing the previous restriction that caused a panic.

Both the safe and unsafe tree implementations now handle root reparenting by detaching the node from roots before inserting it under the new parent.


PoignardAzur (approved):
Sorry, I forgot about this.

The logic makes sense, and I appreciate the added test. Thanks!

PR #1469, opened 2025-11-10
Remove or improve `#[allow(...)]` annotations

PoignardAzur:


xilem_core/src/views/any_view.rs
```diff
@@ -177,7 +177,6 @@ where
 
 /// The state used by [`AnyView`].
 #[doc(hidden)]
-#[allow(unnameable_types)] // reason: Implementation detail, public because of trait visibility rules
```
DJMcNab, line 180: Thanks for calling this out - I'll fix the use of the re-export.

PR #1470, opened 2025-11-10
Rename MessageContext to MessageCtx

PoignardAzur:
This is more consistent with the naming scheme used elsewhere in the repo.

This is a rename-only PR with no behavior changes.


jaredoconnell (commented):
This does appear to make sense. It looks like a few more spots need to be updated, and it needs to be made up to date with main.

PR #1482, opened 2025-11-16
Add `Passthrough` widget and use it for Xilem `AnyView` 

PoignardAzur:
Introduce `masonry::widgets::Passthrough`, a pass‑through single‑child
container that hosts exactly one child and supports dynamic replacement.

In Xilem, we replace the bespoke `DynWidget` with `Passthrough`.
`AnyWidgetView` now uses `Pod<Passthrough>`.

### DIFF
--- a/masonry/src/widgets/passthrough.rs
+++ b/masonry/src/widgets/passthrough.rs
@@ -109,7 +109,7 @@ impl Widget for Passthrough {
 mod tests {
     use super::*;
     use crate::testing::{TestHarness, assert_render_snapshot};
-    use crate::theme::default_property_set;
+    use crate::theme::test_property_set;
     use crate::widgets::Label;
     use vello::kurbo::Size;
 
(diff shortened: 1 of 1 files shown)


PoignardAzur:
Based on #1446.

masonry/src/widgets/passthrough.rs
```diff
@@ -0,0 +1,159 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::testing::{TestHarness, assert_render_snapshot};
+    use crate::theme::default_property_set;
+    use crate::widgets::{Button, Label};
+    use vello::kurbo::Size;
+
+    #[test]
+    fn content_host_replaces_child() {
```
RagibHasin: Shouldn't the tests and screenshots be renamed to passthrough?

PoignardAzur: Good catch!

RagibHasin (approved):

PR #1484, opened 2025-11-21
Split android examples into separate files

PoignardAzur:
This change stops cargo from printing a dozen warning every time we run anything.

### DIFF
--- a/xilem/examples/android/to_do_mvc_android.rs
+++ b/xilem/examples/android/to_do_mvc_android.rs
@@ -3,7 +3,7 @@
 
 //! A to-do-list app, loosely inspired by todomvc.
 
-#[cfg(target_os = "android")]
+#[expect(dead_code, reason = "original main is unused")]
 #[path = "../to_do_mvc.rs"]
 mod to_do_mvc;
 


xilem/examples/emoji_picker.rs
DJMcNab: You've changed this from a suffix to a prefix, so these reasons also need to be updated.

Also, the `android_` version is in a different file, so I suspect this message should be further changed/clarified.

xilem/examples/android/readme.md
DJMcNab, line 1: I'd normally expect this file name to be capitalised, i.e. `README.md`

xilem/examples/calc.rs
```diff
@@ -336,26 +336,7 @@ fn run(event_loop: EventLoopBuilder) -> Result<(), EventLoopError> {
 
 // Boilerplate code: Identical across all applications which support Android
 
-#[expect(clippy::allow_attributes, reason = "No way to specify the condition")]
-#[allow(dead_code, reason = "False positive: needed in not-_android version")]
-// This is treated as dead code by the Android version of the example, but is actually live
-// This hackery is required because Cargo doesn't care to support this use case, of one
-// example which works across Android and desktop
+#[cfg_attr(target_os = "android", allow(dead_code, reason = "see android_main"))]
```
DJMcNab: I'm not sure that "see android_main" is a helpful message here, as `android_main` doesn't exist in this file. Maybe `unused when imported by examples/android`?

xilem/examples/android/calc_android.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright 2024 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! A simple calculator example
+
+#[cfg(target_os = "android")]
+#[path = "../calc.rs"]
+mod calc;
```
DJMcNab: There are *arguments* for using `include!("../calc.rs")` here. That would avoid the need for the `pub(crate)` in the examples. It somewhat depends on whether Rust Analyzer properly links things if you use `include!`. This isn't blocking; this way around is also fine, though.

DJMcNab: An alternative interesting thing we could try is having an `#[expect(dead_code)]` *here*, which should actually remove the need for the equivalent in the non `_android` examples.  I don't think that can be combined with `include!()` meaningfully, though. This also isn't blocking.

PoignardAzur: Ooh, I like that solution.

PoignardAzur: I'd rather avoid using a string include. It's less intuitive for users and it breaks things like go-to-definition.

xilem/examples/android/readme.md
PoignardAzur: This is deliberate. See also `masonry/examples/readme.md`.

My rule of thumb is I use uppercase names for project readmes and lowercase names for folder readmes.

PoignardAzur:
Merging now. If the `xilem/examples/android/readme.md` capitalization is a problem, we should address it in a future PR along with `masonry/examples/readme.md`.

PR #1491, opened 2025-11-25
Remove duplicate flag updates.

xStrom:
As [discussed on Zulip](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Explicit.20vs.20implicit.20state.20change.20flags) we don't need these explicit duplicate flag updates. The noise they add to the code isn't worth the benefits.

### DIFF
--- a/masonry/src/widgets/resize_observer.rs
+++ b/masonry/src/widgets/resize_observer.rs
@@ -68,9 +68,6 @@ impl ResizeObserver {
     pub fn set_child(this: &mut WidgetMut<'_, Self>, child: NewWidget<impl Widget + ?Sized>) {
         let old_child = mem::replace(&mut this.widget.child, child.erased().to_pod());
         this.ctx.remove_child(old_child);
-        // Force a re-send when the layout changes.
-        // This might be unnecessary, but it also shouldn't hurt.
-        this.widget.size = None;
     }
 
     /// Force this layout observer to send a new action.


masonry/src/widgets/resize_observer.rs
```diff
@@ -71,8 +71,6 @@ impl ResizeObserver {
         // Force a re-send when the layout changes.
```
waywardmonkeys: Is this referring to resending children_changed / request_layout? (Is this comment still accurate?)

masonry/src/widgets/flex.rs
```diff
@@ -373,8 +373,10 @@ impl Flex {
         let child = this.widget.children.remove(idx);
         if let Child::Fixed { widget, .. } | Child::Flex { widget, .. } = child {
             this.ctx.remove_child(widget);
+        } else {
+            // We need to explicitly request layout in case of spacer removal
+            this.ctx.request_layout();
         }
-        this.ctx.request_layout();
```
waywardmonkeys, line 377: Not thrilled about this small part of the change, but I understand why.

masonry/src/widgets/resize_observer.rs
DJMcNab: This comment refers to resending a `LayoutChanged`? I agree that "when the layout changes" is not a very clear comment. We should say something like "Send a LayoutChanged when the contained widget changes. This might be unnecessary, but it also shouldn't hurt."

DJMcNab (approved):
I'm happy approving this, whether or not the the comment is changed. It's not clear if Bruce had any blockers.

masonry/src/widgets/resize_observer.rs
waywardmonkeys: Right, but was that re-send done by which of the 3 lines below:  Setting the widget size to none or the children_changed or the request_layout? The latter 2 are gone, so the comment now feels questionable. That's all ...

No blockers though.

xStrom: No, this is referring to `LayoutChanged` which is a `ResizeObserver` thing. I'll at minimum update the comment to be more clear about this.

masonry/src/widgets/flex.rs
xStrom: It's a nano-optimization! 😈

masonry/src/widgets/resize_observer.rs
xStrom: After reading through the code in more detail I decided to just remove that line.

`ResizeObserver` docs state: *A widget which sends a `LayoutChanged` whenever its size changes.*

Well, that contract is true and upheld by its `layout` method. I don't think we need to also send `LayoutChanged` when the size stays the same but the contents changed. That is a different contract!

PR #1492, opened 2025-11-25
Create FocusedBorderColor property

PoignardAzur:


masonry/src/widgets/checkbox.rs
```diff
@@ -280,14 +286,14 @@ impl Widget for Checkbox {
...
-        if ctx.is_focus_target() || is_hovered {
+        if is_focused || is_hovered {
             let widget_size = ctx.size();
 
             let focus_rect = Rect::new(0.0, 0.0, widget_size.width, widget_size.height);
 
             let focus_rect = focus_rect.inflate(2.0, 2.0);
 
-            let focus_color = theme::FOCUS_COLOR;
+            let focus_color = props.get::<FocusedBorderColor>().0.color;
```
waywardmonkeys, line 296: If you can impl `Into<BrushRef<'b>>` for `FocusedBorderColor`, then you can just:

```
let focus_color = props.get::<FocusedBorderColor>();
```

waywardmonkeys: Actually, maybe better to just do it for `BorderColor` and then you can simplify some other stuff as well later.

waywardmonkeys (approved):
Overall, looks good. I would probably impl `Into<BrushRef>` for `BorderColor` to make some code nicer as I suggest in comments.

PR #1494, opened 2025-11-25
Add efficient child replacement methods to widgets.

xStrom:
The only way to update children of `Flex`, `Grid`, `IndexedStack`, and `ZStack` is with `push` / `insert` / `remove`. This is quite inefficient. In order to replace the 0-index child, you have to first call `remove` and then immediately combo it with `insert`. Both methods are `O(n)` as they will shift the whole collection.

This PR introduces new `set_child` methods to these widgets, allowing for `O(1)` replacement of children.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -362,6 +362,84 @@ impl Flex {
         this.ctx.request_layout();
     }
 
+    /// Replace the child widget at the given index with a new non-flex one.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is out of bounds.
+    pub fn set_child(
+        this: &mut WidgetMut<'_, Self>,
+        idx: usize,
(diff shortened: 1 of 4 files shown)


masonry/src/widgets/flex.rs
```diff
@@ -1262,18 +1340,27 @@ mod tests {
...
             harness.edit_root_widget(|mut flex| {
+                Flex::set_child(&mut flex, 0, Label::new("a").with_auto_id());
+                // -> abwd
+                Flex::set_flex_child(
+                    &mut flex,
+                    2,
+                    Label::new("c").with_auto_id(),
+                    FlexParams::default(),
+                );
+                // -> abcd
```
PoignardAzur, line 1363: It might be interesting to replace the children with a different type of widget to show that it's possible.

xStrom: It's an important property, for sure. I'll punt on it myself right now though.

PR #1499, opened 2025-11-28
Split off `WindowView` code into `MasonryRoot`

PoignardAzur:
`MasonryRoot` holds all the winit-agnostic parts of `WindowView`. This is a pre-requisite to splitting off `xilem_masonry`.

### DIFF
--- a/xilem/src/window_view.rs
+++ b/xilem/src/window_view.rs
@@ -12,7 +12,7 @@ use crate::{AnyWidgetView, InitialRootWidget, MasonryRoot, ViewCtx, WidgetView, 
 pub struct WindowView<State: 'static> {
     pub(crate) id: WindowId,
     pub(crate) options: WindowOptions<State>,
-    pub(crate) render_root_view: MasonryRoot<State>,
+    pub(crate) masonry_root: MasonryRoot<State>,
     /// The base color of the window.
     pub(crate) base_color: Color,
 }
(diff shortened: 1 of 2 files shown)


xilem/src/render_root_view.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use masonry::app::RenderRoot;
+use masonry::widgets::Passthrough;
+use xilem_core::Edit;
+
+use crate::core::{Arg, MessageCtx, Mut, View, ViewElement, ViewMarker};
+use crate::{AnyWidgetView, Pod, ViewCtx, WidgetView};
+
+/// A view representing a [`RenderRoot`].
+pub struct RenderRootView<State: 'static> {
```
Philipp-M: I'm still not super happy about the name, I think something like `MasonryRootView` or just `MasonryRoot` might be better as this seems to be the (future) purpose of it. IMO `Render` is too generic (and often has a more graphics related meaning) for it and not really clear.

xilem/src/window_view.rs
```diff
@@ -1,22 +1,21 @@
...
-use crate::{AnyWidgetView, ViewCtx, WidgetView, WindowOptions};
+use crate::{AnyWidgetView, InitialRootWidget, MasonryRoot, ViewCtx, WidgetView, WindowOptions};
 
 /// A view representing a window.
 pub struct WindowView<State: 'static> {
     pub(crate) id: WindowId,
-    options: WindowOptions<State>,
-    root_widget_view: Box<AnyWidgetView<Edit<State>, ()>>,
+    pub(crate) options: WindowOptions<State>,
+    pub(crate) render_root_view: MasonryRoot<State>,
```
xStrom: If we're changing it to `MasonryRoot` then should probably rename this field too.

```suggestion
    pub(crate) masonry_root: MasonryRoot<State>,
```

PR #1500, opened 2025-11-28
Remove `ViewCtx::state_changed`

PoignardAzur:
Field was set in a few places but never read.


xilem/src/view_ctx.rs
```diff
@@ -90,14 +84,6 @@ impl ViewCtx {
         self.widget_map.insert(id, path);
     }
 
-    /// Whether the app's state changed since the last rebuild.
-    ///
-    /// This is useful for views whose current value depends on current app state.
-    /// (That is, currently only virtual scrolling)
```
xStrom, line 96: Seems like this was added for the benefit of `VirtualScroll`. @DJMcNab is that no longer the case or is the current `VirtualScroll` just buggy in not using this flag?

PR #1503, opened 2025-11-29
Split off code into new `xilem_masonry` crate

PoignardAzur:
This will be useful for:

- Embedding Xilem in non-winit targets.
- Build parallelism of Xilem.
- Adding testing features in the future.

### DIFF
--- a/xilem_masonry/README.md
+++ b/xilem_masonry/README.md
@@ -2,7 +2,7 @@
 
 <div align="center">
 
-**Masonry frontend for the Xilem Rust UI framework**
+**Masonry backend for the Xilem Rust UI framework**
 
 <!-- TODO Add shields for crates.io and docs.rs -->
 [![Apache 2.0 license.](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](#license)
--- a/xilem_masonry/Cargo.toml
+++ b/xilem_masonry/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "xilem_masonry"
 version.workspace = true
-description = "Masonry frontend for the Xilem Rust UI framework."
+description = "Masonry backend for the Xilem Rust UI framework."
 keywords = ["gui", "ui", "native", "gpu", "performance"]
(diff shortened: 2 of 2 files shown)


xilem_masonry/README.md
```diff
@@ -0,0 +1,65 @@
+# Xilem Masonry
+
+<div align="center">
+
+**Masonry frontend for the Xilem Rust UI framework**
```
RagibHasin: Should this not be _backend_?

PoignardAzur: Thank you, corrected.

xilem_masonry/src/view/button.rs
```diff
@@ -34,7 +34,9 @@ use crate::{Pod, ViewCtx, WidgetView};
 /// To create a simple button with styled text:
 ///
 /// ```
-/// use xilem::{view::{button, label}, FontWeight};
+/// # use xilem_masonry as xilem;
```
Philipp-M, line 37:
```suggestion
```

Can we avoid these changes in all the doc-examples? They shouldn't be necessary, and will likely be out of sync at some time. Btw. (as I think I was the author of at least a few of those examples), I intentionally have put the `use` statements all on one line (if not unnecessarily long) to shift focus to the actual interesting part, as these are often autocompleted/imported anyway.

Philipp-M: Hmm, I just noticed it's not easily possible, well... unfortunate... We probably need a way to enforce that the API works as expected...

PoignardAzur: > I intentionally have put the use statements all on one line (if not unnecessarily long) to shift focus to the actual interesting part, as these are often autocompleted/imported anyway.

Interesting. This is a good idea, and I think it should be documented policy in our xilem/ARCHITECTURE.md file. I'll open a PR.

PoignardAzur:
Since this is a pretty big change, I'll wait until ~Tuesday to leave everyone a chance to see it, and then merge it unless anyone objects.

PR #1506, opened 2025-11-30
Add Canvas widget - take 3

PoignardAzur:
Based on @richard-uk1's https://github.com/linebender/xilem/pull/875 and @philocalyst's https://github.com/linebender/xilem/pull/1445.

Add Canvas widget to Masonry.
Add canvas element to Xilem.

### DIFF
--- a/xilem/src/view/canvas.rs
+++ b/xilem/src/view/canvas.rs
@@ -13,8 +13,10 @@ use crate::{Pod, ViewCtx};
 /// # Example
 ///
 /// ```
-/// use xilem::masonry::{kurbo::{Rect, Size}, peniko::Fill, vello::Scene};
-/// use xilem::{Affine, view::canvas};
+/// use xilem::masonry::kurbo::{Rect, Size, Affine};
+/// use xilem::masonry::peniko::Fill;
+/// use xilem::masonry::vello::Scene;
+/// use xilem::view::canvas;
 /// use std::sync::Arc;
 /// # use xilem::WidgetView;
 ///


RagibHasin (approved):
Looks good to me. Sorry for the late review.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,231 @@
...
+use vello::Scene;
+use vello::kurbo::Size;
+
+use crate::core::{
+    AccessCtx, AccessEvent, BoxConstraints, EventCtx, LayoutCtx, PaintCtx, PointerEvent,
+    PropertiesMut, PropertiesRef, RegisterCtx, TextEvent, Update, UpdateCtx, Widget, WidgetId,
+    WidgetMut,
+};
+
+// TODO - Add background color?
```
RagibHasin, line 20: I think it is better done in the `draw` function?

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,231 @@
...
+impl Widget for Canvas {
+    type Action = NoAction;
+
+    fn on_pointer_event(
+        &mut self,
+        _ctx: &mut EventCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        _event: &PointerEvent,
+    ) {
+    }
```
RagibHasin, line 107: Shouldn't methods with default impls be omitted?

masonry/src/widgets/canvas.rs
Philipp-M: Yeah I would just remove TODO, I think it's a good default to draw nothing when nothing is specified in `draw`.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,231 @@
...
+// TODO - Add background color?
+
+/// A widget allowing custom drawing.
+///
+/// A canvas takes a painter callback; every time the canvas is repainted, that callback
+/// in run with a [`Scene`].
+/// That Scene is then used as the canvas' contents.
+pub struct Canvas {
+    draw: Arc<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>,
+    alt_text: Option<String>,
```
Philipp-M, line 29: Doesn't hurt to have an `Option` here, but it's not necessary, we can just use the property `is_empty` for that.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,231 @@
...
+        Self::set_painter_arc(this, Arc::new(draw));
+    }
+
+    /// Update the draw function.
+    pub fn set_painter_arc(
+        this: &mut WidgetMut<'_, Self>,
+        draw: Arc<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>,
+    ) {
+        this.widget.draw = draw;
+        this.ctx.request_render();
```
Philipp-M, line 77: Something I'm wondering for a longer time already, is whether we should check within these setters whether really something changed (i.e. in this case if the arc pointer has changed), and only then request a render.

As we often (always?) do this in Xilem it's not necessary here for Xilem at least, but I think it might be nice to have this built into masonry?

See also https://github.com/ricvelozo/xilem/pull/1/files#diff-7d717e7fd640e50cef624a6fa2afb9c20968470e5849c8b3b0abe9782499814eR67-R72
where I did this.

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,231 @@
...
+        this.ctx.request_accessibility_update();
+    }
+
+    /// Remove the canvas' alt text.
+    ///
+    /// See [`Canvas::with_alt_text`] for details.
+    pub fn remove_alt_text(mut this: WidgetMut<'_, Self>) {
+        this.widget.alt_text = None;
+        this.ctx.request_accessibility_update();
+    }
```
Philipp-M, line 94: See the comment below, I think we can just sum this up in `set_alt_text` with `alt_text: String` (or `ArcStr`)

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,112 @@
...
+        _: Arg<'_, State>,
+    ) {
+        if !Arc::ptr_eq(&self.draw, &prev.draw) {
+            widgets::Canvas::set_painter_arc(&mut element, self.draw.clone());
+        }
+        if self.alt_text != prev.alt_text
+            && let Some(alt_text) = &self.alt_text
+        {
+            widgets::Canvas::set_alt_text(element, alt_text.clone());
+        }
```
Philipp-M, line 94: I don't think this is safe?

When alt_test was `Some` and changed to `None` it won't be updated in masonry AFAICS.
So in the current state of this PR `remove_alt_text` needs to be called (but as described above, I think we can simplify this and the masonry code by not using an `Option` at all)

Philipp-M (changes requested):
There's an issue in the `View::rebuild` implementation, otherwise looks good.

xilem/src/view/canvas.rs
```diff
@@ -0,0 +1,112 @@
...
+///         Affine::IDENTITY,
+///         xilem::palette::css::AQUA,
+///         None,
+///         &Rect::new(0.0, 0.0, size.width, size.height),
+///     );
+/// }));
+/// # my_canvas
+/// # }
+/// ```
+pub fn canvas(draw: Arc<dyn Fn(&mut Scene, Size) + Send + Sync + 'static>) -> Canvas {
```
Philipp-M, line 37: Btw. I think we should probably specify in the doc-comment, that this `Arc<dyn Fn>` should be ideally saved within the app state, otherwise it always results in a rerender. Ideally show how this is done better in the doc-comment example

Philipp-M: Maybe we can also use a function pointer instead, to avoid allocs and (mostly) unnecessary generics?

masonry/src/widgets/canvas.rs
PoignardAzur: The thing is, "alt text is unset" and "alt text is set to empty string" have different semantic meanings, though I'm not sure whether screen readers really interpret them that differently in practice.

masonry/src/widgets/canvas.rs
PoignardAzur: I don't know. In general, that kind of reasoning is how you end up with multiple layers of caching in your app that all check whether some parameter changed.

In practice, repainting is usually cheap and having Masonry always repaint something when a parameter is set is more predictable, so I'd lean towards the current design.

masonry/src/widgets/canvas.rs
PoignardAzur: I think we might just want to pass an `Option<String>` and merge them.

PR #1507, opened 2025-11-30
Write guidelines for writing Xilem doc examples

PoignardAzur:
Note that the "include `fn view() -> impl View` in examples" guideline goes against current practice, but I think current practice should change. The two lines saved aren't worth not being able to copy-paste examples into live code.

### DIFF
--- a/xilem/ARCHITECTURE.md
+++ b/xilem/ARCHITECTURE.md
@@ -52,6 +52,22 @@ The Xilem project includes these crates:
 - **`xilem`:** Natively compiled backend for Xilem. Depends on `xilem_core`, `masonry` and `masonry_winit`.
 
 
+## Doc examples
+
+Code snippets in docs comments in both `xilem` and `xilem_masonry` should be as self-contained as possible.
+
+There should only be a single (visible) import statement at the beginning, listing all the needed imports.
+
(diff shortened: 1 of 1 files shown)


xilem/ARCHITECTURE.md
```diff
@@ -52,6 +52,22 @@ The Xilem project includes these crates:
...
+There should only be a single (visible) import statement at the beginning, listing all the needed imports.
+
+That import statement should start with `use xilem::`; in other words, all examples should assume that Xilem is the only available dependency (aside from the standard library).
+
+It should be possible for users to copy-paste the example code into an empty module and have it compile.
+As such, we should avoid `# ` prefixes when possible.
+View examples should include the whole `fn view() -> impl View<...> {` and `}` parts.
+
+*Note:* Some examples in `xilem` don't currently respect these guidelines.
+We should fix those examples, and make sure future examples do.
```
Philipp-M, line 68: I don't have a strong opinion on that, but I think having short focused examples that are obvious in where they can be used are good enough (i.e. within component functions). The Rust standard library also doesn't wrap every example in functions. I just like to include the `fn view() -> impl View<...>` to have the CI type check that the returned type is really a view (as we sometimes have rather complicated type-bounds there).

PR #1510, opened 2025-12-03
Add `CollectionWidget` and implement it for collection widgets.

xStrom:
## Summary

The primary goal of this PR is to add the `CollectionWidget` trait to `masonry_core/src/core/widget.rs`.

The motivation is to be able to write generic collection widget management code on top of Masonry. Having a trait that guarantees this common functionality achieves that.

## `CollectionWidget`

The trait is essentially the following:

```rust
pub trait CollectionWidget<Params>: Widget {
    fn len(&self) -> usize;
    fn is_empty(&self) -> bool;
    fn get_mut<'t>(this: &'t mut WidgetMut<'_, Self>, idx: usize) -> WidgetMut<'t, dyn Widget>;
    fn add(this: &mut WidgetMut<'_, Self>, child: NewWidget<impl Widget + ?Sized>, params: impl Into<Params>);
    fn insert(this: &mut WidgetMut<'_, Self>, idx: usize, child: NewWidget<impl Widget + ?Sized>, params: impl Into<Params>);
    fn set(this: &mut WidgetMut<'_, Self>, idx: usize, child: NewWidget<impl Widget + ?Sized>, params: impl Into<Params>);
    fn set_params(this: &mut WidgetMut<'_, Self>, idx: usize, params: impl Into<Params>);
    fn swap(this: &mut WidgetMut<'_, Self>, a: usize, b: usize);
    fn remove(this: &mut WidgetMut<'_, Self>, idx: usize);
    fn clear(this: &mut WidgetMut<'_, Self>);
}
```
## Implementation

Some collection widgets like `Flex` already implemented all of this functionality, while others like `Grid` were missing some parts.

I had recently already standardized some of these function names in #1497 and this trait takes it even further by making it guaranteed by the compiler. Also, I have chosen to shorten the names. I don't think there is a lot of value to have a `_child` suffix on all of these functions. The most common use case of `add` into a `Flex` is indeed with a child. Other specifics can have qualifiers like `add_spacer`.

Because this trait requries the vanilla `add` to take `params`, this meant that `Flex` had to swap around some function names. `Flex` used to have `add_child` and `add_flex_child`, and now it has `add_fixed` and `add`.

`Flex::get_mut` now also panics when the passed index is for a spacer. Other functions like `update_child_flex_params` already had this panic behavior, expecting callers to know which indices are safe. Indeed we can see that e.g. Xilem was always immediately unwraping anyway.

Not all collection widgets actually need `params` and so they will use the unit type `()` as their params type. It's a bit quirky to have to do `IndexedStack::add(this, child, ())` but it's a small cost to pay for getting the genericity. Especially as usually these functions will be used by the UI framework on top of Masonry, and not by app developers directly.

Similar rationale applies for why having to import `CollectionWidget` to use these functions is fine. The import will be done by the UI framework.

### DIFF
--- a/masonry/README.md
+++ b/masonry/README.md
@@ -109,7 +109,7 @@ impl AppDriver for Driver {
             });
             render_root.edit_widget_with_tag(LIST_TAG, |mut list| {
                 let child = Label::new(self.next_task.clone()).with_auto_id();
-                Flex::add_child(&mut list, child);
+                Flex::add_fixed(&mut list, child);
             });
         } else if action.is::<TextAction>() {
             let action = action.downcast::<TextAction>().unwrap();


xStrom:
`Flex` has a bunch of quriks with the spacers. I feel like the spacer functionality could be achieved with some margin style props and that would simplify the API quite a bit. However that's a topic for a different PR.

PR #1512, opened 2025-12-03
Fix scrollbar rendering bug

PoignardAzur:
This is a very unprincipled fix, but the short version is that reducing the scrollbar radius seems to stop the bug from occurring.


PoignardAzur:
See also https://github.com/linebender/vello/issues/1314

PoignardAzur:
> Opus 4.5 says the sparse strips stuff would fix this and make a hacky fix not needed:

For the record, I really don't find these "here's what the LLM told me" comments helpful.

DJMcNab:
> Is the LLM wrong? 

I firmly believe so. Conflation artifacts are only related to the pixels where two paths (or subpaths) intersect, which the screenshots in the linked issues clearly do not show. The sparse strips work doesn't fundamentally change how paths are rendered in a way which makes watertightness bugs impossible; the fact that it's a new implementation is the main reason that the sparse strips work is less likely to show this issue, but obviously it's likely to have its own issues.

Everyone agrees that we should fix the underlying bug, but I find it unlikely that stating distruths about the cause will help with that. What would help is making a reproduction from pure Vello commands, and then minimising that to the smallest geometry which still reproduces it (i.e. inlining the path segments output by Kurbo's iterator and eliminating those which aren't needed); it would also help to see whether this still happens with `use_cpu` set to true. As a potential point for reproducing, we think it's possible that there's an interaction with clipping/layers causing this (although I personally feel that's unlikely).

PoignardAzur:
Keeping it as a draft for now. If we still have the problem by February, I'll merge it.

PoignardAzur:
> Is the LLM wrong?

Unless you're very confident the answer is "no", giving someone some LLM-generated text and asking them to fact-check it or to make decisions based off it... I don't know, I don't think it's the rudest thing ever, but I really don't like the "effort to post / effort to address" ratio of it.

And in the worst cases, it's basically like dumping garbage in front of someone's front door and telling them their mail is somewhere in there.

PR #1514, opened 2025-12-05
Update to git Vello for wgpu 27

waywardmonkeys:
To test things out and prepare for new releases, update to the current git Vello which uses wgpu 27.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,9 @@ xilem = { version = "0.4.0", path = "xilem", default-features = false }
 tree_arena = { version = "0.2.0", path = "tree_arena" }
 
 anymore = "1.0.0"
-vello = { version = "0.6.0", default-features = false, features = ["wgpu"] }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -50,7 +50,7 @@ xilem = { version = "0.4.0", path = "xilem", default-features = false }
 tree_arena = { version = "0.2.0", path = "tree_arena" }
 
 anymore = "1.0.0"
-vello = { version = "0.6.0", default-features = false, features = ["wgpu"] }
+vello = { version = "0.6.0", git = "https://github.com/linebender/vello", rev = "efa9c21d42c3b321464eb7cf91a73e1b1103f8d6" }
```
DJMcNab: This is re-enabling wgpu's default features for all our users, right (i.e. undoing https://github.com/linebender/xilem/pull/1467?) Is that intentional?

DJMcNab (commented):
If I'm right with my understanding, I think the one concern is a blocker

Cargo.toml
waywardmonkeys: You're right.

PR #1515, opened 2025-12-05
Use rustfmt to reorder imports throughout repo

PoignardAzur:


PoignardAzur:
Note that these options sometimes do things I don't like (e.g. mixing private and pub imports) so I applied them selectively.

PR #1519, opened 2025-12-07
Add Canvas widget

Philipp-M:
Alternative to #1506 (actually it builds on top of it again and applies the review comments/suggestions, quite a group effort :))

I was bugged by the `Arc<dyn Fn>` and the way it was handled.

IMO it's a little bit better layered this way and more predictable when the canvas draw function is actually changed, and allows the app state in the canvas draw function by using `Fn(Arg<'_, State>, &mut Scene, Size)` instead of just `Arc<dyn Fn(&mut Scene, Size)>` which would require capturing in any realistic use-case (accessing the user state) and thus the `Arc` would be somewhat useless, as always a new one needs to be created.

Additionally I've added an example (as I think this is good practice when adding new widgets).
(also to check whether this results in latency, i.e. when pressing the button, that the scene is updated immediately instead of a frame-delay (fortunately this doesn't result in frame-delay))


xilem_masonry/src/view/canvas.rs
```diff
@@ -0,0 +1,119 @@
...
+        _app_state: Arg<'_, State>,
+    ) -> MessageResult<Action> {
+        debug_assert!(
+            message.remaining_path().is_empty(),
+            "id path should be empty in Canvas::message"
+        );
+        match message.take_message::<CanvasSizeChanged>() {
+            Some(_) => MessageResult::RequestRebuild,
+            None => {
+                tracing::error!("Wrong message type in Checkbox::message, got {message:?}.");
```
RagibHasin:
```suggestion
                tracing::error!("Wrong message type in Canvas::message, got {message:?}.");
```

Philipp-M: Whoops good catch

masonry/src/widgets/canvas.rs
```diff
@@ -0,0 +1,188 @@
...
+};
+
+/// A widget allowing custom drawing.
+///
+/// A canvas takes a painter callback; every time the canvas is repainted, that callback
+/// in run with a [`Scene`].
+/// That Scene is then used as the canvas' contents.
+#[derive(Default)]
+pub struct Canvas {
+    alt_text: ArcStr,
```
PoignardAzur: As mentioned in #1519, "empty alt text" and "no alt text" are different semantic values, so we should make this an option.

Philipp-M: Should be solved

PoignardAzur:
Well it seems like the widget would send one action per time the layout pass is ran. If the layout pass is ran multiple times in a frame, the Xilem view will get multiple actions and render multiple scenes, even though only one is used.

Philipp-M:
> If the layout pass is ran multiple times in a frame, the Xilem view will get multiple actions and render multiple scenes, even though only one is used

Unless the canvas size changes multiple times it should still only result in one event because of:

```rust
if self.size != size {
    self.size = size;
    ctx.submit_action::<Self::Action>(CanvasSizeChanged { size });
}
```

We could think about batching or coalescing events (similar as `pointerrawupdate` vs `pointermove` for pointer events in the web), but that's orthogonal to this PR.

I'll keep this PR open until the end of the week to give others the chance to review as I don't think a lot of merge-conflicts will occur anytime soon.

PR #1523, opened 2025-12-08
Simplify Example in Docs

nilsmartel:
The simplification is only minimal.
I did this because I noticed jumping back and forth ("huh, default? where does that come from? Oh, annotation there, annotation derives default an so on, now back to reading the docs") while reading the example code.

I.e. less annotations / derived code makes things a little more obivious to read.

Maybe this is negligible shaving

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -34,7 +34,6 @@
 //! use xilem::{EventLoop, WindowOptions, WidgetView, Xilem};
 //! use xilem::core::Edit;
 //!
-
 //! struct Counter(i32);
 //!
 //! fn app_logic(data: &mut Counter) -> impl WidgetView<Edit<Counter>> + use<> {


xilem/src/lib.rs
```diff
@@ -3,6 +3,8 @@
 
 // After you edit the crate's doc comment, run this command, then check README.md for any missing links
 // cargo rdme --workspace-project=xilem
+// if the command is not recognized
+// cargo install cargo-rdme
```
PoignardAzur:
```suggestion
```

If we want to change that comment, I'd want us to change it in all the `lib.rs` files where it appears at once.

(And also, I think looking up "cargo rdme" online isn't that hard.)

xilem/README.md
```diff
@@ -80,20 +80,17 @@ use xilem::view::{text_button, flex_col, label};
...
-struct Counter {
-    num: i32,
-}
+struct Counter(i32);
 
 fn app_logic(data: &mut Counter) -> impl WidgetView<Edit<Counter>> + use<> {
-    flex_col((
-        label(format!("{}", data.num)),
-        text_button("increment", |data: &mut Counter| data.num += 1),
+    flex_col(Axis::Vertical, (
```
Philipp-M:
```suggestion
    flex_col((
```
What happened here?

xilem/src/lib.rs
```diff
@@ -34,20 +34,18 @@
...
-//!     num: i32,
-//! }
+
+//! struct Counter(i32);
 //!
 //! fn app_logic(data: &mut Counter) -> impl WidgetView<Edit<Counter>> + use<> {
-//!     flex_col((
-//!         label(format!("{}", data.num)),
-//!         text_button("increment", |data: &mut Counter| data.num += 1),
+//!     flex_col(Axis::Vertical, (
```
Philipp-M:
```suggestion
//!     flex_col((
```

xilem/README.md
PoignardAzur: I'm guessing "wrong guesses while solving merge conflicts after rebasing on top of #1534".

nilsmartel: yeah, sorry to cause more work here. I thought it was just easy enough to do this all without leaving github

xilem/src/lib.rs
```diff
@@ -34,20 +34,18 @@
 //! use xilem::{EventLoop, WindowOptions, WidgetView, Xilem};
 //! use xilem::core::Edit;
 //!
-//! #[derive(Default)]
-//! struct Counter {
-//!     num: i32,
-//! }
+
```
Philipp-M:
```suggestion
```

PR #1525, opened 2025-12-09
Reword the WindowId documentation

PoignardAzur:


xStrom (approved):
It might be possible to recycle them, but given the `u64` size it's easier for us to just say don't do it - and not have to worry about having stale ids somewhere in our event queue.

Thanks!

PR #1527, opened 2025-12-10
Add `Gap` property to `Flex` and `Grid`

### LINKED ISSUE #1067: Add Gap or FlexGap property
See https://github.com/linebender/xilem/pull/1055#discussion_r2137514457

PoignardAzur:
Replace existing gap/spacing attributes and related methods.

Fixes #1067.

### DIFF
--- a/masonry/src/properties/gap.rs
+++ b/masonry/src/properties/gap.rs
@@ -14,13 +14,13 @@ use crate::properties::types::Length;
 /// ## Note on spacers and `Flex` widgets
 ///
 /// This gap is between any two children, including `Flex` spacers.
-/// As such, using a non-zero gap and also adding may lead to counter-intuitive results.
+/// As such, using a both a gap and a spacer may lead to counter-intuitive results.
 /// You should usually pick one or the other.
 ///
 /// [gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap
-#[expect(missing_docs, reason = "field names are self-descriptive")]
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct Gap {
+    /// Gap value.
     pub gap: Length,
 }
 
@@ -47,7 +47,7 @@ impl Gap {
     /// Zero-sized gap.
     pub const ZERO: Self = Self { gap: Length::ZERO };
 
-    /// Create new `Gap` with given value.
+    /// Creates a new `Gap` with given value.
     pub const fn new(gap: Length) -> Self {
         Self { gap }
     }


masonry/src/properties/gap.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// The distance between two adjacent widgets in a [`Flex`](crate::widgets::Flex) or
+/// a [`Grid`](crate::widgets::Grid).
+///
+/// Equivalent to the css [gap] property.
+///
+/// ## Note on spacers and `Flex` widgets
+///
+/// This gap is between any two children, including `Flex` spacers.
+/// As such, using a non-zero gap and also adding may lead to counter-intuitive results.
```
xStrom: Also adding? I guess this is about using spacers.

```suggestion
/// As such, using both a gap and a spacer may lead to counter-intuitive results.
```

masonry/src/properties/gap.rs
```diff
@@ -0,0 +1,62 @@
...
+/// You should usually pick one or the other.
+///
+/// [gap]: https://developer.mozilla.org/en-US/docs/Web/CSS/gap
+#[expect(missing_docs, reason = "field names are self-descriptive")]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Gap {
+    pub gap: Length,
```
xStrom: Self-descriptive or not, I think it's less noisy to actually have a short doc comment.

```suggestion
#[derive(Clone, Copy, Debug, PartialEq)]
pub struct Gap {
    /// The distance between two adjacent widgets.
    pub gap: Length,
```

masonry/src/properties/gap.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn from(gap: Length) -> Self {
+        Self { gap }
+    }
+}
+
+impl Gap {
+    /// Zero-sized gap.
+    pub const ZERO: Self = Self { gap: Length::ZERO };
+
+    /// Create new `Gap` with given value.
```
xStrom: [Rust RFC 1574](https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#summary-sentence) recommends using third person singular for API docs. *Creates* instead of *Create*.

```suggestion
    /// Creates a new `Gap` with the given value.
```

xilem_masonry/src/style.rs
```diff
@@ -195,6 +197,14 @@ pub trait Style<State: ViewArgument, Action: 'static>: WidgetView<State, Action>
         self.prop(padding.into())
     }
 
+    /// Set the gap between the element's consecutive children.
```
xStrom, line 200:
```suggestion
    /// Sets the gap between the element's consecutive children.
```

PoignardAzur: All other methods in this file use `Set`, and I'd rather not switch style in the same module.

We can apply the writing convention everywhere in a future PR.

PR #1528, opened 2025-12-10
Add alt-text to Image widget

PoignardAzur:
Code is roughly copy-pasted from the canvas PRs.

### DIFF
--- a/masonry/src/widgets/image.rs
+++ b/masonry/src/widgets/image.rs
@@ -7,15 +7,14 @@
 use std::any::TypeId;
 
 use accesskit::{Node, Role};
-use masonry_core::core::HasProperty;
 use tracing::{Span, trace_span};
 use vello::Scene;
 use vello::kurbo::{Affine, Size};
 use vello::peniko::{BlendMode, ImageBrush};
 
 use crate::core::{
-    AccessCtx, BoxConstraints, ChildrenIds, LayoutCtx, NoAction, PaintCtx, PropertiesMut,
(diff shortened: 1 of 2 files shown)


masonry/src/widgets/image.rs
```diff
@@ -141,9 +163,11 @@ impl Widget for Image {
         &mut self,
         _ctx: &mut AccessCtx<'_>,
         _props: &PropertiesRef<'_>,
-        _node: &mut Node,
+        node: &mut Node,
     ) {
-        // TODO - Handle alt text and such.
+        if let Some(alt_text) = &self.alt_text {
+            node.set_description(&**alt_text);
+        }
```
Philipp-M, line 170:
```suggestion
        if let Some(alt_text) = &self.alt_text {
            node.set_description(&**alt_text);
        } else {
            node.clear_description();
        }
```

PoignardAzur: Nope, each time the `accessibility` method is called, you get a new node from scratch. You don't need to clear anything.

Philipp-M: Ah ok, good to know, then I need to correct that in #1519 too

xStrom: What's the purpose of `clear_description` then? To correct mistakes you did earlier in the same method? Or do you sometimes get mutable node references after all?

PoignardAzur: You'd have to ask Matt.

PR #1531, opened 2025-12-11
Remove all remaining unintended `#[expect(missing_docs)]` annotations.

### LINKED ISSUE #1346: Tracking issue for missing docs
The following items have missing docs in Xilem:

- `WidgetView`
- `ViewCtx`
- `Pod`

---

The following items have missing docs in masonry_winit:

- `MasonryState`
- `run`
- `run_with`

---

We should document them and remove the last `#![expect(missing_docs)]` annotations from the code.

PoignardAzur:
Remove `EventLoop` argument from `masonry_winit::app::run`.
Remove `ViewCtx::with_leaf_action_widget` method.

Fixes #1346.

### DIFF
--- a/xilem_masonry/src/view/canvas.rs
+++ b/xilem_masonry/src/view/canvas.rs
@@ -76,13 +76,16 @@ where
     type ViewState = ();
 
     fn build(&self, ctx: &mut ViewCtx, _: Arg<'_, State>) -> (Self::Element, Self::ViewState) {
-        ctx.with_leaf_action_widget(|ctx| {
-            let widget = match &self.alt_text {
-                Some(alt_text) => widgets::Canvas::default().with_alt_text(alt_text.clone()),
-                None => widgets::Canvas::default(),
-            };
-            ctx.create_pod(widget)
-        })
+        (
+            ctx.with_action_widget(|ctx| {
+                let widget = match &self.alt_text {
+                    Some(alt_text) => widgets::Canvas::default().with_alt_text(alt_text.clone()),
+                    None => widgets::Canvas::default(),
+                };
+                ctx.create_pod(widget)
+            }),
+            (),
+        )
     }
 
     fn rebuild(


masonry/src/widgets/portal.rs
```diff
@@ -239,13 +242,15 @@ impl<W: Widget + FromDynWidget + ?Sized> Portal<W> {
...
 
-    #[expect(missing_docs, reason = "TODO")]
+    /// Translate the scrolling "position" of the container.
     pub fn pan_viewport_by(this: &mut WidgetMut<'_, Self>, translation: Vec2) -> bool {
         Self::set_viewport_pos(this, this.widget.viewport_pos + translation)
     }
 
-    #[expect(missing_docs, reason = "TODO")]
-    // Note - Rect is in child coordinates
+    /// Change the scrolling "position" of the container to that target is scrolled into view.
```
xStrom:
```suggestion
    /// Change the scrolling "position" of the container so that target is scrolled into view.
```

masonry_winit/src/event_loop_runner.rs
```diff
@@ -34,10 +32,20 @@ use crate::app::{AppDriver, DriverCtx, masonry_resize_direction_to_winit, winit_
...
 
+/// The custom event type that we inject into winit's [`EventLoop`](winit::event_loop::EventLoop).
+///
+/// This represents the types that can be emitted during the event loop, but aren't emitted
+/// by winit.
 #[derive(Debug)]
 pub enum MasonryUserEvent {
+    /// An accessibility API emitted an event.
+    ///
+    /// This is how [`accesskit_winit`] route its events to us.
```
xStrom:
```suggestion
    /// This is how [`accesskit_winit`] routes its events to us.
```

xilem_masonry/src/pod.rs
```diff
@@ -19,8 +19,8 @@ use crate::core::{Mut, SuperElement, ViewElement};
 ///
 /// [`transformed`]: crate::view::Transformed
 /// [`WidgetView::transform`]: crate::view::transformed
-#[expect(missing_docs, reason = "TODO - Document these items")]
 pub struct Pod<W: Widget + FromDynWidget + ?Sized> {
+    #[expect(missing_docs, reason = "Self-explanatory")]
```
xStrom: Don't be afraid of simple docs.

```suggestion
    /// A yet to be inserted widget.
```

PoignardAzur: I honestly hate those when I read them.

I read so many docs where everything is like "`foobar_handler` - The handler that handles the foobars", YES, THANK YOU, I COULD INFER THAT FROM THE NAME, why are you wasting my time with this?

Philipp-M: I mean it still provides *some* info, i.e. that the widget is not yet inserted, which *might* not be obivous?
There could of course be even more info, like that this is just configuration for the new widget which will be used to construct the actual widget or something like that,

PoignardAzur: > that the widget is not yet inserted, which might not be obivous?

That's what the first line of the type's doc says. (This is another trope of those "well the linter says we *have* to have a docstring" texts, they tend to restate what the parent doc already says.)

xStrom: I read those docstrings by hovering over a field when it is accessed. The parent type's doc is not shown there.

xilem_masonry/src/pod.rs
```diff
@@ -19,8 +19,8 @@ use crate::core::{Mut, SuperElement, ViewElement};
 ///
 /// [`transformed`]: crate::view::Transformed
 /// [`WidgetView::transform`]: crate::view::transformed
-#[expect(missing_docs, reason = "TODO - Document these items")]
 pub struct Pod<W: Widget + FromDynWidget + ?Sized> {
+    /// The payload.
```
xStrom: This seems deliberately spiteful. It is clearly less useful than just reusing the *a yet to be inserted widget* description.

On the obviousness front, docs are read also by people who are far less experienced than you. The mentalities like "you can just intuit this" or "well it's obvious, because it's documented somewhere else" are not beginner-friendly.

xilem_masonry/src/pod.rs
PoignardAzur: > I read those docstrings by hovering over a field when it is accessed.

Fair enough. I'm only used to docs.rs, but in your context it does seem relevant.

xilem_masonry/src/pod.rs
PoignardAzur: > This seems deliberately spiteful.

I'm not trying to be spiteful. This seemed to me like a suitably minimalist description. I do take your point in the other comment about IDE usage though.

PR #1537, opened 2025-12-15
Fix `ScrollBar` behavior with large content sizes.

xStrom:
The scrollbar would go off screen if `Portal` content is large. That's because the old code only accounted for the minimum bar length when deciding how large it should be visually, but did not account for this minimum length for any positioning calculations.

This PR ensures the positioning math takes the minimum size into account and also refactors it to a separate method instead of having this brittle code duplicated.

Also:
* Renamed `get_cursor_rect` -> `cursor_rect` as in Rust the convention is to not have a `get_` prefix for getters.
* Renamed `z0` & `z1` to `c0` & `c1`. These are either X or Y coords and definitely not Z, so C for coordinate seems better.


PoignardAzur (approved):
Oh man, how I hate this file.

LGTM, makes the code a bit more readable. Thanks!

PR #1538, opened 2025-12-16
Fix `Portal` size constraints.

xStrom:
`Portal` already had `constrain_vertical` and `constrain_horizontal` properties but these were only used to hide the scrollbar. They didn't actually affect the child layout in any way.

Indeed, the `Portal` layout implementation had a simple hardcoded `let max_child_size = bc.max()`, meaning that the child was always fully constrained.

This PR implements actual constrainment logic, configurable with the aforementioned properties. When not set to constrain, the child gets unbounded constraints. After all, that is the main point of `Portal` - to be able to scroll over giant children.

Also, when an axis is constrained, `Portal` now also ignores mouse wheel scrolling in that direction, in addition to hiding the visual scrollbar.

### DIFF
--- a/masonry/examples/to_do_list.rs
+++ b/masonry/examples/to_do_list.rs
@@ -74,7 +74,7 @@ pub fn make_widget_tree() -> NewWidget<impl Widget> {
         ))
         .with_fixed_spacer(WIDGET_SPACING);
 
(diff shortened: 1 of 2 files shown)


PoignardAzur:
Is this superseded by  #1560?

xStrom:
Alright, this is now fully superseded by #1560 and #1583.

PR #1540, opened 2025-12-16
Improve `Flex` child constraint accuracy.

xStrom:
`Flex` contained a hack in the form of always having a phantom `0.0001` flex factor present. This caused the child constraints to be awkward numerically. Even if rounding could prevent issues in a lot of cases, it still made debugging layout annoying due to getting e.g. 123.8567689512667 instead of 124.

This PR removes the phantom flex factor and guards against division by zero explicitly.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -651,7 +651,6 @@ impl Widget for Flex {
 
         let axis = self.direction;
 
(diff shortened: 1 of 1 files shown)


PoignardAzur:
IIRC, "comparing to zero" was the previous behavior, and I deliberately added the initial flex factor to fix a bug; I might be misrembering this, though, because I don't remember what the exact reason was.

Issue #1343, opened 2025-08-15
Tracking issue for layers and menus


PoignardAzur:
We should add a basic layer system to Masonry.

While we've talked before about hardware acceleration and using the platform composer, these are advanced features. We can cover basic use cases much faster by adding a root widget to Masonry which acts as a stack of layers.

(See #1275)

## Layer types

Layers beside the base "regular app" layer could have a few specific types (probably an enum):

- Menu: disappears when the user clicks outside of it.
- Tooltip: disappears when the cursor leaves a given hitbox.
- Dialog: inhibits interactions with any other layer.

We should have `RenderRootSignal` variants and `***Ctx` methods to create each of those.

User code would then call:

```rust
ctx.create_menu(NewWidget::new(...));
```

Which would lead to RenderRoot emitting a `RenderRootSignal::CreateMenu(widget)` signal.

The driver could react either by asking Winit / the platform / etc to create a new child window, or by calling `RenderRoot::add_menu_layer(widget)`.

Layers may also take an `accesskit::Role` for their root.

## New widgets

We should also add helper widgets that apps could pass to the `ctx.create_***` methods.

Useful widgets may include:

- `Menu` (list of children and visual separators)
- `MenuItem` (shiny background when hovered)
- `Tooltip` (specific default border and background)

PR #1541, opened 2025-12-17
Add Tooltip layer

PoignardAzur:
See https://github.com/linebender/xilem/issues/1343.

### DIFF
--- a/masonry_core/src/core/mod.rs
+++ b/masonry_core/src/core/mod.rs
@@ -18,7 +18,7 @@ mod widget_tag;
 
 pub use contexts::*;
 pub use events::*;
-pub use layer::{Layer, LayerType};
+pub use layer::*;
 pub use properties::*;
 pub use text::*;
 pub use widget::*;


masonry/examples/layers.rs
```diff
@@ -74,16 +74,27 @@ impl Widget for OverlayBox {
...
-        if let PointerEvent::Move(PointerUpdate { current, .. }) = event
-            && ctx.is_hovered()
-        {
-            let position = current.logical_point();
-            if let Some(overlay_id) = self.layer_root_id {
-                ctx.reposition_layer(overlay_id, position);
-            } else {
-                let overlay = (self.overlayer)();
+        if let PointerEvent::Move(PointerUpdate { current, .. }) = event {
+            self.tooltip_position = current.logical_point();
```
jaredoconnell: Using the pointer position is resulting in the tooltip getting covered by the cursor. So either we need a fixed offset, or we need to measure the size of the tooltip itself so that it's offset by that amount to prevent the overlap. As a test I found `current.logical_point().add((5.0, -25.0))` to be a good temporary offset.

masonry_core/src/doc/masonry_concepts.md
```diff
@@ -187,6 +187,20 @@ It's the one in which almost all content (buttons, texts, images) will be drawn.
...
+### Adding a layer
+
+<!-- TODO - Flesh this out -->
+
+Most context methods have a `create_layer(layer_type, fallback_widget, pos)` method.
+
+`layer_type` and `fallback_widget` both represent the same content:
+
+- `layer_type` is an enum which represents the layer's "semantic" content, with variants for common layer types.
+- `fallback_widget` represents a widget which should be drawn at the root of the new layer.
```
jaredoconnell: I think the fallback vs built-in behavior is a little unclear in the documentation. Can you clarify it a bit?

jaredoconnell (commented):
I'm glad to see a layer system being added. I haven't reviewed this enough to approve, but I can confirm that it's working.

masonry/src/layers/tooltip.rs
```diff
@@ -0,0 +1,251 @@
...
+use crate::util::{fill, stroke};
+
+/// A [`Layer`] representing a simple tooltip showing some content until the mouse moves.
+pub struct Tooltip {
+    child: WidgetPod<dyn Widget>,
+}
+
+// --- MARK: BUILDERS
+impl Tooltip {
+    /// TODO - Doc
```
xStrom:
```suggestion
    /// Creates a new `Tooltip`.
```

masonry_core/src/doc/masonry_concepts.md
PoignardAzur: The next paragraph feels informative enough to me:

> These two values are sent to the Masonry driver running the app; if the driver has built-in behavior for the given `layer_type`, this behavior will be used.
Otherwise, the driver will add a new layer to the current [`RenderRoot`] with `fallback_widget` as its root.

Do you think something is missing? Or do you think having these two items before the paragraph is confusing?

masonry/examples/layers.rs
PoignardAzur: Done.

masonry/src/layers/tooltip.rs
PoignardAzur: Done.

PoignardAzur:
Made some adjustments to account for reviews. Will merge this next week unless someone objects.

PR #1548, opened 2025-12-29
Remove default padding from label.

### LINKED ISSUE #1530: `TextInput` placeholder has jarring offset.
The placeholder text should be located at the same spot as the main text if they both have the same font size. This is true e.g. on the web.

However the Masonry widget currently has like a 1px offset pushing the placeholder text slightly away from the start.

One easy way to confirm this is with the xilem `to_do_mvc` example. Although you'll need to comment out the `.text_size(16.)` line to get equal font size. After that, try cut/pasting in the same text as the placeholder. You will see that the placeholder is slightly offset.

I'm not sure what causes this, because `TextInput::layout` calls `ctx.place_child` on both the main text and the placeholder with the same position. So the offset is introduced by something else.

PoignardAzur:
Fixes #1530.

### DIFF
--- a/xilem/examples/to_do_mvc.rs
+++ b/xilem/examples/to_do_mvc.rs
@@ -51,17 +51,12 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<Edit<TaskList>> + use<
             task_list.next_task = new_value;
         },
     )
-    //.text_size(16.)
+    .text_size(16.)
     .placeholder("What needs to be done?")
     .insert_newline(InsertNewline::OnShiftEnter)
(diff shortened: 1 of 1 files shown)


xilem/examples/to_do_mvc.rs
```diff
@@ -51,13 +51,18 @@ fn app_logic(task_list: &mut TaskList) -> impl WidgetView<Edit<TaskList>> + use<
             task_list.next_task = new_value;
         },
     )
-    .text_size(16.)
+    //.text_size(16.)
```
nicoburns: Commented code

nicoburns (approved):
Always happy to see default padding removed!

Issue #1521, opened 2025-12-08
UI Flickering/Shaking on Window Resize (macOS)


elgar328:
**Description:**

When resizing the application window for the `to_do_mvc` and `slider_demo` examples, the internal UI elements exhibit severe flickering and shaking, making the resizing experience visually disruptive.

https://github.com/user-attachments/assets/51b0a577-8375-4f83-81b2-f47ea76a9ddd

https://github.com/user-attachments/assets/5db58ad8-056a-449d-bc36-e38bf59ce490

**Reproduction Steps:**

1.  Run either `cargo run --example to_do_mvc` or `cargo run --example slider_demo`.
2.  Resize the application window by dragging an edge or corner.
3.  Observe the severe UI flickering/shaking.

**Expected Behavior:**

When resizing the window, the UI elements should smoothly adjust their layout and size without any visual artifacts such as flickering or shaking.

**Observed Behavior:**

UI elements within the application window appear to flicker, jump, or shake erratically during the resizing operation. The visual instability persists as long as the window is being resized.

**Environment:**

*   **Operating System:** macOS Sequoia 15.6.1
*   **Hardware:** Apple M4 Max
*   **Rust Version:** rustc 1.89.0 (29483883e 2025-08-04)

**Additional Context:**

I have observed similar flickering/shaking behavior during window resizing in other Rust GUI libraries like `egui` and `slint` when tested on the same macOS environment. However, this particular issue was *not* present when testing applications built with `dioxus` on the same machine.

elgar328:
> If egui and slint have the same issue, then it might be a winit bug?

I suspect this issue might be related to macOS HiDPI (Retina) scaling, specifically regarding the scale factor or the handling of logical vs. physical pixels.

noelgbl:
Might be related to [this issue in GPUI](https://github.com/emilk/egui/pull/7641)

PR #1551, opened 2026-01-03
Fix jittering during window resize on MacOS for WGPU/Metal

noelgbl:
This update fixes a widely known jitter issue on Metal/MacOS #1521 that occurs during window resizing. It introduces a new `resized_window` field in `MasonryState` to track the state of window resizing. The `on_window_resize_state_change` method in `RenderContext` is added to handle the resizing behavior for surfaces.

### DIFF
--- a/masonry_winit/Cargo.toml
+++ b/masonry_winit/Cargo.toml
@@ -36,7 +36,7 @@ copypasta = "0.10.2"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 # We need Metal to interact with CoreAnimation during window resizing.  
-wgpu = { version = "27.0.1", default-features = false, features = ["metal"] }
+wgpu = { version = "28.0.0", default-features = false, features = ["metal"] }
 
 [dev-dependencies]
 # We don't use the "workspace" dependency here, because this makes a loop in publishing.
--- a/masonry_winit/src/vello_util.rs
+++ b/masonry_winit/src/vello_util.rs
@@ -222,20 +222,14 @@ impl RenderContext {
             return;
         }
 
-        // SAFETY: We only mutate a backend-specific flag on the Metal HAL surface when the
-        // runtime backend matches Metal. This mirrors the eframe/egui-wgpu fix.
-        //
-        // TODO: Update to new API once Vello is on v28 of WGPU
-        // https://github.com/linebender/xilem/pull/1551#issuecomment-3864370306
         #[allow(
             unsafe_code,
             reason = "We only mutate a backend-specific flag on the Metal HAL surface when the runtime backend matches Metal"
         )]
         unsafe {
             if let Some(hal_surface) = surface.surface.as_hal::<::wgpu::hal::api::Metal>() {
-                let raw =
-                    std::ptr::from_ref::<::wgpu::hal::metal::Surface>(&*hal_surface).cast_mut();
-                (*raw).present_with_transaction = resizing;
+                let guard = hal_surface.render_layer().lock();
+                guard.set_presents_with_transaction(resizing);
             }
         }
 


masonry_winit/src/event_loop_runner.rs
```diff
@@ -215,6 +216,7 @@ pub struct MasonryState<'a> {
     /// Windows that are scheduled to be created in the next resumed event.
     new_windows: Vec<NewWindow>,
     need_first_frame: Vec<HandleId>,
+
```
PoignardAzur, line 226: Remove extra line.

masonry_winit/src/vello_util.rs
```diff
@@ -164,6 +165,32 @@ impl RenderContext {
...
+
+        // SAFETY: We only mutate a backend-specific flag on the Metal HAL surface when the
+        // runtime backend matches Metal. This mirrors the eframe/egui-wgpu fix.
+        #[allow(unsafe_code, reason = "We only mutate a backend-specific flag on the Metal HAL surface when the runtime backend matches Metal")]
+        unsafe {
+            if let Some(hal_surface) = surface.surface.as_hal::<wgpu::hal::api::Metal>() {
+                let raw = std::ptr::from_ref::<wgpu::hal::metal::Surface>(&*hal_surface).cast_mut();
+                (*raw).present_with_transaction = resizing;
+            }
+        }
```
PoignardAzur: I looked more into that unsafe block and looked up the wgpu side to see how safe this was, and even if it works in practice, I'm not convinced this isn't undefined behavior, though I could be wrong.

Aside from that, I don't think this change is forward-compatible. In the current head commit of wgpu, [`wgpu::hal::metal::Surface`](https://github.com/gfx-rs/wgpu/blob/6be68cbc6591ecff9865b06c35841627f2eb56b7/wgpu-hal/src/metal/mod.rs#L438-L442) does not publicly expose `present_with_transaction`. Ordinarily, I wouldn't block a PR on a future dependency, but combined with the unclear safety, it's giving me pause.

masonry_winit/Cargo.toml
```diff
@@ -34,6 +34,14 @@ accesskit_winit.workspace = true
 wgpu-profiler = { optional = true, version = "0.25.0", default-features = false }
 copypasta = "0.10.2"
 
+[target.'cfg(target_os = "macos")'.dependencies]
+# Enables masonry_core/default -> vello -> wgpu -> wgpu_hal -> metal,
+# which we need to interact with CoreAnimation during window resizing.
+# This is not a very elegant fix and also enables all other HAL's.
+# Remove this once a feature gating solution is found.
+# See: https://github.com/linebender/xilem/pull/1551
+masonry_core = { workspace = true, features = ["default"] }
```
xStrom: Can't we just enable the `metal` feature directly?

```suggestion
[target.'cfg(target_os = "macos")'.dependencies]
# We need Metal to interact with CoreAnimation during window resizing.
wgpu = { version = "27.0.1", default-features = false, features = ["metal"] }
```

noelgbl: Also thought of this, but wanted to avoid having to sync the version with Vello. But you probably know better than me if that's the way to go.

xStrom: Keeping `wgpu` version in sync is a low cost to pay and we do that already anyway in some places. So I'd say yes, we should only enable the minimum features that we actually need.

* Keeping `wgpu` in sync is an occasional administration cost.
* Compiling features we don't use is a continuous performance cost.

masonry_winit/src/vello_util.rs
```diff
@@ -207,6 +208,37 @@ impl RenderContext {
...
+        if surface.resizing == resizing {
+            return;
+        }
+
+        #[allow(
+            unsafe_code,
+            reason = "We only mutate a backend-specific flag on the Metal HAL surface when the runtime backend matches Metal"
+        )]
+        unsafe {
+            if let Some(hal_surface) = surface.surface.as_hal::<::wgpu::hal::api::Metal>() {
```
PoignardAzur, line 230: (Non-blocking) As far as I'm aware, only the `.as_hal` call is unsafe, right? I'd prefer if you could narrow the unsafe block to that call, with a `// SAFETY: we do not destroy the returned handle` comment.

PR #1552, opened 2026-01-03
Provide ctx in the draw closure of `Canvas`

DaraJKong:
In a project I'm working on, I needed to draw text inside a `Canvas` widget. I initially created a `FontContext` and a `LayoutContext` inside the draw closure. I used `masonry::core::render_text` to render my text layout to the Vello `Scene`. While theoretically functionnal, my solution was extremely slow and wrong.

After realizing that both `FontContext` and `LayoutContext` are meant to be global resources created only once per application, I set out to tinker with the source code of the `Canvas` widget. What I came up with allowed me to do this:
```rust
canvas(|state: &mut AppState, ctx, scene, size| {
    let (fcx, lcx) = ctx.text_contexts();

    let mut text_layout_builder = lcx.ranged_builder(fcx, "Text", 1., true);
    text_layout_builder.push_default(StyleProperty::FontStack(FontStack::Single(
        FontFamily::Generic(GenericFamily::Serif),
    )));
    text_layout_builder.push_default(StyleProperty::FontSize(24.));

    let mut text_layout = text_layout_builder.build("Text");
    text_layout.break_all_lines(None);
    text_layout.align(None, TextAlign::Start, TextAlignOptions::default());

    masonry::core::render_text(
        scene,
        Affine::IDENTITY,
        &text_layout,
        &[css::WHITE.into()],
        true,
    );
})
```

I updated the `FnOnce` closure to accept a new ctx parameter. Then in Masonry, inside the `Canvas::update_scene` function, I got the ctx from the `WidgetMut<'_, Canvas>` and passed it as a mutable reference to the closure. In Xilem, I simply pass the ctx from the closure to the draw closure.

This PR includes the addition of a new screenshot test. The text_canvas test renders a 200x200 image with the text 'Canvas' stretched to fit inside.

### DIFF
--- a/xilem_masonry/src/view/canvas.rs
+++ b/xilem_masonry/src/view/canvas.rs
@@ -97,7 +97,7 @@ where
         state: Arg<'_, State>,
     ) {
         widgets::Canvas::update_scene(&mut element, |ctx, scene, size| {
-            (self.draw)(state, ctx, scene, size)
+            (self.draw)(state, ctx, scene, size);
         });
         if self.alt_text != prev.alt_text {
             widgets::Canvas::set_alt_text(&mut element, self.alt_text.clone());


masonry/src/widgets/canvas.rs
```diff
@@ -53,9 +53,12 @@ impl Canvas {
 // --- MARK: WIDGETMUT
 impl Canvas {
     /// Updates the canvas scene.
-    pub fn update_scene(this: &mut WidgetMut<'_, Self>, f: impl FnOnce(&mut Scene, Size)) {
+    pub fn update_scene(
+        this: &mut WidgetMut<'_, Self>,
+        f: impl FnOnce(&mut MutateCtx<'_>, &mut Scene, Size),
+    ) {
```
PoignardAzur, line 58: I think using `PaintCtx` would make more sense?

`MutateCtx` can do a lot of things that this widget wouldn't need.

xStrom: I agree, but that isn't really a problem with this PR per-se. This whole `update_scene` method seems weird. Why isn't this user-provided function a field of the widget struct that then gets called in `paint`?

I haven't read through the four canvas PRs, maybe I should do that to find an answer.

Philipp-M: > I haven't read through the four canvas PRs, maybe I should do that to find an answer.

I've explained it in https://github.com/linebender/xilem/pull/1519#issue-3703663811

Regarding context type, I think it might even make sense to create a separate type specifically for this function that has a limited API.

But to use a `PaintCtx` here, something like this is necessary in masonry_core (as the fields are private):
```rust
impl MutateCtx<'_> {
    pub fn paint_mut(&mut self) -> PaintCtx<'_> {
        PaintCtx {
            global_state: self.global_state,
            widget_state: self.widget_state,
            children: self.children.reborrow_mut(),
        }
    }
}
```

PoignardAzur: > But to use a PaintCtx here, something like this is necessary in masonry_core (as the fields are private):

I... kind of don't want to add that.

If we don't have an immediate solution, we might want to accept this PR now and rethink the canvas API later. I do think the current API is a bit weird (why pass a callback instead of a scene directly?), but this PR isn't responsible for it.

xStrom:
The PR has a merge conflict that needs to be resolved. I suggest rebasing on `main`.

xStrom (approved):
Yeah let's merge this and open a new issue for the context type.

PR #1557, opened 2026-01-07
Raise default log level in dev mode

PoignardAzur:
Previous log level was extremely verbose and noisy.


xStrom:
One big offender of the spam is pointer events. Perhaps we should remove (or lower to `trace`) those messages?

PR #1559, opened 2026-01-07
Add runtime benchmark to Masonry

PoignardAzur:
Use divan for benchmarking.


jaredoconnell (approved):
Looks reasonable. I don't have any concerns because this is self-contained in benchmark code.

Issue #37, opened 2023-01-17
Proposal for layout / child widget sizing in Xilem


nicoburns:
The proposals here came from me looking into what it would take to integrate [Taffy](https://github.com/DioxusLabs/taffy) layout into Xilem. But nothing proposed here is really specific to the CSS style layout modes (Flexbox and CSS Grid) that Taffy implements. Nor would they commit Xilem to CSS style layout. Rather, I believe they would enable Taffy layout modes to implemented in Xilem as widgets (which could live in an external crate), in much the same way that the existing `Flex` widget is implemented in Druid.

I suspect that we could make a much more streamlined system if support for associating arbitrary data (e.g. "styles") with elements such that a parent widget could access them on a child widget the chidl widget having to add support for them was implemented (ala https://github.com/linebender/druid/issues/2207). But that's a much more significant change, which I think can wait.

I have also written a [prototype integration of Taffy with Iced](https://github.com/nicoburns/iced_taffy) (Iced also uses a similar layout mechanism to Druid and Xilem). And despite having to work around some limitation of Iced's system (like no `measure` method, and `layout` taking `&self` rather than `&mut self`), the integration actually ended up being relatively straightforward (you can see the implementation of Iced's `layout` method [here](https://github.com/nicoburns/iced_taffy/blob/main/src/grid.rs#L320) (calling into Taffy from Iced), and the implementation of Taffy's `perform_child_layout` method [here](https://github.com/nicoburns/iced_taffy/blob/main/src/grid.rs#L144)  (calling back into Iced from Taffy)).

## Review of Existing Systems

Here I lay out the state of things as they are in Xilem, Druid, and Taffy.

### Prerequisite Type Defintions

A `Size<T>` in Taffy is defined as:

```rust
struct Size<T> {
    width: T,
    height: T,
}
```

A `Size` in Druid/Xilem (kurbo) is a `Size<f64>` using the above definition. For the remainder of this post I will translate this to `Size<f64>` in the function signatures below for clarity.

A `BoxConstraints` in Druid is defined as:

```rust
struct BoxConstraints {
    min: Size<f64>,
    max: Size<f64>,
}
```

An `AvailableSpace` in Taffy is defined as:

```rust
enum AvailableSpace {
	MinContent,
	MaxContent,
	Definite(Size<f32>),
}
```

A `SizingMode` in Taffy is defined as:

```rust
enum SizingMode {
	ContentSize,  // Size ignoring explicit styles
[...]

PR #1560, opened 2026-01-07
Add a new layout system.

xStrom:
# Summary

This PR introduces a new way of doing layout in Masonry. The new approach enables more complex layouts while still supporting the old designs. All widgets received a rewrite of their layout code.

# Background

The old Masonry layout system was fast but limited in capability. Just passing down constraints in a single pass isn't verbose enough to describe a bunch of real world layout strategies. For example, `Portal` having infinite space for its child, because infinite constraints were effectively removed from Masonry.  A more fully featured `Flex` with basis, grow, and shrink support that requires learning about the child's size before making decisions. Sure, it was possible to run speculative full layout passes, but that would make the complexity explode and so wasn't really viable. Even Flutter, from which the old Masonry layout system was derived from, has intrinsic measurement and a double layout pass (`getDryLayout`), with all the nice complexity warnings it deserves.

Changing the layout system has been talked about for years. There's a great old issue by @nicoburns in #37, @jaredoconnell worked on #682 and #701. There's been a ton of good discussion on Zulip, e.g. [1], [2], [3], [4], [5], [6]. The desire to make the layout system more feature rich is there, but a solution has evaded us due to the complexity of the problem, the fear of losing too much performance, and just due to the sheer amount of code changes needed.

# Time for a change

The longer we put this off, the harder it will get to make this massive rework happen. So I have decided to take this task on. I analyzed all the previous discussion the Linebender community has had on this topic, read through the CSS spec, investigated other systems like Flutter, WPF, etc. Spent some weeks iterating and testing. I think the PR currently contains a good set of features to prove that the new system works well. Of course there are still things that need improvement. However, we're already looking at 137 changed files with 5,203 additions and 2,501 deletions. It's already in a state where meaningful reviews are hard. I'll see if I can maybe extract a few more isolated bits to separate PRs, but unfortunately most of this is quite intertwined. Give the changes a glance and feel free to comment without being committed to reviewing the whole thing. If we as a group don't find any significant issues I think we should just merge and move on. There's follow-up work already waiting to happen. Best to not get stuck here.

# Goals

* Good compatibility with existing assumptions, i.e. don't lose features.
* Ability for parent widgets to have total control over their children, enabling complex layout strategies.
* Support for infinite bounds in cases like `Portal`.
* Ergonomic yet powerful way to "style" dimensions of widgets. No `SizedBox` bonanza, please.
* Easy to reason about, while steering people away from accidents like `NaN`s or O(N^2).
* Minimal regression in performance.

# The new system

In addition to still having a `layout` method, widgets now also have a `measure` method. This PR also introduces the idea of core properties that every widget has. The first one is `Dimensions` which specifies the size of the widget.

The general flow of layout is as follows:

* Parent asks Masonry, how big is this child? The parent also provides values for automatic fallback.
* Masonry checks if there are explicitly defined dimensions and resolves them. If there were no explicit dimensions, the parent provided fallback is resolved instead.
* When resolving dimensions (either explicit or parent's fallback), Masonry will either immediately return a result or it may need to measure the child and then return that measurement.
* The parent gets back an answer from Masonry about the size of the child and can decide whether to modify it or use it as-is.
* The parent asks Masonry to lay out the child using the final parent decided size.

## Dimensions

```rust
/// New core property that every widget will have.
struct Dimensions {
    width: Dim,
    height: Dim,
}

enum Dim {
    /// Automatically determine a reasonable length with a strategy chosen by the parent widget.
    #[default]
    Auto,
    /// Specific fixed length.
    Fixed(Length),
    /// Multiple of context length.
    Ratio(f64),
    /// Mimics the context length.
    Stretch,
    /// Intrinsic minimum length.
    MinContent,
    /// Intrinsic preferred length.
    MaxContent,
    /// The widget should attempt to fit into the context length.
    FitContent,
}

/// This is how a parent specifies [`Dim::Auto`] behavior for its children.
struct SizeDef {
    width: LenDef,
    height: LenDef,
}

/// An intermediate representation of widget length,
/// used after resolving Dim but before potentially measuring the widget.
enum LenDef {
    /// Specific fixed length.
    Fixed(f64),
    /// Intrinsic minimum length.
    MinContent,
    /// Intrinsic preferred length.
    MaxContent,
    /// The content should fit in the specified available space.
    FitContent(f64),
}

/// Layout width and height.
/// A length may be missing if it has not been computed yet, i.e. it depends on the child.
struct LayoutSize {
    width: Option<f64>,
    height: Option<f64>,
}
```

## Measure

```rust
/// Widget length measurement algorithm request.
/// It is up to the widget itself to define how it responds to these requests.
enum LenReq {
    /// The widget should measure its minimum intrinsic length.
    MinContent,
    /// The widget should measure its maximum intrinsic length.
    MaxContent,
    /// The widget should attempt to fit into the specified available space.
    FitContent(f64),
}

/// Computes the length that the widget wants to be on the given `axis`.
///
/// The goal of this method is for a parent to learn how its children want to be sized.
/// All the inputs are hints towards what the parent is planning for its child.
/// These hints should be followed, but that is not strictly required. For example,
/// it is completely valid for the child to always return the same size regardless of space.
/// Which is to say, at the end of the day, a widget chooses how it measures itself.
///
/// It's a question of intrinsic size. As `measure` will only be called when there is
/// no defined length present for the `axis`. So `measure` is not about reading the widget's
/// [`Dimensions`] from `props` and returning that. That is handled earlier by Masonry.
/// Instead, it's a task of measuring the contents of the widget. Combining both local state
/// and child measurements to come up with a answer for the total length of this `axis`.
///
/// Call [`MeasureCtx::compute_length`] to measure children and to get their `axis` length.
/// Then account for the child's positioning and any other unique layout factors you might have
/// to determine the total length of this widget on the given `axis`.
fn Widget::measure(
    &mut self,
    ctx: &mut MeasureCtx<'_>,
    props: &PropertiesRef<'_>,
    axis: Axis,
    len_req: LenReq,
    cross_length: Option<f64>,
) -> f64;

/// Computes the length that the `child` widget wants to be on the given `axis`.
///
/// `auto_length` specifies the fallback behavior if the child's dimension is [`Dim::Auto`].
/// If you're calling this from within [`measure`] then you usually want to derive
/// this from `len_req`, probably using [`LenReq::reduce`], i.e. you would call
/// `len_req.reduce(used_space_on_this_axis).into()`. However, if you're calling this
/// from within [`layout`] then you usually want to use use [`LenDef::FitContent`]
/// to ask the child to fit inside the available space. Sometimes a different fallback
/// makes more sense, e.g. `Grid` uses [`LenDef::Fixed`] to fall back to the exact
/// allocated child area size.
///
/// `context_size` is the size, in device pixels, that is used to resolve relative sizes.
/// For example [`Ratio(0.5)`] will result in half the context size.
/// This is usually the container widget's size, excluding its borders and padding.
/// Examples of exceptions include `Grid` which will provide the child's area size,
/// i.e. the union of cell sizes that the child occupies, and `Portal` which will provide
/// its viewport size.
///
/// `cross_length` is the length of the cross axis and is critical information for certain
/// widgets, e.g. for text max advance or to keep an aspect ratio.
fn MeasureCtx::compute_length(
    &mut self,
    child: &mut WidgetPod<impl Widget + ?Sized>,
    auto_length: LenDef,
    context_size: LayoutSize,
    axis: Axis,
    cross_length: Option<f64>,
) -> f64;
```

## Layout

```rust
/// Lays out the widget with the given `size`.
///
/// For every child widget, as defined by [`children_ids`], the container must:
///
/// 1. (Optionally) Call [`LayoutCtx::compute_size`] to get the size the child wants to be.
///    If the container is doing complex layout, it might call `compute_size` multiple times.
///    However, that should be done with great care, as it's a quick path to slow performance.
///    Especially when such complex layout widgets are nested, layout complexity will explode.
///    If the container has somehow already decided on the child length on one axis, then it
///    should instead call [`LayoutCtx::compute_length`] with the correct `cross_length`.
/// 2. Decide on a final [`Size`] that the child should be. The parent is in control.
/// 3. Call [`LayoutCtx::run_layout`] on the child with the chosen size.
/// 4. Call [`LayoutCtx::place_child`] to give the child a location, relative to the parent.
fn Widget::layout(&mut self, ctx: &mut LayoutCtx<'_>, props: &PropertiesRef<'_>, size: Size);

/// Computes the size that the `child` widget wants to be.
///
/// `auto_size` specifies the fallback behavior if the child has any dimension as [`Dim::Auto`].
/// `context_size` is the size, in device pixels, that is used to resolve relative sizes.
fn LayoutCtx::compute_size(
    &mut self,
    child: &mut WidgetPod<impl Widget + ?Sized>,
    auto_size: SizeDef,
    context_size: LayoutSize,
) -> Size;

/// Computes the length that the `child` widget wants to be on the given `axis`.
fn LayoutCtx::compute_length(...)
```

## Performance

Fixed lengths always short-circuit. We want to skip measuring whenever we can. Masonry has a small built-in LRU cache per widget for measurement results to help reduce real calls even more. Costly layout widgets like text should have their own domain specific caches inside the widget. Especially because the Masonry cache gets invalidated according to layout requests, which may not always affect the costly part. The masonry cache is bypassed and verified with debug assertions enabled, which will help catch non-pure `measure` implementations. We'll see if we need to claw back that performance and require an opt-in to that cache verification.

The per-axis `Widget::measure` is designed to be calling only per-axis `MeasureCtx::compute_length` on its children. Single axis computation can be a lot less work, depending on the widget. Also, the widgets that need to measure more of their children need to really opt-in to that extra cost. No accidental calling of `layout` from `measure`.

## Safety and debugging

All the new layout functions sanitize their inputs and outputs. If a `NaN` or `f64::INFINITY` makes an appearance, there will be a debug assertion and a safe fallback.

## Future-proofing

Following the plan ([#1264]) to make the insides of widgets and Masonry core operate on device pixels I wrote all the new layout code with a `scale` variable, which is currently a constant `1.0` but can be easily changed once the scale rework goes forward.

The core parts of the new system are built with the expectation that we'll eventually support more than the classic [`horizontal-tb`] writing mode, e.g. Japanese. So some functions have defined `inline` and `block` axis constants at the top with the rest of the function being agnostic. That said, I did exactly zero testing in other writing modes because our text widgets don't support it yet.

## Test screenshots

Most tests in general continue to pass as before, however a bunch of screenshots have changed. All have good reasons which I outline below.

* `flex_row_cross_axis_baseline`
  New layout flex baseline alignment starts from the top, same as web.

* `example_grid_masonry_initial`, `grid_with_changed_spacing`, `portal_button_list_no_scroll`, `portal_button_list_scroll_to_item_13`, `portal_button_list_scroll_to_item_3`, `portal_button_list_scrolled`
  New layout aligns button text to center even when button is forced below its min intrinsic size.

* `image_layout_fitheight`, `image_layout_none`
  New layout positions image to center, same as web.

* `label_label_alignment_flex`
  New layout no longer has quirk of label always having narrow width with TextAlign::Start. Fixes test case.

* `slider_disabled`, `slider_drag_initial_at_25`, `slider_drag_released_at_75`, `slider_drag_to_75`, `slider_initial_state`, `slider_keyboard_focused`,  `slider_keyboard_moved`
  New layout accepts higher than max-intrinsic height, same old paint code centers vertically.

* `rect_big_red_rectangle`, `rect_blue_rectangle`,  `rect_hovered_rectangle`
  New layout respects the chosen size of the rectangle. Fixes test case.

* `flex_col_cross_axis_baseline`, `flex_col_cross_axis_center`, `flex_col_cross_axis_end`, `flex_col_cross_axis_stretch`, `flex_col_cross_axis_start`, `flex_row_main_axis_center`, `flex_row_main_axis_end`, `flex_row_main_axis_spaceAround`, `flex_row_main_axis_spaceBetween`, `flex_row_main_axis_spaceEvenly`, `flex_row_main_axis_start`
  New layout gives flexible children their correct sizes, previously the children shrunk on their own initiative.

* `flex_col_fill_main_axis`, `flex_row_fill_main_axis`
  New layout flex always fills main axis, same as web. Explicit dimensions are available to limit this.

## Minor notes

* ~~`Length` and `AsUnit` needed to migrate to `masonry_core`. I know the goal outlined in [#1264] was to not deal with logical pixels in the core. However, the core needs to be resolving `Length`. It's a big ergonomics hit to not do it in core.~~ Moved this question into #1565.
* `Widget::layout` no longer gets mutable properties. None of the code actually used it and it felt like a mistake to allow editing these in `layout`. If there is value in it, then maybe we should allow mutating only those which don't affect layout.

# Follow-up work

* `Label` and other text widgets need cleanup and further optimizations. For now I focused on getting things working, but they could be far less messy and more efficient.

* We should probably introduce min and max dimensions as properties. These open up some useful styling choices. They should probably override a parent's "final" choice. However, this needs more thinking.

* The border and padding boilerplate code is out of control. Plus, not every widget even supports both of these! I think we should make these core properties as well. Then Masonry would guarantee their presence for every widget and handle the layout and drawing for them. The `Widget::layout` method would receive the inner content box `Size` instead. Life would be so much easier. The one counterpoint I can think of is irregularly shaped widgets like a circle. However, that sort of thing can probably be solved with a `border_shape(props)` or `paint_border(props)` added to the `Widget` trait.

* The `Property::prop_changed(ctx, property_type)` boilerplate is fast getting out of control too. Could we move the common cases to the framework instead? I haven't looked closely yet, but it is definitely a fragile and verbose situation.

* Are `SizedBox` days numbered? I replaced a bunch of its usage in Xilem already. Having a whole widget (with its own state and everything) for what can be achieved with a simple `Dimensions` property seems like overkill. We'll see, but I think its utility is shrinking fast.

* `Flex` should gain grow and shrink options, plus more basis options. It can be far more powerful now.

* Word wrap and overflow/clip properties need to be split as they are orthogonal and can be mixed in any combination.

* More screenshot tests of more complex layouts to make sure the system stays functional.

[1]: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Layout.20Discussion/with/478177146
[2]: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Olivier's.20thoughts.20about.20Layout/with/479859089
[3]: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Brainstorming.20layout.20changes/with/504675294
[4]: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Design.20proposal.20for.20layout/with/515100278
[5]: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Layout.20refactor.3A.20surgical.20edition/with/534963761
[6]: https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Unbounded.20layout.20constraints/with/565200382

[#1264]: https://github.com/linebender/xilem/issues/1264
[`horizontal-tb`]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/writing-mode

### DIFF
--- a/masonry_core/src/passes/layout.rs
+++ b/masonry_core/src/passes/layout.rs
@@ -247,10 +247,7 @@ pub(crate) fn resolve_size(
     if let Some(inline_length) = inline_length
         && let Some(block_length) = block_length
     {
-        return match inline {
-            Axis::Horizontal => Size::new(inline_length, block_length),
-            Axis::Vertical => Size::new(block_length, inline_length),
-        };
+        return inline.pack_size(inline_length, block_length);
     }
 
     // Otherwise fall back to measurement
@@ -288,10 +285,7 @@ pub(crate) fn resolve_size(
         .sanitize("measured block length")
     });
 
-    match inline {
-        Axis::Horizontal => Size::new(inline_length, block_length),
-        Axis::Vertical => Size::new(block_length, inline_length),
-    }
+    inline.pack_size(inline_length, block_length)
 }
 
 // --- MARK: RUN LAYOUT
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -985,10 +985,7 @@ impl Widget for Flex {
                     Some(child_main_length),
                 );
 
-                match main {
-                    Axis::Horizontal => Size::new(child_main_length, child_cross_length),
-                    Axis::Vertical => Size::new(child_cross_length, child_main_length),
-                }
+                main.pack_size(child_main_length, child_cross_length)
             };
 
         // Helper function to lay out children
--- a/masonry/src/widgets/split.rs
+++ b/masonry/src/widgets/split.rs
@@ -532,44 +532,26 @@ where
 
     fn layout(&mut self, ctx: &mut LayoutCtx<'_>, _props: &PropertiesRef<'_>, size: Size) {
         let bar_area = self.bar_area();
-        let space = match self.split_axis {
-            Axis::Horizontal => Size::new((size.width - bar_area).max(0.), size.height),
-            Axis::Vertical => Size::new(size.width, (size.height - bar_area).max(0.)),
-        };
+        let split_space = (size.get_coord(self.split_axis) - bar_area).max(0.);
+        let cross_space = size.get_coord(self.split_axis.cross());
 
         // Update our effective split point to respect our size
-        let split_space = space.get_coord(self.split_axis);
         self.split_point_effective = self.calc_effective_split_point(split_space);
 
-        let (child1_size, child2_size) = match self.split_axis {
-            Axis::Horizontal => {
-                let child1_width = (space.width * self.split_point_effective).floor().max(0.0);
-                let child2_width = (space.width - child1_width).max(0.0);
-                (
-                    Size::new(child1_width, space.height),
-                    Size::new(child2_width, space.height),
-                )
-            }
-            Axis::Vertical => {
-                let child1_height = (space.height * self.split_point_effective).floor().max(0.0);
-                let child2_height = (space.height - child1_height).max(0.0);
-                (
-                    Size::new(space.width, child1_height),
-                    Size::new(space.width, child2_height),
-                )
-            }
-        };
+        let child1_split_space = (split_space * self.split_point_effective).floor().max(0.);
+        let child2_split_space = (split_space - child1_split_space).max(0.);
+
+        let child1_size = self.split_axis.pack_size(child1_split_space, cross_space);
+        let child2_size = self.split_axis.pack_size(child2_split_space, cross_space);
 
         ctx.run_layout(&mut self.child1, child1_size);
         ctx.run_layout(&mut self.child2, child2_size);
 
-        // Top-left align for both children, out of laziness.
-        // Reduce our unsplit direction to the larger of the two widgets
+        // Top-left align both children.
         let child1_origin = Point::ORIGIN;
-        let child2_origin = match self.split_axis {
-            Axis::Horizontal => Point::new(child1_size.width + bar_area, 0.0),
-            Axis::Vertical => Point::new(0.0, child1_size.height + bar_area),
-        };
+        let child2_origin = self
+            .split_axis
+            .pack_point(child1_split_space + bar_area, 0.);
         ctx.place_child(&mut self.child1, child1_origin);
         ctx.place_child(&mut self.child2, child2_origin);
     }


nicoburns:
I mostly think this looks good. However I have some questions / feedback which I'll probably drip-feed over time rather than trying to write everything up at once.

### `auto_size` / `auto_len` paremeters

```rust
///`auto_length` specifies the fallback behavior if the child's dimension is [`Dim::Auto`].
/// Most widgets should use [`LenDef::FitContent`] to ask the child to fit inside the
/// available space. However sometimes a different fallback makes more sense, e.g.
/// `Grid` uses [`LenDef::Fixed`] to fall back to the exact allocated child area size.
/// `auto_length` values must be finite, non-negative, and in device pixels.
/// An invalid `auto_length` will fall back to [`LenDef::MaxContent`].
```

I'm not convinced by the `auto_size` / `auto_len` parameters to `compute_size` / `compute_length`. Usually you would be calling these functions in order to compute *inputs* to the computation of the "auto size" of a node. So I would not necessarily expect the caller (parent widget) to have an appropriate auto size/len available when calling this function.

For the example given in the doc comment I would expect the parent widget to simply ignore the size returned from the child widget if it "doesn't like it" (I'm presuming the parent will call `WidgetPod::set_size` or similar to actually set the final size of the child?). I would also expect the Grid widget to pass the child area size as the `context_size` / `context_length` so that the child widget can resolve against that rather than the total size of the grid.

An example of this would be something like a HTML-style `<table>` widget that wants to automatically size it's columns to fit the data (CSS Grid also supports this). In that case you don't know the size of the table cell to use as a fallback until *after* you've called this method. And by the time you do know, you don't need to call the method again because you already have all the information you need to determine a finite pixel size.

masonry/examples/custom_widget.rs
```diff
@@ -72,31 +74,35 @@ impl Widget for CustomWidget {
...
+        // Return a result based on the parent's request
+        match len_req {
+            LenReq::MinContent => min_length,
+            LenReq::MaxContent => max_length,
+            LenReq::FitContent(space) => space.clamp(min_length, max_length),
         }
     }
 
+    fn layout(&mut self, _layout_ctx: &mut LayoutCtx<'_>, _props: &PropertiesRef<'_>, _size: Size) {
+        // Here is a good place to calculate size dependant values
```
PoignardAzur:
```suggestion
        // This method is a good place to calculate size-dependent values
```

masonry/src/doc/color_rectangle.rs
```diff
@@ -134,15 +126,29 @@ impl Widget for ColorRectangle {
...
+        //       https://github.com/linebender/xilem/issues/1264
+        let scale = 1.0;
+
+        match len_req {
+            LenReq::MinContent | LenReq::MaxContent => match axis {
+                Axis::Horizontal => 200. * scale,
+                Axis::Vertical => 100. * scale,
+            },
+            LenReq::FitContent(space) => space,
+        }
```
PoignardAzur, line 147: [Post-MVP] This seems really complex for an entry-level tutorial. I hope we can come up with an abstraction to better express this in a single line.

masonry/src/doc/color_rectangle.rs
```diff
@@ -9,15 +9,12 @@
 // TODO - Find some way to check that code chunks in docs
 // are up to date with this file.
 
-#![expect(missing_docs, reason = "This is example code")]
-
```
PoignardAzur, line 13: [Post-MVP] Why is this removed?

masonry/src/doc/implementing_widget.md
```diff
@@ -49,7 +50,9 @@ In the course of a frame, Masonry will run a series of passes over the widget tr
 - `on_pointer_event`, `on_text_event` and `on_access_event` are called once after a user-initiated event (like a mouse click or keyboard input).
 - `on_anim_frame` is called once per frame for animated widgets.
 - `update` is called many times during a frame, with various events reflecting changes in the widget's state (for instance, it gets or loses text focus).
-- `layout` is called during Masonry's layout pass. It takes size constraints and returns the widget's desired size.
+- `measure` and `layout` are called during Masonry's layout pass.
+  `measure` computes the intrinsic size of the widget one a single axis.
+  `layout` receives a chosen size and lays out its children accordingly.
```
PoignardAzur, line 55: The doc refers to intrinsic sizes a few times, but it's not defined anywhere AFAICT. We should probably add at least a TODO entry to `masonry_core/src/doc/masonry_concepts.md`.

masonry/src/doc/implementing_widget.md
```diff
@@ -167,31 +169,45 @@ impl Widget for ColorRectangle {
...
+    fn layout(&mut self, _ctx: &mut LayoutCtx<'_>, _props: &PropertiesRef<'_>, _size: Size) {}
+
     // ...
 }
 ```
 
-Our size is static, and doesn't perform complex operations like line-breaking, or depend on context information like "the widget is currently hovered", so it can be written as a one-liner.
+Our intrinsic size is always 200x100 in logical pixels.
+If we receive a request to fit into a specific space, we always greedily take all of it.
+We don't perform complex operations like line-breaking, or depend on context information like "the widget is currently hovered".
```
PoignardAzur: [Post-MVP] Overall I think this section of the tutorial doc will need a rewrite.

The tutorial needs to give new users just enough info to know how to implement layout for a widget, while hinting that there's more complex systems at play, while letting them know that they don't need to worry about these systems to write a basic widget. That's a fine line to walk, and outside the scope of this PR.

I think we'll probably write an entire chapter about layout eventually, because the new architecture has a lot of interactions to explain (e.g. if I understand correctly, the size passed to a widget's `layout()` depends on both the sized pass by the parent to `run_layout()` and the widget's sizing properties).

masonry/src/properties/border_width.rs
```diff
@@ -39,27 +39,56 @@ impl BorderWidth {
         ctx.request_layout();
     }
 
-    /// Shrinks the box constraints by the border width.
+    /// Returns the [`Length`] on the given `axis`.
+    pub fn length(&self, _axis: Axis) -> Length {
+        Length::px(self.width * 2.)
+    }
```
PoignardAzur: [Post-MVP] The name and doc comment are a little unclear. This returns the amount of space taken by both borders on the given axis.

masonry/src/properties/border_width.rs
```diff
@@ -39,27 +39,56 @@ impl BorderWidth {
...
+
+    /// Shrinks the `size` by the border width.
+    ///
+    /// The returned [`Size`] will be non-negative and in device pixels.
+    ///
+    /// The provided `size` must be in device pixels.
```
PoignardAzur, line 54: [Post-MVP] If we end-up having to document this in a lot of places, we might want to use newtypes more heavily instead.

masonry/src/tests/accessibility.rs
```diff
@@ -3,9 +3,9 @@
 
 use accesskit::NodeId;
 use assert_matches::assert_matches;
-use masonry_core::core::{NewWidget, Widget, WidgetTag};
 use masonry_testing::{ModularWidget, Record, TestHarness, TestWidgetExt, assert_any, assert_none};
 
+use crate::core::{NewWidget, Widget, WidgetTag};
```
PoignardAzur: These changes could be split in a separate PR for readability.

masonry/src/tests/layout.rs
```diff
@@ -190,20 +187,32 @@ fn skip_layout_when_cached() {
...
     // The button did not request layout and its input constraints are the same:
-    // Nothing should happen to it.
-    assert_matches!(harness.take_records_of(button_tag)[..], []);
+    // Nothing besides measurements should happen to it.
+    let button_records: Vec<_> = harness
+        .take_records_of(button_tag)
+        .into_iter()
+        .filter(|r| !matches!(r, Record::Measure(_)))
+        .collect();
+    assert_matches!(button_records[..], []);
```
PoignardAzur, line 199: I'm confused why this test had to be changed. Feels like with caching, measurements should be skipped as well?

masonry/src/tests/layout.rs
```diff
@@ -190,20 +187,32 @@ fn skip_layout_when_cached() {
...
-        ctx.place_child(child, pos);
-        size
-    });
+    let parent = ModularWidget::new_parent(child)
+        .measure_fn(|child, ctx, _props, axis, len_req, cross_length| {
+            let auto_length = len_req.into();
+            let context_size = LayoutSize::maybe(axis.cross(), cross_length);
+
+            ctx.compute_length(child, auto_length, context_size, axis, cross_length)
+        })
```
PoignardAzur: I think this one is already the default implementation in new_parent, so you can remove it.

masonry/src/tests/update.rs
```diff
@@ -246,6 +245,10 @@ fn stash_parent() {
         [
             Record::Update(Update::StashedChanged(false)),
             // Un-stashing also requests a layout pass.
+            Record::Measure(_),
+            Record::Measure(_),
+            Record::Measure(_),
+            Record::Measure(_),
```
PoignardAzur, line 251: [Post-MVP] I'm a little curious, why do we get four measure calls per layout pass?

masonry/ARCHITECTURE.md
```diff
@@ -2,19 +2,31 @@
...
-Developers trying to write immediate-mode GUIs, Elm-architecture GUIs, functional reactive GUIs, etc, can import Masonry and get a platform to create windows (often using Winit as a backend) each with a tree of widgets. Each widget has to implement the `Widget` trait that Masonry provides.
+Developers trying to write immediate-mode GUIs, Elm-architecture GUIs, functional reactive GUIs, etc, can import Masonry and get a platform to create windows (often using Winit as a backend) each with a tree of widgets.
+Each widget has to implement the `Widget` trait that Masonry provides.
 
 
 ## High-level goals
 
 Masonry has some opinionated design goals:
 
-- **Be dumb.** As a general rule, Masonry doesn't do "algorithms". It has no reconciliation logic, no behind-the-scenes dataflow, no clever optimizations, etc. It tries to be efficient, but that efficiency comes from well-designed interfaces and well-placed abstraction boundaries. High-level logic should be implemented in downstream crates.
```
PoignardAzur: I'd rather not have the line-breaking changed as a drive-by fix. It can be its own PR if necessary.

(But also, I'd rather not apply the usual markdown one-line-per-sentence style to ARCHITECTURE.md files. I think it's fine for markdown files meant to be rendered to html, but I expect ARCHITECTURE.md to be most often read from a text editor, and the one-line-per-sentence style hurts readability there.)

masonry_core/src/core/contexts.rs
```diff
@@ -510,11 +524,195 @@ impl AccessCtx<'_> {
     }
 }
 
+// --- MARK: COMPUTE LENGTH
+impl_context_method!(MeasureCtx<'_>, LayoutCtx<'_>, {
+    /// Computes the length that the `child` widget wants to be on the given `axis`.
```
PoignardAzur, line 535: [Post-MVP] A few notes:

- Referring to both  `measure` and `layout` in this doc comment might be a little confusing, given that it'll end up copy-pasted in both `MeasureCtx` and `LayoutCtx`.
- Referring to the Grid widget inside masonry_core isn't exactly *forbidden*, but it's awkward given that Grid is defined in a reverse-dependency.
- The doc uses a voice which is *kinda* similar to passive voice, where it describes things that "will" happen but doesn't describe what sections of the code are responsible for those things, and where they will be observed. This isn't *bad* exactly, but it's not ideal for functions that recurse to user code. When A calls B and B calls A', it's important to explicitly lay out what happens in A/A' vs B.

Overall I think some of this will be moved to a dedicated module in the `docs` folder.

masonry_core/src/core/contexts.rs
```diff
@@ -510,11 +524,195 @@ impl AccessCtx<'_> {
...
+    /// the result of the measurement, then the widget is responsible for requesting layout
+    /// when any of that data changes. For properties, this is handled in [`property_changed`].
+    /// For any other data you reference, the exact mechanism of detecting changes is up to you.
+    /// If you can't detect changes of the referenced data, disable the cache with this method.
+    ///
+    /// [`measure`]: Widget::measure
+    /// [`property_changed`]: Widget::property_changed
+    pub fn cache_result(&mut self, cache_result: bool) {
+        self.cache_result = cache_result;
+    }
```
PoignardAzur, line 672: I'm not sure about this method. Reading this doc, it's not super obvious to me when users are supposed to call it, and I *really* don't trust users to get it right.

Overall, I'd rather have the caching be handled purely by Masonry, transparently to the user, and have `request_layout` be the only user-controlled knob.

masonry_core/src/core/contexts.rs
```diff
@@ -510,11 +524,195 @@ impl AccessCtx<'_> {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if `cross_length` is non-finite or negative and debug assertions are enabled.
+    ///
+    /// [`measure`]: Widget::measure
+    /// [`compute_length`]: Self::compute_length
+    /// [`auto_length`]: Self::auto_length
+    /// [`context_size`]: Self::context_size
+    pub fn redirect_measurement(
```
PoignardAzur, line 698: [Post-MVP] I think "redirect_measurement" doesn't quite communicate the right idea for what the function does. Not sure what a better name would be.

PoignardAzur (commented):
EDIT: Crap, I misclicked. Well, consider this a draft of a review.

masonry_core/src/core/widget.rs
```diff
@@ -229,37 +230,129 @@ pub trait Widget: AsDynWidget + Any {
     /// Handles a property being added, changed, or removed.
     fn property_changed(&mut self, ctx: &mut UpdateCtx<'_>, property_type: TypeId) {}
 
-    /// Computes layout and return the widget's size.
+    /// Computes the length that the widget wants to be on the given `axis`.
```
PoignardAzur, line 231: [Post-MVP] That documentation is way too long. Some of it will probably be moved to a dedicated doc file as well.

masonry_core/src/core/widget.rs
```diff
@@ -229,37 +230,129 @@ pub trait Widget: AsDynWidget + Any {
...
+
+    /// Lays out the widget with the given `size`.
+    ///
+    /// A container widget must ensure all its direct children are laid out in this method.
+    ///
+    /// For every child widget, as defined by [`children_ids`], the container must:
+    ///
+    /// 1. (Optionally) Call [`LayoutCtx::compute_size`] to get the size the child wants to be.
+    ///    If the container is doing complex layout, it might call `compute_size` multiple times.
+    ///    However, that should be done with great care, as it's a quick path to slow performance.
```
PoignardAzur: [Post-MVP] I think we should avoid making such sweeping statements in the doc about performance, especially since we haven't really measured layout-related performance drops "in the wild" and we're just assuming they'd be likely to happen.

If we want to talk about performance, it should be in a specific chapter of the tutorial doc, with a focus on measurements instead of heuristics like "don't call `measure` too much". Otherwise people tend to treat these heuristics as gospel and skip the measurement part altogether.

masonry/src/doc/color_rectangle.rs
xStrom: Because CI was complaining about it. I don't know why, I couldn't replicate it locally.

masonry_core/src/core/widget_state.rs
```diff
@@ -6,7 +6,10 @@ use std::any::TypeId;
 use tracing::Span;
 use vello::kurbo::{Affine, Insets, Point, Rect, Size, Vec2};
 
-use crate::core::{LayoutCache, WidgetId, WidgetOptions};
+use crate::{
+    core::{WidgetId, WidgetOptions},
```
PoignardAzur: Should split those imports.

masonry/ARCHITECTURE.md
xStrom: Our policy isn't limited to *files meant to be rendered to html*, it's for markdown files.

Also, this has been an accepted policy for many years now. I suggest you just come to terms with it, but if not, Zulip is a great place to discuss the potential of changing the policy. PR review is not really an appropriate place to start pushing back against accepted policy.

As for splitting it out to a different PR, seems like busywork to me but sure I'll do that: #1569.

masonry_core/src/doc/pass_system.md
```diff
@@ -203,20 +205,24 @@ External mutation is how Xilem applies any changes to the widget tree produced b
 Some notes about pass context types:
 
 - Render passes should be pure and can be skipped occasionally, therefore their context types ([`PaintCtx`] and [`AccessCtx`]) can't set invalidation flags or send signals.
-- The `layout` and `compose` passes lay out all widgets, which are transiently invalid during the passes, therefore [`LayoutCtx`]and [`ComposeCtx`] cannot access the size and position of the `self` widget.
+- The `layout` and `compose` passes lay out all widgets, which are transiently invalid during the passes, therefore [`MeasureCtx`], [`LayoutCtx`], and [`ComposeCtx`] cannot access the size and position of the `self` widget.
```
PoignardAzur, line 208: [Post-MVP] This isn't quite true anymore (since size is passed down) but gestures at something true (the layout rect and global transform aren't available in these passes). Will need to reformulate.

masonry_core/src/layout/dim.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::layout::{LenDef, Length};
+
+/// Specifies how a widget dimension's length is derived.
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
+pub enum Dim {
```
PoignardAzur, line 8: [Post-MVP] Overall the naming scheme for all these types (`Dim`, `LenDef`, `LenReq`, etc) needs improvement. In particular, they don't communicate whether these are input or output types.

masonry_core/src/layout/len_def.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{layout::LenReq, util::Sanitize};
```
PoignardAzur: These imports should be split.

masonry_core/src/core/contexts.rs
xStrom: There is a debug assertion for the case where widget builders get it wrong, so it's not like such a mistake can be easily hidden.

It's an escape hatch to allow writing `Widget::measure` in a quirky way. It is not currently called by any built-in widgets.

The case where you need to set this is when you reference some data inside your `measure` implementation that has changed but you didn't request layout for. Definitely an anti-pattern in general, but I'm not sure that it should be made completely impossible.

masonry_core/src/layout/length.rs
```diff
@@ -3,7 +3,9 @@
 
 /// A value representing a width, height, or similar distance value.
 ///
-/// Its value is always finite and non-negative.
+/// It is in logical pixels.
```
PoignardAzur: [Post-MVP] I don't think that comment should be included. In practice, yes, the inner fields stores an amount of logical pixels. But from the point of view of the users, it could store anything.

The parts that are in logical/physical pixels are the constructor and getter methods.

masonry_core/src/core/contexts.rs
xStrom: I mean, it's a very matter-of-fact naming. The measurement is being redirected to a child, just like a HTTP redirect.

masonry_core/src/layout/measurement_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// Implemented as a linear search LRU cache over [`Vec`],
+/// because we expect the dataset to be tiny.
+///
+/// We don't expect any `NaN`s to be in [`MeasurementInputs`],
+/// but even if there are that is fine, because we clear the cache regularly.
+#[derive(Clone, Debug)]
+pub(crate) struct MeasurementCache {
+    entries: Vec<(MeasurementInputs, f64)>,
+}
```
PoignardAzur: [Post-MVP] Overall my preferred layout for this file would be:

```rust
struct MeasurementCache { ... }

struct MeasurementInputs { ... } 

impl MeasurementInputs { ... }

// everything else
```

While it doesn't make a huge different for IDE users relying on AST outlines, for text users starting with types (and especially with the eponymous type) makes the file more readable.

masonry_core/src/layout/mod.rs
```diff
@@ -4,7 +4,46 @@
...
+pub use len_def::*;
+pub use len_req::*;
 pub use length::*;
+pub(crate) use measurement_cache::*;
+pub use size_def::*;
+
+use vello::kurbo::Size;
+
+/// Ergonomic layout helper methods for [`Size`].
+pub trait LayoutCalc {
```
PoignardAzur: [Post-MVP] Not sure about this. Trait have a heavy cognitive cost for new users and we already have a lot of them in Masonry. This one is only used in two places in the PR, so I'm not convinced it's pulling its weight.

masonry_core/src/layout/size_def.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// All the [`LenDef`] values must be finite, non-negative, and in device pixels.
+    /// Invalid [`LenDef`] values will fall back to [`LenDef::MaxContent`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if `width` or `height` contain non-finite or negative values
+    /// and debug assertions are enabled.
+    pub fn new(width: LenDef, height: LenDef) -> Self {
+        let width = width.sanitize("SizeDef width");
+        let height = height.sanitize("SizeDef height");
```
PoignardAzur, line 51: [Post-MVP] Something that just occurred to me: if you make sanitize a non-trait function, you could make this constructor `const`.

masonry_core/src/passes/layout.rs
```diff
@@ -3,33 +3,321 @@
...
+    // Return immediately if we already have a fixed size
+    let inline_length = inline_def.fixed();
+    let block_length = block_def.fixed();
+    if let Some(inline_length) = inline_length
+        && let Some(block_length) = block_length
+    {
+        return match inline {
+            Axis::Horizontal => Size::new(inline_length, block_length),
+            Axis::Vertical => Size::new(block_length, inline_length),
+        };
```
PoignardAzur: [Post-MVP] Could be rewritten as a one-liner with the new Size constructor.

masonry_core/src/properties/dimensions.rs
```diff
@@ -0,0 +1,179 @@
...
+
+// Every widget has dimensions.
+impl<W: Widget> HasProperty<Dimensions> for W {}
+
+/// The size of the widget, including borders and padding.
+///
+/// Generally this is meant to be used as a write-only property.
+/// Masonry will automatically resolve the dimensions to actual numbers during the layout pass.
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
+pub struct Dimensions {
```
PoignardAzur, line 21: [Post-MVP] This should probably be two separate properties, especially once we implement property cascade.

masonry/src/tests/accessibility.rs
xStrom: Moved these unrelated import changes to #1570.

masonry/ARCHITECTURE.md
PoignardAzur: > seems like busywork to me

It's not busywork, it's to avoid having three paragraphs of content change be hidden in twenty paragraphs of formatting change.

masonry/src/doc/color_rectangle.rs
xStrom: I guess, maybe, for an entry-level tutorial we could just return `100. * scale` regardless of `len_req`. Need to think about it though if it's a good idea to hide reality that much.

masonry/src/doc/implementing_widget.md
xStrom: Yeah "intrinsic" is a bit too complex for not a lot of gain. Most docs could probably just use the term "preferred size" which is a lot clearer to regular humans. I'll do some updating.

nicoburns: There is also the term "content-based" size, which is what I generally use whenI try to explain what the term "intrinsic" size (which the CSS spec uses) actually means.

masonry/src/doc/implementing_widget.md
xStrom: Yes rewriting the tutorial bits around layout makes a lot of sense. I would wait with that until some further follow-up work has landed too, because things will still change.

masonry/src/properties/border_width.rs
xStrom: I'll add clarifying statements to these docs.

masonry/src/properties/border_width.rs
xStrom: Yes, longer term newtypes are probably the answer. However, I wanted to make sure that things are documented in the interim.

masonry/src/tests/layout.rs
xStrom: With caching they are skipped, however with debug assertions enabled we verify the cache -- by measuring. I'll refactor this test to enable the filter only with debug assertions and add a comment.

masonry/src/tests/layout.rs
xStrom: It is indeed, I'll simplify.

masonry/src/tests/update.rs
xStrom: There's both Horizontal and Vertical calls, so that cuts this down to two.

The remaining two sets are explained as follows:
* `ModularWidget::measure` measures its child to determine its own length
* `ModularWidget::layout` computes the child size (`measure` under the hood) given the actual constraints

`ModularWidget` is implemented as a generic container like `Flex`, not as a shrinkwrap like `Passthrough`. Now, maybe `ModularWidget` should be more of a hybrid between those two? Not sure. However this gets into a side question and the four measurements will still appear in other cases like `Button` measuring its child `Label`.

Overall, this is expected behavior and shouldn't be an issue.

masonry_core/src/core/contexts.rs
xStrom: * Yeah it's not ideal that the docs have to cater to both `MeasureCtx` and `LayoutCtx`, however I think duplicating the implementation is even less ideal.

* Responded regarding the `Grid` topic in a general reply in the PR.

* Regarding "will" happen, not exactly sure which part you mean. The stuff in e.g. `MeasureCtx::auto_length`? In any case, we can certainly consider improvements. I did find it especially tricky with using terms like "parent" because it could reference the current widget (because it has children!) or the grandparent! So I removed the ambigious places and tried to use "parent" only when it applied to both the current widget and the grandparent.

masonry_core/src/core/widget.rs
xStrom: I personally like when documentation is self-sufficient, but I guess I can see it being noisy for experienced devs looking just for keywords regarding certain inputs/outputs.

The issue with external docs is that they go out of sync way more easily. They are out of sight during development and use.

Anyway, this is something we can discuss as a follow-up.

masonry_core/src/core/widget.rs
xStrom: This is an odd comment from you, considering you were afraid of O(N^2) on Zulip.

Repeated calls on the same child are a well known complexity explosion cause. It's not speculation. Look at e.g. [Flutter docs](https://api.flutter.dev/flutter/widgets/IntrinsicHeight-class.html) having a similar warning:

> This class is relatively expensive, because it adds a speculative layout pass before the final layout phase. Avoid using it where possible. In the worst case, this widget can result in a layout that is O(N²) in the depth of the tree.

Anyway, I removed the performance note for now.

masonry_core/src/doc/pass_system.md
xStrom: Yeah I avoided doing too much doc work here because I expect things to change even more, soon.

masonry_core/src/layout/dim.rs
xStrom: Umm, that's not a common property of our types? `Size`, `Point` etc don't communicate input/output either, or if you wish for more Masonry specific, `MainAxisAlignment` and `Length`.

One of my key goals with the names was being concise.

`Dim` instead of `Dimension`
`LenDef` instead of `LengthDefinition`
`LenReq` instead of `LengthRequest`

Just like Rust has `Vec` instead of `Vector` and `std` instead of `standard`.

I'm not super attached to the chosen names, but I wouldn't be a fan of a much more verbose replacement.

masonry_core/src/layout/length.rs
xStrom: Okay, but isn't the same true for the finite and non-negative part? It could store things negatively and then just flip the sign on getters.

Anyway, I'll remove the logical pixels line and leave the other stuff for now.

masonry_core/src/layout/measurement_cache.rs
xStrom: Sure, I'll reorder those.

masonry_core/src/layout/mod.rs
xStrom: Yeah this trait ended up not being needed too much right now, I wrote it in the beginning of this work. I'll just delete it for now. We can add it back later if the pattern gets more popular. Adding a `Size::with_axis()` builder into Kurbo would also solve half the problem.

masonry_core/src/layout/size_def.rs
xStrom: Yes, but it uses formatting.

To quote `Length`:

```rust
// TODO - Make const once const formatting is allowed.
// (aka see you in 2030)
debug_panic!("Invalid length value '{value}'");
```

masonry_core/src/passes/layout.rs
xStrom: Which new constructor? Not sure what you're referring to. Kurbo, even on `main`, doesn't seem to have any other constructors.

masonry_core/src/properties/dimensions.rs
xStrom: Perhaps, but one immediate thought is that it would introduce a choice between two downsides:
* We copy-paste `Dim` to be both `Width` and `Height` - code/doc duplication.
* We lose the ability for automatic `.into()` resolving, because Rust doesn't know if `Dim` gets converted to `Width` or `Height`.

Maybe there's a way to split them up while preserving ergonomics?

Yeah, this is definitely a follow-up exploration.

masonry_core/src/core/widget.rs
Philipp-M: FWIW I think long doc-comments are *fine* and even something that I think is good (as far as it's descriptive).
I can decide myself whether I want to skim non-relevant parts for me. I rather have this in a doc-comment (which I can quickly look at via e.g. rust-analyzer) than having to look up a separate doc file (which can of course still be additional to this).

xilem/examples/http_cats.rs
```diff
@@ -90,9 +88,10 @@ impl HttpCats {
             flex_col((
                 // Add padding to the top for Android. Still a horrible hack
                 FlexSpacer::Fixed(40.px()),
-                split(left_column, portal(sized_box(info_area).expand_width())).split_point(0.4),
-            ))
-            .must_fill_major_axis(true),
+                split(portal(left_column), info_area)
+                    .split_point(0.4)
+                    .flex(1.),
+            )),
```
jaredoconnell, line 94: I won't block on this, but there is a change in behavior for this section of code here. I don't think it's due to changes at this exact spot, but I'm guessing it's due to other layout changes.

Before, the left column properly (for the most part) resized its contents. Now it isn't.

It appears that the xilem implementation of portal doesn't currently support a way to call `constrain_horizontal` on portal, but that's what appears to be necessary to partially revert old behavior. With that hacked in (I just added it in a non-portable way), the column flex grows properly. But it doesn't shrink to any less than the largest list item. That's less important, but will need to be addressed at some point.

masonry/ARCHITECTURE.md
jaredoconnell: Avoiding reformatting unchanged lines is a good policy. If done in the same PR, it's best done in its own commit.
It was a little harder for me to review since I had to scan each line to see if the content itself changed.

Since it sounds like this isn't an official policy, this is definitely something we should discuss and formalize.

masonry_core/src/core/widget.rs
jaredoconnell: Honestly those types of tips in the code are probably beneficial. It's important for developers to realize the costs. Maybe a more generic statement could be used, like "minimize calls to measure, as each one has a cost".

masonry_core/src/passes/layout.rs
xStrom: Okay, on further inspection I realized that this can be replaced with `inline.pack_size(inline_length, block_length)`.

xilem/examples/http_cats.rs
xStrom: This is due to `Portal` now providing actual unbounded constraints, an explicit goal of the new layout system.

You've correctly identified that `constrain_horizontal` can be used to introduce bounded constraints, which matches how `Portal` previously used to work.

I have plans to improve `Flex`, with granular grow and shrink support. However, that's a good chunk of work that isn't suitable for this already big PR. Note also that I already made `Flex` more powerful and principled here than it previously was. It had a lot of inconsistent quirks previously, that examples stumbled into depending on. I updated examples to roughly match the previous intent, but yeah the examples can use more polish for sure.

Also, the "shrink" support that the old cats example had was kind of hilarious, because of course the old `Flex` didn't support any shrinking either.

cats

PoignardAzur:
Some thoughts re: documentation:

- I think the concept of a "preferred size" (or "size the widget wants") is a bit bespoke and arbitrary, which is fine, but it means we'll need to special care to document it.
- `context_size` is also a bit bespoke. It's strongly inspired by the web's box model, which we should be explicit about.
- We'll especially want to document edge cases and overflow behavior.

PoignardAzur (approved):
LGTM.

Thanks for the follow-up changes, sorry for the late approval.

PR #1564, opened 2026-01-08
Add `Sanitize` util trait.

xStrom:
This is such a common check that needs to happen when dealing with floats that it's worth improving the ergonomics of it to a method call.

Now, one quirk of this implementation is that the docs talk about non-negative not being allowed but the actual check is for below zero. That is to say `-0.` is allowed through.

I'm not sure if it's worth forbidding `-0.`, because:

```
5. + -0. == 5.
5. - -0. == 5.
5. * -0. == -0.
5. / -0. == -inf
5_f64.powf(-0.) == 1.
```

We don't want to be dividing by zero anyway and have checks for that, so the only real sign carryover happens with multiplication, but the result will always be zero, so the problem doesn't "grow" into other negatvie numbers.

We could check for the sign bit instead and then if it's negative, do a special check for `-0.` and silently convert it to positive. Maybe worth doing?

### DIFF
--- a/masonry_core/src/util.rs
+++ b/masonry_core/src/util.rs
@@ -36,6 +36,13 @@ macro_rules! debug_panic {
 pub use crate::debug_panic;
 
 /// Provides sanitization of values.
+///
+/// This is a generic trait that doesn't specify what sanitization exactly means,
+/// as that will be implementations specific per type.
+///
+/// Right now it is also implemented for `f64` and `Option<f64>` in a way
+/// where it forbids non-finite and negative values. This is immediately useful for
+/// Masonry itself but is likely to go away as we migrate our float usage to newtypes.
 pub trait Sanitize {
     /// Returns the sanitized value.
     ///


PoignardAzur (commented):
It's not obvious from the trait documentation alone why negative values would panic. It makes more sense in the layout PR for types like `LenReq` where non-negative values are part of the type "invariant", but there's nothing obviously wrong about a negative float value.

As a compromise, I'd approve this PR if the Sanitize trait explicitly mentioned that it always expects values to be positive, with a TODO comment mentioning possible improvements.

(In general, I'm partial to the "parse, don't validate" approach, but that's a post-MVP concern.)

xStrom:
The layout newtypes where it's relevant (like `LenDef`) do have their own implementations of this trait.

It's also required for floats because all device pixels are currently held in floats. There is no `Length` equivalent for device pixels. It might be in the cards later, but it's a bit tricky because a browser-like fractional integer type might make sense but you also don't necessarily want to lose Kurbo which operates on floats. So I've just punted on it and kept using float for now, but still want to make guarantees about the state of these numbers.

The sanitize trait itself is meant to be quite generic. However, yes the float implementations are quirks of the current state. Their docs also already mentioned exactly what they do with the floats. I guess I can also add a note to the trait itself that mentions that about the current float implementations.

PoignardAzur:
> Their docs also already mentioned exactly what they do with the floats. I guess I can also add a note to the trait itself that mentions that about the current float implementations.

Right, I'd rather have the documentation in the trait itself. Nobody reads trait impl docs.

(Well, not *literally* nobody, but it's quite rare. The rustdoc collapses them by default, and people usually don't expect to find useful info in them.)

PR #1565, opened 2026-01-08
Move `Length` and `AsUnit` to `masonry_core`.

xStrom:
`Length` needs to be in core in order for the core layout engine to use it in core properties. See #1560 for context.

I also moved `AsUnit` because it's essentially a companion trait to `Length`. Now core doesn't actually need this trait right now, so this one is a bit more of an open question. I like the code locality of it next to `Length`, but if the goal is to keep `masonry_core` as slim as possible then this would be a candidate for staying in `masonry`.

I created a new module in `masonry_core` called `layout`. There will be a bunch more layout types coming, so I think this organization makes sense.

This moves some noise out of #1560.

---

I didn't do it yet, but I can see `MainAxisAlignment`, `CrossAxisAlignment`, and `UnitPoint` also better suited to living in this new core layout module. These are all lightweight and fairly foundational. No rush with these though, and I guess it partially depends on what people think of `AsUnit` living in core.

### DIFF
--- a/masonry/README.md
+++ b/masonry/README.md
@@ -73,9 +73,9 @@ The to-do-list example looks like this, using Masonry Winit as the backend:
 ```rust
 use masonry::core::{ErasedAction, NewWidget, Properties, Widget, WidgetId, WidgetTag};
 use masonry::dpi::LogicalSize;
+use masonry::layout::Length;
 use masonry::peniko::color::AlphaColor;
 use masonry::properties::Padding;
-use masonry::layout::Length;
 use masonry::theme::default_property_set;
 use masonry::widgets::{Button, ButtonPress, Flex, Label, Portal, TextAction, TextArea, TextInput};
 use masonry_winit::app::{AppDriver, DriverCtx, NewWindow, WindowId};


Philipp-M:
> I didn't do it yet, but I can see MainAxisAlignment, CrossAxisAlignment, and UnitPoint also better suited to living in this new core layout module. These are all lightweight and fairly foundational. No rush with these though, and I guess it partially depends on what people think of AsUnit living in core.

As layout is fairly independent on drawing backends or OS I think it makes sense to put it in core anyway.

DJMcNab:
> As layout is fairly independent on drawing backends or OS I think it makes sense to put it in core anyway.

This is a non-sequitur; the split between Masonry Core and Masonry isn't about drawing backends or OS backends (e.g. Masonry doesn't depend on Winit), it's about separating the core Masonry runtime from the things built on top of it.

I'm not sure that `MainAxisAlignment` and `CrossAxisAlignment` are that broadly useful (they assume a main and a cross axis, which is pretty specialised to flex, I think). On `UnitPoint` however, I would be in favour of moving it to Masonry Core.

masonry_core/src/layout/length.rs
```diff
@@ -1,8 +1,6 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use masonry_core::debug_panic;
-
 /// A value representing a width, height, or similar distance value.
 ///
 /// Its value is always finite and non-negative.
```
DJMcNab, line 6: Whilst we're here, it's probably worth pointing to `AsUnit` in these docs.

xStrom: I'll do it in a follow-up with a few other changes. Wanted to keep this PR strictly about moving.

PR #1571, opened 2026-01-13
Make `WidgetId::next()` private and remove most uses.

PoignardAzur:
Part of #1255.

### DIFF
--- a/masonry_core/src/core/contexts.rs
+++ b/masonry_core/src/core/contexts.rs
@@ -509,7 +509,7 @@ impl AccessCtx<'_> {
         self.tree_update
     }
 
-    /// Return an id which is guaranteed not to collide with [`WidgetId`]s or with previous ids returned by this function.
+    /// Returns an id which is guaranteed not to collide with [`WidgetId`]s or with previous ids returned by this function.
     pub fn next_node_id() -> NodeId {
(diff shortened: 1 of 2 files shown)


masonry_core/src/core/contexts.rs
```diff
@@ -508,6 +508,12 @@ impl AccessCtx<'_> {
     pub fn tree_update(&mut self) -> &mut TreeUpdate {
         self.tree_update
     }
+
+    /// Return an id which is guaranteed not to collide with [`WidgetId`]s or with previous ids returned by this function.
```
xStrom:
```suggestion
    /// Returns an id which is guaranteed not to collide with [`WidgetId`]s or with previous ids returned by this function.
```

masonry_core/src/core/widget.rs
```diff
@@ -499,6 +487,7 @@ pub fn find_widget_under_pointer<'c>(
 pub trait AllowRawMut: Widget {}
 
 impl WidgetId {
+    // TODO doc can't create use WidgetTag instead
```
xStrom: If you intend to leave this TODO, I think it should be expanded to full sentences so that when someone else reads it they know what it means.

PoignardAzur: Oh, crap, I thought I'd removed that.

PR #1574, opened 2026-01-14
Add `include_doc_path!` proc macro

### LINKED ISSUE #1080: Need a `file_exists!()` macro
In the places where we include images in the doc, it would be nice if we could use a macro to programmatically confirm that the images exist at build time.

It could be based on `include_str!`, though we'd want to do that in a way that doesn't hurt regular build performance. Maybe using `cfg(doc)`.

PoignardAzur:
Create new `linebender_include_doc_path` crate.
Add proc macro which checks if a local path exists, and fallback to a Github URL (or a compile error).

Use this proc macro everywhere the doc imports images. Remove existing `include_screenshot!` macros.

Add a flag in CI to check for invalid doc paths.
Fix outdated screenshot path.

Fixes #1080.

### DIFF
--- a/include_doc_path/src/logic.rs
+++ b/include_doc_path/src/logic.rs
@@ -67,7 +67,16 @@ pub(crate) fn include_doc_path_impl(
     let Some(manifest_dir) = env("CARGO_MANIFEST_DIR") else {
         return cargo_env_fallback;
     };
-    let doc_file_path = Path::new(&manifest_dir).join(&relative_path);
+    let mut doc_file_path = Path::new(&manifest_dir)
+        .join(&relative_path)
+        .to_str()
+        .unwrap()
+        .to_string();
+    if cfg!(windows) {
+        // On Windows, both "/" and "\\" are valid separators, but "\\" will be
+        // encoded as %5C by rustdoc.
+        doc_file_path = doc_file_path.replace("\\", "/");
+    }
 
     let Some(repo_url) = env("CARGO_PKG_REPOSITORY") else {
         return cargo_env_fallback;
@@ -93,7 +102,7 @@ pub(crate) fn include_doc_path_impl(
     // FLAG PRE-FALLBACK LOGIC
 
     if force_local {
-        return literal_string(doc_file_path.to_str().unwrap(), span);
+        return literal_string(&doc_file_path, span);
     }
 
     if force_url {
@@ -104,12 +113,11 @@ pub(crate) fn include_doc_path_impl(
 
     // First, check if the local file exists
     if std::fs::metadata(&doc_file_path).is_ok() {
-        return literal_string(doc_file_path.to_str().unwrap(), span);
+        return literal_string(&doc_file_path, span);
     }
 
     // If not and CHECK_DOC_PATHS is set, error out.
     if check_doc_paths {
-        let doc_file_path = doc_file_path.to_string_lossy();
         let error_message = format!("File '{doc_file_path}' not found");
         return compile_error(&error_message, span);
     }


PoignardAzur:
(By the way, "Having broken screenshots in our published docs" was the main factor holding us back from widely advertising the last Masonry release. Once this (and #1560) is merged we should start planning for the next release.)

include_doc_path/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//!
+//! - If the path points to an existing local file, it returns that path.
+//! - Otherwise, it returns a `raw.githubusercontent.com` URL to the file.
+//!
+//! This is useful for pointing to files such as screenshots and videos which you might want to exclude from a crate's package, but still include in the doc.
+//!
+//! If the fallback is taken, the generated URL is
+//! `https://raw.githubusercontent.com/{owner}/{name}/{tag_prefix}{crate_version}/{crate_name}/{relative_path}`. This has a few baked-in assumptions:
+//!
+//! - Your Git repository forge is Github.
```
xStrom:
```suggestion
//! - Your Git repository forge is GitHub.
```

include_doc_path/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//! - The file you're linking to is in that folder.
+//! - Your repository gets a Git tag with the crate version number every time the crate releases.
+//!
+//! The `tag_prefix` string defaults to `v` (e.g. `v1.2.3`), but can be changed by setting the `CRATE_TAG_PREFIX` env variable.
+//!
+//! Otherwise, those assumptions are fairly opinionated, as this crate is meant primarily to be used by Linebender projects.
+//!
+//! # Error checking
+//!
+//! The URL fallback is needed when publishing to `docs.rs`, but in cases like CI you might want to emit a compile error when the path isn't found instead.
```
xStrom:
```suggestion
//! The URL fallback is needed when publishing to `docs.rs`, but in cases like CI you might instead want to emit a compile error when the path isn't found.
```

include_doc_path/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//!
+//! To do so, the `CHECK_DOC_PATHS` env variable to `true` (or anything non-empty) when running the macro.
```
xStrom: I think that mentioning the non-empty part removes flexibility for future variants other than `true`, so lets keep it strict in the documentation at least.

```suggestion
//! To do so, set the `CHECK_DOC_PATHS` env variable to `true` when running the macro.
```

include_doc_path/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+/// ```
+#[proc_macro]
+pub fn include_doc_path(input: TokenStream) -> TokenStream {
+    let lit = parse_macro_input!(input as LitStr);
+    let relative_path = lit.value();
+
+    logic::include_doc_path_impl(relative_path, lit.span(), false, false).into()
+}
+
+/// Helper macro which always returns the same string is the path was found.
```
xStrom:
```suggestion
/// Helper macro which always returns the same string as if the path was found.
```

include_doc_path/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+/// Useful for debugging.
+#[proc_macro]
+pub fn helper_include_local_path(input: TokenStream) -> TokenStream {
+    let lit = parse_macro_input!(input as LitStr);
+    let relative_path = lit.value();
+
+    logic::include_doc_path_impl(relative_path, lit.span(), true, false).into()
+}
+
+/// Helper macro which always returns the same string is the path not was found.
```
xStrom:
```suggestion
/// Helper macro which always returns the same string as if the path was not found.
```

include_doc_path/src/logic.rs
```diff
@@ -0,0 +1,106 @@
...
+    relative_path: String,
+    span: Span,
+    force_local: bool,
+    force_url: bool,
+) -> TokenStream {
+    // PARSE ENV VARIABLES
+
+    let Some(manifest_dir) = env("CARGO_MANIFEST_DIR") else {
+        return compile_error("Could not read env variable 'CARGO_MANIFEST_DIR'", span);
+    };
```
xStrom, line 69: I'm not sure that this is the correct choice, to throw a compile error.

These env variables are set by Cargo but not everyone uses Cargo. Some setups have their own build managers that call `rustc` directly.

I think we should graciously fall back to the original string if Cargo wasn't used. Also the CI env could force the requirement of Cargo so that we don't get any fallbacks in CI.

include_doc_path/src/logic.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    // If not and CHECK_DOC_PATHS is set, error out.
+    let check_doc_paths = env("CHECK_DOC_PATHS").is_some();
+    if check_doc_paths {
+        let doc_file_path = doc_file_path.to_string_lossy();
+        let error_message = format!("File '{doc_file_path}' not found");
+        return compile_error(&error_message, span);
+    }
+
+    // Else, return Github URL
```
xStrom:
```suggestion
    // Else, return GitHub URL
```

xStrom (commented):
Looks like a promising approach, but I have some concerns over introducing a hard requirement for Cargo.

include_doc_path/src/logic.rs
```diff
@@ -0,0 +1,106 @@
...
+
+fn github_to_raw(repo: &str) -> Result<String, ()> {
+    const GH_PREFIX: &str = "https://github.com/";
+    if !repo.starts_with(GH_PREFIX) {
+        return Err(());
+    }
+
+    // Handle some common variants such as https://github.com/<owner>/<repo>.git
+    let mut rest = repo[GH_PREFIX.len()..].trim_end_matches('/').to_string();
+    rest = rest.trim_end_matches(".git").to_string();
```
DJMcNab: I think we should be able to do something like:

```suggestion
    let Some(mut rest) = repo.strip_prefix(GH_PREFIX) else {
        return Err(());
    }

    // Handle some common variants such as https://github.com/<owner>/<repo>.git
    rest = rest.trim_end_matches('/');
    rest = rest.trim_end_matches(".git");
```
(Although I haven't checked this with the compiler)

include_doc_path/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+//!
+//! If the fallback is taken, the generated URL is
+//! `https://raw.githubusercontent.com/{owner}/{name}/{tag_prefix}{crate_version}/{crate_name}/{relative_path}`. This has a few baked-in assumptions:
+//!
+//! - Your Git repository forge is Github.
+//! - Your crate is in a `crate_name` folder at the root of the repository.
+//! - The file you're linking to is in that folder.
+//! - Your repository gets a Git tag with the crate version number every time the crate releases.
+//!
+//! The `tag_prefix` string defaults to `v` (e.g. `v1.2.3`), but can be changed by setting the `CRATE_TAG_PREFIX` env variable.
```
DJMcNab: (This doesn't block)

This being configurable through an environment variable doesn't help much, because (as far as I can tell) you can't set environment variables on docs.rs, and otherwise this then becomes a global setting, which isn't what you want.

The way I'd expect to see this configured would be as part of `package.metadata`, however that would require bringing in either a TOML or JSON parser, which stinks. As such, I'd probably rather it just weren't configurable, rather than being "fake" configurable

include_doc_path/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license.workspace = true
+repository.workspace = true
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.103"
+quote = "1.0.42"
+syn = { version = "2.0.111", features = ["parsing"] }
```
DJMcNab, line 18: I'd like for us to not depend on syn here, however that can easily happen as a follow-up (i.e. once we merge this, we should make an issue for it)

include_doc_path/README.md
DJMcNab, line 1: Ideally we'd use cargo-rdme here, but this crate isn't exactly extremely public-facing. (I appreciate that it uses the `linebender_resource_handle` naming convention for non-external-facing crates)

DJMcNab (commented):
This looks good; I agree that we can't avoid proc macros here, although I still wish we could.
For awareness, I haven't tested this locally.

Nothing newly blocking in my review, just some notes on code style, and some follow-ups I'd like to see made into issues.
I'd probably like to treat Kaur's review comments about needing to not break people who don't use Cargo as non-blocking to this PR, because we can always make that into a follow-up. But I'm not going to override his review with an approval. (To be clear, I agree with the sentiment, that we should address it before the next release, but I just don't think it needs to be this PR). The reason I want it to be a follow-up is that this PR is proving to be a merge-conflict magnet, unsurprisingly given how many docs it touches

Should this be marked as fixing #1080?

include_doc_path/src/logic.rs
PoignardAzur: > I think we should graciously fall back to the original string if Cargo wasn't used. 

Mhh... doing so moves the problem from "the user gets a build error" to "the user gets an invalid image for opaque reasons", which I guess it better anyway since more people will build Masonry than build its doc.

PoignardAzur: The thing is, the macro uses the env variables defined by Cargo for both the local path case and the fallback case. If they're not defined, the macro basically can't do anything useful.

I'll just have these values default to `<unknown>` to make it extra clear something has gone wrong in that case.

EDIT: Wait no, that also leads to error. I guess returning the original string is the simplest option.

xStrom: Yeah I don't think it's worth generating anything useful when these env variables aren't set. So `<unknown>` or whatever is fine.

It is ok to depend on the env variables for reasonable results. It's just that, it should still be possible to compile Masonry without them in the edge cases where that is needed.

include_doc_path/src/lib.rs
PoignardAzur: I've removed the option, documented it as a future possibility.

PR #1577, opened 2026-01-16
Update to Vello 0.7.0, ui-events 0.3.0

waywardmonkeys:
An old version of `kurbo` is still brought in by `ui-events`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1181,6 +1181,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "font-types"
+version = "0.11.0"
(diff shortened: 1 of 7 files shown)


PoignardAzur (approved):

waywardmonkeys:
Might need to do a `ui-events` 0.3 but there's something else I wanted in that ... so let me see ...

waywardmonkeys:
I went and updated ui-events and published a new version, so added that here too to fix the Kurbo issues.

I will rely on the previous approval from @PoignardAzur and land this later today if CI is passing.

PoignardAzur (approved):

PR #1578, opened 2026-01-16
Clean up `Split` widget.

xStrom:
There are effectively no functionality changes in this PR.

* Renamed *size* references to *length* when it was a reference to a single axis length.
* Removed some overly mouse-specific language as we support arbitrary pointer devices.
* Removed code duplication where `Axis` matching could be replaced with `Axis` packing.
* Refactored methods that used to take `Size` / `Point` but only cared about the split axis to instead require only the input they actually use.
* Added `scale` usage to a bunch of places in preparation of the shift to do everything inside widgets in device pixels.

I did notice a few bugs but fixing those will be a follow-up in separate PR for easier review of logic changes.

### DIFF
--- a/masonry/src/widgets/split.rs
+++ b/masonry/src/widgets/split.rs
@@ -34,7 +34,7 @@ where
     split_point_chosen: f64,
     split_point_effective: f64,
     min_lengths: (Length, Length), // Integers only
-    bar_length: Length,            // Integers only
+    bar_thickness: Length,         // Integers only
     min_bar_area: Length,          // Integers only
     solid: bool,
     draggable: bool,
@@ -55,7 +55,7 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
             split_point_chosen: 0.5,
             split_point_effective: 0.5,
             min_lengths: (Length::ZERO, Length::ZERO),
-            bar_length: 6.px(),
+            bar_thickness: 6.px(),
             min_bar_area: 6.px(),
             solid: false,
             draggable: true,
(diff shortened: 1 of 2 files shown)


masonry/src/widgets/split.rs
```diff
@@ -33,14 +33,14 @@ where
     split_axis: Axis,
     split_point_chosen: f64,
     split_point_effective: f64,
-    min_size: (Length, Length), // Integers only
-    bar_size: Length,           // Integers only
-    min_bar_area: Length,       // Integers only
+    min_lengths: (Length, Length), // Integers only
+    bar_length: Length,            // Integers only
```
PoignardAzur: Feels like `bar_length` implies that it's storing the length of the bar, instead of its width/breadth.

xStrom: It's following the choice of naming `Length`. I'm going for consistency in term usage.

However, I agree that in natural language, bar length would mean the longer one (whether height or width, whichever is longer) but in our case it is always referring to the shortest length.

How about `thickness`, e.g. `bar_thickness`? It's a bit more verbose but it communicates intent better.

Could also go with `breadth` I guess, although that's a bit jargony.

PoignardAzur: > How about thickness, e.g. bar_thickness? It's a bit more verbose but it communicates intent better.

Oh, I hadn't thought of that!
That's much better, yes.

> It's following the choice of naming Length. I'm going for consistency in term usage.

We can think about renaming Length, but overall I don't think everything that's stored in a Length needs to be called xxx_length. The Rust ecosystem rarely uses the hungarian notation.

xStrom: Updated it to thickness.

As for `Length` naming, I'm not the happiest with it but it's also a very minor issue in terms of what problems Masonry has, so I'd just leave that as-is.

PR #1580, opened 2026-01-16
Fix `Split` bar area.

xStrom:
At some point the intended functionality of the bar area was lost in `Split`  and it started being treated as the minimum bar thickness. This is despite the docs being quite clear:

```rust
/// The minimum splitter bar area defines the minimum thickness of the area
/// where pointer hit detection is done for the splitter bar.
/// The final hit detection area thickness is either this minimum
/// or the splitter bar thickness, whichever is greater.
///
/// This can be useful when you want to use a very narrow visual splitter bar,
/// but don't want to sacrifice user experience by making it hard to click on.
```

This PR restores that functionality. Only hit detection is done using the area. Child space calculation and bar painting are done based on the visual bar thickness.

I also renamed some of the local methods to make it more clear whether they operate on the area or not.

I also added a refinement to `bar_area_edges()` to always keep the defined bar area clickable. So when the bar gets shifted all the way to one side (which you want, to hide one child), you don't want half the accessbility bar area to get lost too. The new code will progressively shift the bar area to the side that has space for it.

### DIFF
--- a/masonry/src/widgets/split.rs
+++ b/masonry/src/widgets/split.rs
@@ -35,8 +35,8 @@ where
     min_bar_area: Length,
     solid: bool,
     draggable: bool,
-    /// Offset from the split point (bar center) to the actual position where the
-    /// bar was clicked. This is used to ensure a click without moving is a no-op,
+    /// Offset from the bar center to the actual position where the bar was clicked.
+    /// This is used to ensure a click without moving is a no-op,
     /// instead of re-centering the bar to the click position.
     click_offset: f64,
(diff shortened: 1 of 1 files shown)


DJMcNab:
Just want to double check that this doesn't interact with https://github.com/linebender/xilem/pull/1580. I don't *think* it does from the description, but they seem very related.

If nothing else, it can't hurt to have a backlink to this PR from it (added by this comment)

DJMcNab:
Eep yes. I meant #1581.

PR #1586, opened 2026-01-20
masonry_winit: add AppDriver::on_wgpu_ready callback

waywardmonkeys:
Expose a new `AppDriver::on_wgpu_ready(window_id, &WgpuContext)` hook that fires when Masonry creates a WGPU device. The provided `WgpuContext` gives access to the shared `wgpu::Instance`, `Adapter`, `Device`, and `Queue` so applications can create GPU resources using the same device as Masonry.

This is intended as the first commit in a series of follow-up PRs adding additional WGPU-related features and improvements.

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -29,7 +29,9 @@ use winit::event::{DeviceEvent as WinitDeviceEvent, DeviceId, WindowEvent as Win
 use winit::event_loop::ActiveEventLoop;
 use winit::window::{Window as WindowHandle, WindowAttributes, WindowId as HandleId};
 
-use crate::app::{AppDriver, DriverCtx, masonry_resize_direction_to_winit, winit_ime_to_masonry};
+use crate::app::{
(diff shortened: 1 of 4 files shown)


xStrom (approved):
Great, thanks!

masonry_winit/src/event_loop_runner.rs
```diff
@@ -534,9 +536,24 @@ impl MasonryState<'_> {
...
+
+            if self.render_cx.devices.len() != devices_before {
+                let device_handle = &self.render_cx.devices[dev_id];
+                let wgpu = WgpuContext {
+                    instance: &self.render_cx.instance,
+                    adapter: &device_handle.adapter,
+                    device: &device_handle.device,
+                    queue: &device_handle.queue,
+                };
+                app_driver.on_wgpu_ready(window.id, &wgpu);
```
DJMcNab, line 553: I'm not sure that passing `window.id` is right here; this is effectively one arbitrary window for which the device is applicable. I think we need to call this closure for all applicable windows. It might be fine to do it unconditionally when we make a new surface, i.e. remove the innermost condition.

The user app can then key off the identity of the device they are passed (although maybe we'd also want to give them the id to use themselves?)

PR #1589, opened 2026-01-20
masonry_winit: add persistent Vello image overrides

waywardmonkeys:
Add `DriverCtx::{set_image_override, clear_image_override}` and `ImageOverride` to associate a `peniko::ImageData` with a GPU-resident `wgpu::Texture` on a per-window basis. Overrides are applied during rendering via `Renderer::override_image` and restored afterward to avoid leaking state between windows/frames.

### DIFF
--- a/masonry_winit/src/app_driver.rs
+++ b/masonry_winit/src/app_driver.rs
@@ -8,6 +8,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 
 use masonry_core::app::RenderRoot;
 use masonry_core::core::{ErasedAction, WidgetId};
+use masonry_core::peniko::ImageData;
 use masonry_core::vello::wgpu;
 use tracing::field::DisplayValue;
 use winit::event_loop::ActiveEventLoop;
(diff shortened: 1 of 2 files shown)


masonry_winit/src/event_loop_runner.rs
```diff
@@ -641,25 +667,41 @@ impl MasonryState<'_> {
...
+                    queue,
+                )
+                .unwrap_or(renderer.profiler);
+                renderer.profiler = new_profiler;
+            }
+            renderer
+        });
+
+        // Apply any persistent image overrides for this window.
+        let mut prev_overrides = Vec::new();
```
xStrom:
```suggestion
        let mut prev_overrides = Vec::with_capacity(window.image_overrides.len());
```

masonry_winit/src/app_driver.rs
```diff
@@ -65,6 +66,22 @@ pub struct WgpuContext<'a> {
...
+/// A persistent image override that maps an [`ImageData`] to a GPU-resident texture.
+///
+/// This is used with [`DriverCtx::set_image_override`] and [`DriverCtx::clear_image_override`].
+///
+/// When set, Vello will copy from `texture` into its internal image atlas whenever the `image` is
+/// drawn in the UI scene.
+///
+/// The texture must be `Rgba8Unorm` and include `COPY_SRC` usage.
+#[derive(Debug)]
+pub struct ImageOverride {
```
xStrom: Looks like `ImageOverride` can only be used by internal methods, so perhaps it also makes sense to make this `pub(crate)` and move its useful warnings about texture requirements to the `set_image_override` method.

masonry_winit/src/app_driver.rs
```diff
@@ -145,6 +162,26 @@ impl DriverCtx<'_, '_> {
         self.state.close_window(window_id);
     }
 
+    /// Set a persistent Vello image override for a given window.
+    ///
+    /// This associates the given [`ImageData`] with the provided GPU texture.
+    ///
+    /// Overrides persist until cleared with [`DriverCtx::clear_image_override`].
```
xStrom, line 156:
```suggestion
    /// This associates the given [`ImageData`] with the provided GPU texture.
    ///
    /// Correct behaviour is not guaranteed if the texture does not have the same
    /// dimensions as the image.
    ///
    /// Overrides persist until cleared with [`DriverCtx::clear_image_override`].
```

masonry_winit/src/event_loop_runner.rs
```diff
@@ -669,6 +711,10 @@ impl MasonryState<'_> {
             )
             .expect("failed to render to surface");
 
+        for (image, prev) in prev_overrides {
+            renderer.override_image(&image, prev);
+        }
```
xStrom: This back-and-forth dance seems weird to me but I don't understand the internals enough to really know.

Couldn't we be more granular with only doing the override once, storing the prev, and then restoring it only when it has been cleared via `clear_image_override`?

DJMcNab (commented):
So to clarify, the reason that this is per-window is because when you have multiple windows, each window can have a different renderer, and so you need to associate them properly? Even still, I don't think it's reasonable to expect a given `ImageData` to have more than one texture for a given device, so constantly setting and resetting it seems very odd to me. As such, I'd be inclined to just make this a simple passthrough to the `Renderer` method.

We currently never change the device associated with a window, right? If we did, the app would need to know about that to update their overrides. I think we will need to support that at some point, but not in this PR if it doesn't already exist

masonry_winit/src/event_loop_runner.rs
waywardmonkeys: Gone.

masonry_winit/src/app_driver.rs
waywardmonkeys: Gone.

masonry_winit/src/app_driver.rs
waywardmonkeys: Not sure an edit like this is needed now...

masonry_winit/src/event_loop_runner.rs
waywardmonkeys: It is gone now.

waywardmonkeys:
@DJMcNab Based on that, I've simplified this a lot ... hopefully you and @xStrom like it better now!

masonry_winit/src/app_driver.rs
DJMcNab: I believe that the dimension requirement still applies, and doesn't seem to be captured in the docs.

xStrom: As the requirement still applies, let's add the comment.

Issue #1592, opened 2026-01-21
`Background` tracking issue


xStrom:
There are at least two ways that `Background` painting should improve.
 
* We should skip painting fully transparent backgrounds. It's useless work. However, for some reason we can't skip them yet. There's some issue with blend modes or something, I couldn't immediately figure it out. Hence, future work.

* By the time we introduce new border styles like dashed we'll need to expand the background painting rect to the outer edge of the border. Didn't do that immediately because it will be visible beyond the border. Future work is to figure out how to make it work. Perhaps by drawing a thicker border but clipping it at the intended size.

PR #1593, opened 2026-01-21
Implement `Background` in Masonry Core.

xStrom:
`Background`, `ActiveBackground`, and `DisabledBackground` will now be painted for all widgets.

There is a new `Widget::pre_paint` method that has a default implementation with reusable helper methods. This gives widgets an escape hatch if they want to handle their core property painting differently.

This does have a bloating effect on the per-widget scene cache, which now holds three `Scene` structs. Future optimizations could include a pool of empty scenes from where we take one and immediately give it back to the pool if the widget didn't draw anything. Then the per-widget cache could contain `None`.

Active and disabled state changes also request pre-paint for the related widgets, making the default `ActiveBackground` / `DisabledBackground` handling truly automatic.

Existing background painting behavior is replicated here. However, there are two improvements that should happen at a later date, for which I created a tracking issue at #1592.

Additionally, I discovered a bug in our previous scene caching code that I also fixed. Previously we had effectively this:

```rust
if state.request_paint || state.request_post_paint {
    // Always reset both scene types
    scene.reset();
    postfix_scene.reset();

    // But only redraw them piecemeal,
    // leaving one of them potentially incorrectly empty.
    if state.request_paint { draw_scene() }
    if state.request_post_paint { draw_postfix_scene() }
}

```

This PR is a follow-up to #1588.

### DIFF
--- a/masonry/src/widgets/switch.rs
+++ b/masonry/src/widgets/switch.rs
@@ -217,6 +217,15 @@ impl Widget for Switch {
 
     fn layout(&mut self, _ctx: &mut LayoutCtx<'_>, _props: &PropertiesRef<'_>, _size: Size) {}
 
+    fn pre_paint(
+        &mut self,
+        _ctx: &mut PaintCtx<'_>,
+        _props: &PropertiesRef<'_>,
+        _scene: &mut Scene,
+    ) {
+        // TODO: Make Switch painting work with generic shadow/background/border
+    }
+
     fn paint(&mut self, ctx: &mut PaintCtx<'_>, props: &PropertiesRef<'_>, scene: &mut Scene) {
         // TODO: Remove HACK: Until scale factor rework happens, just pretend it's always 1.0.
         //       https://github.com/linebender/xilem/issues/1264


masonry/src/widgets/slider.rs
```diff
@@ -314,11 +313,13 @@ impl Widget for Slider {
...
 
     fn property_changed(&mut self, ctx: &mut UpdateCtx<'_>, property_type: TypeId) {
-        Background::prop_changed(ctx, property_type);
         BarColor::prop_changed(ctx, property_type);
         TrackThickness::prop_changed(ctx, property_type);
         ThumbColor::prop_changed(ctx, property_type);
         ThumbRadius::prop_changed(ctx, property_type);
+        if property_type == TypeId::of::<Background>() {
+            ctx.request_paint_only();
+        }
```
PoignardAzur, line 320: It seems a little off to remove `Background::prop_changed()` only to add code which does the same thing.

I get that this is because the slider widget does a special thing with a gradient instead of using the usual background color, but... this seems maybe worth a rethink?

For instance, I'll note that with the current design, if you apply an `alpha=0.5` color as the background, that color will be applied twice (and thus end up as a de factor `alpha=0.5` color).

I'm not sure there's an immediate way to solve this, but I'd like at least some comments noting the issue, beyond the "TrackColor property" todo.

xStrom: It's not exactly the same anymore though, because core requests pre-paint but this requests paint. These are different paint calls now. This was a key motivation for adding `Property::matches` and moving away from `Property::prop_changed`. Because if we want to allow properties that are reusable, then different widgets will use them in ways that require different passes.

PoignardAzur: Alright. I guess it will make sense once more properties are transitioned to this new model.

PR #1595, opened 2026-01-21
Add `Property::matches` method.

xStrom:
This is a slightly different flavor of the existing `Property::prop_changed` method. The difference being that `matches` doesn't decide what flag must be set. This is important for more generic properties that can have varied uses.

For example, one widget might use a thickness property in a way that affects layout, but for another widget it will only affect paint. This will become even more relevant with #1593 introducing `Widget::pre_paint` as then we will have three different paint methods. Widgets can disagree which of the three need to re-run if `ContentColor` changes.

I am guessing that it will make sense to standardize on `matches` instead of `prop_changed`, however this PR does not do that. For now I just want to introduce it so we can get some real use of it in and see how it goes.

---

I'm aware that it's a slim method that barely does anything, however I think it's still a sufficient ergonomics win.

`property_type == TypeId::of::<ContentColor>()` has less tab completion in it and is awkward to remember.
`ContentColor::matches(property_type)` is easier to remember and can be fully auto-completed with the input parameter not even needing a tab.

Whether it should be part of the `Property` trait is a question. The benefit is that it'll definitely be implemented for every property. The downside is that widgets will need to import the `Property` trait for the method to be available. I'm leaning towards the trait, because the import needs to happen only once per widget so isn't that bad. When a property is missing this method it will be a bad experience for someone who is used to it and has never directly written the `TypeId::of` turbofish.

### DIFF
--- a/masonry/examples/calc_masonry.rs
+++ b/masonry/examples/calc_masonry.rs
@@ -11,7 +11,6 @@
     reason = "Don't matter for example code"
 )]
 
-use std::any::TypeId;
 use std::str::FromStr;
 
 use masonry::core::{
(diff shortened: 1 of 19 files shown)


waywardmonkeys (commented):
It looks like every implementation is the same yet we don't provide a default impl?

PR #1600, opened 2026-01-22
masonry_winit: allow configuring WGPU device request

waywardmonkeys:
- Add `WgpuLimits` and expose it via `masonry_winit::app`
- Add `MasonryState::{set_wgpu_device_options, add_wgpu_features, set_wgpu_limits}` to configure requested `wgpu::Features` and limits before device creation
- Thread requested features/limits through `RenderContext::new_device` (best-effort features via intersection with adapter features)

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -30,7 +30,8 @@ use winit::event_loop::ActiveEventLoop;
 use winit::window::{Window as WindowHandle, WindowAttributes, WindowId as HandleId};
 
 use crate::app::{
-    AppDriver, DriverCtx, WgpuContext, masonry_resize_direction_to_winit, winit_ime_to_masonry,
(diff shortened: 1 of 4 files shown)


masonry_winit/src/event_loop_runner.rs
```diff
@@ -392,6 +393,37 @@ impl MasonryState<'_> {
         }
     }
 
+    /// Configure how Masonry requests the WGPU device (features and limits).
+    ///
+    /// This must be called before the first surface/device is created (i.e. before the first
+    /// redraw). A good place is [`AppDriver::on_start`].
```
xStrom, line 409: These three new public methods should document what the returned bool actually means.

```suggestion
    /// This must be called before the first surface/device is created (i.e. before the first
    /// redraw). A good place is [`AppDriver::on_start`]. This method will return `false`
    /// if it was called too late and a device had already been created.
```

PR #1602, opened 2026-01-23
masonry: improve Split interaction + add playground example

waywardmonkeys:
- Add absolute split-point modes (from start/end) alongside fractional mode.
- Make the divider focusable and operable via keyboard and accessibility actions.
- Improve divider visuals (hover/active/focus feedback + focus ring) and fix bar hit-area padding.
- Add interaction tests for drag and keyboard movement.
- Add `split_playground` example to demo and exercise the new behavior.

### DIFF
--- a/masonry/src/widgets/split.rs
+++ b/masonry/src/widgets/split.rs
@@ -208,7 +208,7 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
     fn bar_area_edges(&self, length: f64, scale: f64) -> (f64, f64) {
         let (edge1, edge2) = self.bar_edges(length, scale);
         let (space1, space2) = (edge1.max(0.), (length - edge2).max(0.));
-        let padding = (self.bar_area(scale) - self.bar_thickness.dp(scale)).max(0.);
+        let padding = self.bar_area(scale) - self.bar_thickness.dp(scale);
 
         // Half the padding to the first edge
         let pad1 = (0.5 * padding).min(space1);


masonry/src/widgets/split.rs
```diff
@@ -193,32 +235,87 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
         (min_limit, max_limit)
     }
 
+    fn chosen_child1_len(&self, split_space: f64, scale: f64) -> f64 {
```
jaredoconnell: I'm not a fan of the name of this function.

waywardmonkeys: Inlined the function into the single caller.

masonry/src/widgets/split.rs
```diff
@@ -3,23 +3,41 @@
...
+pub enum SplitPoint {
+    /// Split by a fraction of the available space.
+    ///
+    /// `0.0` means the first child gets no space, `1.0` means the second child gets no space.
+    /// Values outside `0.0..=1.0` are clamped when set.
+    Fraction(f64),
+    /// Split by an absolute distance from the start (in logical pixels).
+    FromStart(Length),
+    /// Split by an absolute distance from the end (in logical pixels).
+    FromEnd(Length),
```
xStrom: I think we can omit the logical pixel mention. It's true but the idea is that this requirement is made clear by the `Length` encapsulation already.

```suggestion
    /// Split by an absolute distance from the start.
    FromStart(Length),
    /// Split by an absolute distance from the end.
    FromEnd(Length),
```

xilem_masonry/src/view/split.rs
```diff
@@ -96,14 +96,31 @@ impl<ChildA, ChildB, State, Action> Split<ChildA, ChildB, State, Action> {
...
+    ///
+    /// This is the size of the first child along the split axis, in logical pixels.
+    pub fn split_point_from_start(mut self, split_point: Length) -> Self {
+        self.split_point = widgets::SplitPoint::FromStart(split_point);
+        self
+    }
+
+    /// Set the split point as an absolute distance from the end.
+    ///
+    /// This is the size of the second child along the split axis, in logical pixels.
```
xStrom: Same story here.

```suggestion
    /// This is the size of the first child along the split axis.
    pub fn split_point_from_start(mut self, split_point: Length) -> Self {
        self.split_point = widgets::SplitPoint::FromStart(split_point);
        self
    }

    /// Set the split point as an absolute distance from the end.
    ///
    /// This is the size of the second child along the split axis.
```

masonry/src/widgets/split.rs
```diff
@@ -80,14 +98,40 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
...
+    /// This can be useful when one side should have a stable pixel size, even when the split
+    /// container is resized.
+    pub fn split_point_from_start(mut self, split_point: Length) -> Self {
+        self.split_point_chosen = SplitPoint::FromStart(split_point);
+        self
+    }
+
+    /// Builder-style method to set the split point as an absolute distance from the end.
+    ///
+    /// This is the size of the second child along the split axis, in logical pixels.
```
xStrom: Ditto.

```suggestion
    /// This is the size of the first child along the split axis.
    /// This can be useful when one side should have a stable pixel size, even when the split
    /// container is resized.
    pub fn split_point_from_start(mut self, split_point: Length) -> Self {
        self.split_point_chosen = SplitPoint::FromStart(split_point);
        self
    }

    /// Builder-style method to set the split point as an absolute distance from the end.
    ///
    /// This is the size of the second child along the split axis.
```

masonry/src/widgets/split.rs
```diff
@@ -203,27 +245,78 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
...
+                } else {
+                    child1_len / length
+                });
+            }
+            SplitPoint::FromStart(_) => {
+                let logical = if scale <= f64::EPSILON {
+                    child1_len
+                } else {
+                    child1_len / scale
+                };
```
xStrom: Scale will come from Masonry, so we can safely assume it won't be epsilon or below. Even if it is, things will probably blow up before `Split` gets a word in.

```suggestion
                let logical = child1_len / scale;
```

masonry/src/widgets/split.rs
```diff
@@ -203,27 +245,78 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
...
+                self.split_point_chosen = SplitPoint::FromStart(Length::px(logical));
+            }
+            SplitPoint::FromEnd(_) => {
+                let child2_len = (length - child1_len).max(0.0);
+                let logical = if scale <= f64::EPSILON {
+                    child2_len
+                } else {
+                    child2_len / scale
+                };
```
xStrom:
```suggestion
                let logical = child2_len / scale;
```

masonry/src/widgets/split.rs
```diff
@@ -400,37 +499,145 @@ where
                             .get_coord(self.split_axis);
                         let length = ctx.size().get_coord(self.split_axis);
                         // If widget has pointer capture, assume always it's hovered
-                        let effective_pos = pos - self.click_offset;
-                        self.update_split_point(length, effective_pos, scale);
+                        let effective_center = pos - self.click_offset;
+                        self.update_split_point_from_bar_center(length, effective_center, scale);
                         ctx.request_layout();
+                        ctx.request_render();
```
xStrom: Layout implicitly requests render already and we have a habbit of not explicitly requesting every pass that gets implicitly called. Keeps code less noisy.

```suggestion
```

masonry/src/widgets/split.rs
```diff
@@ -400,37 +499,145 @@ where
...
+            }
+            Key::Named(NamedKey::End) => {
+                child1_len = self.split_side_limits(split_space, scale).1;
+            }
+            _ => return,
+        }
+
+        self.set_chosen_from_child1_len(split_space, child1_len, scale);
+        ctx.request_layout();
+        ctx.request_render();
```
xStrom: Ditto.

```suggestion
```

masonry/src/widgets/split.rs
```diff
@@ -400,37 +499,145 @@ where
...
+                    }
+                }
+                _ => return,
+            },
+            _ => return,
+        }
+
+        self.set_chosen_from_child1_len(split_space, child1_len, scale);
+        ctx.request_layout();
+        ctx.request_render();
```
xStrom: Here too.

```suggestion
```

masonry/src/widgets/split.rs
```diff
@@ -531,10 +738,25 @@ where
...
+
+        if ctx.is_focus_target() && self.draggable && !ctx.is_disabled() {
+            let size = ctx.size();
+            let length = size.get_coord(self.split_axis);
+            let cross_length = size.get_coord(self.split_axis.cross());
+            let (edge1, edge2) = self.bar_area_edges(length, scale);
+
+            let p1 = self.split_axis.pack_point(edge1, 0.);
+            let p2 = self.split_axis.pack_point(edge2, cross_length);
+            let rect = Rect::from_points(p1, p2).inset(1.0);
```
xStrom: The focus ring should be around the visual element, not around the hit detection area. We can increase the insets to 2px to make it a bit more appealing than a 1px hug.

```suggestion
            let (edge1, edge2) = self.bar_edges(length, scale);

            let p1 = self.split_axis.pack_point(edge1, 0.);
            let p2 = self.split_axis.pack_point(edge2, cross_length);
            let rect = Rect::from_points(p1, p2).inset(2.0);
```

masonry/src/widgets/split.rs
```diff
@@ -566,10 +788,35 @@ where
...
     ) {
+        // TODO: Remove HACK: Until scale factor rework happens, just pretend it's always 1.0.
+        //       https://github.com/linebender/xilem/issues/1264
+        let scale = 1.0;
+
+        let total_length = ctx.size().get_coord(self.split_axis);
+        let bar_thickness = self.bar_thickness.dp(scale);
+        let split_space = (total_length - bar_thickness).max(0.0);
+        let (min_limit, max_limit) = self.split_side_limits(split_space, scale);
+        let child1_len = (split_space * self.split_point_effective).clamp(min_limit, max_limit);
```
xStrom: The effective split point calculation already took min/max limits into account.

```suggestion
        let child1_len = split_space * self.split_point_effective;
```

masonry/src/widgets/split.rs
```diff
@@ -566,10 +788,35 @@ where
...
+
+        node.set_orientation(match self.split_axis {
+            Axis::Horizontal => accesskit::Orientation::Horizontal,
+            Axis::Vertical => accesskit::Orientation::Vertical,
+        });
+        node.set_value(child1_len.to_string());
+        node.set_numeric_value(child1_len);
```
xStrom, line 799: I wish AccessKit was actually documented to know whether setting both the string value and numeric value is correct here. As it stands, I have no idea.

@mwcampbell Should numeric values also get set as string values?

masonry/src/widgets/split.rs
```diff
@@ -166,7 +208,7 @@ impl<ChildA: Widget + ?Sized, ChildB: Widget + ?Sized> Split<ChildA, ChildB> {
     fn bar_area_edges(&self, length: f64, scale: f64) -> (f64, f64) {
         let (edge1, edge2) = self.bar_edges(length, scale);
         let (space1, space2) = (edge1.max(0.), (length - edge2).max(0.));
-        let padding = (self.min_bar_area.dp(scale) - self.bar_thickness.dp(scale)).max(0.);
+        let padding = (self.bar_area(scale) - self.bar_thickness.dp(scale)).max(0.);
```
xStrom: This "fix" adds more work. If you want to do it this way, then let's at least remove the `.max(0.)`.

```suggestion
        let padding = self.bar_area(scale) - self.bar_thickness.dp(scale);
```

Old logic:

`min_bar_area` - `bar_thickness` => correct positive padding if `min_bar_area` > `bar_thickneess`, while negative result will get clamped with `.max(0.)`

New logic:

`bar_thickness.max(min_bar_area)` - `bar_thickness` => correct positive padding, negative results impossible but gets clamped with `.max(0.)` anyway

PR #1603, opened 2026-01-23
masonry: add Clip widget

waywardmonkeys:
Add `widgets::Clip`, a single-child container that clips its child to the widget’s layout rect.


waywardmonkeys:
Better to do clips through other means rather than a wrapper widget.

PR #1604, opened 2026-01-23
masonry_winit: Remove window ID from on_wgpu_ready

waywardmonkeys:
Per review comment from Daniel, this isn't needed as we move forward, and if it is, we'll have more work to do fix things. In the meantime, having this here results in incorrect assumptions.

### DIFF
--- a/masonry_winit/src/app_driver.rs
+++ b/masonry_winit/src/app_driver.rs
@@ -100,11 +100,8 @@ pub trait AppDriver {
         ctx.exit();
     }
 
-    /// Called when Masonry has created a WGPU device.
-    ///
-    /// This may be called more than once if multiple devices are created (e.g. multiple windows
-    /// requiring different compatible adapters).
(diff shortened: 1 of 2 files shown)


masonry_winit/src/app_driver.rs
```diff
@@ -101,10 +101,7 @@ pub trait AppDriver {
     }
 
     /// Called when Masonry has created a WGPU device.
```
DJMcNab:
```suggestion
    /// Called when Masonry has created the WGPU device.
```
or 
```suggestion
    /// Called when Masonry has created its WGPU device.
```

(As far as I can tell, the proper thing to call `wgpu` is either "`wgpu`" or possibly "Wgpu", but not "WGPU"). It might be worth asking Connor if there's a preferred name stylisation.

waywardmonkeys: Will revisit "wgpu" vs "WGPU" later as there are other occurrences of it.

PR #1609, opened 2026-01-23
masonry: focus Button on pointer down

waywardmonkeys:
Request focus on `PointerEvent::Down` so clicking a button also moves text focus. Add a non-rendering regression test for click-to-focus.

### DIFF
--- a/masonry/src/widgets/button.rs
+++ b/masonry/src/widgets/button.rs
@@ -125,6 +125,7 @@ impl Widget for Button {
     ) {
         match event {
             PointerEvent::Down(..) => {
+                ctx.request_focus();
                 ctx.capture_pointer();
                 // Changes in pointer capture impact appearance, but not accessibility node
                 ctx.request_paint_only();
@@ -389,6 +390,19 @@ mod tests {
             harness.pop_action(),
             Some((ButtonPress { button: None }, button_id))
         );
+    }
+
+    #[test]
+    fn mouse_down_requests_focus() {
+        let widget = NewWidget::new(Button::with_text("Hello"));
+        let mut harness = TestHarness::create(test_property_set(), widget);
+        let button_id = harness.root_id();
+
+        harness.focus_on(None);
+        harness.mouse_move_to(button_id);
+        harness.mouse_button_press(PointerButton::Primary);
+
+        assert_eq!(harness.focused_widget_id(), Some(button_id));
     }
 
     #[test]


masonry/src/tests/event.rs
```diff
@@ -243,11 +243,11 @@ fn click_anchors_focus() {
...
     let other_id = harness.get_widget(other).id();
 
-    // Clicking a button doesn't focus it.
+    // Clicking a button focuses it.
     harness.mouse_click_on(child_3_id);
-    assert_eq!(harness.focused_widget_id(), None);
+    assert_eq!(harness.focused_widget_id(), Some(child_3_id));
 
-    // But the next tab event focuses its neighbor.
+    // The next tab event focuses its neighbor.
```
PoignardAzur: I think the change of behavior is an improvement, but also that test was checking something specific that the new test doesn't check: specifically, that if you click something, even if it isn't focused, the click moves a "focus anchor" so that the next tab event will focus something "close" to what you clicked.

PoignardAzur: I think the proper change would be to replace the button with a label, in which case I think the rest of the test stays the same.

xStrom: Makes sense to keep the anchor test functional.

waywardmonkeys: I see. I will update this tomorrow.

waywardmonkeys: Just changing this to a `Label` isn't enough ... because the you get a panic that the widget doesn't accept pointer events.

waywardmonkeys: Using a Prose results in a different assert ... but this whole way of doing the focus anchor doesn't really make sense to me.

PoignardAzur: > Just changing this to a Label isn't enough ... because the you get a panic that the widget doesn't accept pointer events.

Oh, right, *that's* why I wrote the test this way.

> Using a Prose results in a different assert

Which one?

> but this whole way of doing the focus anchor doesn't really make sense to me.

Worst case, you can add an `#[ignore]` annotation to the test to unblock this PR.

waywardmonkeys: The Prose assert was that the widget isn’t visible, I think. Not at computer.

waywardmonkeys: I can probably do something like using a `ModularWidget`. (That was what Codex wanted to do but I thought might be overkill.)

PoignardAzur: Quick note: I figured where the "Widget #XX is not visible" error is was coming from.

See [#masonry > &#96;Harness::mouse_click_on&#96; on a &#96;Prose&#96; panics](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/.60Harness.3A.3Amouse_click_on.60.20on.20a.20.60Prose.60.20panics/with/570396588).

For now I recommend adding `#[ignore]` to the test.

PR #1615, opened 2026-01-27
Fix scrolling a nested child into view.

xStrom:
The scroll-into-view mechanics were quite busted.

Previously the scroll request code used the target's rect in the target's parent's coordinate space and then sent that as an update payload to the target itself (useless rect), the parent (useful rect), and all the other ancestors (useless rect).

This PR changes it so the initial request rect is in the target's local coordinate space, as the target itself will get this update too. Then as we move up the tree to each ancestor we will translate the rect into the respective ancestor's coordinate space.

---

I modified the button list test in a way where I replaced a bunch of the spacers and gaps with padding inside of a `ModularWidget`. This ensures that the button that we're requesting to scroll to won't be at `(0,0)` of its parent and thus actually tests scrolling to a nested child.

---

This PR does not add any compose passes or scroll translations, which means that the nested `Portal` scenario is still broken.

### DIFF
--- a/masonry/src/widgets/portal.rs
+++ b/masonry/src/widgets/portal.rs
@@ -851,14 +851,12 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
 // --- MARK: TESTS
 #[cfg(test)]
 mod tests {
-    use masonry_testing::ModularWidget;
-
     use super::*;
     use crate::core::keyboard::{Key, NamedKey};
     use crate::core::{WidgetOptions, WidgetTag};
     use crate::layout::AsUnit;
     use crate::properties::Dimensions;
-    use crate::testing::{TestHarness, assert_render_snapshot};
+    use crate::testing::{ModularWidget, TestHarness, assert_render_snapshot};
     use crate::theme::test_property_set;
     use crate::widgets::{Button, Flex, SizedBox};
 
--- a/masonry_core/src/core/events.rs
+++ b/masonry_core/src/core/events.rs
@@ -98,7 +98,7 @@ pub enum Update {
     /// [`set_stashed`]: crate::core::EventCtx::set_stashed
     StashedChanged(bool),
 
-    /// Called when a descendant widgets requests to be scrolled to.
+    /// Called when a descendant widget requests to be scrolled to.
     ///
     /// The included [`Rect`] is in the receiving widget's border-box coordinate space.
     ///


masonry_core/src/core/events.rs
```diff
@@ -98,8 +98,14 @@ pub enum Update {
     /// [`set_stashed`]: crate::core::EventCtx::set_stashed
     StashedChanged(bool),
 
-    /// Called when a child widgets uses
-    /// [`EventCtx::request_scroll_to_this`](crate::core::EventCtx::request_scroll_to_this).
+    /// Called when a descendant widgets requests to be scrolled to.
```
PoignardAzur:
```suggestion
    /// Called when a descendant widget requests to be scrolled to.
```

masonry_core/src/passes/update.rs
```diff
@@ -725,18 +725,23 @@ pub(crate) fn run_update_scroll_pass(root: &mut RenderRoot) {
...
             // translations are updated and the rect passed to parents
-            // is more accurate.
+            // is more accurate. Until then we don't add scroll_translation
+            // at all and only support a single scrolling parent.
 
+            // Before continuing to the parent, we need to convert the target_rect from this
+            // widget's border-box coordinate space to the parent's border-box coordinate space.
             let state = &ctx.widget_state;
-            target_rect = target_rect + state.scroll_translation + state.origin.to_vec2();
+            target_rect = target_rect + state.origin.to_vec2();
```
PoignardAzur, line 744: I'm a little confused. Is this the change the PR describes?

It seems to me that, if you want to go from child coordinate space to parent coordinate space, you'd also want to account for at least `WidgetState::transform`.

masonry/src/widgets/portal.rs
```diff
@@ -851,18 +851,49 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
 // --- MARK: TESTS
 #[cfg(test)]
 mod tests {
+    use masonry_testing::ModularWidget;
+
```
PoignardAzur: Should be

```rust
    use crate::testing::ModularWidget;
```

masonry_core/src/passes/update.rs
xStrom: One piece of info that could be insightful is that `target_rect` was previously unused, i.e. it was write-only but never used anywhere. I sort of repurposed that dead code, with this modification.

As for accounting for transform, well maybe, but it gets tricky. Currently it doesn't even account for scroll translation in nested portals. I wanted to get a simple nested widget working for now, other stuff can be follow-up as this stuff  needs non-trivial attention.

PoignardAzur: > One piece of info that could be insightful is that target_rect was previously unused, i.e. it was write-only 

Oh, you mean

```rust
            let event = Update::RequestPanToChild(target_rect);
```

Alright, makes sense now.

PR #1617, opened 2026-01-27
Fix baseline pixel snapping.

xStrom:
Baseline is stored as distance from the bottom of the widget. The old pixel snapping code treated it like distance from the top of the widget, which is wrong.

### DIFF
--- a/masonry/src/tests/layout.rs
+++ b/masonry/src/tests/layout.rs
@@ -208,7 +208,7 @@ fn pixel_snapping() {
         let child_size = ctx.compute_size(child, SizeDef::fit(size), size.into());
         ctx.run_layout(child, child_size);
         ctx.place_child(child, pos);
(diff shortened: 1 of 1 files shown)


masonry/src/tests/layout.rs
```diff
@@ -208,16 +208,20 @@ fn pixel_snapping() {
         let child_size = ctx.compute_size(child, SizeDef::fit(size), size.into());
         ctx.run_layout(child, child_size);
         ctx.place_child(child, pos);
+        ctx.set_baseline_offset(5.);
     });
```
PoignardAzur: Seems like you'd want to set it to a non-integer number to test the snapping behavior?

xStrom: Makes sense, I adjusted it to `5.1`.

PR #1619, opened 2026-01-27
Remove `WidgetState::layout_rect`.

xStrom:
This "layout rect" was the pixel aligned version of the layout rect in the parent widget's coordinate space. We don't currently need it and there are wishful comments about its removal.

It may make sense to add it back at a later date, we'll see, but for now we can get rid of it.

The only use of this was in the simple layout test which doesn't really need it as it used a `(0,0)` origin in the parent's coordinate space anyway.

### DIFF
--- a/masonry/src/tests/layout.rs
+++ b/masonry/src/tests/layout.rs
@@ -39,13 +39,11 @@ fn layout_simple() {
 
     let harness = TestHarness::create(test_property_set(), widget);
 
-    let first_box_rect = harness.get_widget(tag_1).ctx().size().to_rect();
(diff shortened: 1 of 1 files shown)


masonry/src/tests/layout.rs
PoignardAzur: I'd suggest something like:

```rust
let first_box_size = harness.get_widget(tag_1).ctx().size();
let first_box_paint_rect = harness.get_widget(tag_1).ctx().paint_rect();

assert_eq!(first_box_size.width, BOX_WIDTH);
assert_eq!(first_box_size.height, BOX_WIDTH);
```

(Github suggestion UI messed it up)

PR #1623, opened 2026-01-27
Describe the box model in Masonry Concepts.

xStrom:
After having spent a good amount of time in the Masonry internals figuring out how all the various bits work, I figured I'd make it easier for the next person (*and future me, who may forget details*) by documenting more of what is going on.

In this PR I focus on defining terms in the concepts document, with the primary focus being on two key groupings:

1. The names for different boxes: *content, border, paint,* and *bounding*.
2. The names for box lifecycle: *preferred, chosen, layout, aligned,* and *effective*.

The main goal is to be able to precicely document what is what in the code. So that if someone learns what these terms mean, they can understand the exact nature of a variable/method without having to start investigating the internals of how it is derived and updated. It will greatly help with choosing the correct method and it will also help avoid mixing up variables of different lifecycles.

---

In follow-up PRs I will be improving the actual field/method documentation while using these terms.

### DIFF
--- a/masonry_core/src/doc/masonry_concepts.md
+++ b/masonry_core/src/doc/masonry_concepts.md
@@ -156,23 +156,45 @@ All widgets have associated data of arbitrary types called "properties".
 These properties are mostly used for styling and event handling.
 
 
-## Bounding rect
+## Box model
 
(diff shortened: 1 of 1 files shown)


jaredoconnell (approved):
Looks good. I see that a TODO item is left in there. That may be nice. If you do that, feel free to request another review.

PR #1627, opened 2026-01-28
Replace set_clip_path with set_clips_contents method

PoignardAzur:
This changes clipping behavior to a boolean flag, which better matches how it was used so far.

A follow-up PR will add set_clip_shape to set arbitrary (even non-Rect) shapes.

### DIFF
--- a/masonry_core/src/core/widget_state.rs
+++ b/masonry_core/src/core/widget_state.rs
@@ -436,15 +436,12 @@ impl WidgetState {
     /// A widget's clip shape will have two effects:
     /// - It serves as a mask for painting operations of this widget and its children.
     ///   Note that while all painting done by children will be clipped by this path,
-    ///   only the painting done in [`paint`] by this widget itself will be clipped.
-    ///   The remaining painting done in [`pre_paint`] and [`post_paint`] will not be clipped.
+    ///   only the painting done in `Widget::paint` by this widget itself will be clipped.
+    ///   The remaining painting done in `Widget::pre_paint` and `Widget::post_paint` will not be clipped.
     /// - Pointer events must be inside that shape to reach the widget's children.
     ///
     /// This currently returns the border-box rect if `clips_contents` is true and `None` otherwise, but in the future we may want to support more complex clip shapes, in which case this method would need to be updated.
     ///
-    /// [`paint`]: Widget::paint
-    /// [`pre_paint`]: Widget::pre_paint
-    /// [`post_paint`]: Widget::post_paint
     /// [clip shape]: crate::doc::masonry_concepts#clip-shape.
     pub(crate) fn clip_shape(&self) -> Rect {
         self.border_box_size().to_rect() - self.border_box_translation()


masonry_core/src/core/contexts.rs
```diff
@@ -1054,12 +1038,12 @@ impl_context_method!(
             self.widget_state.bounding_rect()
         }
 
-        /// The clip path of the widget, if any was set.
+        /// Whether the children clips its own contents and that of its children.
```
waywardmonkeys: I think this should be something like:

```
/// Whether the widget clips its own contents and that of its children.
```

masonry_core/src/core/widget_state.rs
```diff
@@ -109,10 +109,14 @@ pub(crate) struct WidgetState {
...
-    // Currently Kurbo doesn't really provide a type that lets us
-    // efficiently hold an arbitrary shape.
-    pub(crate) clip_path: Option<Rect>,
+    // TODO - Add clip_shape
+    // TODO - Use more efficient type and avoid allocating by default.
+    // /// Shape that explicitly restricts pointer events.
+    // pub(crate) clip_shape: Option<BezPath>,
+    //
+    /// Whether the widget and its children's scenes are clipped by the
+    /// [clip shape](crate::doc::masonry_concepts#clip-shape).
```
waywardmonkeys, line 131: Should this say "clip shape" yet?

masonry_core/src/doc/masonry_concepts.md
```diff
@@ -175,6 +175,20 @@ Previous versions of Masonry had a concept of a widget's "layout rect", composed
...
 
+## Clip shape
+
+Widgets have a shape, usually one that matches their visual appearance, which has two purposes:
+
+- Pointer events outside of that shape will not affect the pointer.
+- If the widget is set to clip its contents, pointer events outside the clip shape won't affect the children either.
+- If the widget is set to clip its contents, its scene and the children's scenes will be painted inside of the clip shape.
+
+Currently, the clip shape is hardcoded to be the layout rect.
```
waywardmonkeys: Maybe better to say "the widgets local rect, transformed for rendering" or something since we just say that a layout rect doesn't make sense above.

masonry_core/src/core/contexts.rs
```diff
@@ -892,36 +892,20 @@ impl LayoutCtx<'_> {
...
     ///
-    /// See [`LayoutCtx::set_clip_path`] for details.
-    pub fn clear_clip_path(&mut self) {
-        trace!("clear_clip_path");
-        self.widget_state.clip_path = None;
-        // TODO - Updating the clip path may have
-        // other knock-on effects we'd need to document.
+    /// [clip shape]: crate::doc::masonry_concepts#clip-shape.
+    pub fn set_clips_contents(&mut self, clips: bool) {
+        self.widget_state.clips_contents = clips;
```
waywardmonkeys: Maybe check to see if it has actually changed before doing all o this.

masonry_core/src/core/widget_state.rs
PoignardAzur: Yeah, the documentation link is already valid.

masonry_core/src/doc/masonry_concepts.md
PoignardAzur: True. I think I need to start a discussion about this on Zulip.

masonry_core/src/core/widget.rs
```diff
@@ -562,9 +564,9 @@ pub fn find_widget_under_pointer<'c>(
 
     let local_pos = ctx.window_transform().inverse() * pos;
 
-    if let Some(clip) = ctx.clip_path()
-        && !clip.contains(local_pos)
-    {
+    let is_inside_clip_shape = ctx.content_box().contains(local_pos);
+
+    if ctx.clips_contents() && !is_inside_clip_shape {
```
xStrom, line 569: A lot of widgets won't have `ctx.clips_contents()` as `true`, right? So we could avoid doing the content box fetching and position checking.

```suggestion
    if ctx.clips_contents() && !ctx.content_box().contains(local_pos) {
```

masonry_core/src/core/widget_state.rs
```diff
@@ -405,24 +408,45 @@ impl WidgetState {
...
+    ///
+    /// A widget's clip shape will have two effects:
+    /// - It serves as a mask for painting operations of this widget and its children.
+    ///   Note that while all painting done by children will be clipped by this path,
+    ///   only the painting done in `Widget::paint` by this widget itself will be clipped.
+    ///   The remaining painting done in `Widget::pre_paint` and `Widget::post_paint` will not be clipped.
+    /// - Pointer events must be inside that shape to reach the widget's children.
+    ///
+    /// This currently returns the border-box rect if `clips_contents` is true and `None` otherwise, but in the future we may want to support more complex clip shapes, in which case this method would need to be updated.
+    ///
```
xStrom, line 444: I would prefer to keep these paint methods short and clickable links. Shortness also helps keep the line length under a 100 and links help keep docs fresh.

Similarly, let's break the long sentence into multiple lines so it's below 100.

```suggestion
    ///   only the painting done in [`paint`] by this widget itself will be clipped.
    ///   The remaining painting done in [`pre_paint`] and [`post_paint`] will not be clipped.
    /// - Pointer events must be inside that shape to reach the widget's children.
    ///
    /// This currently returns the border-box rect if `clips_contents` is true and `None` otherwise,
    /// but in the future we may want to support more complex clip shapes,
    /// in which case this method would need to be updated.
    ///
    /// [`paint`]: crate::core::Widget::paint
    /// [`pre_paint`]: crate::core::Widget::pre_paint
    /// [`post_paint`]: crate::core::Widget::post_paint
```

masonry_core/src/core/widget_state.rs
```diff
@@ -405,24 +408,45 @@ impl WidgetState {
...
+    /// [clip shape]: crate::doc::masonry_concepts#clip-shape.
+    pub(crate) fn clip_shape(&self) -> Rect {
+        self.border_box_size().to_rect() - self.border_box_translation()
```
xStrom, line 447: This is wrong, that translation means that it returns the border box rect in content-box coordinate space.

To return the border box rect in border-box coordinate space, you just do this:

```suggestion
        self.border_box_size().to_rect()
```

If instead you wanted to actually clip to the content-box rect, then it would be:

```suggestion
        self.border_box_size().to_rect() - self.border_box_insets
```

However, then a bunch of comments would need updating as well.

xStrom (commented):
If you want to land this sooner than later, then I suggest making the `bool` mean border-box (matches the old behavior). I mention in an inline comment how to get that.

That will get the same old results for `Prose` and `Label`. It will (temporarily) regress `TextInput` but that would be ok.

---

More generally, I think we should provide easy clip support with more options. I think the three choices from CSS's [`overflow-clip-margin`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-clip-margin) are all very useful.
* content-box
* padding-box
* border-box

Essentially I can imagine having:

```rust
enum Clip {
    None,
    ContentBox,
    PaddingBox,
    BorderBox,
    Custom(Shape),
}
```

Whether you think it should happen already in this PR I'll leave for you to decide.

I am also willing to implement the various pre-defined box clips myself, given I'm very familiar with the box conversions right now.

PR #1632, opened 2026-01-28
Add `TestHarness::print_render_snapshot()` method

PoignardAzur:
This should be useful for debugging tests.

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -1067,7 +1067,7 @@ impl<W: Widget> TestHarness<W> {
     // --- MARK: SNAPSHOT
 
     /// Renders the current widget tree to a pixmap, and writes it to a temporary PNG file.
-    pub fn save_render_screenshot(&mut self) {
+    pub fn save_render_snapshot(&mut self) {
         let image = self.render();
 
         let mut buffer = Cursor::new(Vec::new());


masonry_testing/src/harness.rs
```diff
@@ -1065,6 +1066,24 @@ impl<W: Widget> TestHarness<W> {
...
+
+        let id = std::time::SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis();
+        let tmp_path = std::env::temp_dir().join(format!("masonry-{id:016}-screenshot.png"));
+
+        std::fs::write(&tmp_path, image_data).unwrap();
+        debug!("Screenshot written to {}", tmp_path.display());
+    }
```
xStrom: I would keep the terminology more consistent. Saving screenshots seems like an ok choice.

```suggestion
    pub fn save_render_screenshot(&mut self) {
        let image = self.render();

        let mut buffer = Cursor::new(Vec::new());
        image.write_to(&mut buffer, ImageFormat::Png).unwrap();
        let image_data = buffer.into_inner();

        let id = std::time::SystemTime::now()
            .duration_since(UNIX_EPOCH)
            .unwrap()
            .as_millis();
        let tmp_path = std::env::temp_dir().join(format!("masonry-{id:016}-screenshot.png"));

        std::fs::write(&tmp_path, image_data).unwrap();
        debug!("Screenshot saved to {}", tmp_path.display());
    }
```

xStrom (approved):
Seems useful.

I have some inline non-blocking bikeshedding.

PR #1633, opened 2026-01-28
Add `TestHarness::accessibility_click_on()` method

PoignardAzur:
This method is meant for cases where you want to emulate a click without dealing with visibility issues.

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -781,10 +781,10 @@ impl<W: Widget> TestHarness<W> {
     /// Clicks the given widget.
     ///
     /// This will send an accesskit [`Click`] action to the widget,
-    /// which may have different behavior to a normal mouse click.
+    /// which may have different behavior compared to a normal mouse click.
     ///
     /// Unlike [`Self::mouse_click_on`], the event will be routed to the target widget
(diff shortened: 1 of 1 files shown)


masonry_testing/src/harness.rs
```diff
@@ -778,6 +778,28 @@ impl<W: Widget> TestHarness<W> {
         });
     }
 
+    /// Clicks the given widget.
+    ///
+    /// This will send an accesskit [`Click`] action to the widget,
+    /// which may have different behavior to a normal mouse click.
```
xStrom:
```suggestion
    /// which may have different behavior compared to a normal mouse click.
```

masonry_testing/src/harness.rs
```diff
@@ -778,6 +778,28 @@ impl<W: Widget> TestHarness<W> {
...
+    ///
+    /// Unlike [`Self::mouse_click_on`], the event will be routed to the target widget
+    /// event if it's not visible, or hidden behind other widgets.
```
xStrom: It could be behind a transparent widget.

```suggestion
    /// even if it's not visible or behind other widgets.
```

PoignardAzur: I'll keep the current wording, it's evocative even if it's not quite accurate.

PR #1635, opened 2026-01-29
masonry: badges + badged overlays

waywardmonkeys:
* Add `widgets::Badge` and `widgets::Badged` (overlay badge on arbitrary content)
* Add count helpers with overflow + hide-at-zero support
* Extend gallery `Badge` demo (including interactive count)
* Add `xilem_masonry::view::{badge*, badged*}` wrappers


masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+#[doc = concat!(
+    "![Badge with text](",
+    include_doc_path!("screenshots/badge_with_text.png"),
+    ")",
+)]
+///
+/// This widget is non-interactive (it emits no actions). It uses the theme's per-widget
+/// defaults for [`Background`], [`DisabledBackground`], [`Padding`], [`CornerRadius`], and
+/// border properties.
+///
```
PoignardAzur: I think that part doesn't need to be included, it does the same thing as every other widget.

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// Always show the exact count value.
+    Exact,
+    /// If the count exceeds `max`, show `max` or `max+` depending on `show_plus`.
+    Cap {
+        /// The maximum value to display.
+        max: u32,
+        /// Whether to append a `+` when `count > max`.
+        show_plus: bool,
+    },
+}
```
PoignardAzur, line 63: I think this part is a little too opinionated for Masonry. It should probably be moved to Xilem.

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+                }
+            }
+        };
+        Self::with_text(text)
+    }
+
+    /// Creates a numeric badge only when `count` is non-zero.
+    pub fn count_nonzero(count: u32) -> Option<Self> {
+        (count != 0).then(|| Self::count(count))
+    }
```
PoignardAzur, line 136: Same here, that's too many options. I think `Badge::with_text` already provides what users need.

I think you might want to include a `text_for_count` function like this:

```rust
impl Badge {
    pub fn text_for_count(count: u32, max_count: Option<u32>) -> String {
        if let Some(max_count) = max_count
            && count > max_count
        {
            format!("{max_count}+")
        } else {
            format!("{count}")
        }
    }
}
```

and then let Xilem and other users call that function.

(Which is already more or less what you're doing when Xilem needs to update the count.)

masonry/src/widgets/badged.rs
```diff
@@ -0,0 +1,293 @@
...
+        content: NewWidget<impl Widget + ?Sized>,
+        badge: Option<NewWidget<dyn Widget>>,
+    ) -> Self {
+        Self {
+            content: content.erased().to_pod(),
+            badge: badge.map(|b| b.to_pod()),
+            placement: BadgePlacement::TopRight,
+            offset: Vec2::ZERO,
+        }
+    }
```
PoignardAzur, line 93: I get the way this composes with the `Badge::count_nonzero` constructors, but adding and removing children from widgets is a lot of hassle, as the rest of this PR shows.

Have you considered having a `badge_visible: bool` flag instead? You could then stash and unstash the badge when needed, and not have to deal with `Option::map` throughout the code. (Though note that Masonry currently doesn't let you create a widget already stashed, so you'd need to update it in the layout method.)

xilem_masonry/src/view/badge.rs
```diff
@@ -0,0 +1,255 @@
...
+    BadgeCount {
+        count,
+        overflow: widgets::BadgeCountOverflow::default(),
+    }
+}
+
+/// A badge displaying a numeric count with explicit overflow formatting.
+pub fn badge_count_with_overflow(count: u32, overflow: widgets::BadgeCountOverflow) -> BadgeCount {
+    BadgeCount { count, overflow }
+}
```
PoignardAzur, line 43: I think you could merge those two by writing this:

```rust
pub fn badge_count(count: u32, max_count: Option<u32>) -> BadgeCount {
    BadgeCount {
        count,
        overflow: ...,
    }
}
```

xilem_masonry/src/view/badge.rs
```diff
@@ -0,0 +1,255 @@
...
+    Badge {
+        child,
+        phantom: PhantomData,
+    }
+}
+
+/// A badge containing a styled text label.
+pub fn badge_text(text: impl Into<ArcStr>) -> BadgeText {
+    BadgeText { text: text.into() }
+}
```
PoignardAzur, line 28: You could make this return `Badge<xilem::views::Label>` and remove the `BadgeText` type.

xilem_masonry/src/view/badged.rs
```diff
@@ -0,0 +1,310 @@
...
+use crate::core::{
+    Arg, MessageCtx, MessageResult, Mut, View, ViewArgument, ViewId, ViewMarker, ViewPathTracker,
+};
+use crate::{Pod, ViewCtx, WidgetView};
+
+/// Decorate `content` by overlaying a badge widget on top of it.
+pub fn badged<State, Action, Content, BadgeV>(
+    content: Content,
+    badge: BadgeV,
+) -> Badged<Content, BadgeV, State, Action>
```
PoignardAzur, line 18: The documentation should refer to `badge()`.

(Also, I find it a little odd that BadgeV ends up being generic, but I guess it matches how the `Badged` widget accepts any widget as a badge.)

PoignardAzur (commented):
Looks good overall.

I think the code is trying a little too hard to provide ergonomic default options, in a way that leads to a lot of code/API duplication, but I guess this is a widget for which default options are good.

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+    ///
+    /// # Examples
+    /// ```
+    /// use masonry::core::Widget;
+    /// use masonry::widgets::Badge;
+    ///
+    /// let badge = Badge::with_text("New");
+    /// ```
+    pub fn with_text(text: impl Into<Arc<str>>) -> Self {
+        use crate::parley::style::FontWeight;
```
xStrom: I'd move this import to the top of the file.

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+        let child_length = ctx.compute_length(
+            &mut self.child,
+            auto_length,
+            context_size,
+            axis,
+            cross_space,
+        );
+
+        child_length + border_length + padding_length
+    }
```
xStrom, line 187: The `measure` implementation can be much simpler now.

```suggestion
    fn measure(
        &mut self,
        ctx: &mut MeasureCtx<'_>,
        _props: &PropertiesRef<'_>,
        axis: Axis,
        len_req: LenReq,
        cross_length: Option<f64>,
    ) -> f64 {
        let auto_length = len_req.into();
        let context_size = LayoutSize::maybe(axis.cross(), cross_length);

        ctx.compute_length(
            &mut self.child,
            auto_length,
            context_size,
            axis,
            cross_length,
        )
    }
```

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+        let child_origin = ((size - child_size).to_vec2() * 0.5).to_point();
+        ctx.place_child(&mut self.child, child_origin);
+
+        let child_baseline = ctx.child_baseline_offset(&self.child);
+        let child_baseline = border.baseline_up(child_baseline, scale);
+        let child_baseline = padding.baseline_up(child_baseline, scale);
+        let child_bottom = child_origin.y + child_size.height;
+        let bottom_gap = size.height - child_bottom;
+        ctx.set_baseline_offset(child_baseline + bottom_gap);
+    }
```
xStrom: `layout` can be simpler too.

```suggestion
    fn layout(&mut self, ctx: &mut LayoutCtx<'_>, _props: &PropertiesRef<'_>, size: Size) {
        let child_size = ctx.compute_size(&mut self.child, SizeDef::fit(size), size.into());
        ctx.run_layout(&mut self.child, child_size);

        let child_origin = ((size - child_size).to_vec2() * 0.5).to_point();
        ctx.place_child(&mut self.child, child_origin);

        let child_baseline = ctx.child_baseline_offset(&self.child);
        let child_bottom = child_origin.y + child_size.height;
        let bottom_gap = size.height - child_bottom;
        ctx.set_baseline_offset(child_baseline + bottom_gap);
    }
```

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+use vello::Scene;
+
+use crate::core::{
+    AccessCtx, ChildrenIds, LayoutCtx, MeasureCtx, NewWidget, NoAction, PaintCtx, Properties,
+    PropertiesMut, PropertiesRef, RegisterCtx, StyleProperty, Update, UpdateCtx, Widget, WidgetId,
+    WidgetMut, WidgetPod,
+};
+use crate::kurbo::{Axis, Size};
+use crate::layout::{LayoutSize, LenReq, SizeDef};
+use crate::properties::{BorderWidth, ContentColor, DisabledContentColor, Padding};
```
xStrom: These imports are no longer needed.

```suggestion
use crate::properties::{ContentColor, DisabledContentColor};
```

masonry/src/widgets/badge.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+// --- MARK: TESTS
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use masonry_testing::TestHarnessParams;
+
+    use crate::testing::{TestHarness, assert_render_snapshot};
```
xStrom:
```suggestion

    use crate::testing::{TestHarness, TestHarnessParams, assert_render_snapshot};
```

masonry/src/widgets/badged.rs
```diff
@@ -0,0 +1,293 @@
...
+            this.widget.offset = offset;
+            this.ctx.request_layout();
+        }
+    }
+
+    /// Removes the badge entirely.
+    pub fn clear_badge(this: &mut WidgetMut<'_, Self>) {
+        if let Some(old) = this.widget.badge.take() {
+            this.ctx.remove_child(old);
+            this.ctx.children_changed();
```
xStrom: `remove_child` already calls `children_changed` and that is guaranteed and documented, so we can omit it.

```suggestion
            this.ctx.remove_child(old);
```

masonry/src/widgets/badged.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        let content_anchor = match self.placement {
+            BadgePlacement::TopLeft => Point::new(0.0, 0.0),
+            BadgePlacement::TopRight => Point::new(content_size.width, 0.0),
+            BadgePlacement::BottomLeft => Point::new(0.0, content_size.height),
+            BadgePlacement::BottomRight => Point::new(content_size.width, content_size.height),
+        };
+
+        let badge_origin = content_anchor + self.offset
+            - Vec2::new(badge_size.width / 2.0, badge_size.height / 2.0);
```
xStrom: Can nano-optimize this.

```suggestion
        let badge_origin = content_anchor + self.offset
            - Vec2::new(badge_size.width * 0.5, badge_size.height * 0.5);
```

masonry/src/widgets/badged.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+}
+
+// --- MARK: TESTS
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use masonry_testing::TestHarnessParams;
+
+    use crate::testing::{TestHarness, assert_render_snapshot};
```
xStrom:
```suggestion

    use crate::testing::{TestHarness, TestHarnessParams, assert_render_snapshot};
```

masonry/src/widgets/badged.rs
```diff
@@ -0,0 +1,293 @@
...
+pub enum BadgePlacement {
+    /// Place the badge at the content's top-left corner.
+    TopLeft,
+    /// Place the badge at the content's top-right corner.
+    TopRight,
+    /// Place the badge at the content's bottom-left corner.
+    BottomLeft,
+    /// Place the badge at the content's bottom-right corner.
+    BottomRight,
+}
```
xStrom, line 27: I'd remove this bespoke `BadgePlacement` type and use the existing `UnitPoint` type we have. We'd get more alignments and reuse of code.

masonry/src/widgets/badged.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        let badge_size = ctx.compute_size(badge, SizeDef::MAX, size.into());
+        ctx.run_layout(badge, badge_size);
+
+        let content_anchor = match self.placement {
+            BadgePlacement::TopLeft => Point::new(0.0, 0.0),
+            BadgePlacement::TopRight => Point::new(content_size.width, 0.0),
+            BadgePlacement::BottomLeft => Point::new(0.0, content_size.height),
+            BadgePlacement::BottomRight => Point::new(content_size.width, content_size.height),
+        };
```
xStrom: When using `UnitPoint` for placement we could simplify this part.

```suggestion
        let content_anchor = self.placement.resolve(content_size.to_rect());
```

masonry/src/widgets/badge.rs
waywardmonkeys: I don't use Xilem.

masonry/src/widgets/badge.rs
waywardmonkeys: I don't use XIlem.

If Masonry can't have some ergonomics, then I'm not sure what the point is in me using it ... I'd be more inclined to just do this widget in my own layer on top of Masonry and not bother having anything here?

masonry/src/widgets/badged.rs
waywardmonkeys: A `badge_visible` flag might work. Apart from it not working as you seem to indicate when you say that I can't create a widget already stashed ...

masonry/src/widgets/badged.rs
waywardmonkeys: But that allows for illegal stuff as well ...

xStrom: Only in the "looks bad" sense though, so if someone wants to opt-in to that, I'd say go for it.

Anyway, I don't think this is a blocker or anything, so if you want to keep the API restricted so be it.

masonry/src/widgets/badge.rs
xStrom: I definitely think that Masonry should provide the less opinionated escape hatch, as this does with `with_text`.

Providing extra functionality on top doesn't seem an issue to me, at least when it doesn't introduce a heavy cost, which doesn't seem to be the case here.

What are the primary issues with having these options? Is it the maintenance cost of extra code or some principled  threshold of how non-specific functionality should be or ..?

PoignardAzur: I guess there isn't a strong downside.

masonry/src/widgets/badge.rs
waywardmonkeys: I interpret the response to Kaur below as this needn't change.

masonry/src/widgets/badge.rs
waywardmonkeys: Removed in next push.

xilem_masonry/src/view/badge.rs
waywardmonkeys: That's not possible with the current design (and would lead to some code duplication, I think).

masonry/src/widgets/badged.rs
waywardmonkeys: I added the same function extracting this to `BadgePlacement`, so at least the call site looks good now.

masonry/src/widgets/badged.rs
waywardmonkeys: I'd like to keep it for now ... I'm unsure of the best direction.

PR #1637, opened 2026-01-30
Skip painting transparent background.

xStrom:
The issues with skipping transparent background painting I described in #1592 seem to have disappeared, which is great.

We can now skip transparent background painting, enabling more efficient scene caching. No need to store the draw call and if it was the only pre-paint draw call then in the future we could skip caching the scene as a whole.

I also added a test to make sure transparent layering stays working.


masonry/screenshots/paint_transparency.png
PoignardAzur, line 1: I think you should maybe make the test image larger, give it space to breathe, make it clearer that it's four separate versions of the same thing?

Maybe make the shape be squares, with each pair of squares having about a fourth of their respective surfaces overlap?

xStrom: I guess I can introduce some gaps and padding.

As for the overlap, reducing it from 75% to 25% would really hide what's being primarily tested here - the overlap!

PR #1638, opened 2026-01-30
xilem_web: Add components example

flosse:
I am trying to demonstrate how to build components that are generic,
but as soon as I define the event handler

```rust
 html::button("click").on_click(|s: &mut AppState, _| { s.clicks += 1;  }),
```

I get this error message that I don't understand:

```
error[E0161]: cannot move a value of type `impl xilem_web::interfaces::Element<&mut AppState, CardAction<()>>`
  --> xilem_web/web_examples/components/src/main.rs:52:9
   |
52 | /         card(
53 | |             "Card Example",
54 | |             state.card_collapsed,
55 | |             html::div((
...  |
60 | |             )),
61 | |         ),
   | |_________^ the size of `impl xilem_web::interfaces::Element<&mut AppState, CardAction<()>>` cannot be statically determined
   ```
   
   Can anyone help me?

### DIFF
--- a/xilem_web/web_examples/components/src/date_picker.rs
+++ b/xilem_web/web_examples/components/src/date_picker.rs
@@ -1,3 +1,6 @@
+// Copyright 2026 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
 use time::{Date, Duration, OffsetDateTime};
 use xilem_web::{elements::html, interfaces::Element, modifiers::style};
 

### CODE REVIEW

**Comment by PoignardAzur** [comment]
Welp, by now I've spent more than five hours looking at the code and I can't figure out how to fix the general problem except "remove the ViewArgument trait".

On the shorter term, you probably want to use the trick mentioned in the Zulip thread Philipp linked; in our case that means specifying `html::button::<Edit<AppState>, _, _>(...)` :

```rust
            html::div((
                "Some content ...",
                html::button::<Edit<AppState>, _, _>("click").on_click(|s: &mut AppState, _| {
                    s.clicks += 1;
                }),
            )),
```

Doing that makes the code compile on my end.

flosse:
> Welp, by now I've spent more than five hours looking at the code and I can't figure out how to fix the general problem

:open_mouth: 

> except "remove the ViewArgument trait".

Okay, and that would have a lot of consequences, right?

> On the shorter term, you probably want to use the trick mentioned in the Zulip thread Philipp linked;

Well, it's not the most beautiful code, but it could be much worse :wink: 
Then I'll finish the PR with the current workaround, right?

Thank you for investing your time!

RagibHasin:
https://github.com/RagibHasin/scrolled-quran/blob/be9a78953f94b327920b0758bee5460c74adf86c/src/view.rs#L48-L64

https://github.com/RagibHasin/scrolled-quran/blob/be9a78953f94b327920b0758bee5460c74adf86c/src/view.rs#L225-L229

https://github.com/RagibHasin/scrolled-quran/blob/be9a78953f94b327920b0758bee5460c74adf86c/src/view.rs#L251-L348

This is the place where I extract only the relevant part of app state for the child view.

I guess this can be done using an Adapt view too, have to see.

flosse:
Sorry, I still don't understand the problem. What do you mean by

> I see no option other than index-unwrap in the child.

Are you talking about this [line](https://github.com/RagibHasin/scrolled-quran/blob/be9a78953f94b327920b0758bee5460c74adf86c/src/view.rs#L52)?
```rust
let (idx, reader_state) = state.reader.as_mut().unwrap();
```
Your [`reader` field is a `Option`](https://github.com/RagibHasin/scrolled-quran/blob/be9a78953f94b327920b0758bee5460c74adf86c/src/model.rs#L152C5-L152C16):
```rust
pub reader: Option<(usize, ScrollingReader)>
 ```
So you either `unwrap` it (I wouldn't do that) or you `map` it.
But this has nothing to do with Xilem or `ViewArgument`.

I assume you mean something else, but please specify exactly what you mean!

PR #1644, opened 2026-02-08
xilem_web: Pass `State` to `after_build` and `after_rebuild` methods

flosse:
Motivated by https://github.com/linebender/xilem/pull/780#issuecomment-3862054876 this PR passes the `State` as argument to the `after_build` and `after_rebuild` methods.

### DIFF
--- a/xilem_web/web_examples/open_and_save_file/src/main.rs
+++ b/xilem_web/web_examples/open_and_save_file/src/main.rs
@@ -116,6 +116,7 @@ fn hidden_save_link() -> impl Element<Edit<AppState>> + use<> {
             if !state.start_saving {
                 return;
             }
+            state.start_saving = false;
             let blob = Blob::new(&*state.text);
             let url = ObjectUrl::from(blob);
             el.set_href(&url);


xilem_web/web_examples/open_and_save_file/src/main.rs
```diff
@@ -129,24 +101,25 @@ fn open_file_input(app_state: &mut AppState) -> impl Element<Edit<AppState>> + u
...
-            }
-        })
-        .before_teardown({
-            let el_ref = Rc::clone(&state.raw_save_link);
-            move |_| {
-                *el_ref.borrow_mut() = None;
+        .after_rebuild(|state: &mut AppState, el| {
+            if !state.start_saving {
+                return;
             }
```
Philipp-M, line 118:
```suggestion
            }
            state.start_saving = false;
```

Otherwise a file is saved every rebuild...

Generally it would be nice if we could avoid this component entirely with something like https://developer.mozilla.org/en-US/docs/Web/API/File_System_API#writing_to_files

But that API is not yet supported everywhere unfortunately. Maybe there are other ways too, but I've not looked deep into that topic/issue yet...

xilem_web/web_examples/raw_dom_access/src/main.rs
```diff
@@ -21,34 +18,33 @@ use xilem_web::{App, DomView, document_body};
...
                 })
-                .after_rebuild(move |el| {
+                .after_rebuild(move |app_state: &mut AppState, el| {
                     log::debug!("element was re-build");
-                    if focus.get() {
+                    if app_state.focus {
                         let _ = el.focus();
                         // Reset `focus` to avoid calling `el.focus` on every rebuild.
-                        focus.set(false); // NOTE: this does NOT trigger a rebuild.
+                        app_state.focus = false; // NOTE: this does NOT trigger a rebuild.
```
Philipp-M, line 40: It might be worth just saving the input element as `Option<...>` in the app state via `after_build`/`before_teardown` and act on it directly below.

Philipp-M (approved):
Apart of the file-saving issue as commented, LGTM.

Nice, thanks for implementing that. Definitely cleans things up a little bit.

xilem_web/web_examples/open_and_save_file/src/main.rs
flosse: > Otherwise a file is saved every rebuild...

:see_no_evil: 
 
> Generally it would be nice if we could avoid this component entirely with something like https://developer.mozilla.org/en-US/docs/Web/API/File_System_API#writing_to_files

I had no idea such a thing existed.

> But that API is not yet supported everywhere unfortunately. Maybe there are other ways too, but I've not looked deep into that topic/issue yet...

I can't promise anything, but if I'm feeling bored again, I'll take a look at it :relaxed:

xilem_web/web_examples/raw_dom_access/src/main.rs
flosse: > It might be worth just saving the input element as `Option<...>` in the app state via `after_build`/`before_teardown` and act on it directly below.

I think the problem is, that `View::teardown` does not provide access to the app state :thinking:

Philipp-M: Right, forgot about that we removed app state access in teardown...
Well it's orthogonal to this change anyway, we should address these state-access issues in different PRs.

Issue #1639, opened 2026-02-01
Text input has a bigger background than actual text area.


ricvelozo:
On the `main` branch.

<img width="480" height="270" alt="Image" src="https://github.com/user-attachments/assets/96f0557f-1796-4d74-b361-b5739905e8c0" />
<img width="480" height="270" alt="Image" src="https://github.com/user-attachments/assets/e2c25d44-43f0-40f8-a650-c41f6748f2b6" />

PoignardAzur:
This is as fixed as it's going to be in the near future. Closing this issue.

PR #1647, opened 2026-02-10
Give focus to TextArea when clicking TextInput padding

PoignardAzur:
Fixes part of #1639.

It's not a perfect fix (mouse cursor isn't changed, click-and-dragging won't start a text selection) but it's makes the jank less user-visible.

### DIFF
--- a/masonry/src/widgets/text_input.rs
+++ b/masonry/src/widgets/text_input.rs
@@ -4,16 +4,15 @@
 use std::any::TypeId;
 
 use accesskit::{Node, Role};
-use masonry_core::core::{PointerButton, PointerButtonEvent};
 use tracing::{Span, trace_span};
 use vello::Scene;
 
 use crate::TextAlign;
 use crate::core::{
     AccessCtx, ArcStr, ChildrenIds, EventCtx, HasProperty, LayoutCtx, MeasureCtx, NewWidget,
-    NoAction, PaintCtx, PointerEvent, PrePaintProps, PropertiesMut, PropertiesRef, RegisterCtx,
-    Update, UpdateCtx, Widget, WidgetId, WidgetMut, WidgetPod, paint_background, paint_border,
-    paint_box_shadow,
+    NoAction, PaintCtx, PointerButton, PointerButtonEvent, PointerEvent, PrePaintProps,
+    PropertiesMut, PropertiesRef, RegisterCtx, Update, UpdateCtx, Widget, WidgetId, WidgetMut,
+    WidgetPod, paint_background, paint_border, paint_box_shadow,
 };
 use crate::kurbo::{Axis, Point, Size};
 use crate::layout::{LayoutSize, LenReq};


masonry/src/widgets/text_input.rs
```diff
@@ -4,14 +4,16 @@
...
 use crate::TextAlign;
 use crate::core::{
-    AccessCtx, ArcStr, ChildrenIds, HasProperty, LayoutCtx, MeasureCtx, NewWidget, NoAction,
-    PaintCtx, PrePaintProps, PropertiesMut, PropertiesRef, RegisterCtx, Update, UpdateCtx, Widget,
-    WidgetId, WidgetMut, WidgetPod, paint_background, paint_border, paint_box_shadow,
+    AccessCtx, ArcStr, ChildrenIds, EventCtx, HasProperty, LayoutCtx, MeasureCtx, NewWidget,
+    NoAction, PaintCtx, PointerEvent, PrePaintProps, PropertiesMut, PropertiesRef, RegisterCtx,
+    Update, UpdateCtx, Widget, WidgetId, WidgetMut, WidgetPod, paint_background, paint_border,
+    paint_box_shadow,
 };
```
xStrom:
```suggestion
use tracing::{Span, trace_span};
use vello::Scene;

use crate::TextAlign;
use crate::core::{
    AccessCtx, ArcStr, ChildrenIds, EventCtx, HasProperty, LayoutCtx, MeasureCtx, NewWidget,
    NoAction, PaintCtx, PointerButton, PointerButtonEvent, PointerEvent, PrePaintProps,
    PropertiesMut, PropertiesRef, RegisterCtx, Update, UpdateCtx, Widget, WidgetId, WidgetMut,
    WidgetPod, paint_background, paint_border, paint_box_shadow,
};
```

masonry/src/widgets/text_input.rs
```diff
@@ -156,6 +158,25 @@ impl HasProperty<UnfocusedSelectionColor> for TextInput {}
...
+        &mut self,
+        ctx: &mut EventCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        event: &PointerEvent,
+    ) {
+        match event {
+            PointerEvent::Down(PointerButtonEvent {
+                button: None | Some(PointerButton::Primary),
+                ..
+            }) => {
```
xStrom, line 170: This matches `TextArea` so I guess it's fine for now. Longer term, however, I think focus should definitely also happen with other buttons.

masonry/src/widgets/text_input.rs
PoignardAzur: I really need to look into fixing this in rust-analyzer, this is driving me crazy.

xStrom: I looked into it at one point and found that rust-analyzer doesn't support configuring this. It will always prefer the shortest import path.

Maybe we should stop worrying about this? Not sure, for now I've just been suggesting these changes to keep things consistent.

PR #1651, opened 2026-02-11
Refactor how Xilem handles async actions

PoignardAzur:
Add `AppDriver::on_async_action` method.
Remove `ASYNC_MARKER_WIDGET` (fake widget id used to route events).
Remove `WidgetId::reserved` method.

### DIFF
--- a/masonry_winit/src/app_driver.rs
+++ b/masonry_winit/src/app_driver.rs
@@ -98,7 +98,7 @@ pub trait AppDriver {
 
     /// A hook which will be executed for async actions sent outside the widget tree.
     ///
-    /// This is called when the winit event loops gets a [`MasonryUserEvent::AsyncAction`] event.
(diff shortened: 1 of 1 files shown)


masonry_winit/src/app_driver.rs
```diff
@@ -96,6 +96,19 @@ pub trait AppDriver {
         action: ErasedAction,
     );
 
+    /// A hook which will be executed for async actions sent outside the widget tree.
+    ///
+    /// This is called when the winit event loops gets a [`MasonryUserEvent::AsyncAction`] event.
```
xStrom:
```suggestion
    /// This is called when the winit event loop gets a [`MasonryUserEvent::AsyncAction`] event.
```

xStrom (approved):
Seems like an improvement.

PR #1653, opened 2026-02-14
Add a Xilem stress test to measure stacking properties

PoignardAzur:


xilem/stress_tests/property_stack.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright 2026 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Stress test for the build speed a Xilem component with lots of styles applied.
```
xStrom:
```suggestion
//! Stress test for the build speed of a Xilem component with lots of styles applied.
```

xStrom (approved):
I guess repeating the same style works because we don't do any optimizations for overrides. It does feel a bit silly though.

More importantly, does 9 styles really count as a stress test? I feel like I'd have more than that in regular non-extreme use.

Overall the skeleton seems fine, so approving.

PR #1655, opened 2026-02-14
Add `decorative` option for Images with no accessibility node

### LINKED ISSUE #1136: `Image` should have a way to not generate an accessibility node
Some images are decorative and don't need to be present in the AccessKit node tree. There should be a way to prevent it from appearing.

I think this needs changes to the accessibility pass to provide a way to opt out of this process (so I haven't flagged it right now as a good first issue / help wanted).

PoignardAzur:
Fixes #1136

### DIFF
--- a/masonry/src/widgets/image.rs
+++ b/masonry/src/widgets/image.rs
@@ -59,7 +59,8 @@ impl Image {
         }
     }
 
-    /// Specifies whether the image is decorative, meaning it doesn't have meaningful content and is only for visual presentation.
+    /// Specifies whether the image is decorative, meaning it doesn't have meaningful content
+    /// and is only for visual presentation.
     ///
     /// If `is_decorative` is `true`, the image will be ignored by screen readers.
     pub fn decorative(mut self, is_decorative: bool) -> Self {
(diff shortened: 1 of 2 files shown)


masonry/src/widgets/image.rs
```diff
@@ -53,10 +54,19 @@ impl Image {
     pub fn new(image_data: impl Into<ImageBrush>) -> Self {
         Self {
             image_data: image_data.into(),
+            decorative: false,
             alt_text: None,
         }
     }
 
+    /// Specifies whether the image is decorative, meaning it doesn't have meaningful content and is only for visual presentation.
```
xStrom: Let's keep the comment lines also under 100 chars.

```suggestion
    /// Specifies whether the image is decorative, meaning it doesn't have meaningful content
    /// and is only for visual presentation.
```

masonry/src/widgets/image.rs
```diff
@@ -80,6 +90,14 @@ impl Image {
         this.ctx.request_layout();
     }
 
+    /// Sets whether the image is decorative, meaning it doesn't have meaningful content and is only for visual presentation.
```
xStrom:
```suggestion
    /// Sets whether the image is decorative, meaning it doesn't have meaningful content
    /// and is only for visual presentation.
```

xilem_masonry/src/view/image.rs
```diff
@@ -49,6 +51,14 @@ impl Image {
         self.prop(fill)
     }
 
+    /// Specifies whether the image is decorative, meaning it doesn't have meaningful content and is only for visual presentation.
```
xStrom:
```suggestion
    /// Specifies whether the image is decorative, meaning it doesn't have meaningful content
    /// and is only for visual presentation.
```

PR #1657, opened 2026-02-15
Add README and screenshots for xilem examples

sikma-dev:
- Add `xilem/examples/README.md` with screenshot previews.
- Add a short one-line description for each example.
- Store the screenshot assets under `xilem/screenshots`.

If this direction is not desired, I can adjust or close this PR.


Philipp-M:
If these were automated + snapshot tests I think this would provide value as additional CI checks for more "advanced" usage of Xilem/Masonry (although for some interactive examples this is probably a little bit of effort to do effectively, like the canvas example).
This also poses the question for a testing suite in Xilem...

I think otherwise it's probably fairly quickly out of sync (and/or maintenance burden), although it's nice to have for a quick initial impression. Additionally it seems that for the transforms example a font is missing.

PR #1659, opened 2026-02-16
Add `Widget::propagates_pointer_interaction` method.

PoignardAzur:
This is a cleaner way to make Button and similar widgets have exclusively control of their pointer events, rather than making every widget that might be inside a button never accept them.

### DIFF
--- a/masonry_testing/src/modular_widget.rs
+++ b/masonry_testing/src/modular_widget.rs
@@ -74,7 +74,7 @@ impl<S> ModularWidget<S> {
             state,
             icon: CursorIcon::Default,
             accepts_pointer_interaction: true,
-            propagates_pointer_interaction: false,
+            propagates_pointer_interaction: true,
             accepts_focus: false,
             accepts_text_input: false,
             on_pointer_event: None,


xStrom (approved):
Functionality makes sense.

I'm not too sure about the name though, it's not immediately clear at all what's the difference between `accepts_pointer_interaction` and `contains_pointer_interactions`. Also why is only one of them plural?

`captures_` seems a bit better, but I guess it has a conflict with the existing pointer capture functionality.

`terminates_` or `concludes_` or even just `ends_` make it more clear that the interaction will end but don't make it clear whether the widget itself will still get the event.

Perhaps the `bool` should be flipped and then it could be `propagates_pointer_interaction`. Because that's what it actually controls, right? Propagation of the pointer interaction to its children.

PoignardAzur:
Yeah, I'm not too sure about the naming.

> Perhaps the bool should be flipped and then it could be propagates_pointer_interaction. Because that's what it actually controls, right? Propagation of the pointer interaction to its children.

Good idea, thanks!

PR #1660, opened 2026-02-16
Add CollapsePanel and DisclosureButton widgets

PoignardAzur:


masonry/src/theme.rs
```diff
@@ -29,7 +31,7 @@ pub const BACKGROUND_COLOR: Color = Color::from_rgb8(0x1D, 0x1D, 0x1D);
 
 pub const BORDER_WIDTH: f64 = 1.;
 
-// Zync color variations from https://tailwindcss.com/docs/colors
+// Zync color variations from https://tailwindcoYDIM_GRAY/docs/colors
```
xStrom: What's going on here?

```suggestion
// Zync color variations from https://tailwindcss.com/docs/colors
```

PoignardAzur: Downside of using multi-cursor editors: sometimes you write stuff without realizing it.

PR #1663, opened 2026-02-17
Add concept of "attached layers" to Masonry

PoignardAzur:


masonry_core/src/core/contexts.rs
```diff
@@ -1804,11 +1810,81 @@ impl_context_method!(
             ));
         }
 
+        /// Creates a new [layer] at a specified `position`, and ties it to the current widget.
+        ///
+        /// The layer will be removed when the current widget is removed from the tree.
+        /// A widget can only have one layer of any given type `W` at a time: if a widget creates two layers of the same type, the old layer will be removed.
```
xStrom: Let's keep line lengths under 100.

```suggestion
        /// A widget can only have one layer of any given type `W` at a time:
        /// if a widget creates two layers of the same type, the old layer will be removed.
```

PR #1667, opened 2026-02-18
Rename "Properties" to "PropertySet"

PoignardAzur:
Split `properties.rs` into (roughly) one file per struct/trait.

This is a moving-and-renaming-only PR, with no behavior changes.


PoignardAzur:
Right, I was skipping some context there. We're still discussing when and how we'll integrate features from understory_properties and your experimental branch using them.

I'd still like to merge this now. "Properties" is a bad name, and I'll still want to change it even (especially?) if we depend on understory_properties. I don't think this PR will make your work harder to integrate.

PR #1669, opened 2026-02-19
masonry_core: Add first-class pseudo-class state on WidgetState

waywardmonkeys:
## Motivation

Masonry tracks interaction states (hovered, active, focused, disabled) as individual booleans on `WidgetState`. When these change, the framework sets `request_pre_paint` and `needs_paint`, then calls `widget.update()`. But `request_paint` (for `Widget::paint()`) is **not** set by the framework— widgets must call `ctx.request_paint_only()` in their `update()` to get `paint()` re-run. This is fragile: every widget that changes appearance on hover/focus/active must remember to do this, and forgetting causes rendering bugs that are easy to miss.

There's also no unified state representation that a future style/selector system could match against. CSS-like selectors (`:hover`, `:focus`, `:disabled`, `:checked`) need a single queryable bitfield, not scattered booleans.

## Changes

Introduce `PseudoId(u8)` and `PseudoSet(u64)` as a compact bitfield on `WidgetState` that mirrors the existing boolean interaction flags:

- **Built-in pseudo IDs:** `HOVER(0)`, `ACTIVE(1)`, `FOCUS(2)`, `FOCUS_WITHIN(3)`, `DISABLED(4)`. Bits 5–63 are available for widget-defined states.
- **Context API:** `ctx.set_pseudo(id, value)` on mutable contexts auto-invalidates `pre_paint` + `paint` when state changes. `ctx.pseudos()` and `ctx.has_pseudo(id)` on all contexts for reads.
- **Framework wiring:** The update pass now sets pseudo bits alongside the existing boolean flags at all five interaction sites (hover, active, focus, focus-within, disabled). Also adds the missing `request_paint = true` at each site, fixing the root fragility.
- **Widget-defined state:** `PSEUDO_TOGGLED(5)` for checkboxes and switches. `Checkbox::set_checked` and `Switch::set_on` now use `ctx.set_pseudo()` instead of `request_render()`, and their `update()` methods no longer need manual `request_paint_only()` calls.

## Next steps

- **Style/selector matching:** With pseudo state unified in a bitfield, a selector engine can match `:hover`, `:active`, `:focus`, `:disabled`, `:checked` etc. against `widget_state.pseudos` in a single bitmask comparison.
- **Pseudo registry:** A runtime registry for widget/app-defined pseudo IDs (allocating from bits 5+) so names can be used in stylesheets.
- **Migrate remaining booleans:** The existing `is_hovered`, `is_active`, etc. booleans could eventually be replaced by pseudo-set queries, reducing `WidgetState` size.
- **Broader widget adoption:** Other toggle-like widgets (disclosure panels, radio buttons) can adopt `PSEUDO_TOGGLED` or define their own pseudo states.

### DIFF
--- a/masonry/src/widgets/badge.rs
+++ b/masonry/src/widgets/badge.rs
@@ -160,10 +160,12 @@ impl Widget for Badge {
         ctx.register_child(&mut self.child);
     }
 
-    fn update(&mut self, ctx: &mut UpdateCtx<'_>, _props: &mut PropertiesMut<'_>, event: &Update) {
-        if matches!(event, Update::DisabledChanged(_)) {
-            ctx.request_paint_only();
-        }
+    fn update(
+        &mut self,
+        _ctx: &mut UpdateCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        _event: &Update,
+    ) {
     }
 
     fn measure(
--- a/masonry/src/widgets/label.rs
+++ b/masonry/src/widgets/label.rs
@@ -488,9 +488,6 @@ impl Widget for Label {
                 self.clear_cache();
                 ctx.request_layout();
             }
-            Update::DisabledChanged(_) => {
-                ctx.request_paint_only();
-            }
             _ => {}
         }
     }
--- a/masonry/src/widgets/split.rs
+++ b/masonry/src/widgets/split.rs
@@ -624,16 +624,14 @@ where
         ctx.register_child(&mut self.child2);
     }
 
-    fn update(&mut self, ctx: &mut UpdateCtx<'_>, _props: &mut PropertiesMut<'_>, event: &Update) {
-        match event {
-            Update::FocusChanged(_)
-            | Update::HoveredChanged(_)
-            | Update::ActiveChanged(_)
-            | Update::DisabledChanged(_) => {
-                ctx.request_paint_only();
-            }
-            _ => {}
-        }
+    fn update(
+        &mut self,
+        _ctx: &mut UpdateCtx<'_>,
+        _props: &mut PropertiesMut<'_>,
+        _event: &Update,
+    ) {
+        // Interaction state changes (hover, active, focus, disabled) now automatically
+        // invalidate paint via the framework's update passes.
     }
 
     fn measure(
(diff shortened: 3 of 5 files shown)


waywardmonkeys:
@PoignardAzur This is a demonstration of what we talked about yesterday.

masonry/src/widgets/checkbox.rs
```diff
@@ -70,8 +71,8 @@ impl Checkbox {
     /// Checks or unchecks the box.
     pub fn set_checked(this: &mut WidgetMut<'_, Self>, checked: bool) {
         this.widget.checked = checked;
-        // Checked state impacts appearance and accessibility node
-        this.ctx.request_render();
+        this.ctx.set_pseudo(PSEUDO_TOGGLED, checked);
+        this.ctx.request_accessibility_update();
```
waywardmonkeys, line 75: This was calling request_render before, but now relies on what set_pseudo does, the accessibility (and a post-paint) are the missing bits. This widget doesn't need a post-paint.

masonry_core/src/core/contexts.rs
```diff
@@ -1353,11 +1353,37 @@ impl_context_method!(
...
+    /// If the value actually changes, this automatically requests
+    /// `pre_paint` and `paint` passes so the widget is re-rendered.
+    pub fn set_pseudo(&mut self, id: PseudoId, value: bool) {
+        if self.widget_state.pseudos.contains(id) != value {
+            self.widget_state.pseudos.set(id, value);
+            self.widget_state.request_pre_paint = true;
+            self.widget_state.request_paint = true;
+            self.widget_state.needs_paint = true;
+        }
+    }
```
waywardmonkeys, line 1385: In the future, it would be up to the style engine to decide what needs requesting. This is just a step along the way.

masonry/src/lib.rs
```diff
@@ -223,5 +223,11 @@ pub use vello::{kurbo, peniko};
...
 
 pub use masonry_core::{app, core, layout, ui_events, util};
+
 #[cfg(any(feature = "testing", test))]
 pub use masonry_testing as testing;
+
+/// Pseudo-class for toggled widgets (checkboxes, switches, etc.).
+///
+/// Widget-defined pseudo state at bit 5.
+pub const PSEUDO_TOGGLED: core::PseudoId = core::PseudoId(5);
```
waywardmonkeys, line 233: This is an example of doing it outside of `PseudoId` and would, in the future, be done with a registry. But not sure where to hang that registry from.

PR #1670, opened 2026-02-20
Fix Slider widget not working with non-mouse pointers

RagibHasin:


masonry/src/widgets/slider.rs
```diff
@@ -162,8 +162,13 @@ impl Widget for Slider {
...
                 state,
-                ..
-            }) => {
+                pointer:
+                    PointerInfo {
+                        pointer_type: PointerType::Mouse,
+                        ..
+                    },
+            })
+            | PointerEvent::Down(PointerButtonEvent { state, .. }) => {
```
PoignardAzur: The way we've usually allowed non-mouse pointers is by using `button: Some(PointerButton::Primary) | None`.

I think that should also apply here, unless I'm missing something.

RagibHasin: Alright. I didn’t know we had a convention for this.

PR #1676, opened 2026-02-24
Add `Svg` widget.

xStrom:
The `Svg` widget is very similar to the existing `Image` widget. With the data source being an SVG instead of an already rasterized image. It has the `ObjectFit` property and thus implements its measurement and layout with the help of that.

SVG support is based on Resvg. It has not yet been ported over to our other libraries, so unfortunately we're not doing direct draw calls based on the SVG. Instead the SVG is rasterized via Tiny Skia and then we draw that image. Text stack is also different, but at least there are already open PRs that are trying to port parts of that.

So the current implementation state kind of sucks in terms of performance, bloat, and expected behavior. At the same time it isn't too bad for non-animated static size stuff, so this widget will still unlock a whole range of use cases even before Resvg becomes more aligned with the Linebender ecosystem.


waywardmonkeys:
This was discussed in Office Hours today with a question about it using `resvg` rather than `vello_svg`.

Is part of that decision made to push things along with improving the integration of `resvg` into our ecosystem?

PR #1677, opened 2026-02-26
Add `action` pass.

xStrom:
This PR adds a new action pass to the pass system.

The key motivation is widget composability. This will allow e.g. a container widget to include a `Button` widget in a way where it overrides its `ButtonPress` actions into something more specific to the container. This will also enable refactoring `TextInput` in such a way that it produces its own actions, instead of documenting that the inner `TextArea` will emit actions.

Other use cases might include containers reacting to actions but still letting those exact actions propagate upwards.

Instead of actions going straight to `AppDriver::on_action` they first bubble up from the source widget through every ancestor widget's `Widget::on_action` where those widgets can call `ActionCtx::set_handled`.

### DIFF
--- a/masonry_core/src/passes/action.rs
+++ b/masonry_core/src/passes/action.rs
@@ -1,7 +1,7 @@
 // Copyright 2026 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use tracing::warn;
+use tracing::debug;
 
 use crate::app::{RenderRoot, RenderRootSignal};
 use crate::core::{ActionCtx, ErasedAction, Handled, PropertiesMut, WidgetId};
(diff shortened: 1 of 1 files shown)


masonry_core/src/passes/action.rs
```diff
@@ -0,0 +1,75 @@
...
+            // While partially processed actions can lead to incredibly hard to debug subtle issues
+            // For example if a container widget monitors for actions and generates its own
+            // companion action for the app driver. Then the app driver expecting the companion
+            // action to always follow the initial action. Partial propagation would break that
+            // assumption in the rare cases where the initial action source has just been deleted.
+            warn!(
+                "Aborting action {} propagation because the source widget {source} \
+				has been removed from the tree.",
+                action.type_name(),
+            );
```
PoignardAzur: I'm not sure how useful this warning is. If a developer's widget emits that warning, what are they supposed to do about it?

(In general, I think we have too many warnings in Masonry. I'm worried in real-world apps most of them will be unactionable for users.)

xStrom: Having a clear call to action would be awesome indeed. Problem is that there is no generally applicable suggestion, it really depends on the exact scenario.

I think we can reduce the severity from warn to debug to combat the too many warnings concern. It's a useful log message when trying to figure out why some action didn't register, but nothing in Masonry itself breaks because of it, so it's not that severe.

PR #1680, opened 2026-03-02
Paint each layer into its own Scene

waywardmonkeys:
Replace the single `Scene` return from `run_paint_pass()` with a `PaintResult` struct carrying one `Scene` per layer. Consumers call `PaintResult::composite()` to flatten back into a single `Scene` for Vello, so rendering is visually unchanged.

Overlay layer scenes use layer-local coordinates with an associated transform, so repositioning a layer only changes its composite transform without requiring a repaint.

This is the foundational step toward system compositor integration (Subduction). Once each layer has its own `Scene` in local coordinates, they can be rendered to individual textures and handed to a platform compositor instead of being flattened in software.

### DIFF
--- a/masonry_core/src/app/layer_stack.rs
+++ b/masonry_core/src/app/layer_stack.rs
@@ -64,7 +64,6 @@ impl LayerStack {
     pub(crate) fn layer_id(&self, idx: usize) -> WidgetId {
         self.layers[idx].widget.id()
     }
-
 }
 
 // --- MARK: IMPL WIDGETMUT
--- a/masonry_core/src/passes/paint.rs
+++ b/masonry_core/src/passes/paint.rs
@@ -211,7 +211,13 @@ pub(crate) fn run_paint_pass(root: &mut RenderRoot) -> PaintResult {
     for (idx, &layer_widget_id) in layer_ids.iter().enumerate() {
         if idx == 0 {
             // Base layer: paint directly in window space.
-            paint_layer(root, &mut base_scene, &mut scene_cache, root_id, layer_widget_id);
+            paint_layer(
+                root,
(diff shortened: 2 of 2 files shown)


masonry_core/src/app/layer_stack.rs
```diff
@@ -65,6 +64,7 @@ impl LayerStack {
     pub(crate) fn layer_id(&self, idx: usize) -> WidgetId {
         self.layers[idx].widget.id()
     }
+
```
xStrom: Could omit this new empty line.

```suggestion
```

masonry_core/src/passes/paint.rs
```diff
@@ -14,6 +14,45 @@ use crate::core::{DefaultProperties, PaintCtx, PropertiesRef, WidgetArenaNode, W
...
+///
+/// The base layer contains the main application content in window coordinate space.
+/// Overlay layers contain tooltips, menus, and other popups in layer-local coordinate
+/// space, ordered from bottom to top (painter's order).
+pub struct PaintResult {
+    /// The base layer scene (main application content) in window coordinate space.
+    pub base: Scene,
+    /// Overlay layer scenes in z-order (bottom to top), each in layer-local coordinates.
+    pub overlays: Vec<PaintedLayer>,
+}
```
xStrom, line 43: I think it's easier to reason if we omit making this distinction about spaces of base vs. other layers. All layers can be considered to be in the layer-local coordinate space. How to map the base layer into the window (generally NOP) can exist as a question outside of the layer system itself.

waywardmonkeys: I think that might be good for followups and see if that's true or not as we iterate on this.

PR #1711, opened 2026-04-02
Add common LLM files to .gitignore

PoignardAzur:


.gitignore
```diff
@@ -2,3 +2,13 @@
...
+
+# Common LLM agent files
+AGENT.md
+AGENTS.md
+CLAUDE.md
+GEMINI.md
+research.md
+plan.md
+.mcp.json
+.github/copilot-instructions.md
```
HigherOrderLogic, line 14:
```suggestion
.github/copilot-instructions.md
.codex
```

Codex recently start to produce these files, maybe it's a good idea to gitignore them.

xStrom (approved):
These are common user specific files so it makes sense to ignore them.

DJMcNab:
This is outside of our prior policy on this class of file (I.e. editor settings). Our prior approach has been to avoid an ever-growing list to sync between repositories.

Like, I'm not going to try and block this, but I don't find this compellingly different.

xStrom:
I don't think we should be syncing these across repos at this point, my approval was for this repo only.

I guess we could also block this on [website#137](https://github.com/linebender/linebender.github.io/pull/137) which mentions adding these `.gitignore` entries.

PR #1712, opened 2026-04-03
Refactor NewWidget constructors

PoignardAzur:
Use a single `NewNewidget::new` constructor with builder methods for the other variants.

The changes in `widget_pod.rs` were made by hand. Almost everything else was AI-generated. I did go through each diff chunk to check the edits made sense.

### DIFF
--- a/masonry/src/tests/transforms.rs
+++ b/masonry/src/tests/transforms.rs
@@ -34,8 +34,7 @@ fn blue_box(inner: impl Widget) -> impl Widget {
 #[test]
 fn transforms_translation_rotation() {
     let translation = Vec2::new(100.0, 50.0);
-    let transformed_widget = NewWidget::new_with_options(
-        blue_box(Label::new("Background")),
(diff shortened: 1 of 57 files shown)


masonry/src/tests/transforms.rs
```diff
@@ -34,8 +34,7 @@ fn blue_box(inner: impl Widget) -> impl Widget {
 #[test]
 fn transforms_translation_rotation() {
     let translation = Vec2::new(100.0, 50.0);
-    let transformed_widget = NewWidget::new_with_options(
-        blue_box(Label::new("Background")),
+    let transformed_widget = NewWidget::new(blue_box(Label::new("Background"))).with_options(
```
xStrom, line 37: This `with_options` is wrong, but so are other things in this file because apparently this `transforms` module isn't listed in `mod.rs` 😅

I guess we can merge as-is and deal with this module in a follow-up.

xStrom (approved):
I guess the extra flexibility makes this change worth it.

PR #1714, opened 2026-04-05
Extract masonry_imaging backend adapters

waywardmonkeys:
Add masonry_imaging as the Masonry-owned bridge between paint output and concrete imaging backends.

This change:
- extracts shared Masonry render-source and frame helpers into masonry_imaging
- splits backend integration into image rendering and texture rendering capabilities
- supports imaging_vello, imaging_vello_hybrid, imaging_vello_cpu, and imaging_skia
- updates masonry_testing to render through masonry_imaging
- updates masonry_winit to use masonry_imaging for runtime rendering while keeping window and surface management local
- replaces vello::wgpu usage with direct wgpu
- aligns runtime rendering with target-oriented backend renderers where available

The result is a cleaner boundary:
- masonry_core owns paint output
- masonry_imaging owns Masonry-to-backend rendering integration
- host crates such as masonry_winit own presentation and platform integration

### DIFF
--- a/masonry_imaging/README.md
+++ b/masonry_imaging/README.md
@@ -1,0 +1,11 @@
+# Masonry Imaging
+
+Masonry-owned imaging helpers and backend adapters.
+
+This crate currently provides:
+
+- shared helpers for preparing retained scenes from Masonry-style base content plus overlays
+- a `vello` module backed by `imaging_vello`
+- a `vello_hybrid` module backed by `imaging_vello_hybrid`
+- a `vello_cpu` module backed by `imaging_vello_cpu`
+- a `skia` module backed by `imaging_skia`
--- a/masonry_imaging/src/lib.rs
+++ b/masonry_imaging/src/lib.rs
@@ -1,0 +1,354 @@
+// Copyright 2026 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/linebender/xilem/main/docs/assets/masonry-logo.svg"
+)]
+//! Imaging helpers owned by Masonry.
+//!
+//! `masonry_imaging` owns the bridge between Masonry paint output and concrete imaging backends.
+//! In this first slice that means:
+//!
+//! - preparing flattened Masonry frame content from base content plus overlays
+//! - exposing backend-specific renderer modules for `imaging_vello`,
+//!   `imaging_vello_hybrid`, `imaging_vello_cpu`, and `imaging_skia`
+//! - exposing host-neutral texture rendering helpers for writing into caller-provided WGPU
+//!   targets
+//!
+//! This crate does not own window integration, surfaces, or compositor policy.
+//!
+//! # Feature flags
+//!
+//! - `default`: Enables the `vello` module.
+//! - `imaging_vello`: Enables the `vello` module and texture rendering support.
+//! - `imaging_vello_hybrid`: Enables the `vello_hybrid` module and texture rendering support.
+//! - `imaging_vello_cpu`: Enables the `vello_cpu` module for headless image rendering only.
+//! - `imaging_skia`: Enables the `skia` module and texture rendering support on non-wasm targets.
+
+// LINEBENDER LINT SET - lib.rs - v3
+// See https://linebender.org/wiki/canonical-lints/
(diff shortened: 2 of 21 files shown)


PoignardAzur:
Will have to review more in-depth once I'm not sick, but overall I'm not sure there's much of an interest giving masonry_winit two different imaging backends, especially given the amount of cfg feature plumbing this requires.

I'd rather switch masonry_winit to vello_hybrid, and if someone wants vello classic, well, masonry_winit is small and easy to fork.

waywardmonkeys:
`vello_hybrid` doesn't yet do everything that `vello_cpu` and `vello` can do. It also uses more memory than I expected and I haven't fully tracked this down yet. There's still a bunch of trade-offs for each of the Vello backends.

I've also been running with the Skia backend and it tends to perform well *and* use much less memory. For something else (not Masonry), I've been using `vello_cpu` + `softbuffer` and that's both fast and even smaller in memory.

But the overall point of some of this is that things will be shrinking even further as `masonry_winit` still has room over the coming weeks to shrink as other things happen.

waywardmonkeys:
But yes, the feature plumbing and stuff was the single most annoying thing in doing this really. And CI issues.

masonry_imaging/src/lib.rs
```diff
@@ -0,0 +1,355 @@
...
+
+#[cfg(any(feature = "imaging_vello", feature = "imaging_vello_hybrid"))]
+mod headless_wgpu;
+
+/// Masonry helpers for rendering retained scenes with `imaging_skia`.
+#[cfg(all(feature = "imaging_skia", not(target_arch = "wasm32")))]
+pub mod skia;
+/// Host-neutral texture rendering helpers for texture-capable backends.
+#[path = "runtime.rs"]
+pub mod texture_render;
```
xStrom: What's the story behind this name mismatch? Why couldn't they both be the same? (*I don't even care which one, just curious about the mismatch.*)

masonry_imaging/README.md
```diff
@@ -0,0 +1,9 @@
+# Masonry Imaging
+
+Masonry-owned imaging helpers and backend adapters.
+
+This crate currently provides:
+
+- shared helpers for preparing retained scenes from Masonry-style base content plus overlays
+- a `vello` module backed by `imaging_vello`
+- a `vello_hybrid` module backed by `imaging_vello_hybrid`
```
xStrom, line 9: I understand that this is a temporary `README.md` and that `lib.rs` has the better docs (and we should eventually probably set up `cargo-rdme`). Even so, I think it's worth mentioning just briefly that the other modules already exist as well.

```suggestion
- a `vello` module backed by `imaging_vello`
- a `vello_hybrid` module backed by `imaging_vello_hybrid`
- a `vello_cpu` module backed by `imaging_vello_cpu`
- a `skia` module backed by `imaging_skia`
```

xStrom (commented):
For the sanity of Windows users, we'll need to use pre-built Skia binaries. [imaging#33](https://github.com/forest-rs/imaging/pull/33) as the quickest fix, and we'll hopefully have a better fix later.

Otherwise this generally seems like a really nice step forward. I can successfully run Masonry on Windows 11 with Vello Classic, Vello Hybrid, and Skia.

There are plenty of issues, like Hybrid crashing with `AtlasLimitReached` when resizing the Tiger example in `gallery` but issues like that aren't strictly related to this PR here and can be addressed in follow-up work.

masonry_imaging/src/lib.rs
waywardmonkeys: Leftover from compat stuff ... removing.

masonry_imaging/README.md
waywardmonkeys: Applied locally.

waywardmonkeys:
I'd like to land this ... and find a better way to deal with the features and stuff in follow-ups.

xStrom (approved):
This is a great step forward. I've been successfully running derivatives of this PR locally for a few days now. While there are clearly plenty of things to still solve, I agree that those can be addressed in follow-up PRs.

Let's move towards merging so we can be more focused on the remaining issues. As this PR is already quite large.

xStrom:
Also, for the record, the `AtlasLimitReached` crash issues got solved in [imaging#37](https://github.com/forest-rs/imaging/pull/37) which is now part of this PR here.

PR #1718, opened 2026-04-08
Fix IME composition baseline jitter for multi-line text

Cupnfish:
## Summary

Fixes vertical jitter during IME composition (e.g. CJK input) for multi-line text in `TextArea`.

When composing text via IME, font fallback from Latin to Chinese fonts changes line metrics, causing visible vertical jitter of existing text and the cursor. The previous code only tracked line 0's baseline, so multi-line text on lines 2+ still jittered during composition.

### Changes

- **`masonry_core/src/core/text.rs`**: Add `render_text_with_line_offsets()` which accepts a `Fn(usize) -> f64` closure to apply per-line Y offsets during rendering. `render_text()` now delegates to it with `|_| 0.0`.
- **`masonry/src/widgets/text_area.rs`**: 
  - Save all lines' baselines in `compose_baselines: Option<Vec<f32>>` before composition starts
  - Compute per-line offsets via `compute_line_offsets()` during paint
  - Apply offsets to text rendering, cursor, and selection geometry
  - Use `request_render()` instead of `request_layout()` for preedit to avoid unnecessary relayout during composition

### Testing

Tested manually with Chinese IME input on both single-line and multi-line `TextArea` widgets — text and cursor no longer jitter during composition.

## Notes

This PR was generated with assistance from GLM 5.1 (by Zhipu AI) and has **not yet been reviewed**. A manual review is planned before marking as ready for merge.

### DIFF
--- a/masonry/src/widgets/text_area.rs
+++ b/masonry/src/widgets/text_area.rs
@@ -12,11 +12,13 @@ use crate::core::{
     AccessCtx, AccessEvent, BrushIndex, ChildrenIds, CursorIcon, EventCtx, Ime, LayoutCtx,
     MeasureCtx, PaintCtx, PointerButton, PointerButtonEvent, PointerEvent, PointerUpdate,
     PropertiesMut, PropertiesRef, QueryCtx, RegisterCtx, StyleProperty, TextEvent, Update,
-    UpdateCtx, Widget, WidgetId, WidgetMut, render_text, set_accesskit_brush_properties,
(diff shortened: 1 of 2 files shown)


PoignardAzur:
It seems like you made the same PR twice with two AI agents? Did you intend to compare the two?

Cupnfish:
> It seems like you made the same PR twice with two AI agents? Did you intend to compare the two?

Yes, but after a deeper investigation, I found that modifying the source code of Parley was actually a more suitable approach.

PR #1719, opened 2026-04-08
Change highlighting from INFO to NOTE

rohitaryal:
> [!NOTE]
> `[!NOTE]` is supported in GitHub markdown blockquote

> [!INFO]
> `[!INFO]` is not supported in GitHub markdown blockquote


xStrom:
On closer review, maybe it's worth merging anyway.

xStrom (approved):

PoignardAzur:
Thanks! It indeed looks better.

PR #1720, opened 2026-04-10
fix: correct copy-paste error in Svg view error message

hobostay:
## Summary

The error message in `Svg::message` (`xilem_masonry/src/view/svg.rs:109`) incorrectly said "Image doesn't consume any messages" instead of "Svg doesn't consume any messages". This was a copy-paste error introduced when the SVG view was created from the Image view in #1708.

## Test plan

- [x] `cargo check -p xilem_masonry` passes
- [ ] String literal grep confirms all other `*::message` error messages are self-referential

🤖 Generated with [Claude Code](https://claude.com/claude-code)


waywardmonkeys (approved):

waywardmonkeys:
Thank you!

PR #1723, opened 2026-04-12
Visual layer plan

waywardmonkeys:
This PR introduces `VisualLayerPlan` as Masonry’s shared visual-layer model and uses it to carry explicit paint boundaries from `masonry_core` through host presentation.

The main outcome is that Masonry can now represent and preserve an ordered layer sequence like:

1. Masonry scene content below
2. a host-managed external surface slot
3. Masonry scene content above

That gives us a cleaner path toward compositor-driven presentation, instead of flattening everything into one scene and treating external content as a special case.

This PR does **not** claim that the compositor model is finished.

In particular:

* `VisualLayerPlan` is still a flat ordered plan, not a grouped/tree compositor model
* the `subduction` path is an early host integration, not the final cross-platform presentation architecture
* scene-layer caching / frame pacing / platform-native presenter choices re follow-up work


masonry_winit/examples/external_surface_subduction.rs
PoignardAzur, line 1: Let's keep all the examples in `masonry/examples/`.

waywardmonkeys:
This was done with assistance from Codex (gpt-5.4, high). But the overall design is mine. (But probably still evolving ...)

masonry_core/src/passes/paint/tests.rs
PoignardAzur, line 1: This test file should be moved to `masonry/src/tests/`.

There's a few reasons we do a lot of tests in `masonry` instead of `masonry_core`. The main one is that `masonry` defines a lot of widgets that are useful for tests. For instance, `PaintLeaf`, `TripleRow`, `OffsetBox` and `MixedTripleRow` could trivially be replaced by `SizedBox` and `Flex`.

Overall, tests are supposed to be a kind of maintainer-only documentation explaining what we expect from the code. You shouldn't need to read the implementation of four custom widgets to get to read a test.

masonry_core/src/app/tracing_backend.rs
```diff
@@ -15,13 +15,18 @@
 
 use std::error::Error;
 use std::fmt;
+#[cfg(not(target_arch = "wasm32"))]
 use std::fs::File;
+#[cfg(not(target_arch = "wasm32"))]
 use std::time::UNIX_EPOCH;
```
PoignardAzur, line 21: These seem like drive-by changes? I don't think they're needed for this PR.

masonry_core/src/app/render_layers.rs
PoignardAzur, line 1: Alright, so this file defines *breathes in* `VisualLayerId`, `ExternalLayerKind`, `VisualLayerBoundary`, `VisualLayerKind`, `VisualLayer`, `SceneVisualLayer`, `ExternalVisualLayer`, and `VisualLayerPlan`.

This is my problem with these AI-generated PRs. Even if all of the code is correct, they add *a lot* of API surface that you then need to learn and maintain.

Also, general comment: while this isn't documented anywhere (but it should be!), Masonry modules usually adhere to a coding style where all types are defined first, then implementations. This helps identify which types are composed of which other types, etc.

masonry_core/src/app/render_layers.rs
```diff
@@ -0,0 +1,302 @@
...
+///
+/// Scene-backed layers contain retained `imaging` content in the layer's local coordinate space.
+/// External layers preserve a host-managed layer boundary. In both cases, apply
+/// [`transform`](Self::transform) to place the layer in window space.
+pub struct VisualLayer {
+    id: VisualLayerId,
+    /// The content realization of this layer.
+    pub kind: VisualLayerKind,
+    /// Where this visual layer boundary originated.
+    pub boundary: VisualLayerBoundary,
```
PoignardAzur: We've had a discussion about AI-generated doc lately, where Raph posted [a parody of AI doc](https://gist.github.com/raphlinus/0567bb617724a65db24ba7242af4026a). Reading a lot of doc in this file reminds me of Raph's parody; this line is a good example.

When we have a type named `VisualLayer` with a field named `boundary` of type `VisualLayerBoundary`, a doc that says "Its the visual layer boundary" is extremely unhelpful.

masonry_core/src/app/render_layers.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Returns the axis-aligned clip in window coordinates, if any.
+    pub fn window_clip_bounds(&self) -> Option<Rect> {
+        self.clip
+            .map(|clip| self.transform.transform_rect_bbox(clip))
+    }
+}
+
+/// Narrowed view of a scene-backed visual layer.
+#[derive(Clone, Copy, Debug)]
+pub struct SceneVisualLayer<'a> {
```
PoignardAzur: This struct is unused and should be removed.

masonry_core/src/app/render_layers.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Returns the axis-aligned clip in window coordinates, if any.
+    pub fn window_clip_bounds(self) -> Option<Rect> {
+        self.clip
+            .map(|clip| self.transform.transform_rect_bbox(clip))
+    }
+}
+
+/// Narrowed view of an externally realized visual layer.
+#[derive(Clone, Copy, Debug)]
+pub struct ExternalVisualLayer {
```
PoignardAzur: This struct is *technically* used, but the uses could trivially be refactored away. It should be removed.

masonry_core/src/app/render_layers.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn replay_into<S>(&self, sink: &mut S)
+    where
+        S: PaintSink + ?Sized,
+    {
+        for layer in &self.layers {
+            if let VisualLayerKind::Scene(scene) = &layer.kind {
+                replay_transformed(scene, sink, layer.transform);
+            }
+        }
+    }
```
PoignardAzur: If the layer plan has any non-scene layers, this is almost never going to do what the user wants/expects, and so is kind of a footgun.

If the user *does* want to do this, they can trivially implement it themselves. This should be removed.

masonry_core/src/core/layer.rs
```diff
@@ -25,6 +25,34 @@ pub enum LayerType {
...
+pub enum LayerRealization {
+    /// Masonry paints this layer into a retained `imaging` scene.
+    #[default]
+    Scene,
+    /// Masonry preserves this layer boundary for host-managed realization.
+    ///
+    /// This is intended for content such as foreign surfaces, 3D viewports, or
+    /// platform-native compositor layers.
+    External,
+}
```
PoignardAzur, line 39: I'm not sure "realization" is a very expressive name here. I know it's used in computer graphics, but it doesn't convey "either a local description or an external handler".

masonry_core/src/core/layer.rs
```diff
@@ -25,6 +25,34 @@ pub enum LayerType {
...
+    #[default]
+    Inline,
+    /// Paint the subtree into its own retained scene layer while preserving painter order.
+    IsolatedScene,
+    /// Reserve the subtree as a host-managed external layer.
+    ///
+    /// Masonry still traverses the subtree to clear paint invalidation flags, but its retained
+    /// scene output is discarded and the host is expected to realize the layer separately.
+    External,
+}
```
PoignardAzur, line 54: This PR provides two enums (`LayerRealization`, `PaintLayerMode`) which are essentially two ways of doing the same thing (inserting external layers). Yes, I understand they're used in different places, but they still overlap a lot (since layers also implement the `Widget` trait, nothing stops you from implementing an external layer that returns `PaintLayerMode::External`).

This is probably okay for a prototype, but it's not very ergonomic.

masonry_core/src/lib.rs
```diff
@@ -75,7 +75,10 @@
 #![cfg_attr(not(debug_assertions), expect(unused, reason = "Deferred: Noisy"))]
 #![expect(missing_debug_implementations, reason = "Deferred: Noisy")]
 #![expect(clippy::cast_possible_truncation, reason = "Deferred: Noisy")]
-#![expect(clippy::large_enum_variant, reason = "FIXME: NewWidget is too large")]
+#![cfg_attr(
+    not(target_arch = "wasm32"),
+    expect(clippy::large_enum_variant, reason = "FIXME: NewWidget is too large")
+)]
```
PoignardAzur, line 81: This is also a drive-by change. I'd rather you add wasm32 support in a separate PR.

masonry_imaging/src/lib.rs
PoignardAzur, line 1: Another unofficial formatting rule I should document: in general, in Masonry code, we try to keep code outside of `mod.rs` / `lib.rs` files, because these are kind of annoying to search. So ideally these types should be in dedicated files.

masonry_winit/src/app_driver.rs
```diff
@@ -127,6 +184,51 @@ pub trait AppDriver {
...
+    /// real compositor override seam.
+    ///
+    /// Masonry Winit does not realize host-managed layers itself. If the application ignores
+    /// external layers in this plan, those surfaces will be absent from the final presentation.
+    fn on_visual_layers(
+        &mut self,
+        window_id: WindowId,
+        ctx: &mut DriverCtx<'_, '_>,
+        layers: &VisualLayerPlan,
+    ) {
```
PoignardAzur: I'm not super clear on how this is supposed to be used, from reading documentation.

Is the expectation that apps with custom layers will render them to wgpu from this method? Or just update them?

Also, `on_visual_layers` is a confusing name.

masonry_winit/src/subduction_presenter.rs
```diff
@@ -0,0 +1,756 @@
...
+    /// Render the given scene layer into `target.view`.
+    fn render_scene_layer(
+        &mut self,
+        target: SceneLayerTarget<'_>,
+        layer: SceneVisualLayer<'_>,
+    ) -> Result<(), BoxError>;
+}
+
+/// Realize host-owned external layers into the textures allocated by the subduction presenter.
+pub trait ExternalLayerRenderer {
```
PoignardAzur: I think at some point I'm going to put out an official moratorium on adding more traits to the Xilem codebase, because the trait inflation is getting way out of control.

I'm pretty sure these two could just be `impl Fn`s.

masonry_winit/src/subduction_presenter.rs
```diff
@@ -0,0 +1,756 @@
...
+    pub device: &'a wgpu::Device,
+    /// Queue used for uploads and submission.
+    pub queue: &'a wgpu::Queue,
+    /// Texture view owned by the compositor for this layer.
+    pub view: &'a wgpu::TextureView,
+    /// Output size in physical pixels.
+    pub output_size: winit::dpi::PhysicalSize<u32>,
+    /// Window scale factor used to convert logical layer geometry into pixels.
+    pub scale_factor: f64,
+}
```
PoignardAzur: Okay, so *this* struct is exactly the same as `ExternalLayerTarget`, except without the `format` field. I suspect they could have been merged.

masonry_winit/src/subduction_presenter.rs
```diff
@@ -0,0 +1,756 @@
...
+        let state = self
+            .layers
+            .get_mut(&layer_id)
+            .ok_or_else(|| std::io::Error::other("external layer state was not prepared"))?;
+        self.renderer.render_layer(target, layer, state)
+    }
+}
+
+/// Host-specific renderer for external layers with optional per-layer state.
+pub trait ManagedExternalLayerRenderer {
```
PoignardAzur: Oh hey, a third trait!

masonry_winit/src/subduction_presenter.rs
```diff
@@ -0,0 +1,756 @@
...
+    ) -> Result<ExternalLayerRenderResult, BoxError>;
+}
+
+#[derive(Clone, Copy)]
+struct ExternalLayerSlot<Id = VisualLayerId, Layer = ExternalVisualLayer> {
+    id: Id,
+    layer: Layer,
+}
+
+trait ExternalLayerStateDriver<Layer, State> {
```
PoignardAzur: Oh hey, a fourth trait!

PoignardAzur (changes requested):
I'm going to be honest, I gave up halfway through `subduction_presenter.rs`, because at some point I don't even know if I'm giving feedback to Bruce or to some LLM, and this doesn't feel like a good use of my time.

Anyway, **the** guiding principle of Masonry is that we make a lot of effort to keep the code readable for complete beginners. This principle is relaxed for Masonry Winit because platform glue code is inherently messy, but even then, this PR does not remotely meet our standards.

masonry_winit/examples/external_surface_subduction.rs
waywardmonkeys: This one is here for now because parts of it are very specific to `masonry_winit` and with other stuff that I had locally, it was annoying to keep other things building ... but I'll move it before this lands.

masonry_core/src/passes/paint/tests.rs
waywardmonkeys: If we didn't have to implement a bunch of empty methods for every widget, this would've been a lot easier as well.

But the things that it is testing aren't as easily done with SizedBox / Flex given that they make things less easily deterministic ..

But I'll move it after other things are resolved.

masonry_core/src/app/tracing_backend.rs
waywardmonkeys: They were required to fix CI since with this, some stuff started getting compiled for Wasm in the existing CI that wasn't happening before.

They can probably be done in a separate PR at some point.

masonry_core/src/app/render_layers.rs
waywardmonkeys: Some of the are because I used to have support for actual child windows on Windows and macOS before, not just a single external layer. I've removed that though thinking I might add it back, but realistically, I'll probably never actually re-implement that within Masonry.

The problem here with these types isn't that it was AI generated, it is that at various points (over the last more than a week), this PR has been doing a lot of other things along the way that I removed.

Also, some parts of this are actually complex things to model and require some detail.

But the `VisualLayerPlan` is the actually important thing here and where most of the attention has gone.

masonry_core/src/app/render_layers.rs
waywardmonkeys: Would you rather the type be called `Boundary`?

It is fair to call out that the doc comment could use more work.

Or I guess we could mock the work that I'm doing instead of something productive.

masonry_core/src/app/render_layers.rs
waywardmonkeys: Removed and used `VisualLayer` where it was actually used.

masonry_core/src/app/render_layers.rs
waywardmonkeys: Applied the same thing as for `SceneVisualLayer` to remove it and replace with `VisualLayer`.

waywardmonkeys: (There used to be larger differences with them when there were other types of external layers.)

masonry_core/src/app/render_layers.rs
waywardmonkeys: Removed. It was vestigial from an earlier state.

masonry_core/src/core/layer.rs
waywardmonkeys: Okay.

masonry_core/src/lib.rs
waywardmonkeys: It is something that changed as a result of this PR. I didn't add wasm32 support in this PR.

masonry_winit/src/app_driver.rs
waywardmonkeys: It is for letting a host inspect the frame without taking over presentation. I'll just remove it.

waywardmonkeys:
> I'm going to be honest, I gave up halfway through `subduction_presenter.rs`, because at some point I don't even know if I'm giving feedback to Bruce or to some LLM, and this doesn't feel like a good use of my time.

It is fair to give up in parts of that version of the presenter. I've done a massive simplification on it now that I've removed some other things and will probably never re-implement them within Masonry (like OS child windows and a form of native control integration).

masonry_core/src/lib.rs
waywardmonkeys: Ahh, it is because `masonry_imaging` isn't in `NO_WASM_PKGS` in the CI config.

PoignardAzur: Oh yeah. We get trapped by this every time we add a crate. We should probably start rethinking how we do our builds.

waywardmonkeys:
Replaced partially by #1746 and PRs that will follow.

PR #1726, opened 2026-04-12
Refactor example gallery

PoignardAzur:
The new `on_action` method was written manually, everything else was generated by Claude Code, with manual verification.

Also added comment to `slider.rs`.

### DIFF
--- a/masonry/examples/gallery/demo.rs
+++ b/masonry/examples/gallery/demo.rs
@@ -28,7 +28,10 @@ pub(crate) trait DemoPage {
         render_root: &mut RenderRoot,
         action: &ErasedAction,
         widget_id: WidgetId,
-    ) -> Handled;
+    ) -> Handled {
+        #![expect(unused_variables, reason = "Default impl")]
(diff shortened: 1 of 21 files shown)


masonry/examples/gallery/demo.rs
```diff
@@ -23,64 +23,12 @@ pub(crate) trait DemoPage {
...
-        _render_root: &mut RenderRoot,
-        _widget_id: WidgetId,
-        _toggled: bool,
-    ) -> bool {
-        false
-    }
+        render_root: &mut RenderRoot,
+        action: &ErasedAction,
+        widget_id: WidgetId,
+    ) -> Handled;
```
xStrom: How about we add a default implementation of `Handled::No` here instead of having it for a bunch of demos like `Spinner`?

PoignardAzur: Good idea.

xStrom (approved):
This makes things more manageable, nice.

PR #1731, opened 2026-04-13
Fix wasm-target hygiene in masonry_core

waywardmonkeys:
This doesn't yet enable them in CI as that requires more work for tests and the feature matrix.

### DIFF
--- a/masonry_core/Cargo.toml
+++ b/masonry_core/Cargo.toml
@@ -32,13 +32,15 @@ kurbo.workspace = true
 parley.workspace = true
 peniko.workspace = true
 smallvec.workspace = true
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
In #1723, `masonry_imaging` adds a dependency on `masonry_core` and then these fixes for compilation were needed.

xStrom (approved):
We do want to see if we can get Wasm support to be a thing, so I guess this makes sense.

waywardmonkeys:
Yeah, have to change the use of `block_on` in `masonry_winit` and set up font data ...

waywardmonkeys:
And fix feature matrix issues.

PR #1735, opened 2026-04-13
Add Visible built-in property

PoignardAzur:
This property lets you skip a widget and its children from the paint tree.

The test was generated with Claude, then edited a bit.


masonry_core/src/passes/paint.rs
```diff
@@ -138,7 +152,7 @@ fn paint_widget(
     let content_box_to_layer_transform =
         border_box_to_layer_transform.pre_translate(state.border_box_translation());
     let has_clip = state.clip_path.is_some();
-    if !is_stashed {
+    if !is_stashed && visible {
```
waywardmonkeys, line 155: This is after the request_pre_paint / request_paint / request_post_paint above, so invisible widgets are still running paint callbacks and rebuilding caches, I think.

waywardmonkeys (changes requested):
Widgets that aren't visible still participate in hit-testing, I think as well as focus / pointer capture and accessibility.

PoignardAzur:
> Widgets that aren't visible still participate in hit-testing

Oh right, good catch.

PoignardAzur:
On second thought, I don't feel like dealing with all the knock-on effects of adding this right now.

PR #1736, opened 2026-04-13
Box NewWidget in cases in some enums

PoignardAzur:


xStrom (approved):

waywardmonkeys:
This does add new allocations though, so making sure that this doesn't impact things that call `WidgetPod::to_pod` or `FlexElement::Child` construction would be good.

PoignardAzur:
We already do a lot of allocations per-widget, so I don't think this specific change will add much overhead.

I'm planning to do a lot more benchmarking in the future though, because it's not great to rely on intuition.

PR #1737, opened 2026-04-13
Rename `Widget::with_auto_id()` to `prepare()`

PoignardAzur:
`Widget::with_auto_id()` was an unclear name. This changes the name to `prepare()`, which hints at the "builder pattern" nature of `NewWidget`.

Remove `Widget::with_props()`, and replace it with `my_widget.prepare().with_props()` to have only one set of builder methods.

### DIFF
--- a/masonry/README.md
+++ b/masonry/README.md
@@ -107,7 +107,7 @@ impl AppDriver for Driver {
                 TextArea::reset_text(&mut text_area, "");
             });
             render_root.edit_widget_with_tag(LIST_TAG, |mut list| {
-                let child = Label::new(self.next_task.clone()).with_auto_id();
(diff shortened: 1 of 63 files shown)


xStrom (approved):
```rust
// I like that this is now shorter.
w.with_auto_id()
w.prepare()

// Not a fan of this being now longer,
// but I understand wanting to have only one set of builder methods.
w.with_props()
w.prepare().with_props()
```

This approach has its pros and cons, but probably not worth bikeshedding over it much.

waywardmonkeys (commented):
This needs an explanation or something in the commit message or the PR description ... anywhere.

PoignardAzur:
> This needs an explanation or something in the commit message or the PR description ... anywhere.

I've fleshed it out a bit.

PR #1744, opened 2026-04-15
Tweak `NewWidget::with_props` to add props instead of overwriting them

PoignardAzur:
This removes a potential footgun where calling `new_widget.with_props(A).with_props(B)` would get the propset `(B)` instead of `(A, B)`.

### DIFF
--- a/masonry_core/src/core/widget_pod.rs
+++ b/masonry_core/src/core/widget_pod.rs
@@ -154,6 +154,9 @@ impl<W: Widget + ?Sized> NewWidget<W> {
 
     /// Applies the given [properties] to this widget.
     ///
+    /// Calling this method multiple times will merge the properties together,
+    /// with later properties taking precedence over earlier ones with the same type.
+    ///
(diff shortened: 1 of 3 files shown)


masonry_core/src/core/widget_pod.rs
```diff
@@ -128,9 +128,18 @@ impl<W: Widget + ?Sized> NewWidget<W> {
         self
     }
 
-    /// Sets the [`PropertySet`] for this widget.
+    /// Applies the given [properties] to this widget.
+    ///
+    /// [properties]: crate::doc::masonry_concepts#properties
```
xStrom, line 133: This comment should more clearly state that the method will extend any existing properties.

xStrom (commented):
This would greatly benefit from a simple test that does `.with_props(A1,B).with_props(A2,C)` and confirms the result is `A2,B,C`.

PoignardAzur:
Added a short test.

xStrom (approved):

PR #1745, opened 2026-04-16
Remove Vello image override support

waywardmonkeys:
Image overrides were a Vello-only workaround that Masonry does not use and does not want to encourage downstream use of.

Keeping them around had two costs:

* they exposed a public driver API for a backend-specific escape hatch
* they forced `masonry_imaging` to keep a separate Vello texture-render path that Skia, Vello Hybrid, and Vello CPU could not share

Remove that support end-to-end:

* drop the `DriverCtx` image-override API
* remove the event-loop forwarding
* remove the Vello renderer override state

With that workaround gone, the Vello texture path can use the same shared `TextureRenderer`-based flow as the other backends. `masonry_imaging` no longer needs a direct `vello` dependency or a special native-scene helper for this case.

This simplifies the runtime rendering code and keeps the architecture pointed at the real replacement: proper layer/external-texture handling rather than a backend-specific image substitution hook.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2460,7 +2460,6 @@ dependencies = [
  "imaging_vello_hybrid",
  "kurbo",
  "peniko",
- "vello",
  "wgpu",
 ]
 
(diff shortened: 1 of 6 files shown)


xStrom (approved):
Makes sense and significantly simplifies the code. Thanks!

PR #1746, opened 2026-04-16
masonry_core: introduce VisualLayerPlan and scene-layer paint splits

waywardmonkeys:
This PR introduces `VisualLayerPlan` as the paint output of `masonry_core` and adds the first in-tree scene-layer split mechanism.

The scope is intentionally narrow:

* core-only
* no host or `masonry_winit` presentation changes
* no `ExternalSurface`
* no Subduction integration
* no behavior change in current rendering paths

Current consumers still flatten the plan back into the same rendered output they had before.

What changed

### 1. Replace `PaintResult` with `VisualLayerPlan`

`RenderRoot::redraw()` now returns a `VisualLayerPlan` instead of the old `PaintResult` / `PaintedLayer` split.

`VisualLayerPlan` is now the semantic paint output of Masonry:

* painter-ordered visual layers
* each layer carries a retained scene plus its transform into window space
* current consumers use compatibility helpers to replay/flatten those layers

This changes the representation, but not runtime behavior.

### 2. Add scene-only paint splits inside a root

This PR adds a small paint-pass seam:

* `PaintLayerMode`
* `PaintCtx::set_paint_layer_mode(...)`

For now it supports:

* `Inline`
* `IsolatedScene`

`PaintLayerMode::IsolatedScene` records the current widget subtree as its own scene layer in the `VisualLayerPlan`.

Current hosts still flatten those layers back together, so this does **not** change final rendering yet. What it changes is that the core paint output can now represent in-tree scene boundaries explicitly, not just top-level `LayerStack` roots.

## Why this is split out first

The goal of this PR is to establish the core paint model before introducing:

* external layers
* widget-level external surface APIs
* host-side realization
* Subduction integration

That gives us a cleaner sequence:

1. core can describe visual layering
2. core can describe in-tree scene boundaries
3. later work can add external placeholders
4. later still, hosts can realize the plan differently

This PR only handles the first two steps.

## Compatibility / behavior

There should be no user-visible behavior change.


masonry_core/src/app/render_root.rs
```diff
@@ -15,6 +15,7 @@ use parley::{FontContext, LayoutContext};
 use tracing::{debug, info_span, warn};
 use tree_arena::{ArenaMut, TreeArena};
 
+use super::VisualLayerPlan;
```
xStrom: As noted in `ARCHITECTURE.md` we don't do `super` imports outside of tests, so this import should be updated.

masonry/src/tests/paint.rs
```diff
@@ -1,17 +1,25 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::Arc;
+
+use masonry_core::app::{RenderRoot, RenderRootOptions, WindowSizePolicy};
+use masonry_core::dpi::PhysicalSize;
+use masonry_core::peniko::Blob;
```
xStrom: For consistency these should use the `crate::` prefix like the other imports.

masonry_core/src/core/contexts.rs
```diff
@@ -2229,6 +2229,16 @@ impl Drop for RawCtx<'_> {
 
 // --- MARK: DEBUG PAINT
 impl PaintCtx<'_> {
+    /// Controls how this widget subtree is recorded in the current paint pass.
+    ///
+    /// This is reset to [`PaintLayerMode::Inline`](crate::core::PaintLayerMode::Inline)
+    /// at the start of each paint pass for the widget. Widgets that want isolated
+    /// scene layers should set this during `pre_paint`, `paint`, or `post_paint`
+    /// each time they paint.
+    pub fn set_paint_layer_mode(&mut self, mode: crate::core::PaintLayerMode) {
```
xStrom: Let's import `PaintLayerMode` instead of using the fully qualified name.

```suggestion
    pub fn set_paint_layer_mode(&mut self, mode: PaintLayerMode) {
```

masonry_core/src/passes/paint.rs
```diff
@@ -8,57 +8,53 @@ use peniko::{Color, Fill};
...
+    fn scene_mut(&mut self) -> &mut Scene {
+        &mut self.current_scene
+    }
+
+    fn finish_current_layer(&mut self, allow_empty: bool) {
+        if !allow_empty && self.current_scene == Scene::new() {
+            return;
         }
+
+        let scene = std::mem::replace(&mut self.current_scene, Scene::new());
```
xStrom: Checking for an empty scene like this seems inefficient. I guess `Scene::is_empty` doesn't exist?

Also, given that we'll create another empty scene for the mem replace, perhaps we could only create one empty scene and then use that for both the comparison and replacement?

masonry_core/src/app/visual_layers.rs
```diff
@@ -0,0 +1,112 @@
+// Copyright 2026 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::core::WidgetId;
+use crate::imaging::PaintSink;
+use crate::imaging::record::{Scene, replay_transformed};
+use kurbo::Affine;
```
xStrom: StdExtCrate group ordering.

```suggestion
use kurbo::Affine;

use crate::core::WidgetId;
use crate::imaging::PaintSink;
use crate::imaging::record::{Scene, replay_transformed};
```

masonry_core/src/app/visual_layers.rs
```diff
@@ -0,0 +1,112 @@
...
+            layers: vec![
+                VisualLayer {
+                    scene: root_scene.clone(),
+                    transform: Affine::IDENTITY,
+                    root_id: crate::core::WidgetId::next(),
+                },
+                VisualLayer {
+                    scene: overlay_scene.clone(),
+                    transform: Affine::translate((20.0, 5.0)),
+                    root_id: crate::core::WidgetId::next(),
```
xStrom: Let's import `WidgetId` instead of fully qualifying it.

```suggestion
                    root_id: WidgetId::next(),
                },
                VisualLayer {
                    scene: overlay_scene.clone(),
                    transform: Affine::translate((20.0, 5.0)),
                    root_id: WidgetId::next(),
```

xStrom (commented):
Looks like good step overall. I have mostly import related nits.

xStrom (approved):
Approving with the hope that the code style will be adjusted before merging.

masonry_core/src/passes/paint.rs
waywardmonkeys:
```
-        if !allow_empty && self.current_scene == Scene::new() {
+        let empty_scene = Scene::new();
+        if !allow_empty && self.current_scene == empty_scene {
             return;
         }

-        let scene = std::mem::replace(&mut self.current_scene, Scene::new());
+        let scene = std::mem::replace(&mut self.current_scene, empty_scene);
```

And I'll add a `Scene::is_empty()` upstream.

masonry/src/tests/paint.rs
```diff
@@ -150,6 +156,71 @@ fn paint_clipping() {
...
+        RenderRootOptions {
+            default_properties: Arc::new(test_property_set()),
+            use_system_fonts: false,
+            size_policy: WindowSizePolicy::User,
+            size: PhysicalSize::new(40, 20),
+            scale_factor: 1.0,
+            test_font: Some(test_font),
+        },
+    )
+}
```
PoignardAzur, line 198: We can probably skip this if we add a TestHarness method that returns the layer plan without rasterizing it. Putting that on my todo-list.

masonry_core/src/app/visual_layers.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// Snapshot of Masonry's current visual layers in painter order.
+///
+/// This is the semantic paint output of Masonry. Current hosts still consume it through
+/// compatibility helpers that flatten or reinterpret it, but the plan is the source of truth.
+#[derive(Debug)]
+pub struct VisualLayerPlan {
+    /// Layers in painter order, back to front.
+    pub layers: Vec<VisualLayer>,
+}
```
PoignardAzur, line 18: I like that this isn't exported from the `passes` mod anymore.

masonry_core/src/app/visual_layers.rs
```diff
@@ -0,0 +1,114 @@
...
+/// The retained scene is stored in layer-local coordinates. Apply [`transform`](Self::transform)
+/// to composite it into window space. The root layer uses the identity transform.
+#[derive(Debug)]
+pub struct VisualLayer {
+    /// The retained `imaging` scene for this layer in layer-local coordinates.
+    pub scene: Scene,
+    /// Transform from layer-local space to window space.
+    pub transform: Affine,
+    /// The root widget that owns this layer.
+    pub root_id: WidgetId,
```
PoignardAzur, line 62: Is there a specific reason this field is included? As far as I can tell, it's not read anywhere.

masonry_core/src/core/contexts.rs
```diff
@@ -2229,6 +2230,15 @@ impl Drop for RawCtx<'_> {
...
 // --- MARK: DEBUG PAINT
 impl PaintCtx<'_> {
+    /// Controls how this widget subtree is recorded in the current paint pass.
+    ///
+    /// This is reset to [`PaintLayerMode::Inline`] at the start of each paint
+    /// pass for the widget. Widgets that want isolated scene layers should set
+    /// this during `pre_paint`, `paint`, or `post_paint` each time they paint.
+    pub fn set_paint_layer_mode(&mut self, mode: PaintLayerMode) {
+        self.widget_state.paint_layer_mode = mode;
+    }
```
PoignardAzur, line 2240: I like that this isn't a Widget method anymore. Having to reset it during every paint pass is probably sensible too.

masonry_core/src/passes/paint.rs
```diff
@@ -219,13 +226,28 @@ fn paint_widget(
...
+            Painter::new(layer_collector.scene_mut())
+                .fill(rect, HOVER_FILL_COLOR)
+                .transform(border_box_to_layer_transform)
+                .draw();
+        }
+    }
+
+    if isolated_scene {
+        layer_collector.finish_current_layer(false);
     }
```
PoignardAzur, line 245: Why is this called twice in the same function? Does that mean that painting an isolated scene explicitly splits the containing layer into "before" and "after" layers?

masonry_core/src/passes/paint.rs
```diff
@@ -8,57 +8,54 @@ use peniko::{Color, Fill};
...
-    /// Apply this transform when compositing the layer's scene into window space.
-    pub transform: Affine,
-    /// The root widget ID of this layer.
-    pub root_id: WidgetId,
+struct LayerCollector {
+    current_scene: Scene,
+    layers: Vec<VisualLayer>,
+    root_id: WidgetId,
+    transform: Affine,
 }
```
PoignardAzur, line 26: I suspect you could remove this struct, skip the middleman and work directly on a VisualLayerPlan.

PoignardAzur (commented):
Overall I like this PR, and I think it pretty well-structured. I'm not quite sure what the broader plan is, though.

masonry_core/src/app/visual_layers.rs
waywardmonkeys: Metadata for now, useful in the near future when we're doing integration into the host stuff (`masonry_winit`, etc).

masonry_core/src/passes/paint.rs
waywardmonkeys: Yes.

This is more important after one of the upcoming PRs which adds external surface stuff.

masonry_core/src/passes/paint.rs
waywardmonkeys: It would be worse. (VLP would have to carry some builder data.)

But anyway, wait a bit until more pieces land ...

PR #1749, opened 2026-04-17
Fix `clippy::collapsible_match` lints

waywardmonkeys:
These show up now when running Rust 1.95.


xStrom (approved):
Not so sure about this lint's value, but the changes seem correct for the lint.

PoignardAzur:
Yeah, I don't like these types of lints too much.

waywardmonkeys:
It is on by default in 1.95 due to the new / improved features in that area. I'll submit the CI update for 1.95 and typos later today (need dinner).

PR #1751, opened 2026-04-17
masonry_imaging: update for new imaging renderer traits

waywardmonkeys:
Update Masonry's imaging integration to match the new upstream renderer trait split and backend types.

This updates the workspace imaging dependency revision, adds the new `imaging_wgpu` dependency, and switches Masonry to the revised renderer APIs exposed by upstream:

- re-export `TextureRenderer` from `imaging_wgpu` instead of `imaging`
- update the no-backend stub to the new `ImageRenderer` contract
- use the new Vello, Vello Hybrid, and Skia texture renderer types and target wrappers
- route texture rendering through the new shared `TextureRendererError` surface

It also folds in the small Masonry-local simplification enabled by the new traits: `PreparedFrame` now implements `RenderSource` directly, so `texture_render` can render frames without a separate `WindowSource` wrapper.

This keeps the external behavior the same while reducing adapter surface inside `masonry_imaging`.

### DIFF
--- a/masonry_imaging/src/lib.rs
+++ b/masonry_imaging/src/lib.rs
@@ -39,35 +39,41 @@ use imaging::{PaintSink, Painter};
 use kurbo::{Affine, Rect};
 use peniko::Color;
 
-#[cfg(any(feature = "imaging_vello", feature = "imaging_vello_hybrid"))]
+#[cfg(all(
+    not(target_arch = "wasm32"),
+    any(feature = "imaging_vello", feature = "imaging_vello_hybrid")
+))]
 mod headless_wgpu;
 
 /// Masonry helpers for rendering retained scenes with `imaging_skia`.
 #[cfg(all(feature = "imaging_skia", not(target_arch = "wasm32")))]
 pub mod skia;
 /// Host-neutral texture rendering helpers for texture-capable backends.
+#[cfg(not(target_arch = "wasm32"))]
 pub mod texture_render;
 /// Masonry helpers for rendering retained scenes with `imaging_vello`.
-#[cfg(feature = "imaging_vello")]
+#[cfg(all(not(target_arch = "wasm32"), feature = "imaging_vello"))]
 pub mod vello;
 /// Masonry helpers for rendering retained scenes with `imaging_vello_cpu`.
 #[cfg(feature = "imaging_vello_cpu")]
 pub mod vello_cpu;
 /// Masonry helpers for rendering retained scenes with `imaging_vello_hybrid`.
-#[cfg(feature = "imaging_vello_hybrid")]
+#[cfg(all(not(target_arch = "wasm32"), feature = "imaging_vello_hybrid"))]
 pub mod vello_hybrid;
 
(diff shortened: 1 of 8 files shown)


masonry_imaging/src/lib.rs
```diff
@@ -39,35 +39,42 @@ use imaging::{PaintSink, Painter};
...
 #[cfg(feature = "imaging_vello_cpu")]
 pub mod vello_cpu;
 /// Masonry helpers for rendering retained scenes with `imaging_vello_hybrid`.
-#[cfg(feature = "imaging_vello_hybrid")]
+#[cfg(all(not(target_arch = "wasm32"), feature = "imaging_vello_hybrid"))]
 pub mod vello_hybrid;
 
-pub use imaging::render::{ImageRenderer, TextureRenderer};
+pub use imaging::render::ImageRenderer;
+/// Re-export of Imaging's WGPU texture-renderer trait after the upstream `imaging_wgpu` split.
```
xStrom, line 65: This comment doesn't seem valuable next to this import.

```suggestion
```

masonry_imaging/src/lib.rs
```diff
@@ -98,7 +106,10 @@ pub mod image_render {
         all(feature = "imaging_skia", not(target_arch = "wasm32"))
     )))]
     mod no_backend {
-        use imaging::{RgbaImage, render::RenderSource};
+        use imaging::render::{
+            ImageBufferFormat, ImageBufferTarget, ImageRendererError, RenderSource,
+        };
+        use std::vec::Vec;
```
xStrom, line 112: `Vec` is already in `std::prelude` and doesn't need to be imported manually.

```suggestion
```

masonry_imaging/src/skia.rs
```diff
@@ -24,17 +24,17 @@ impl std::error::Error for Error {}
 pub const BACKEND_NAME: &str = "imaging_skia";
 
 /// Masonry alias for the selected Skia image renderer type.
-pub type Renderer = imaging_skia::SkiaRenderer;
+pub type Renderer = imaging_skia::SkiaCpuRenderer;
```
xStrom, line 27: Was this GPU before and is now switched to CPU or what is going on here?

xStrom (commented):
There's some weird regression with the Skia backend. When I start resizing the window, it switches to a white rectangle at the size of the initial window, regardless of the current window size at the start of resize.

The white rectangle remains visible for a variable amount of time (up to 100ms) and then gets replaced with a proper render at the correct size. However, this means that as I'm resizing the window, there's a constant light show, with every window resize event showing the white rectangle for ~100ms.

This behavior does not exist on `main`.

Image

masonry_imaging/src/skia.rs
waywardmonkeys: The Skia CPU (raster) was `SkiaRenderer` before, `SkiaCpuRenderer` now.

xStrom (approved):
Now with [imaging#44](https://github.com/forest-rs/imaging/pull/44) landed the Skia regression has been fixed. So we can move forward with this update here.

Thanks!

PR #1752, opened 2026-04-17
Add external placeholders to `VisualLayerPlan`

waywardmonkeys:
Extend the core paint model so `VisualLayerPlan` can represent external boundaries in painter order.

`PaintLayerMode::External` now emits an external placeholder layer between surrounding scene layers, while current compatibility consumers in `masonry_testing` and `masonry_winit` continue to flatten only the scene layers. This keeps runtime rendering unchanged while making the core plan able to describe `scene -> external -> scene` structure.

The new tests cover both sides of this step:
- external placeholders appear in the plan in painter order
- existing scene-only rendering compatibility still works

### DIFF
--- a/masonry/src/tests/paint.rs
+++ b/masonry/src/tests/paint.rs
@@ -181,6 +181,39 @@ fn make_layer_split_tree(isolate_trailing_box: bool) -> NewWidget<impl Widget> {
         .prepare()
 }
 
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
This is another step down the path ...

xStrom (approved):
Looks like a plausible step.

PR #1754, opened 2026-04-18
Edit tree_arena code to better unify safe and unsafe version

PoignardAzur:
Move some code blocks around so the order is the same.
Make the doc comments identical in both versions.
Fix some errors in the signatures and trait impls.

### DIFF
--- a/tree_arena/src/tree_arena_unsafe.rs
+++ b/tree_arena/src/tree_arena_unsafe.rs
@@ -296,6 +296,10 @@ impl<T> TreeArena<T> {
     ///
     /// Returns a mutable reference to the item if present.
     ///
+    /// # Complexity
+    ///
+    /// - O(Depth) in the safe implementation.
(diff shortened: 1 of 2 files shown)


tree_arena/src/tree_arena_unsafe.rs
```diff
@@ -272,20 +275,27 @@ impl<T> TreeArena<T> {
...
+    /// - O(1) in the unsafe implementation.
     pub fn find(&self, id: impl Into<NodeId>) -> Option<ArenaRef<'_, T>> {
         self.data_map.find_inner(id.into())
     }
 
     /// Finds an item in the tree.
     ///
     /// Returns a mutable reference to the item if present.
+    ///
     pub fn find_mut(&mut self, id: impl Into<NodeId>) -> Option<ArenaMut<'_, T>> {
```
xStrom, line 303: Missing complexity section?

xStrom (approved):

PR #1755, opened 2026-04-18
Improve tree_arena's safety documentation

PoignardAzur:
See [#masonry > Refactoring tree arena for disjoint borrows](https://xi.zulipchat.com/#narrow/channel/317477-masonry/topic/Refactoring.20tree.20arena.20for.20disjoint.20borrows/with/586266467) for rationale.

Note that this PR neither changes the behavior of tree_arena_unsafe nor the conditions under which it's sound.

The PR makes `find_mut_inner` unsafe, but I'd argue it was already unsafe in the sense that calling it with the wrong arguments would lead to UB. This was acceptable because the method was private and the rest of the code never called it with wrong arguments, but I would argue that marking the method as unsafe is more accurate and helpful.

Similarly, the PR points out the [box invalidation problem](https://github.com/Storyyeller/stable_deref_trait/issues/15), but that problem already exists in our current implementation.

### DIFF
--- a/tree_arena/src/tree_arena_unsafe.rs
+++ b/tree_arena/src/tree_arena_unsafe.rs
@@ -425,21 +425,18 @@ impl<'arena, T> ArenaRefList<'arena, T> {
     /// O(depth) and the limiting factor for find methods
     /// not from the root
     fn is_descendant(&self, id: NodeId) -> bool {
-        if self.parent_arena.items.contains_key(&id) {
-            // the id of the parent
-            let parent_id = self.parent_id;
-
-            // The arena is derived from the root, and the id is in the tree
-            if parent_id.is_none() {
-                return true;
-            }
-
-            // iff the path is empty, there is no path from id to self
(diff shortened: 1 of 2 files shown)


tree_arena/src/tree_arena_unsafe.rs
```diff
@@ -451,11 +383,50 @@ impl<'arena, T> ArenaRefList<'arena, T> {
...
+
+        // SAFETY
+        //
+        // `node_cell.get().as_ref()` creates a shared borrow of the node
+        // with a lifetime of 'arena.
+        //
+        // This is sound because `self.arena` is a `&DataMap<T>`, and the only other code
+        // path that can access `node_cell` is `ArenaMutList::find_mut_inner`.
+        // `ArenaMutList::parent_arena` is a `&mut DataMap<T>`, which means that an
+        // `ArenaRefList` can't be constructed as long as an `ArenaMutList` is live.
```
PoignardAzur, line 421: Come to think of it, this is already inaccurate because of `ArenaMutList::reborrow`.

Uuuugh.

PR #1760, opened 2026-04-19
Add builder methods to TestHarnessParams

PoignardAzur:
Part of this code was generated with Claude (though I mostly rewrote the output).

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -268,6 +268,57 @@ impl Default for TestHarnessParams {
     }
 }
 
+impl TestHarnessParams {
+    /// Returns parameters for a test harness with custom dimensions.
+    pub fn size_and_padding(window_size: impl Into<Size>, root_padding: u32) -> Self {
+        Self {
+            window_size: window_size.into(),
+            root_padding,
+            ..Self::DEFAULT
+        }
+    }
+
+    /// Builder method to set `window_size`.
+    pub fn with_size(self, window_size: impl Into<Size>) -> Self {
+        Self {
+            window_size: window_size.into(),
+            ..self
+        }
+    }
+
+    /// Builder method to set `background_color`.
+    pub const fn with_background(self, background_color: Color) -> Self {
+        Self {
+            background_color,
+            ..self
+        }
+    }
+
+    /// Builder method to set `root_padding`.
+    pub const fn with_padding(self, root_padding: u32) -> Self {
+        Self {
+            root_padding,
+            ..self
+        }
+    }
+
+    /// Builder method to set `scale_factor`.
+    pub const fn with_scale(self, scale_factor: f64) -> Self {
+        Self {
+            scale_factor,
+            ..self
+        }
+    }
+
+    /// Builder method to set `max_screenshot_size`.
+    pub const fn with_max_screenshot_size(self, max_screenshot_size: u32) -> Self {
+        Self {
+            max_screenshot_size,
+            ..self
+        }
+    }
+}
+
 /// Roboto is our current test font.
 pub const ROBOTO: &[u8] = include_bytes!(concat!(
     env!("CARGO_MANIFEST_DIR"),
--- a/masonry/examples/custom_widget.rs
+++ b/masonry/examples/custom_widget.rs
@@ -236,9 +236,9 @@ mod tests {
     fn screenshot_test() {
         let my_string = "Masonry + Vello".to_string();
 
-        let mut test_params = TestHarnessParams::default();
         // This is a screenshot of an example, so it being slightly larger than a normal test is expected.
-        test_params.max_screenshot_size = 16 * TestHarnessParams::KIBIBYTE;
+        let test_params =
+            TestHarnessParams::default().with_max_screenshot_size(16 * TestHarnessParams::KIBIBYTE);
 
         let mut harness = TestHarness::create_with(
             default_property_set(),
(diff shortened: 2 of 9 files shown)


masonry_testing/src/harness.rs
```diff
@@ -268,6 +268,57 @@ impl Default for TestHarnessParams {
...
+            ..Self::DEFAULT
+        }
+    }
+
+    /// Builder method to set `window_size`.
+    pub fn with_size(self, window_size: impl Into<Size>) -> Self {
+        Self {
+            window_size: window_size.into(),
+            ..self
+        }
```
xStrom, line 286: Interesting choice to allocate a new object instead of mutating the existing one. I don't think we do this anywhere else. I guess for testing code the performance difference doesn't matter too much.

xStrom (approved):

masonry_testing/src/harness.rs
PoignardAzur: "Allocate" seems like an odd choice of words? None of this code will even call malloc, it's all stack variables.

In the general case, I'm 99% sure that

```rust
let new_foo = Foo {
  field: some_value(),
  ..old_foo
};
old_foo
```

and

```rust
old_foo.field = some_value();
old_foo
```

will almost always generate the same machine code. The MIR representation might be slightly different (we're creating an intermediary variable in one case), but once it gets to codegen, the two snippets do the exact same thing and should compile to the same code.

xStrom: [Stack allocation](https://en.wikipedia.org/wiki/Stack-based_memory_allocation) is standard vocabulary, is it not?

Anyway, I played around with [godbolt](https://godbolt.org/) to see how it compiles.

<details>
<summary>Test source</summary>

```rust

use std::hint::black_box;

#[derive(Clone, Copy)]
pub struct Size(pub u64);

#[derive(Clone, Copy)]
pub struct Big {
    pub window_size: Size,
    pub a: u64,
    pub b: u64,
    pub c: u64,
    pub d: u64,
    pub e: u64,
    pub f: u64,
    pub g: u64,
}

impl Big {
    #[inline(never)]
    #[unsafe(no_mangle)]
    pub fn func_a(mut self, window_size: Size) -> Self {
        self.window_size = window_size;
        self
    }

    #[inline(never)]
    #[unsafe(no_mangle)]
    pub fn func_b(self, window_size: Size) -> Self {
        Self {
            window_size,
            ..self
        }
    }
}

#[unsafe(no_mangle)]
pub fn call_a(x: Big, w: Size) -> Big {
    black_box(x).func_a(black_box(w))
}

#[unsafe(no_mangle)]
pub fn call_b(x: Big, w: Size) -> Big {
    black_box(x).func_b(black_box(w))
}
```
</details>

<details>
<summary>-C opt-level=0 -C codegen-units=1 -C target-cpu=x86-64-v3</summary>

```asm
func_a:
        sub     rsp, 24
        mov     rax, rdi
        mov     qword ptr [rsp + 8], rax
        mov     qword ptr [rsp + 16], rdx
        mov     qword ptr [rsi], rdx
        mov     edx, 64
        call    memcpy@PLT
        mov     rax, qword ptr [rsp + 8]
        add     rsp, 24
        ret

func_b:
        mov     rax, rdi
        mov     qword ptr [rsp - 8], rdx
        mov     qword ptr [rdi], rdx
        mov     rcx, qword ptr [rsi + 8]
        mov     qword ptr [rdi + 8], rcx
        mov     rcx, qword ptr [rsi + 16]
        mov     qword ptr [rdi + 16], rcx
        mov     rcx, qword ptr [rsi + 24]
        mov     qword ptr [rdi + 24], rcx
        mov     rcx, qword ptr [rsi + 32]
        mov     qword ptr [rdi + 32], rcx
        mov     rcx, qword ptr [rsi + 40]
        mov     qword ptr [rdi + 40], rcx
        mov     rcx, qword ptr [rsi + 48]
        mov     qword ptr [rdi + 48], rcx
        mov     rcx, qword ptr [rsi + 56]
        mov     qword ptr [rdi + 56], rcx
        ret
```

</details>

<details>
<summary>-C opt-level=3 -C codegen-units=1 -C target-cpu=x86-64-v3</summary>

```asm
func_a:
        mov     rax, rdi
        vmovups ymm0, ymmword ptr [rsi + 32]
        vmovups ymmword ptr [rdi + 32], ymm0
        mov     qword ptr [rsi], rdx
        mov     rcx, qword ptr [rsi]
        mov     qword ptr [rdi], rcx
        vmovups xmm0, xmmword ptr [rsi + 8]
        vmovups xmmword ptr [rdi + 8], xmm0
        mov     rcx, qword ptr [rsi + 24]
        mov     qword ptr [rdi + 24], rcx
        vzeroupper
        ret

func_b:
        mov     rax, rdi
        mov     qword ptr [rdi], rdx
        vmovups ymm0, ymmword ptr [rsi + 8]
        vmovups ymmword ptr [rdi + 8], ymm0
        vmovups xmm0, xmmword ptr [rsi + 40]
        vmovups xmmword ptr [rdi + 40], xmm0
        mov     rcx, qword ptr [rsi + 56]
        mov     qword ptr [rdi + 56], rcx
        vzeroupper
        ret
```

</details>

They do not result in the same code with either O0 or O3. Hard to say which one is actually faster with O0 as the code is too different, but interestingly in the O3 case looks like the `Self { ...self }` approach actually ends up in faster code.

PR #1764, opened 2026-04-19
Rename HasProperty to UsesProperty

PoignardAzur:
Make it a sub-trait of Widget.

### DIFF
--- a/masonry/src/layers/selector_menu.rs
+++ b/masonry/src/layers/selector_menu.rs
@@ -5,9 +5,9 @@ use accesskit::{Node, Role};
 use tracing::{Span, trace_span};
 
 use crate::core::{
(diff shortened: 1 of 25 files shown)


xStrom (approved):
I can see why `HasProperty` can be misleading, because it might not actually have those properties set. However `UsesProperty` isn't guaranteeing that the widget will actually use the property either, e.g. in the case where the flex factor is used as a property. It's going to be used by the container instead. Anyway, I don't have a desire to bikeshed the name, so whatever.

PoignardAzur:
> UsesProperty isn't guaranteeing that the widget will actually use the property either, e.g. in the case where the flex factor is used as a property

My (vague) plan is that widgets wouldn't implement `UsesProperty<FlexFactor>`, so the distinction is indeed meaningful.

> I can see why HasProperty can be misleading, because it might not actually have those properties set

I think it's misleading in the other direction: a widget might have a property set even if it doesn't implement Has/UsesProperty.

PR #1766, opened 2026-04-22
Add `AppCtx` for shared app-wide services.

xStrom:
For some state it makes sense to have only one copy per app. Until now we haven't had a mechanism to store and provide access to such shared app state. Instead we've hacked around it in various ways. Let's look at two examples.

1. Clipboard access. Deciding what to get from the clipboard, if anything, is a widget level concern. Different widgets might want different formats like plain text, rich text, or images. Some widget might want to implement its own custom in-app copy/paste mechanism for some app data and ignore the platform clipboard. However, the current hack we have is that `masonry_winit` detects Ctrl+V and replaces the key event with a plain text paste event. This does not allow for the aforementioned variety of paste logic. Widgets need to be able to decide what to do with the key event or they might want to access the clipboard even from a pointer event.

2. Font database/cache access. This is currently implemented at the `RenderRoot` level, meaning that font resolution, loading, and caching is duplicated for each root. It works but is needlessly inefficient.

This PR introduces a new `AppCtx` type that encapsulates these kind of app-wide services. It does require a fair amount of plumbing at the framework level. I think the benefit of synchronous mutable access at the widget level makes it worth it.

Because the plumbing changes are already quite noisy I migrated only the clipboard access in this PR. Font context migration would be follow-up work.

### DIFF
--- a/masonry_testing/src/harness.rs
+++ b/masonry_testing/src/harness.rs
@@ -26,6 +26,7 @@ use std::path::PathBuf;
 use std::sync::{Arc, mpsc};
 use std::time::UNIX_EPOCH;
 
+use copypasta::ClipboardProvider;
 use image::{DynamicImage, ImageFormat, ImageReader, Rgba, RgbaImage};
 use imaging_vello_cpu::VelloCpuRenderer;
 use masonry_core::imaging::record::{Scene, replay_transformed};
@@ -36,7 +37,7 @@ use tracing::debug;
 use masonry_core::accesskit::{Action, ActionRequest, Node, Role, Tree, TreeId, TreeUpdate};
 use masonry_core::anymore::AnyDebug;
 use masonry_core::app::{
-    RenderRoot, RenderRootOptions, RenderRootSignal, VisualLayerKind, VisualLayerPlan,
+    AppCtx, RenderRoot, RenderRootOptions, RenderRootSignal, VisualLayerKind, VisualLayerPlan,
     WindowSizePolicy, try_init_test_tracing,
 };
 use masonry_core::core::keyboard::{Code, Key, KeyState, NamedKey};
@@ -144,6 +145,7 @@ pub const PRIMARY_MOUSE: PointerInfo = PointerInfo {
 /// [`insta`]: https://docs.rs/insta/latest/insta/
 #[derive(Debug)]
 pub struct TestHarness<W: Widget> {
+    app_ctx: AppCtx,
     signal_receiver: mpsc::Receiver<RenderRootSignal>,
     render_root: RenderRoot,
     access_tree: accesskit_consumer::Tree,
@@ -158,7 +160,6 @@ pub struct TestHarness<W: Widget> {
     action_queue: VecDeque<(ErasedAction, WidgetId)>,
     has_ime_session: bool,
     ime_rect: (LogicalPosition<f64>, LogicalSize<f64>),
-    clipboard: String,
     title: String,
     _marker: PhantomData<W>,
 }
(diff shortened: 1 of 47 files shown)


PoignardAzur:
I need to review this more in-depth, but some scattered thoughts:

- Having some shared state between RenderRoots probably makes sense.
- I'd like to look for a name other than AppCtx. I think we call too many things across the ecosystem ThingCtx and it's often uninformative.
- Passing this global context to *every* render root method and every pass is a massive amount of plumbing, but I guess it's the only design that makes sense.
- While I can see the point of sharing font data, I'm a little uneasy about injecting a clipboard provider directly into a shared context. I'd like to centralize RenderRoot's inputs and outputs as much as possible; though I take your point that the way we do it is at odd with letting widgets customize their clipboard handling.

xStrom:
Regarding the name of `AppCtx`, if you have a better name we can definitely change it.

I'm not sure I understand your point about centralizing RenderRoot's inputs and outputs. Doesn't having a per-app context do just that? What do you mean by centralizing?

PoignardAzur:
> Regarding the name of AppCtx, if you have a better name we can definitely change it.

Mhhh... maybe `SharedRootState` or `SharedAppState`?

xStrom:
Something worth noting as well is that one of the benefits of `AppCtx` is its length, of both the type name and the variable name. Longer names will more easily cause method declarations to be multi-line.

```rust
app_ctx: &mut AppCtx
rstate: &mut SharedRootState
sas: &mut SharedAppState
```

Of those two you provided, if I had to choose I'd go with `SharedAppState` although I don't prefer it over `AppCtx` due to the verbosity. Especially because it's a type that will be frequently encountered.

Some other options could include going even shorter like `Hub`, `Env`, or `Common`. Which would all be fine for shortness at least.

PoignardAzur:
(Am currently on vacation, will reply in a day or two.)

PoignardAzur:
(Well, "a day or two" was optimistic, but I haven't forgotten about this.)

masonry_testing/src/harness.rs
```diff
@@ -341,6 +371,12 @@ pub const ROBOTO: &[u8] = include_bytes!(concat!(
     "/resources/fonts/roboto/Roboto-Regular.ttf"
 ));
 
+/// Creates a new [`AppCtx`] suitable for testing.
+pub fn create_app_ctx() -> AppCtx {
+    let clipboard = SimpleClipboard::new();
+    AppCtx::new(Box::new(clipboard))
+}
```
PoignardAzur, line 378: (Optional) This function should probably be inlined. (It's used in `tests/paint.rs` which is a special case, but even there you could copy the function.)

masonry_testing/src/harness.rs
```diff
@@ -206,6 +207,10 @@ pub struct TestHarnessParams {
     pub max_screenshot_size: u32,
 }
 
+struct SimpleClipboard {
+    contents: String,
+}
```
PoignardAzur, line 212: (Optional) I think this type should be called `MockClipboard` and moved to its own module.

PoignardAzur (approved):
LGTM.

I have misgivings about the proliferation of context types (in one code path you have a DriverCtx that gives you an AppCtx that gives you two text contexts) and I still think we should come up with a better name than AppCtx, but I don't want to block on naming.

This does seem like a thing we'll want sooner or later, and I don't have any better alternatives.

PoignardAzur:
@xStrom Any blocker left?

lvyaqiao:
I opened #1843 with a focused fix for the Windows Ctrl+V clipboard-read panic in masonry_winit.

This standalone change handles get_contents() errors by logging a warning instead of panicking, while preserving normal paste behavior. Since #1766 also touches the clipboard path, I wanted to link the two PRs for coordination. If #1766 includes this fix, #1843 can be closed as superseded.

PoignardAzur:
It's fine, this PR needs rebasing anyway.

PR #1770, opened 2026-04-25
Fix Slider widget properties

RagibHasin:
Follow up of #1691


RagibHasin:
It seems that our image diffing algorithm did not/does not pick up the change in appearance of this and some other widgets.

We should use a high contrasting background color/transparent background color for test harness to reliably catch these.

PoignardAzur:
> It seems that our image diffing algorithm did not/does not pick up the change in appearance of this and some other widgets.

Wait, what?

masonry/src/properties/slider.rs
```diff
@@ -27,6 +27,41 @@ impl TrackThickness {
...
+    /// Color of the inactive portion.
+    pub inactive: Color,
+}
+
+impl Property for TrackColor {
+    fn static_default() -> &'static Self {
+        static DEFAULT: TrackColor = TrackColor {
+            active: theme::ACCENT_COLOR,
+            inactive: theme::ZYNC_800,
+        };
```
PoignardAzur, line 45: The pattern so far has been that default property values are "neutral", usually zero/false/black, and the theme colors stay in the `theme.rs` file.

There's been a few PRs breaking this pattern that I didn't catch, but I'd like to stick to it moving forward.

RagibHasin:
> > It seems that our image diffing algorithm did not/does not pick up the change in appearance of this and some other widgets.
> 
> Wait, what?

Yes, it's wild. Although to be fair the difference is very minimal with very low contrast with background.

masonry/screenshots/slider_drag_initial_at_25.png
PoignardAzur, line 1: The new paint code now has the track start at the beginning of the widget, instead of leaving padding for the thumb size. Is that intentional? Even if it is, I feel like you should change the default padding to replicate the same effect.

Also, (non-blocking) it might be nice to add a new screenshot with a slider at 0%/100% progress.

PoignardAzur (commented):
Thanks for the PR!

I appreciate simplifying the widget code, just not sure about the behavior changes.

xStrom:
> It seems that our image diffing algorithm did not/does not pick up the change in appearance of this and some other widgets.

Which image diffs were not picked up? The one that you updated anyway here or something else? I would like to take a look at the specific test case that doesn't get detected.

RagibHasin:
> I would like to take a look at the specific test case that doesn't get detected.

I would see if I can dig those up.

masonry/screenshots/slider_drag_initial_at_25.png
RagibHasin: Yes, the lack of padding is intentional. It started as the padding being dependent on thumb radius and thus jumping around when the slider is hovered/activated.

Also, I don't think having a default padding to any widget is standard practice among toolkits. I was just now exploring WinUI 3 Gallery and noticed the lack of such padding.

RagibHasin:
It sits ready for review, @xStrom .

xStrom:
I looked into the image differences. So the color changes from `rgb(39, 39, 42)` to `rgb(41, 41, 41)` which isn't a lot, but I can definitely see it just by eyeballing it when I swap between the screenshots.

The current `DEFAULT_SCREENSHOT_TOLERANCE` is `16`, but this difference wouldn't even be noticed with `2`. The tolerance needs to be `1` or `0` for this change to be picked up.

I don't think anything needs to happen in this PR in relation to this tolerance. However, now that we have Vello CPU test rendering, we realy ought to see if we can migrate to a tolerance of `0` in a separate PR. I can look into that at some point myself.

masonry/src/properties/slider.rs
xStrom: So I can definitely understand not importing `theme` for property defaults.

However, I don't think using transparent colors or zero sizes for usable bits makes much sense. It makes parts of the widget disappear.

The question is, what is the purpose of property defaults? Is it merely to avoid crashes or should it still result in a usable widget, just not styled?

To put it another way, should things still be usable (even if ugly) without any theme whatsoever? I think the answer should be yes.

masonry/screenshots/slider_drag_initial_at_25.png
xStrom: Lack of padding is fine. Does the thumb get painted outside of the box at the extremes? Then it should be setting the paint insets.

masonry/src/widgets/slider.rs
```diff
@@ -320,13 +294,10 @@ impl Widget for Slider {
...
 
     fn property_changed(&mut self, ctx: &mut UpdateCtx<'_>, property_type: TypeId) {
-        BarColor::prop_changed(ctx, property_type);
         TrackThickness::prop_changed(ctx, property_type);
+        TrackColor::prop_changed(ctx, property_type);
         ThumbColor::prop_changed(ctx, property_type);
         ThumbRadius::prop_changed(ctx, property_type);
-        if Background::matches(property_type) {
-            ctx.request_paint_only();
-        }
```
xStrom: Instead of adding the `TrackColor::prop_changed` method, you should use `TrackColor::matches` like `Background` was using. The reason is that the older `prop_changed` approach demands a specific implementation of the property, while the newer `matches` approach allows properties to be reused across different widgets which may request different passes.

masonry/screenshots/slider_drag_initial_at_25.png
RagibHasin: No, the thumb border sits at tracks end at extrema.

masonry/src/properties/slider.rs
RagibHasin: What colors would you suggest? Black/dark gray/light gray from the CSS palette?

xStrom: That's a good question. We basically need some very basic high contrast implicit theme for default colors. I guess the main question is whether it should be light or dark theme. A lot of the existing properties use black for content color, so I guess light theme it is.

So background should be assumed white, and so most of these property colors should probably just default to black. For properties that are semantically background, transparent can be reasonable. Otherwise white I guess.

PoignardAzur: Yeah, `#FFF` and `#000` are both acceptable default values.

PoignardAzur (approved):

PR #1771, opened 2026-04-26
Make Scrollbar collapse when not hovered

RagibHasin:


xStrom:
The scrollbar should also not collapse when there is a scroll in progress without hover, i.e. `grab_anchor.is_some()`. That would match Firefox behavior as well.

masonry/screenshots/portal_button_list_no_scroll.png
PoignardAzur, line 1: I don't like how many screenshots are changed here. These screenshots in particular are pretty heavy. Could we make hovered scrollbars the default in tests instead?

PoignardAzur (approved):
Thanks for the PR!

I'm not convinced it looks better, but the code change is idiomatic, and you're clearly more motivated to improve this code than me.

LGTM.

masonry/src/widgets/scroll_bar.rs
```diff
@@ -338,12 +340,14 @@ impl Widget for ScrollBar {
...
-    fn update(
-        &mut self,
-        _ctx: &mut UpdateCtx<'_>,
-        _props: &mut PropertiesMut<'_>,
-        _event: &Update,
-    ) {
+    fn update(&mut self, ctx: &mut UpdateCtx<'_>, _props: &mut PropertiesMut<'_>, event: &Update) {
+        match event {
+            Update::HoveredChanged(received) => {
+                self.expanded = *received;
```
PoignardAzur: Note that you don't to set a bool, you could use `PaintCtx::is_hovered()`.

masonry/src/widgets/scroll_bar.rs
```diff
@@ -384,12 +388,20 @@ impl Widget for ScrollBar {
...
+        let scrollbar_width = theme::SCROLLBAR_WIDTH;
 
         let size = ctx.content_box_size();
-        let (inset_x, inset_y) = self.axis.pack_xy(0.0, cursor_padding);
+        let inset_start = cursor_padding
+            + if self.expanded {
+                0.
+            } else {
+                scrollbar_width / 2.
+            };
```
PoignardAzur: As Kaur points out, you should take the grabbing status too.

Also, this code would be more readable by inlining the `cursor_padding +` part:

```rust
let inset_start = if ... {
    cursor_padding + 0.
} else {
    cursor_padding + scrollbar_width / 2.
};
```

PoignardAzur (changes requested):
Forgot about the "user is grabbing the bar" case. That needs to be fixed before we can merge this.

masonry/screenshots/portal_button_list_no_scroll.png
xStrom: In general, having a collapsable scrollbar is a style choice. So having it be an optional configuration of the widget makes sense. So if we want to reduce the screenshot churn, one option would be to make it configurable right now and then we can have it not be collapsable in those tests.

RagibHasin:
As we are concerned about screenshot churn and repo size bloat, #1773 also touches some of these screenshots (specifically of `Portal` widget).
If that behavior is desired then it would be better to merge these two PRs for better management of screenshot churn.

masonry/screenshots/portal_button_list_no_scroll.png
RagibHasin: We should migrate ScrollBar to use property system. Right now it does not do so at all. We can then have `AutoCollapsible` property that can be false by default.

What do you guys think?

RagibHasin:
Can we have a decision here, so that we can make progress?
@xStrom @PoignardAzur

xStrom:
I don't think it needs to be merged with #1773 because of the screenshot churn. Once that is rebased I would assume zero screenshot changes in that one, because you can just do screenshots with autohide disabled.

masonry/screenshots/portal_button_list_no_scroll.png
xStrom: I think it can be a property, yes. I would also be fine with it being a simple `bool` on the struct just to get it working, if that is easier.

PoignardAzur (approved):
LGTM.

PR #1772, opened 2026-04-26
feat: `label()` letter/word spacing

l0uisgrange:
This PR adds a new `label("Text").letter_spacing(-2.0)` and `label("Text").word_spacing(-2.0)` method to change letter/word spacing for specific texts

### DIFF
--- a/masonry/src/theme.rs
+++ b/masonry/src/theme.rs
@@ -60,7 +60,7 @@ pub fn default_property_set() -> DefaultProperties {
 
     // Badge
     properties.insert::<Badge, _>(Padding::from_vh(3., 5.));
-    properties.insert::<Badge, _>(CornerRadius::all(999.));
+    properties.insert::<Badge, _>(CornerRadius { radius: 999. });
     properties.insert::<Badge, _>(BorderWidth { width: 0. });
     properties.insert::<Badge, _>(Background::Color(ACCENT_COLOR));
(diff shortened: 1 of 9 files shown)


l0uisgrange:
Hey there, some checks are not passing, but I am waiting on some input from Xilem's team, just to be sure I go in a correct direction.

RagibHasin:
We generally have one feature in one PR policy. That is you should split this PR in two, one for `label` improvement and another for `CornerRadius`.

Also, `CornerRadius` should wrap a `kurbo::RoundedRectRadii`, not redefine it. Forwarding methods from the wrapped object when useful is fine.

l0uisgrange:
Yep, totally agree. I've removed CornerRadius, I was doing it wrong. 
This PR is just about adding spacing methods to `label()` then.
Please let me know if you need any additional modifications.

RagibHasin:
We would ideally like to have these attributes as attributes on a span of attributed text, but that is still some ways into the future.

Meanwhile, we have a precedent of accepting such changes if it unblocks someones usecase (in #1673).

Someone with more experience and authority (like @xStrom) should decide here.

PoignardAzur (approved):
This PR is reminding me that I *really* wish we had better integration between Parley's properties and Masonry's Properties. Not something that can be fixed in this specific PR, though.

LGTM, thanks for the PR!

PR #1773, opened 2026-04-26
Make Scrollbar auto-hideable

RagibHasin:


RagibHasin:
Depends on #1771 .

masonry/src/widgets/portal.rs
```diff
@@ -646,6 +700,12 @@ impl<W: Widget + FromDynWidget + ?Sized> Widget for Portal<W> {
         }
     }
 
+    fn property_changed(&mut self, ctx: &mut UpdateCtx<'_>, property_type: TypeId) {
+        if AutoHideScrollBar::matches(property_type) {
+            ctx.request_layout();
```
PoignardAzur: I think this should be `ctx.request_anim_frame()` instead, which would let you centralize the code, among other things.

PoignardAzur (approved):
Overall, LGTM.

PR #1775, opened 2026-04-27
Checkbox padding snapshot

### LINKED ISSUE #1369: Checkbox: Validate that `Padding` property is handled properly
In #1366, we noted that `Checkbox`'s `paint` code doesn't use the `Padding` property. This seems suspicious.
We should check what a `Checkbox` with `Padding` looks like, and add a test for that rendering.

We should also consider whether the focus indicator should be outside or inside the padding. I suspect inside, but I've not thought about it.

WaterWhisperer:
Fixes: #1369


waywardmonkeys:
The markdown changes don't belong in this PR, so you'll want to remove that commit.

PoignardAzur (approved):
I think https://github.com/linebender/xilem/issues/1369 was out of date, so this test might not be necessary, but we might as well merge it.

Thanks for the PR!

PR #1776, opened 2026-04-27
Add option to set default base color to Xilem apps

RagibHasin:


DJMcNab (commented):
I'm on mobile so can't easily read nearby code. But on the whole, I'd be cautious about this approach - it adds a lot of noise to each example, which isn't really self-documenting.
It would also make it easier to tell which example this is being used in!

What do you think about a new `default_base_color` (maybe `with_default_base_color`) on `Xilem`, which is the base colour for every window which doesn't have it set explicitly (perhaps?). Such a method would also provide a good place to document the way to set it dynamically, which is to not use the `new_simple` constructor.

I'm very happy to see this feature restored, though...

RagibHasin:
> What do you think about a new `default_base_color` (maybe `with_default_base_color`) on `Xilem`, which is the base colour for every window which doesn't have it set explicitly (perhaps?).

Yes, that would be better. I would rework the PR.

RagibHasin:
Okay, I've added a new `default_base_color` field to `Xilem` struct beside `default_properties`.

This solution is less intrusive to end-users but forces `xilem::MasonryDriver` to carry the base color too, for injection at window creation time.

DJMcNab (approved):
I'd like to see this used in at least one example, but otherwise this is good to land. Thanks!

PR #1779, opened 2026-04-30
Add widget timers to Masonry and use them for delayed UI   behavior.

waywardmonkeys:
* Add `understory_timing` as a `masonry_core` dependency.
* Store pending widget timers in `RenderRoot`.
* Add `TimerToken` and deliver expired timers through `Update::Timer`.
* Add context APIs:
  * `request_timer(delay) -> TimerToken`
  * `cancel_timer(token)`
* Add backend hooks for timer driving:
  * `RenderRoot::set_timer_time`
  * `RenderRoot::next_timer_deadline`
  * `RenderRoot::handle_timers`
* Update `masonry_winit` to drive timer time and platform wakeups.
* Convert `TextArea` cursor blinking to one-shot timers.
* Convert the layers tooltip-delay example to one-shot timers.
* Document when to use timers versus animation frames.

 ## Rationale

Masonry has several UI behaviors that are delayed but not frame-cadenced. Examples include cursor blinking, tooltip
  delays, debounce, and long press recognition. These fit a timer model better than animation frames.

This PR adds a widget-local timer lifecycle: widgets request a timer, keep the returned `TimerToken`, and receive
  `Update::Timer` when it expires. Widgets can cancel pending timers when the relevant state changes.

Animation frames remain the right tool for continuous visual updates, such as interpolation or motion that should advance with frame cadence.

### DIFF
--- a/masonry/examples/layers.rs
+++ b/masonry/examples/layers.rs
@@ -77,10 +77,10 @@ impl Widget for OverlayBox {
     ) {
         if let PointerEvent::Move(PointerUpdate { current, .. }) = event {
             self.last_cursor_pos = current.logical_point();
-            if let Some(timer) = self.hover_timer.take() {
-                ctx.cancel_timer(timer);
+            if self.layer_root_id.take().is_some() && ctx.is_hovered() && self.hover_timer.is_none()
+            {
+                self.hover_timer = Some(ctx.request_timer(Duration::from_millis(300)));
             }
-            self.hover_timer = Some(ctx.request_timer(Duration::from_millis(300)));
         }
     }
 
@@ -90,13 +90,24 @@ impl Widget for OverlayBox {
 
     fn update(&mut self, ctx: &mut UpdateCtx<'_>, _props: &mut PropertiesMut<'_>, event: &Update) {
         match event {
+            Update::HoveredChanged(true)
+                if self.layer_root_id.is_none() && self.hover_timer.is_none() =>
+            {
+                self.hover_timer = Some(ctx.request_timer(Duration::from_millis(300)));
+            }
             Update::HoveredChanged(false) => {
                 if let Some(timer) = self.hover_timer.take() {
                     ctx.cancel_timer(timer);
                 }
+                if let Some(layer) = self.layer_root_id.take() {
+                    ctx.remove_layer(layer);
+                }
             }
-            Update::Timer(token) if Some(*token) == self.hover_timer => {
+            Update::TimerExpired(token) if Some(*token) == self.hover_timer => {
                 self.hover_timer = None;
+                if !ctx.is_hovered() {
+                    return;
+                }
                 let (overlay, layer_type) = (self.overlayer)();
                 self.layer_root_id = Some(overlay.id());
                 let layer_pos = self.last_cursor_pos + Vec2::new(5., -25.);
--- a/masonry/examples/layers.rs
+++ b/masonry/examples/layers.rs
@@ -10,7 +10,7 @@ use masonry::accesskit::{Node, Role};
 use masonry::core::{
     AccessCtx, ChildrenIds, ErasedAction, EventCtx, LayerType, LayoutCtx, MeasureCtx, NewWidget,
     NoAction, PaintCtx, PointerEvent, PointerUpdate, PropertiesMut, PropertiesRef, PropertySet,
-    RegisterCtx, StyleProperty, Update, UpdateCtx, Widget, WidgetId, WidgetPod,
+    RegisterCtx, StyleProperty, TimerToken, Update, UpdateCtx, Widget, WidgetId, WidgetPod,
 };
 use masonry::imaging::Painter;
 use masonry::kurbo::{Axis, Point, Size, Vec2};
@@ -20,7 +20,7 @@ use masonry::parley::FontWeight;
 use masonry::peniko::Color;
 use masonry::properties::{Background, BorderColor, BorderWidth, ContentColor};
 use masonry::theme::default_property_set;
-use masonry::util::{Duration, Instant};
+use masonry::util::Duration;
 use masonry::widgets::{Flex, Label, Selector};
 use masonry_winit::app::{AppDriver, DriverCtx, NewWindow, WindowId};
 use masonry_winit::winit::window::Window;
@@ -44,7 +44,7 @@ struct OverlayBox {
     child: WidgetPod<dyn Widget>,
     overlayer: Box<dyn Fn() -> (NewWidget<dyn Widget>, LayerType)>,
     layer_root_id: Option<WidgetId>,
-    last_mouse_move: Option<Instant>,
+    hover_timer: Option<TimerToken>,
     last_cursor_pos: Point,
 }
 
@@ -59,7 +59,7 @@ impl OverlayBox {
             child: child.erased().to_pod(),
             overlayer,
             layer_root_id: None,
-            last_mouse_move: None,
+            hover_timer: None,
             last_cursor_pos: Point::ZERO,
         }
     }
@@ -77,27 +77,10 @@ impl Widget for OverlayBox {
     ) {
         if let PointerEvent::Move(PointerUpdate { current, .. }) = event {
             self.last_cursor_pos = current.logical_point();
-            self.last_mouse_move = Some(Instant::now());
-            ctx.request_anim_frame();
-        }
-    }
-
-    fn on_anim_frame(
-        &mut self,
-        ctx: &mut UpdateCtx<'_>,
-        _props: &mut PropertiesMut<'_>,
-        _interval: u64,
-    ) {
-        if let Some(last_mouse_move) = self.last_mouse_move {
-            let now = Instant::now();
-            if now.duration_since(last_mouse_move) > Duration::from_millis(300) {
-                let (overlay, layer_type) = (self.overlayer)();
-                self.layer_root_id = Some(overlay.id());
-                let layer_pos = self.last_cursor_pos + Vec2::new(5., -25.);
-                ctx.create_layer(layer_type, overlay, layer_pos);
-            } else {
-                ctx.request_anim_frame();
+            if let Some(timer) = self.hover_timer.take() {
+                ctx.cancel_timer(timer);
             }
+            self.hover_timer = Some(ctx.request_timer(Duration::from_millis(300)));
         }
     }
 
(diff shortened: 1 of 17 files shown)


waywardmonkeys:
This was done with LLM assistance to avoid typing out the boring stuff.

Draft since I will publish `understory_timing` soon rather than a git dep.

masonry/src/widgets/text_area.rs
```diff
@@ -425,55 +465,6 @@ pub enum TextAction {
...
-        /// For the first half (i.e. currently 0.5s) the cursor is shown, and for the second half,
-        /// it is hidden.
-        const CURSOR_BLINK_TIME: u64 = 1000;
-
-        /// The timeout, in milliseconds, after which the cursor will stop blinking (i.e. stay
-        /// solid).
-        const CURSOR_BLINK_TIMEOUT: u64 = 10_000; // 10 seconds
-
-        // TODO: These should be reading from the system settings, but we currently
-        // aren't aware of a robust way to read that cross-platform.
```
xStrom, line 444: Let's preserve this comment near the top area where these consts now live.

masonry/src/widgets/text_area.rs
```diff
@@ -257,6 +267,36 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
...
+    fn advance_cursor_blink(&mut self) -> bool {
+        self.anim_prev_interval += CURSOR_BLINK_INTERVAL;
+        self.anim_elapsed += CURSOR_BLINK_INTERVAL;
+
+        if self.anim_prev_interval >= CURSOR_BLINK_TIME {
+            self.anim_prev_interval = self.anim_prev_interval.rem_euclid(CURSOR_BLINK_TIME);
+        }
+
+        let should_show = self.anim_elapsed >= CURSOR_BLINK_TIMEOUT
+            || self.anim_prev_interval < CURSOR_BLINK_INTERVAL;
```
xStrom: The `self.anim_prev_interval < CURSOR_BLINK_INTERVAL` check seems needlessly misleading, the only scenario that satisfies this is `self.anim_prev_interval == 0`.

masonry/examples/layers.rs
```diff
@@ -77,37 +77,32 @@ impl Widget for OverlayBox {
...
-                let (overlay, layer_type) = (self.overlayer)();
-                self.layer_root_id = Some(overlay.id());
-                let layer_pos = self.last_cursor_pos + Vec2::new(5., -25.);
-                ctx.create_layer(layer_type, overlay, layer_pos);
-            } else {
-                ctx.request_anim_frame();
+            if let Some(timer) = self.hover_timer.take() {
+                ctx.cancel_timer(timer);
             }
+            self.hover_timer = Some(ctx.request_timer(Duration::from_millis(300)));
```
xStrom: I know this replicated the old logic of showing the layer after the mouse stopped moving for 300ms, but I think this is a good opportunity to improve things.

Canceling and requesting a new timer for every mouse move position is kind of wasteful, including `understory_timing` always looping over the whole set of timers for the cancel. Also, usually tooltips are shown after hovering for a certain amount of time, not after stopping movement. This is certainly how Windows tooltips work.

Thus I think we should update the code. Perhaps just having `Update::HoveredChanged(true)` request the timer and `HoveredChanged(false)` cancel it.

xStrom (commented):
Timers are such a great feature to have, thanks for moving this forward.

The `TextInput` cursor is also much less buggy with this implementation compared to the old one. For example, losing focus when the cursor is not shown doesn't cause it disappear for good anymore. 😅

masonry/examples/layers.rs
PoignardAzur: > Perhaps just having Update::HoveredChanged(true) request the timer and HoveredChanged(false) cancel it.

I'm all for finding a better solution, but I'll note that this does not match how tooltips work in most frameworks.

masonry/src/tests/update.rs
```diff
@@ -65,6 +66,65 @@ fn new_widget() {
...
+        |record| matches!(record, Record::Update(Update::Timer(seen)) if seen == token),
+    );
+
+    let records = harness.take_records_of(parent_tag);
+    assert!(
+        !records
+            .iter()
+            .any(|record| matches!(record, Record::Update(Update::Timer(_)))),
+        "timer update should not bubble to the parent"
+    );
```
PoignardAzur: You should use `assert_all` for this.

masonry/src/tests/update.rs
```diff
@@ -65,6 +66,65 @@ fn new_widget() {
...
+    harness.set_timer_time(Duration::from_millis(10));
+    assert_eq!(harness.handle_timers(), 0);
+
+    let records = harness.take_records_of(target_tag);
+    assert!(
+        !records
+            .iter()
+            .any(|record| matches!(record, Record::Update(Update::Timer(_)))),
+        "cancelled timer should not be delivered"
+    );
```
PoignardAzur: You should use `assert_all` for this.

masonry_core/src/app/render_root.rs
```diff
@@ -37,15 +38,20 @@ use crate::passes::paint::run_paint_pass;
...
+use crate::util::Duration;
 
 /// We ensure that any valid initial IME area is sent to the platform by storing an invalid initial
 /// IME area as the `last_sent_ime_area`.
 const INVALID_IME_AREA: Rect = Rect::new(f64::NAN, f64::NAN, f64::NAN, f64::NAN);
 
+fn duration_to_timer_ticks(duration: Duration) -> TimerInstant {
+    u64::try_from(duration.as_nanos()).unwrap_or(u64::MAX)
+}
+
```
PoignardAzur: The top of `render_root.rs` is valuable "real estate". This should be moved towards the bottom.

masonry_core/src/app/render_root.rs
```diff
@@ -961,6 +975,52 @@ impl RenderRoot {
...
     }
 
+    /// Sets the current host-provided timer time.
+    ///
+    /// The time is measured from an origin chosen by the host. All calls to
+    /// this method and [`next_timer_deadline`](Self::next_timer_deadline) must
+    /// use the same origin.
+    pub fn set_timer_time(&mut self, time: Duration) {
+        self.global_state.timer_now = duration_to_timer_ticks(time);
+    }
```
PoignardAzur, line 985: Is there a benefit to splitting this from `handle_timers`? As far as I can tell, all call sites call them both at once.

This seems like a surprising pattern, we don't do this anywhere else (we don't call one method to set the mouse position and another to trigger all mouse events).

masonry_core/src/app/render_root.rs
```diff
@@ -961,6 +975,52 @@ impl RenderRoot {
...
+
+    /// Returns the next timer deadline, measured from the host's timer origin.
+    pub fn next_timer_deadline(&self) -> Option<Duration> {
```
PoignardAzur, line 988: So as far as I can tell, the reason to use `Duration` instead of `Instant` is that `Instant` can't be mocked in tests? If so, this should probably be documented in either a doc comment or a code comment.

masonry_core/src/app/render_root.rs
```diff
@@ -961,6 +975,52 @@ impl RenderRoot {
...
+        let mut fired = Vec::new();
+        // Pop the due set before dispatch. Widget updates may request or cancel
+        // timers; those mutations should not change which timers fire in this
+        // drain cycle.
+        while let Some(timer) = self.global_state.timers.pop_expired(now) {
+            let token = TimerToken(timer.id());
+            fired.push((timer.into_target(), token));
+        }
+
+        let mut delivered = 0;
```
PoignardAzur: I don't think keeping track of `delivered` is very useful. It's used to short-circuit rewrite passes in RenderRoot and signal handling in EventLoop, but both of those already have short-circuits built in. I'd recommend dropping it.

masonry_core/src/core/events.rs
```diff
@@ -120,6 +136,9 @@ pub enum Update {
     /// [hovered]: crate::doc::masonry_concepts#widget-status
     ChildHoveredChanged(bool),
 
+    /// Called when a timer requested by this widget expires.
+    Timer(TimerToken),
```
PoignardAzur: This should probably be `TimerExpired` or something for consistency with other event names. (Though "expired" kind of suggests an error situation? Idk, naming is hard.)

masonry_core/src/passes/update.rs
```diff
@@ -1248,4 +1248,12 @@ pub(crate) fn run_update_fonts_pass(root: &mut RenderRoot) {
     );
 }
 
+pub(crate) fn run_update_timer_pass(root: &mut RenderRoot, target: WidgetId, token: TimerToken) {
```
PoignardAzur, line 1254:
```suggestion
// ----------------

// --- MARK: TIMERS
pub(crate) fn run_update_timer_pass(root: &mut RenderRoot, target: WidgetId, token: TimerToken) {
```

masonry_testing/src/harness.rs
```diff
@@ -856,6 +856,18 @@ impl<W: Widget> TestHarness<W> {
         self.process_signals();
     }
 
+    /// Sets the simulated timer time.
+    pub fn set_timer_time(&mut self, time: Duration) {
+        self.render_root.set_timer_time(time);
+    }
```
PoignardAzur: As above, I think this should be merged with `handle_timers`.

In fact, I'm thinking maybe this should be merged with the `animate_ms` method above into a `move_time_forward` method, because a situation where animations progress but timers don't is  unrealistic.

PoignardAzur (commented):
Timers are long-overdue in Masonry. Glad to finally have an implementation!

I have doubts about the API, but the internals seem sound.

waywardmonkeys:
Thanks for the feedback. I'm heading out for the evening shortly but will respond with updates within the next day or two.

I didn't know about `assert_all`!

masonry_core/src/app/render_root.rs
waywardmonkeys: Done.

masonry/src/tests/update.rs
waywardmonkeys: Done.

masonry/src/tests/update.rs
waywardmonkeys: Done.

masonry_core/src/app/render_root.rs
waywardmonkeys: I wrote up a whole thing ... and thought it was still complicated ... so going to revisit this code soon even further. I think there's a better / simpler solution (on the surface, it might need some other changes internally, we'll see).

masonry/src/widgets/text_area.rs
waywardmonkeys: Restored.

waywardmonkeys:
This is now using the published version 0.1.2 of `understory_timing`.

masonry_core/src/passes/update.rs
waywardmonkeys: Done.

masonry_core/src/core/events.rs
waywardmonkeys: Done.

masonry_core/src/app/render_root.rs
waywardmonkeys: Yes, that was the reason and there's a comment there. But I may revisit this in a moment due to some other simplifications.

masonry_core/src/app/render_root.rs
waywardmonkeys: Done, I think.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -27,8 +27,9 @@ use ui_events_winit::{WindowEventReducer, WindowEventTranslation};
 use winit::application::ApplicationHandler;
 use winit::dpi::PhysicalSize;
 use winit::error::EventLoopError;
+use winit::event::StartCause;
 use winit::event::{DeviceEvent as WinitDeviceEvent, DeviceId, WindowEvent as WinitWindowEvent};
```
RagibHasin, line 31:
```suggestion
use winit::event::{
    DeviceEvent as WinitDeviceEvent, DeviceId, StartCause, WindowEvent as WinitWindowEvent,
};
```

PoignardAzur (approved):
Again, I have nitpicks about the API, but I'm fine merging now and fixing them later.

LGTM.

PR #1782, opened 2026-05-01
feat: `WindowOptions` for macOS

l0uisgrange:
Adding new `WindowOptions` for macOS

```rust
#[derive(Debug, Clone, Default)]
pub(crate) struct PlatformSpecificInitialWindowAttrs {
    pub(crate) titlebar_hidden: bool,
    pub(crate) title_hidden: bool,
    pub(crate) fullsize_content_view: bool,
    pub(crate) titlebar_transparent: bool,
    pub(crate) movable_by_window_background: bool,
}
```

I'm not sure about moving some to reactive window attributes... what do you think?

### DIFF
--- a/xilem/src/lib.rs
+++ b/xilem/src/lib.rs
@@ -181,3 +181,6 @@ pub use window_view::{PodWindow, WindowView, window};
 
 #[cfg(windows)]
 pub use window_options::WindowOptionsExtWindows;
+
+#[cfg(target_os = "macos")]
+pub use window_options::WindowOptionsExtMacOS;
--- a/xilem/src/window_options.rs
+++ b/xilem/src/window_options.rs
@@ -588,10 +588,10 @@ mod windows {
 
 #[cfg(target_os = "macos")]
 mod macos {
-    use winit::platform::macos::{WindowAttributesExtMacOS, WindowExtMacOS};
+    use winit::platform::macos::WindowAttributesExtMacOS;
     use winit::window::{Window, WindowAttributes};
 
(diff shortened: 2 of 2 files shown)


l0uisgrange:
The last CI is failing, but does not seem to be the PR's fault

PoignardAzur (approved):
Sometimes I read a PR and think "Wow, this is cursed, why did the author write it that way?" and then I look at the existing code and realized the author followed our conventions.

This is one of those cases. I absolutely hate the design patterns `window_options.rs` are using, but I don't see a point blocking this PR until we fix them.

LGTM!

PoignardAzur:
(CI is fixed. Do you have merge powers?)

l0uisgrange:
I will always try to follow your conventions, even if it seems odd.
I don't have merge power (I'm not sure the question was for me though 😂)

RagibHasin:
> I absolutely hate the design patterns `window_options.rs` are using

I would be the one responsible for the cursed-ness of this design, although Daniel approved the PR at that time.

My intention was to closely follow Winit conventions here, as this was new ground for Xilem at that time.

If you have preference for any other organization of this code, I would be amenable to clean it up that way. 🤷🏽

PoignardAzur:
Part of what makes the code cursed is indeed winit-inspired (I don't think we get much value from using ext traits in this module), but the biggest problem is that we're using data structures meant to be re-generated (Xilem views) for something that is supposed to only be read once (window creation parameters).

Unfortunately, I don't think anybody has come up with a better architecture yet.

PR #1783, opened 2026-05-01
feat: `app.with_on_start()` function

l0uisgrange:
A new function `.with_on_start()` is introduced on `Xilem::new` to run a `FnOnce` after the app has started. 

```rust
let app = Xilem::new_simple(state, app_logic, opts)
  .with_on_start(move |_| {
      #[cfg(target_os = "macos")]
      menu.init_for_nsapp();
  })
app.run_in(EventLoop::with_user_event())?;
```

This has a few use cases, and amongst them:

1. Check for update on startup (with some API/GitHub releases/etc.)
2. More specific for macOS, but this allows compatibility with the [`muda`](https://docs.rs/muda/latest/muda/) crate. Without this, it's almost impossible to add items in the macOS menu bar (used by all applications), because `NSApplication` needs to be set **before** using `muda` functions, which is only possible after the initialization of `Xilem`

   If you're not familiar with macOS, the menu bar looks like this

   Capture d’écran 2026-05-01 à 13 59 51

### DIFF
--- a/xilem/src/app.rs
+++ b/xilem/src/app.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use masonry::core::DefaultProperties;
 use masonry::peniko::{Blob, Color};
 use masonry::theme::{BACKGROUND_COLOR, default_property_set};
-use masonry_winit::app::{EventLoopBuilder, MasonryUserEvent, NewWindow, WindowId};
+use masonry_winit::app::{EventLoopBuilder, MasonryState, MasonryUserEvent, NewWindow, WindowId};
 use tokio::runtime::Runtime as TokioRuntime;
 use winit::error::EventLoopError;
 
@@ -27,6 +27,8 @@ pub struct Xilem<State, Logic> {
     default_base_color: Color,
     // Font data to include in loading.
     fonts: Vec<Blob<u8>>,
+    // Callback invoked once on startup, after windows creation.
+    on_start: Option<Box<dyn FnOnce(&mut MasonryState<'_>)>>,
 }
 
 /// State type used by [`Xilem::new_simple`].
@@ -147,6 +149,7 @@ where
             default_properties: None,
             default_base_color: BACKGROUND_COLOR,
             fonts: Vec::new(),
+            on_start: None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


xilem/src/app.rs
```diff
@@ -27,6 +27,8 @@ pub struct Xilem<State, Logic> {
     default_base_color: Color,
     // Font data to include in loading.
     fonts: Vec<Blob<u8>>,
+    // Callback invoked once on startup, after windows creation.
+    on_start: Option<Box<dyn FnOnce(&mut MasonryState<'_>)>>,
```
l0uisgrange, line 31: I do not need the `MasonryState` argument for the use cases I introduced, but maybe someone needs it one day? I can remove it if you think it's unnecessary.

RagibHasin:
Have you tried https://github.com/baltobor/xilem_extras ? It claims to provide application menus using `muda`. Although I have not personally tried it, yet.

l0uisgrange:
Yes, I've found the library in a comment a few days ago and tried it. It is good, indeed, though I would love to bring some of its components to `xilem` directly. 
I could not try the application menu since `xilem_extras` only provides a clean way to declare it, but no way to activate it (after `NSApplication` instantiation). Even `xilem_extras` would need a `.with_on_start()` function.

PoignardAzur (approved):
This seems like a fairly standard feature for a GUI framework to have.

LGTM!

RagibHasin:
Also, it can be argued that other app lifecycle events should be hookable to. We should look into an extensible option for this.

I would still prefer to use the custom event loop option for this usecase though.

l0uisgrange:
@RagibHasin if you manage to find another implementation where it's possible to set these events inside a custom event loop, and it's very compact, I don't mind. The goal of `.with_on_start()` was mainly to give an api very clean and compact.

RagibHasin:
> @RagibHasin if you manage to find another implementation where it's possible to set these events inside a custom event loop, and it's very compact, I don't mind. The goal of `.with_on_start()` was mainly to give an api very clean and compact.

I'm not disputing the brevity or usefulness of this API.
I was wondering how it could be better designed to accommodate other lifecycle events (e.g close, suspend, resume, etc.).

PoignardAzur:
I'm fine carving out a special case for start for now, since it's the one most likely to be used.

I'd welcome a more general lifecycle system, though I wouldn't be super comfortable grafting one on top of Xilem's current API, which I think already needs a lot of rework at the composition root.

Issue #1340, opened 2025-08-15
Examples should have interaction tests


PoignardAzur:
We have unit tests in most of the Masonry examples, but all of them are "Load the window and take a screenshot" tests.

We should add tests that interact with app in at least two instances:

- Click some buttons in the calc example.
- Add a few tasks in the to_do_list example.

PR #1784, opened 2026-05-01
masonry: add interaction tests to calc and to_do_list examples (#1340)

tonybierman:
Replaces the screenshot-only `// TODO` tests in `calc_masonry` and `to_do_list` with interaction tests that click buttons and type text, closing #1340.


xStrom:
These aren't really interaction tests. These tests include a duplicated implementation within the test function. They don't touch the actual `AppDriver::on_action` code. Of course `AppDriver` is a `masonry_winit` thing, so I'm not fully sure if `TestHarness` should even accommodate it. 

@PoignardAzur what do you think? How should we even test `AppDriver`? Some sort of `masonry_winit_testing` harness or ..?

PoignardAzur:
> How should we even test AppDriver?

No, TestHarness already does everything we need, and the API is not that hard to use.

This is clearly vibe-coded, and the author didn't take the time to understand what they submitted. Part of the point of tagging https://github.com/linebender/xilem/issues/1340 as "good first issue" is to encourage people to come in and learn the codebase. If we maintainers have to spend energy giving feedback to an author who's clearly going to pipe that feedback straight to Claude, we might as well use Claude ourselves, which defeats the point of the "good first issue" tag.

This is worse than useless. I'm closing this.

PR #1789, opened 2026-05-03
Fix test harness mouse move.

### LINKED ISSUE #1756: Clicking on a very small button fails in test harness
Code to reproduce:

```rust
let tag = WidgetTag::unique();
let button = Button::with_text("").prepare().with_tag(tag);
let parent = Flex::row().with(button, 0.).prepare();

let params = ...;
let mut harness = TestHarness::create_with(test_property_set(), widget, params);
let button_id = harness.get_widget(tag).id();

assert_render_snapshot!(harness, "button_no_text");

harness.mouse_click_on(button_id, Some(PointerButton::Primary));
assert_matches!(
    harness.pop_action::<ButtonPress>(),
    Some((ButtonPress { .. }, ..))
);
```

I'm still investigating why this happens, but from the printf-debugging I've done, I suspect it's because of errors in coordinate handling.

xStrom:
The test harness uses `widget.ctx().window_transform()` which transforms from window space to content-box space. Doing border-box size division won't lead to correct results. It's better to take the center of the correct rect.

Fixes #1756

### DIFF
--- a/masonry/src/widgets/button.rs
+++ b/masonry/src/widgets/button.rs
@@ -513,10 +513,6 @@ mod tests {
     // or a with a keyboard shortcut indicator.
 
     #[test]
-    #[should_panic = "is not visible"]
-    // FIXME: Somehow the harness can't find the button under the coordinates it gets
-    // for the button center.
-    // See https://github.com/linebender/xilem/issues/1756
     fn textless_button() {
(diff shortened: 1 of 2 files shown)


RagibHasin (approved):
Looks good. Thanks!

PR #1790, opened 2026-05-04
feat: per-frame tracing buffer that dumps on panic

### LINKED ISSUE #1556: [Feature request] Add per-frame tracing subscriber
Masonry's logging strategy can currently be summarized as "yes".

Every Masonry apps starts by printing this to stdout;

```
---
Writing full logs to /tmp/masonry-<some-timestamp>-dense.log
---
```

Then, in addition to what we send to the console, we send every log line to that file. This is way too much overhead by default, given that almost nobody has reported reading that file.

This logging strategy was initially chosen so that developers would have access to detailed fine-grained log after a crash without restarting the app (especially useful for non-deterministic crashes).

There's another way to achieve this: we should instead create a tracing subscriber that only records traces during `run_rewrite_passes` calls, that dumps in contents to a log file in case of a panic.

This would avoid the syscall overhead of logs in most cases while still providing fine-grained info in case of a crash.

tonybierman:
Replaces the always-on debug-mode file log with a per-frame in-memory buffer that captures tracing events emitted during `run_rewrite_passes` and dumps them to a `$TMPDIR/masonry-*-panic.log` file only when the process panics. Closes #1556.

### DIFF
--- a/masonry_core/src/app/panic_log_buffer.rs
+++ b/masonry_core/src/app/panic_log_buffer.rs
@@ -1,0 +1,293 @@
+// Copyright 2025 the Xilem Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! A tracing buffer that records events during rewrite passes and dumps them
+//! to a file only when the process panics.
+//!
+//! # Usage
+//!
+//! 1. At subscriber setup time, call [`register_panic_hook`] once and register
+//!    a `tracing_subscriber::fmt::layer().with_writer(BufferWriter)`.
+//! 2. In [`RenderRoot::run_rewrite_passes`](crate::app::RenderRoot), call
+//!    [`start_frame_recording()`] and hold the returned
+//!    [`FrameRecordingGuard`] for the duration of the call.
+//! 3. If a panic occurs while a guard is live, the panic hook writes the
+//!    buffer to the log path chosen during step 1.
+//!
+//! Only the *most recent* panic before process exit is preserved on disk:
+//! each call to the panic hook truncates the existing log. This is fine for
+//! the common case (panic crashes the process); callers using
+//! `std::panic::catch_unwind` should be aware that earlier panics' buffers
+//! will be overwritten.
+
+use std::collections::VecDeque;
+use std::fs::File;
+use std::io::{self, BufWriter, Write};
+use std::path::PathBuf;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Mutex, OnceLock, TryLockError};
+
+/// Fast-path flag: `true` while a frame recording is active.
+///
+/// `Relaxed` is sufficient *for correctness of buffer accesses*: the [`Mutex`]
+/// in [`EVENT_BUFFER`] provides the happens-before relationship. Both the
+/// recording flip in [`FrameRecordingGuard::drop`] and the verifying load in
+/// [`BufferWriter::write`] happen *under* the mutex, so once drop returns no
+/// subsequent writer can push into the buffer for the just-ended frame.
+///
+/// The fast-path load in [`BufferWriter::write`] is *not* under the mutex; it
+/// is a best-effort optimization to skip the slow path when no frame is
+/// recording. On weak memory architectures a cross-thread writer may briefly
+/// observe a stale `false` just after [`start_frame_recording`] flips the
+/// flag, dropping events emitted in that window. This is acceptable because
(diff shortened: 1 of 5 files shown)


xStrom:
Hi Tony, thanks for the PR. Would you be so kind to describe a bit of the process you went through when making these changes? I'm just wondering if this is a result of Claude's thinking or if you played any role in it too. Basically, what differentiates it from a simple Claude prompt? Just so I know how to approach reviewing it.

tonybierman:
Stage 1
I point claude to the issue #,  ask it define a statement of work in the form of  a concept statement and check list of tasks.  It analyzes the code base and comes back to me with a series of tasks to meet the work requirements.  I try to tackle small issues with well-defined boundaries.  Particularly if I'm not familiar with the code base yet.  We go through the task list until we have a first blush.  I do a walk through of the code (in vscode).  After that we iterate a process that looks something like this:

Stage 2
- Did we create any magic numbers or assumptions?
-  work order, human review
- Did we create multiple versions of the truth or commit DRY violations?
-  work order, human review
- Are all functions, variables and constants named idiomatically?
- work order, human review
- Are all functions, variables and constants scoped correctly?
-  work order, human review
- Are there any stale, unnecessary, turgid, or inaccurate comments?
-  work order, human review
- Is the uncommitted code well-structured?
-  work order, human review

Stage 3
I ask claude to look for code smells.  we usually do that about three times.  Maybe 4 or 5 if I don't like what I see.  Same process as before - checklist, walk-through, repeat.

Stage 4
- Are the uncommitted code and comments in compliance with the repo's standards and practices?
-  work order, human review
- Does the repo have any guidance on submitting PRs?
-  work order, human review

Stage 5
Submit PR

Background
I'm a retired developer, project manager, and exec.  This is roughly the same process I've used with human developers for about two decades.  My background is mostly in C#.  I've been working with Rust for about a year.

I'd add that I like xilem/masonry project.  I've been reading the docs and playing with it.  I like the philosophy behind it, and I think it has been designed very well.  I wanted to contribute.  I hope the PR meets your standards, which I find to be high.  If you want me to fix something, tell me about.  If you want me to close the PR, no problem. No hard feelings.

My opinions about agent assisted coding.  You have to put bumpers on the agent.  You have to iterate, refactor, walk-through.  If you follow a tight regime, you can produce a pretty good product.  If  you feel otherwise about the result, or the process, I know you're busy, but let me know.

https://www.linkedin.com/in/tonybierman/

masonry_core/src/app/panic_log_buffer.rs
```diff
@@ -0,0 +1,294 @@
...
+
+impl Write for BufferWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // Fast path: avoid touching the mutex when no frame is recording.
+        if !IS_RECORDING.load(Ordering::Relaxed) {
+            return Ok(buf.len());
+        }
+        let Ok(line) = std::str::from_utf8(buf) else {
+            return Ok(buf.len());
+        };
```
PoignardAzur, line 269: Alright, here's a fun one: explain this paragraph to me, without asking Claude first. Please include "This is my own understanding, I didn't ask Claude" in your answer (I'll take your word for it).

tonybierman: give me either the buffer content or the length

tonybierman: I didn't ask Claude

tonybierman: *why to do it this way is to bail out gracefully if i don't have valid UTF-8

PoignardAzur: > give me either the buffer content or the length

I don't think this is an accurate description of the chunk I quoted.

> *why to do it this way is to bail out gracefully if i don't have valid UTF-8

I dispute the idea that it bails gracefully. At the very least, it drops data silently.

tonybierman: I didn't say it bails out gracefully.  I said "to bail out," as in the caller would use it to bail out.   Your objection about silent data loss is still valid, but doesn't accurately contradict what I said.  I can provide an example if it would serve to clarify my meaning.

Both parts of my answer are succinct, but accurate.

The buffer content itself — Ok(line) binds the decoded &str.
The length of the buffer — Ok(buf.len()) returns how many bytes were consumed.

Your point about data loss is probably more substantive, though.  

I'll stop here and give you the last word, if that's what you want.

PR #1794, opened 2026-05-13
Track sizing support in Grid widget and view

RagibHasin:
This completely reworks how the Grid widget lays its children out.

The new implementation closely follows CSS grid specification with one intentional deviation, it does not support minmax track size. It has been omitted from track sizing MVP and because I don't find it very useful.


PoignardAzur:
Thanks for the work! Not sure I'll have time to review it in the coming weeks, but I definitely appreciate changes that get us closer to CSS Grid behavior.

At first glance, my impression is that this PR makes a lot of drive-by changes to the API, some of which I don't agree with. For instance, I don't see a reason to remove `GridExt::grid_pos`, and I think replacing `GridParams::new` with tuples makes the code harder to read.

Anyway, I'll do a more in-depth review soon. This is work we really want to get merged!

RagibHasin:
> I don't see a reason to remove `GridExt::grid_pos`

I removed that and unified its functionality into `GridExt::grid` to be similar to `FlexExt::flex`. Also, `GridParams` implements `From<(u16, u16)>` for position items, making `grid_pos` more verbose.


> I think replacing `GridParams::new` with tuples makes the code harder to read.

I would replace the tuples with `GridParams` builder methods if requested. `GridParams::new()` is now a default constructor as it should be, IMO.

PoignardAzur:
(I'm back from RustWeek, will probably review this tomorrow!)

masonry/examples/grid_masonry.rs
```diff
@@ -16,6 +15,7 @@ use masonry::peniko::Color;
 use masonry::properties::{BorderColor, BorderWidth, Gap};
 use masonry::theme::default_property_set;
 use masonry::widgets::{Button, ButtonPress, Grid, GridParams, Prose, SizedBox, TextArea};
+use masonry::{TextAlign, widgets::GridTrackSize};
```
PoignardAzur: Wrong way to group those imports.

xilem_masonry/src/view/grid.rs
```diff
@@ -302,50 +406,25 @@ pub trait GridExt<State: 'static, Action>: WidgetView<State, Action> {
...
-    /// ), 2, 2)
+    ///     text_button("click me", |_| ()).grid((0, 0, 2, ())),
+    ///     prose("a prose").grid((1, 1)),
+    /// ))
+    /// .columns(repeat_tracks(2, GridTrackSize::FRACTION))
+    /// .rows(repeat_tracks(2, GridTrackSize::FRACTION))
     /// # }
     /// ```
-    fn grid_item(self, params: impl Into<GridParams>) -> GridItem<Self, State, Action>
+    fn grid(self, params: impl Into<GridParams>) -> GridItem<Self, State, Action>
```
PoignardAzur: `.grid(...)` seems like a very unclear method name.

I'd lean towards keeping `grid_pos` and renaming `grid_item` to `grid_params`.

PoignardAzur (changes requested):
Some general thoughts:

- I maintain that the `impl From<...>` with tuples are a bad idea. They're way more error-prone than the builder methods, they're less IDE-friendly (IDEs can show you function argument names, but not for tuples), they're less explicit for someone reading the code, etc. Unless there's a very strong argument to have them (other than "it makes code shorter") I think I'll block the feature until they're removed, sorry.
- I think you should keep the `Grid::with_dimensions` builder, since I expect it to be useful for a lot of examples. You could rename it and/or document that it creates tracks with `GridTrackSize::FRACTION`, or even pass the track size as a parameter.
- Overall, I think it would be helpful to split this PR in two, one that deals with the renamed constructors, then one that deals with the logic changes. I won't block on it though.

Sorry to wait so long just to add more negative feedback. I still need to review the grid code itself, but I do think the ergonomics part is important on its own.

masonry/examples/grid_masonry.rs
RagibHasin: IDE auto-imports...

Would fix it.

RagibHasin:
> * I maintain that the `impl From<...>` with tuples are a bad idea. They're way more error-prone than the builder methods, they're less IDE-friendly (IDEs can show you function argument names, but not for tuples), they're less explicit for someone reading the code, etc. Unless there's a very strong argument to have them (other than "it makes code shorter") I think I'll block the feature until they're removed, sorry.

My rationale for this was similar impls for `FlexParams`. I agree that the lack of parameter name hints would hinder unfamiliar users. Also, the `From` impls feel very much like a DSL.

But I don't like the impending verbosity of the builder patterns, too. 🤷🏽

I think if we revert to `GridExt::grid_pos` with `(x: impl Into<Option<u16>>, ..)` as parameters, we can avoid most of the verbosity. Would the `impl Into<Option<u16>>` tank compile times heavily?

> * I think you should keep the `Grid::with_dimensions` builder, since I expect it to be useful for a lot of examples. You could rename it and/or document that it creates tracks with `GridTrackSize::FRACTION`, or even pass the track size as a parameter.

I have no opinion in this regard, so I would re-establish `Grid::with_dimensions`.

> * Overall, I think it would be helpful to split this PR in two, one that deals with the renamed constructors, then one that deals with the logic changes. I won't block on it though.

That would be painful. I would like to skip splitting it.

PoignardAzur:
> I think if we revert to GridExt::grid_pos with (x: impl Into<Option<u16>>, ..) as parameters, we can avoid most of the verbosity. Would the impl Into<Option<u16>> tank compile times heavily?

I think the compile times would be fine, but also I think that's over-engineering. In most cases I expect users to either specify both coordinates or nothing. `grid_pos(u16, u16)` should be fine.
         
> That would be painful. I would like to skip splitting it.

Fair enough.

masonry/src/widgets/grid.rs
```diff
@@ -27,142 +27,638 @@ use crate::util::debug_panic;
...
+    pub fn with_column(mut self, column: GridTrackSize) -> Self {
+        self.columns.push(column);
+        self
+    }
+
+    /// Builder-style method to define the track sizes (heights) of the grid rows.
+    pub fn with_rows(mut self, rows: impl IntoIterator<Item = GridTrackSize>) -> Self {
+        self.rows = rows.into_iter().collect();
+        self
+    }
```
PoignardAzur: The behavior of `with_rows` is inconsistent with `with_row`: the latter adds a new row, while the former replaces existing rows.

I expect users *will* bump into this and be confused.

masonry/src/widgets/grid.rs
```diff
@@ -27,142 +27,638 @@ use crate::util::debug_panic;
...
+                        consecutive += 1;
+                        if consecutive == child.params.height {
+                            let start_row = row as u16 + 1 - child.params.height;
+                            place_and_advance(&mut occupied, child, col, start_row, &mut cursor);
+                            continue 'child;
+                        }
+                    }
+                    debug_panic!(
+                        "Child {child:?} could not be placed in the grid, no row has enough space",
+                    );
```
PoignardAzur, line 225: I'm not sure how I feel about this debug panic.

It doesn't match web behavior, and while Masonry has always been enthusiastic with panics, it seems a little brutal here.

At the very least, it should be mentioned in the documentation.

masonry/src/widgets/grid.rs
```diff
@@ -27,142 +27,638 @@ use crate::util::debug_panic;
...
+                *cursor_col = (col + child.params.width) as usize % n_col;
+                *cursor_row = row as usize + (*cursor_col == 0) as usize;
+            };
+
+        let mut prefix = vec![0; occupied.len()];
+        'child: for child in &mut self.children {
+            let width = child.params.width as usize;
+            let height = child.params.height as usize;
+
+            match (child.params.col, child.params.row) {
```
PoignardAzur, line 205: I think I get what this match block does, but it's hard to read.

I was gonna say the block should be replaced by a function (or three) returning `Option<(u16, u16)>`, but I'm not actually sure that would be an improvement, since it'd spread the code out.

(Honestly, I'd be willing to accept changes that pessimize the performance if they make this code more readable. For instance, I'm not sure reusing the `prefix` Vec if worth the boilerplate it adds.)

masonry/src/widgets/grid.rs
```diff
@@ -27,142 +27,638 @@ use crate::util::debug_panic;
...
+                            place_and_advance(&mut occupied, child, start_col, row, &mut cursor);
+                            continue 'child;
+                        }
+                    }
+                    debug_panic!(
+                        "Child {child:?} could not be placed in the grid, no column has enough space",
+                    );
+                }
+                (None, None) => {
+                    // build prefix sum
```
PoignardAzur: This match arm seems like it follows a specific algorithm. Is there a wikipedia article or something describing it, that you could point to?

Understanding what prefix sum algorithms do just by looking at their accumulator loop is *really* tough.

masonry/src/widgets/grid.rs
```diff
@@ -309,52 +805,32 @@ impl Widget for Grid {
...
-                    auto_length,
-                    context_size,
-                    axis,
-                    cross_area_length,
-                );
-
-                (child_length.get() + gap_length) / area_cells
-            };
-            cell_length = cell_length.max(desired_cell_length);
+        self.assign_cell_to_auto_placed_children();
```
PoignardAzur, line 868: I notice that this implementation of `measure` will re-place cells on every measure call, even when there's no reason to think the cell placement has changed.

Is this deliberate?

masonry/src/widgets/grid.rs
```diff
@@ -27,142 +27,638 @@ use crate::util::debug_panic;
...
+                        }
+                    }
+
+                    debug_panic!("Child {child:?} could not be placed in the grid");
+                }
+            }
+        }
+    }
+
+    fn resolve_track_lengths(
```
PoignardAzur, line 320: I've tried to read the algorithm in `resolve_track_lengths` several times, and it's pretty rough to decipher as someone not familiar with CSS grid layout.

Can you tell me more about how you wrote it? Is it based on a reference implementation? The CSS spec description?

Either way, comments telling "this part of the function is based on this part of the spec" (like taffy does) would be helpful.

PoignardAzur (commented):
Any changes to layout algorithms are daunting to review, but I tried to be thorough and avoid bikeshedding.

I've tried the examples and they seem to render fine.

Overall I lean towards trusting that your implementation works as intended, but there's still some nits I want to address.

masonry/src/widgets/grid.rs
RagibHasin: Do you propose renaming `with_(row/col)` to `add_(row/col)`/`with_added_(row/col)` or changing the behavior of `with_(rows/cols)` to extend existing rows/columns?

masonry/src/widgets/grid.rs
RagibHasin: Yes, it is from the CSS spec, slightly simplified for the fact that this implementation only supports only one track sizing function for a track using it for both min and max size.

I'd add doc comments to explain the steps as I understood them. The spec is dense. 😬

masonry/src/widgets/grid.rs
RagibHasin: Yes. Because my understanding of the CSS spec is that it requires the cell allotment to be fully resolved before the grid size to be measured.

It makes sense too. If you dont know where a definitely sized item is placed how can you measure its containing row's/column's track length if that was set to `Auto`.

There might be optimization opporunity in omitting the work in `layout` but I didn't see any and wont bother looking for it now.

masonry/src/widgets/grid.rs
RagibHasin: Moving the `prefix` vector into the arm would only save one line of code. For very large grids with hundreds of rows/columns, this would have measurable impact, I think.

masonry/src/widgets/grid.rs
RagibHasin: Matching web behavior here is not feasible here, as web extends the grid with more tracks in that case and I really do not want to do that. IMO that would be a strange behavior.

Mentioning it in the docs should definitely be done.

masonry/src/widgets/grid.rs
RagibHasin: I think Claude suggested it to me. From where it got the algorithm is simplification of CSS grid spec.

I'd describe the alogorithm in English in comment.

masonry/src/widgets/grid.rs
PoignardAzur: The latter.

masonry/src/widgets/grid.rs
PoignardAzur: > I think Claude suggested it to me.

I know that we haven't really advertised our AI policy yet, but *please* mention this ahead of time, *especially* for sensitive code like this.

If there's any Claude-generated code in the PR, please mention it in the PR description.

masonry/src/widgets/grid.rs
PoignardAzur: > Because my understanding of the CSS spec is that it requires the cell allotment to be fully resolved before the grid size to be measured.

Right, but if you measure the grid size twice in the same frame, you probably don't need to re-run cell placement twice, right? Unless cells were added or removed or changed?

masonry/src/widgets/grid.rs
```diff
@@ -27,142 +27,685 @@ use crate::util::debug_panic;
...
+        self.columns = columns.into_iter().collect();
+        self
+    }
+
     /// Builder-style method to add a child widget.
+    ///
+    /// # Panics
+    ///
+    /// If the `params` refer to a row/column that does not exist, `measure`/`layout` pass would
+    /// panic in debug mode and silently ignore this child in release mode.
```
PoignardAzur: I don't think this is enough documentation.

The problem isn't just "will panic if params are out of bounds", if "if the params don't specify a specific row/column, an algorithm will try to find a free space and panic if one can't be found".

PoignardAzur (commented):
As an aside, I'm not quite sure that cell auto-placement is something we want in Masonry? I think it would be find for grid to only accept specific coordinates / ranges and let higher-level frameworks (e.g. Xilem) figure it out.

Either way, it's fine to keep it as-is for now.

masonry/src/widgets/grid.rs
RagibHasin: Will clarify, then.

masonry/src/widgets/grid.rs
RagibHasin: Right, I misunderstood that `measure` and `layout` are decoupled, which they are not.
`assign_cell_to_auto_placed_children` can be removed from `layout`.

If you are asking about caching the result, that can be a little tricky.

RagibHasin: I re-miusunderstood. My original understanding was correct, and I'm adding a caching flag.

PoignardAzur (approved):

RagibHasin:
One CI check seems to be duplicated and stuck.

(Now all checks are pending... There might be some issue with GitHub.)

RagibHasin:
(Or, this might be something in our CI script, although I don't see anything that might cause it.)

PR #1796, opened 2026-05-16
Migrate VirtualScroll onto `understory_virtual_list`

RagibHasin:


masonry/src/widgets/virtual_scroll.rs
```diff
@@ -680,256 +726,113 @@ impl Widget for VirtualScroll {
...
+        }
+
+        let mut visible_indices = self.virtual_list.visible_indices();
+        if let Some(anchor_index) =
+            visible_indices.find(|i| self.virtual_list.is_index_partially_visible(*i))
+        {
+            let last_visible_index = visible_indices
+                .rfind(|i| self.virtual_list.is_index_partially_visible(*i))
+                .unwrap_or(anchor_index);
+            let new_range_in_viewport = anchor_index..last_visible_index;
```
waywardmonkeys: `last_visible_index` is inclusive, but `Range.end` is exclusive so this drops the last visible item and becomes empty for a single visible item, I think.

But I think the real fix is to improve some stuff upstream, so we can come back to this.

masonry/src/widgets/virtual_scroll.rs
```diff
@@ -565,34 +611,42 @@ impl Widget for VirtualScroll {
...
+        // For horizontal layouts, PageDown/ArrowRight scroll forward and PageUp/ArrowLeft scroll back.
+        // In both cases "forward" means increasing scroll offset (towards end of list).
+        // For BottomToTop and RightToLeft, direction.appropriate() negates the delta so that
+        // the arrow key which moves visually "forward" still increases the scroll offset correctly.
+        let delta = match (key, self.direction.axis()) {
+            (NamedKey::PageDown, _) => Some((DELTA_PAGE, true)),
+            (NamedKey::PageUp, _) => Some((-DELTA_PAGE, true)),
+            (NamedKey::ArrowDown, Axis::Vertical) => Some((DELTA_LINE, false)),
+            (NamedKey::ArrowUp, Axis::Vertical) => Some((-DELTA_LINE, false)),
+            (NamedKey::ArrowLeft, Axis::Horizontal) => Some((DELTA_LINE, false)),
```
waywardmonkeys, line 612: This and the right arrow are reversed. Left should be negative, right positive.

masonry/src/widgets/virtual_scroll.rs
```diff
@@ -680,256 +726,113 @@ impl Widget for VirtualScroll {
...
-        let mut first_item: Option<i64> = None;
-        let mut last_item: Option<i64> = None;
+
+        let offset_of_anchor_re_viewport = self.scroll_offset_from_anchor();
+
+        let mut total_children_length = 0.;
+        let mut total_children_count = 0_usize;
 
         // Calculate the sizes of all children
+        self.virtual_list.model_mut().clear();
```
waywardmonkeys, line 719: You don't want to clear this ideally ... we should see if there's a way to retain the measured extents unless there's a deliberate reason for invalidating it all here.

masonry/src/widgets/virtual_scroll.rs
```diff
@@ -680,256 +726,113 @@ impl Widget for VirtualScroll {
...
-            }
-        }
-        if was_dense {
-            // For each time we have the falling edge of becoming not dense, we want to warn.
-            self.warned_not_dense = false;
+        self.virtual_list.clamp_scroll_to_content();
+
+        let active_range =
+            self.virtual_list.visible_strip().start..self.virtual_list.visible_strip().end;
+        if self.active_range != active_range {
```
waywardmonkeys: The ffetch action submission no longer checks `action_handled`. A second layout before the driver/Xilem rebuild handles the first fetch can queue stale actions with the same `old_active`; handling them later can trip `will_handle_action` or apply the wrong diff. Reintroduce coalescing/guarding semantics deliberately.

masonry/Cargo.toml
```diff
@@ -32,6 +32,7 @@ smallvec.workspace = true
 tracing = { workspace = true, features = ["default"] }
 include_doc_path.workspace = true
 resvg.workspace = true
+understory_virtual_list = "0.1.0"
```
waywardmonkeys: I've already released a 0.1.1 ... but I'm going to do some other stuff to improve your code here and release a 0.1.2 in a few hours.

masonry/src/widgets/virtual_scroll.rs
```diff
@@ -967,58 +870,60 @@ impl Widget for VirtualScroll {
...
+                ScrollDirection::BottomToTop => accesskit::Action::ScrollDown,
+                ScrollDirection::LeftToRight => accesskit::Action::ScrollLeft,
+                ScrollDirection::RightToLeft => accesskit::Action::ScrollRight,
+            });
         }
-        let at_end = self.anchor_index + 1 == self.valid_range.end && {
-            let max_scroll = (self.anchor_height - ctx.content_box_size().height / 2.).max(0.0);
-            self.scroll_offset_from_anchor >= max_scroll
-        };
+        let last_visible_index = self.virtual_list.last_visible_index();
```
waywardmonkeys: end-of-list accessibility is broken. `last_visible_index()` can only return up to `len - 1`, but the code compares it to `len`, so `at_end` is effectively never `true` and `ScrollDown`/forward actions remain exposed at the end.

waywardmonkeys (commented):
I'll have more comments later, once I improve some stuff upstream.

masonry/src/widgets/virtual_scroll.rs
RagibHasin: By the declared convention of the previous implementation, we treat half-open (`start..end`) ranges as closed (`start..=end`) range on API boundary.

The rationale has been half-open ranges being easier on eyes or something.

I would gladly change that to proper closed ranges, however.

masonry/src/widgets/virtual_scroll.rs
RagibHasin: It gives the correct behavior for me. However, I sense that we should have conventions for scroll info from input devices being translated to logical scroll direction.

Some desktop environments support flipping the scroll direction at system level, too.

masonry/src/widgets/virtual_scroll.rs
RagibHasin: How would we know if a child has changed its size if we don't measure it?
And if we are going to measure it anyway, why not start fresh?
`Model::clear()` does not deallocate anything, so there should not be any performance penalty.

masonry/src/widgets/virtual_scroll.rs
RagibHasin: Right, I messed this up here.

waywardmonkeys:
I took your comments about the naming into mind and https://github.com/forest-rs/understory/pull/171 ... I'll merge it tomorrow unless you have some comments.

PoignardAzur:
@DJMcNab Do you have time to review this, or would you rather leave it to Bruce and I?

(Honestly, I feel out of my depth reading this code.)

PoignardAzur (approved):
Thanks for the PR!

This code is hard for me to review; I can roughly tell what each individual line does, but I honestly can't tell whether the overall implementation is sound.

In the long term, I'd like VirtualScroll to either be much, *much* better structured and documented, or just split out from the Xilem repo. Honestly, a virtual scrolling widget isn't a load-bearing part of a GUI framework and we only added it in the first place to impress Google executives, and we all know how _that_ went.

In the short term, this seems fine? It seems to have roughly the same logic and to pass the same tests as the original, and I'm fine with adding an understory dependency. I'm not clear on how this is an improvement, but I'm fine with the change either way.

Anyway, LGTM.

RagibHasin:
The improvement is in the result of some behavior at the end of the list, and in performance (hugely).

Old implementation was so slow on debug builds it dropped frames all the time.

The wrong behavior was that it would reach the end (end of the last element) and then jump to the start of the last element, in some cases.

Overall, the core logic is faster and more robust with `understory_virtual_list`.

PR #1798, opened 2026-05-19
MacOS: Present next frame in single CATransaction when resized from bottom left

noelgbl:
Resized and moved conceptually are both requesting a redraw and should participate in the same resize transaction, as bottom-left resize moves the window origin, generating moved events along resized. Resizing the window from the bottom left, therefore, caused the same effect we saw in [#1521](https://github.com/linebender/xilem/issues/1521). (you might as well close this issue now). Adding `WindowEvent::Moved` to be treated as transaction-end.

### DIFF
--- a/masonry_winit/src/event_loop_runner.rs
+++ b/masonry_winit/src/event_loop_runner.rs
@@ -880,7 +880,9 @@ impl MasonryState<'_> {
         // visible jitter.
         #[cfg(target_os = "macos")]
         match event {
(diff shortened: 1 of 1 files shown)


PoignardAzur:
@jaredoconnell Could you test the bottom-left resizing behavior?

jaredoconnell:
> @jaredoconnell Could you test the bottom-left resizing behavior?

I just tested this and I can confirm that it works as intended. It fixes the issue.

PoignardAzur (approved):

PR #1801, opened 2026-05-21
docs: document running Android examples with cargo-apk

### LINKED ISSUE #1780: Document how to run examples on Android
It would be nice to add a full android example, with gradle files and all, so we can just run a command to test.

SAY-5:
The `xilem/examples/android/README.md` only had a TODO placeholder asking
for `cargo apk` instructions. The CI workflow already installs
`cargo-apk` and runs `cargo apk check` on every Android example, but the
README left contributors without a documented path from "I want to run
this on my phone" to a working command.

This fills in the README with the prerequisites (Android SDK / NDK envvars,
an Android rustup target, `cargo-apk`), the build invocation, and the
install/run invocation, using the same package and example selectors that
CI uses (`-p xilem --example <name>`).

Fixes #1780

### DIFF
--- a/xilem/examples/android/README.md
+++ b/xilem/examples/android/README.md
@@ -2,4 +2,39 @@
 
 All the examples in this folder are just glue code importing from the respective examples in the parent folder.
 
-(TODO: Include note about how to use `cargo apk` here.)
+## Running on Android
+
+The examples are built and packaged with [`cargo-apk`](https://crates.io/crates/cargo-apk).
+
+### Prerequisites
+
+- The Android SDK and NDK, and the `ANDROID_HOME` and `ANDROID_NDK_ROOT` environment variables set so that `cargo-apk` can find them. See the [`cargo-apk` README](https://github.com/rust-mobile/cargo-apk) for the full requirements.
+- An Android Rust target installed via `rustup`, for example:
+
+  ```sh
+  rustup target add aarch64-linux-android
+  ```
+
+- `cargo-apk` itself:
+
+  ```sh
+  cargo install cargo-apk
+  ```
+
+### Build and run
+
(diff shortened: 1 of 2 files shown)


PoignardAzur (approved):
I think we might want to include a mention of Android Studio and the whole mess of setting up a connected device/emulator, but this is already a huge improvement.

xilem/examples/android/README.md
```diff
@@ -2,4 +2,39 @@
...
+- An Android Rust target installed via `rustup`, for example:
+
+  ```sh
+  rustup target add aarch64-linux-android
+  ```
+
+- `cargo-apk` itself:
+
+  ```sh
+  cargo install cargo-apk
```
DJMcNab:
```suggestion
  cargo install cargo-apk --locked
```

xilem/examples/android/README.md
```diff
@@ -2,4 +2,39 @@
...
+
+### Build and run
+
+From the repository root, build any of the examples in this folder with `cargo apk build --example <name>`, for example:
+
+```sh
+cargo apk build -p xilem --example calc_android
+```
+
+The supported examples are `calc_android`, `emoji_picker_android`, `http_cats_android`, `mason_android`, `stopwatch_android`, `to_do_mvc_android`, and `variable_clock_android`.
```
DJMcNab, line 32: If we're going to list these all here (not unreasonably), then we need a comment in the Cargo.toml which defines these to also update this when adding one.

DJMcNab (approved):
This seems reasonable to me. The `locked` comment is blocking, but otherwise this can land from my perspective.

I'd maybe consider reordering the section about being `cargo apk build`, and instead move it to be similar to how `check` is mentioned. But that's a personal preference thing, and this way around is still an improvement.

PR #1805, opened 2026-05-25
Fix accessibility region in `TextArea`.

xStrom:
This matches [the implementation in `Label`](https://github.com/linebender/xilem/blob/c24197dfd536bf6e63e234640625c156d1b850b7/masonry/src/widgets/label.rs#L626-L642).


waywardmonkeys (approved):
Feel like this could use some testing, but that probably isn't easy.

xStrom:
Yes, accessibility related coordinates could use more testing in general. There's a lot of #yolo going on.

For this fix here, I saw the discrepancy between `Label` and `TextArea` and by just looking into the methods a bit, it seemed like the `Label` approach is the correct one.

PR #1809, opened 2026-05-25
Simplify hit testing transform.

xStrom:
It's the same result but with shorter code that also makes the intent more obvious.

### DIFF
--- a/masonry_core/src/core/widget.rs
+++ b/masonry_core/src/core/widget.rs
@@ -581,7 +581,7 @@ pub fn find_widget_under_pointer<'c>(
         return None;
     }
 
-    let local_pos = ctx.window_transform().inverse() * pos;
+    let local_pos = ctx.to_local(pos);
 
     if let Some(clip) = ctx.clip_path()
         && !clip.contains(local_pos)


masonry_core/src/core/widget.rs
```diff
@@ -581,7 +581,7 @@ pub fn find_widget_under_pointer<'c>(
         return None;
     }
 
-    let local_pos = ctx.window_transform().inverse() * pos;
+    let local_pos = ctx.to_local(pos);
```
waywardmonkeys, line 584: THis adds handling the border box translation ... so not exactly the same, but more correct, right?

waywardmonkeys (approved):

masonry_core/src/core/widget.rs
xStrom: It's actually 100% the same, the old `ctx.window_transform()` also does the border-box to content-box translation, just like `ctx.to_local()`.

This core method should potentially be rewritten to do all of its checks in border-box space, instead of content-box like right now. However, that's ultimately a minor performance question and not high priority.

waywardmonkeys: > It's actually 100% the same, the old `ctx.window_transform()` also does the border-box to content-box translation, just like `ctx.to_local()`.

Oh. That's confusing.

PR #1810, opened 2026-05-25
Improve `assert_layout_done`.

xStrom:
Previously `assert_layout_done` only verified that layout was available (i.e. not "needed"). This meant that it would succeed with a cached layout even if `run_layout` had not been called.

This PR makes use of the debug vist marker we have to ensure that `assert_layout_done` fails if `run_layout` has not been called on the widget, even if it ends up having a cached layout available anyway.

The key motivation here is that layout caching logic is an internal implementation detail. Widgets should always call `run_layout` before accessing those results. Caching rules may change without announcement, and so depending on specific caching behavior can lead to subtle bugs.

### DIFF
--- a/masonry_core/src/core/contexts.rs
+++ b/masonry_core/src/core/contexts.rs
@@ -773,7 +773,7 @@ impl LayoutCtx<'_> {
     #[track_caller]
     fn assert_layout_done(&self, child: &WidgetPod<impl Widget + ?Sized>, method_name: &str) {
         let child_state = self.get_child_state(child);
-        // request_layout is used as a visit marker, but only when debug_assertions are enabled.
+        // If debug_assertions are enabled, then request_layout is always true until visited.
         let child_not_visited = cfg!(debug_assertions) && child_state.request_layout;
         if child_state.needs_layout() || child_not_visited {
             debug_panic!(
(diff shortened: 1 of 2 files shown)


masonry_core/src/core/contexts.rs
```diff
@@ -772,7 +772,10 @@ impl MeasureCtx<'_> {
 impl LayoutCtx<'_> {
     #[track_caller]
     fn assert_layout_done(&self, child: &WidgetPod<impl Widget + ?Sized>, method_name: &str) {
-        if self.get_child_state(child).needs_layout() {
+        let child_state = self.get_child_state(child);
+        // request_layout is used as a visit marker, but only when debug_assertions are enabled.
```
waywardmonkeys: This comment is slightly confusing because what you mean is that it is additionally used as ... I think.

waywardmonkeys (approved):
Minor nit.

masonry_core/src/core/contexts.rs
xStrom: I reworded it as *If debug_assertions are enabled, then request_layout is always true until visited.* The main point being that it is a reliable vist marker only with debug assertions.

PR #1813, opened 2026-05-26
Avoid origin assumptions in box model API.

xStrom:
The layout accessor methods `content_box_size`, `border_box_size`, and `paint_box_size` are encouraging direct usage without considering the origin of these boxes, which leads to subtle bugs.

This PR removes those context methods, plus `WidgetState::border_box_size`.

The size can still be accessed through e.g. `border_box().size()`, while making correct usage in general more likely. For example `border_box().contains(pos)` instead of `border_box_size().to_rect().contains(pos)`, or `border_box().center()` instead of `border_box_size() / 2.`.

A few widgets got updated with the more robust approach of using e.g. `.center()`, but generally e.g. `content_box_size()` just got replaced with `content_box().size()`. Thus, this PR is focused on changing the API in a way that is more conducive to correct use, but analyzing and fixing all existing widget code is out of scope for this PR.

This PR also adds a few more box model tests, to make sure that these APIs actually work as advertised.

### DIFF
--- a/masonry/src/layers/selector_menu.rs
+++ b/masonry/src/layers/selector_menu.rs
@@ -363,7 +363,7 @@ impl Layer for SelectorMenu {
             PointerEvent::Down(PointerButtonEvent { state, .. }) => {
                 let local_pos = ctx.local_position(state.position);
 
-                !ctx.border_box_size().to_rect().contains(local_pos)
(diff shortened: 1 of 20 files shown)


PoignardAzur (approved):
LGTM.

PR #1814, opened 2026-05-26
Fix `Portal::set_viewport_pos` dirty flag choice.

xStrom:
`Portal::set_viewport_pos` was requesting layout, but that isn't actually needed. Just compose is enough for scroll translation. Other methods of `Portal` already correctly request compose.


PoignardAzur (approved):
I think the reason we did that was to force the child scrollbars to run a layout pass for some reason? Either way, probably not necessary anymore.

xStrom:
I think that line might predate the compose pass even? Not sure, but it definitely predates work in this Linebender repo. Either way, theoretically it should work and I haven't encountered issues yet.

PR #1815, opened 2026-05-26
Restore transform testing and add one new transform test.

xStrom:
When #1322 moved the transform tests to a new location, `mod.rs` was not updated to actually include it in the compilation graph. Since then, the transform test code has started rotting.

This PR here restores these tests, with the required modifications to use the latest APIs.

The blue box size was reduced because the larger size is not critical for the tests, but it did cause the pointer test snapshot to be over 8 KB, which is a soft limit we want to adhere to.

This PR also adds a new test to verify that transforms are respected when calculating the translation between content-box and window space.


RagibHasin:
Should we not have a lint that checks for unused .rs files in /src directory?

xStrom:
It would be nice, but there is no Rust / Clippy lint for that. We would have to build our own script for it probably.

PoignardAzur (approved):
I didn't look too hard at the test code, I assume it's fine if it passes.

LGTM.

PR #1816, opened 2026-05-26
Add pixel snap tests to `Flex`.

xStrom:
`Flex` needs its own separate pixel snap tests, because it inherently does a lot of layout math itself. These tests ensure that the no-gap/no-overflow properties of the Masonry pixel snap strategy are also present in `Flex`.

The creation of these tests was motivated by pixel snapping work where I explored various snap strategies and often ran into subtle scenarios where `Flex` went out of sync in behavior from the general snap strategy.

### DIFF
--- a/masonry/src/widgets/flex.rs
+++ b/masonry/src/widgets/flex.rs
@@ -1252,13 +1252,13 @@ impl Widget for Flex {
 mod tests {
     use super::*;
     use crate::core::WidgetTag;
-    use crate::kurbo::{Cap, Line, Stroke};
+    use crate::kurbo::{Cap, Line, Rect, Stroke};
     use crate::layout::AsUnit;
     use crate::palette;
     use crate::properties::{BorderColor, BorderWidth, Dimensions, Padding};
     use crate::testing::{ModularWidget, TestHarness, assert_debug_panics, assert_render_snapshot};
     use crate::theme::{ACCENT_COLOR, test_property_set};
-    use crate::widgets::Label;
+    use crate::widgets::{Label, SizedBox};
 
     #[test]
     fn test_main_axis_alignment_spacing() {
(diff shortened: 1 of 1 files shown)


masonry/src/widgets/flex.rs
```diff
@@ -1428,6 +1428,85 @@ mod tests {
...
+    fn equal_children_snap_shared_edges() {
+        let child_1 = WidgetTag::unique();
+        let child_2 = WidgetTag::unique();
+        let child_3 = WidgetTag::unique();
+
+        let widget = Flex::row()
+            .cross_axis_alignment(CrossAxisAlignment::Stretch)
+            .with(NewWidget::new(SizedBox::empty()).with_tag(child_1), 1.0)
+            .with(NewWidget::new(SizedBox::empty()).with_tag(child_2), 1.0)
+            .with(NewWidget::new(SizedBox::empty()).with_tag(child_3), 1.0)
```
RagibHasin: Why don't you use `prepare` for children?

xStrom: The other surrounding tests didn't and it's just syntactic sugar. It's fine to use `prepare` too, so I changed it now.

RagibHasin: I was just curious about the convention. Having two different ways of doing something in the same scope feels jarring.

RagibHasin (approved):

masonry/src/widgets/flex.rs
xStrom: I don't think there is a strong convention yet. Masonry code is full of both uses. Yes it does make things confusing.

masonry/src/widgets/flex.rs
```diff
@@ -1428,6 +1428,93 @@ mod tests {
...
+        };
+        let widget = Flex::row()
+            .main_axis_alignment(MainAxisAlignment::SpaceEvenly)
+            .with_fixed(child(child_1))
+            .with_fixed(child(child_2))
+            .with_fixed(child(child_3))
+            .prepare()
+            .with_props(Gap::new(0.px()));
+
+        let harness = TestHarness::create_with_size(test_property_set(), widget, (101, 10));
```
PoignardAzur, line 1483: Quick question: is there a specific reason you didn't reuse the `(100, 10)` value from the previous test? As in, are you testing something different?

xStrom: With `101` there's slightly better coverage, because the middle child will also get fractional coordinates.

PR #1817, opened 2026-05-27
Tweak some docs.

xStrom:
* Improved the documentation around the compose pass so that it more accurately reflects reality.
* Migrated some older box model related docs to the newer more precise terms defined in Masonry concepts.
* Replaced misleading use of the term "property" used to reference state that is not related to Masonry properties.
* Fixed a misleading doc link for IME.


RagibHasin (approved):
Looks alright.

PR #1818, opened 2026-05-28
Expose logic returns the current window content size in logical pixels.

phamsongnewl:


xStrom:
What's the use case for this?

I'm skeptical, because these "logical pixels" aren't the same as the pixels used inside the widget, because it doesn't account for transforms. So it would be an easy footgun.

phamsongnewl:
> What's the use case for this?
> 
> I'm skeptical, because these "logical pixels" aren't the same as the pixels used inside the widget, because it doesn't account for transforms. So it would be an easy footgun.

thanks for you replied, i want to get the window width for updating position my dropdown button correctly

```rust
let pos = self.menu_origin(ctx.window_origin(), ctx.window_size());
fn menu_origin(&self, trigger_origin: Point, viewport_size: Size) -> Point {
		popup_menu_origin(
				...
				viewport_size,
		)
}
pub(crate) fn popup_menu_origin(
    anchor: PopupAnchor,
    menu_size: Option<Size>,
    viewport_size: Size,
) -> Point {
    ...
    let right_limit = viewport_size.width - POPUP_VIEWPORT_MARGIN;
    let bottom_limit = viewport_size.height - POPUP_VIEWPORT_MARGIN;
    ...
    let max_x = (viewport_size.width - POPUP_VIEWPORT_MARGIN - menu_size.width).max(0.0);
    let max_y = (viewport_size.height - POPUP_VIEWPORT_MARGIN - menu_size.height).max(0.0);
    x = x.clamp(0.0, max_x);
    y = y.clamp(0.0, max_y);
    Point::new(x, y)
}
```

phamsongnewl:
Agreed, I removed the `public window_size() API`. For popup placement, I moved this downstream: overlay layers receive the `window-sized` layout context from `LayerStack`, so the menu layer can capture that viewport during measurement and report it back to the trigger. The anchor still comes from `ctx.to_window(...)`, so placement stays in window coordinates without exposing raw logical window size on widget contexts.

PR #1820, opened 2026-05-29
Track Xilem transform changes in Xilem.

xStrom:
Xilem wants to know whether a child view has changed the widget's transform.

Currently this dirty flag is managed by Masonry and exposed via `MutateCtx::transform_has_changed`. However, Masonry itself doesn't care about the flag but does care about a similar flag. This means that Xilem's requirements are creating increased complexity inside Masonry. See also #1811.

This PR here adds the necessary dirty flag tracking to Xilem itself. It's quite straightforward and mimics the existing prop change tracking which has the same root motivation.

This will enable the eventual removal of `MutateCtx::transform_has_changed` from Masonry.


PoignardAzur (approved):
Lovely!

I mean, no, it's very awkward on the Xilem side, but that's a problem for the future. I definitely appreciate that we're getting better separation of concerns between Masonry and Xilem.

Issue #1819, opened 2026-05-28
Runtime theme switching: swap DefaultProperties without a restart


mark-ik:
Hi,

I'm a hobbyist making a music practice app using Xilem, among other things. I have an issue with swapping themes at runtime; the widgets and buttons and whatnot only update at init time, so you have to restart (no bueno!).

<details><summary>screencap of the problem in my app</summary><img width="400" height="300" alt="Image" src="https://github.com/user-attachments/assets/4bb05c13-d6ab-4ff5-b6fb-1fccd617068c" />

(Old screenshot, but pictured: a music practice app's settings page. The buttons remain black from the dark theme when I switch to the light theme, which is not the intended behavior.)</details>

I wanted to ask whether I should submit a pull request with a fix. masonry's `render_root.rs` even carries a // TODO: Find better ways to customize default property set. 

I don't want to step on any toes though, so let me be up front in saying I use AI. It's all tools to me, and I couldn't find any AI policy, but no hard feelings if y'all don't want submissions of this nature.

<details><summary>screencap of the problem fixed in my app</summary><img width="400" height="300" alt="Image" src="https://github.com/user-attachments/assets/913ee3e0-5f0a-4584-908c-9ee776f53da5" />

(Pictured: a music practice app's settings page, with more options reflecting additional development. There are more theme presets and configurable primary/secondary/tertiary/neutral colors, and switching themes changes the buttons to the correct color palette because the fix is in this build)</details>

<details><summary>the fix (two small additions)</summary>

`masonry_core`: RenderRoot gains a runtime setter:
```rust
// Replace the tree-wide default properties at runtime and repaint.
pub fn set_default_properties(&mut self, default_properties: Arc<DefaultProperties>) {
    self.property_arena.default_properties = default_properties;
    self.request_render_all();
}
```

`xilem`: WindowView takes them reactively, applied only when the `Arc` identity changes (so steady-state frames don't re-apply):
```rust
pub fn with_default_properties(mut self, props: Arc<DefaultProperties>) -> Self { ... }
// in rebuild(): if the Arc changed, window.render_root().set_default_properties(props.clone());
```
</details>

This is just swapping at runtime, not the broader set of changes required for per-widget `TextInput::background_color` or more granular configuration. But I've run it for weeks, seems to work fine. Apologies if I've made mistakes in this issue, I never really do this >_<

Relevant issues:

#1765 
#1786

PoignardAzur:
[...]

Issue #1786, opened 2026-05-02
Custom selector-based styling


BigWingBeat:
I'm making an app with Xilem, which includes some custom styling of interactive widgets, but this seems very limited at the moment. Xilem has some built-in styling that uses `PropertyStack`s to change the styling of widgets in response to state changes, but as for custom styling, from what I can tell, you can only apply custom styling *globally*, by setting a custom `DefaultProperties`.

I want to apply custom styling to only certain widgets, and I can see a few possible paths to this:
1. Allow setting custom classes on specific widgets. You could then filter for these classes in the selectors in the global `DefaultProperties`.
2. Allow directly setting `PropertyStack`s on specific widgets.
3. Add special-case methods to the widget types for each individual custom styling use-case.

Another thing to note is that one of the things I want to do is style a container `FlexRow` based on the state of a child `TextInput`. This is a bit different than styling a widget just based on its own state, so I'm not sure how easily the above options could be generalized to this use case.

PoignardAzur:
Right, so, this is honestly a massive failure on my part signpost the intended roadmap.

> Allow directly setting `PropertyStack`s on specific widgets.

This is my intended solution for the Xilem side.

To be honest, I haven't worked on it yet or even laid out a design, so any help there would be appreciated.

My broad plan would be to implement some way to assign a shared `PropertyStack` to a xilem element the way we currently assign properties. The API might look like this:

```rust
trait Style {
   fn with_property_stack(stack: Arc<PropertyStack>) -> Self;

  // Or
  fn with_property_stack(stack: PropertyStackId) -> Self;
}
```

Then you'd want some way to plumb it to Masonry. You'd want to do it in a way that easily lets you declare PropertyStacks as static variables or consts or something (the important part is they don't get recomputed every update) and then refer to them in user code.

> Another thing to note is that one of the things I want to do is style a container FlexRow based on the state of a child TextInput.

Implementing this in Masonry would be hard, probably browser-level hard.

BigWingBeat:
> Implementing this in Masonry would be hard, probably browser-level hard.

That's fair, I should be more specific - the state that I care about is the focus state, which I think is more reasonable than arbitrary state? From what I understand, focus is something that *all* widgets have a notion of, and it's typically already propagated to parent widgets, in one way or another.

PoignardAzur:
For this you might want to look up `has_focus_target`. I don't remember if the dynamic property system covers that one.

l0uisgrange:
[...]

PR #1821, opened 2026-05-29
masonry: allow swapping tree-wide default properties at runtime

### LINKED ISSUE #1765: Support for TextInput background_color and light/dark mode theming. What does the roadmap say?
I'm building a desktop app that needs consistent light-mode styling regardless of system theme. 
Currently text_input provides text_color(), caret_color(), selection_color(), and placeholder_color() - but no background_color().                                                                                            
                                                                                                                                                                     
The Background property exists in masonry's theme system (theme.rs:240), but there's no way to override it per-widget in xilem views.                            
                                                                                                                                                                     
My app needs readable text inputs in both system light and dark modes with a white background. I am targeting the light mode as default,

<img width="234" height="48" alt="Image" src="https://github.com/user-attachments/assets/fad929c7-6b06-4009-ba6c-fba740e92d97" />

And by default I get dark text on dark background. Sure, I found ways for workarounds. Connecting the theme system to the systems color set would be nice.                                                         
                                                                                                                                                                     
## Questions:                                                                                                                                                         
    1. Should I open an issue to add background_color() to TextInput?                                                                                                  
    2. Is there a broader plan for widget-level theme overrides (like CSS's ability to set properties per-element)?                                                    
    3. As a workaround, can I modify the default property set at app initialization? 

As I already said, this is no show stopper. I just would like to know the xilem team's plan so I can adopt the architecture of the application.

Related Issues:
#1342 
#1332 
#1386 


## Example

I made my own textfield wrapper like this:

```rust
// Generic over any State type                                                                                                                                     
pub fn styled_text_input<State, Action, F>(
      value: String,                                                                                                                                                 
      placeholder_text: &str,                                                                                                                                        
      on_change: F,                                                                                                                                                  
 ) -> impl WidgetView<State, Action>                                                                                                                                
 where                                                                                                                                                              
      State: 'static,                                                                                                                                                
      Action: 'static,                                                                                                                                               
      F: Fn(&mut State, String) -> Action + Send + Sync + 'static,                                                                                                   
 {                                                                                                                                                                  
      text_input(value, on_change)                                                                                                                                   
          .placeholder(placeholder_text)                                                                                                                             
          .text_color(Color::BLACK)                                                                                                                                  
          .prop(Background::Color(AlphaColor::new([1.0, 1.0, 1.0, 1.0])))                                                                                            
          .prop(BorderColor { color: BORDER_COLOR })                                                                                                                 
          .prop(CaretColor { color: CARET_COLOR })                                                                                                                   
          .prop(PlaceholderColor::new(PLACEHOLDER_COL))                                                                                                              
 }      
```

mark-ik:
Scope
Touches `masonry_core` (the method), `masonry_testing` (a one-line `TestHarness::set_default_properties` passthrough for the test), and `masonry` (the test).

Background
`DefaultProperties` is fixed at construction, so a host that wants to re-theme a running app (e.g. a light/dark theme toggle re-coloring widgets that rely on default `ContentColor` / `Background`) cannot do so today without restarting.

Fix
This PR adds `RenderRoot::set_default_properties`, which swaps the arena's default properties. It iterates over the tree flagging every widget and invalidating each property cache so the update-props pass re-fires `Widget::property_changed`. Each widget's handler (with the built-in `core_property_changed`) then requests whatever paint or layout that property needs.

A defaults swap does not change a defaulted property's resolved stack index (only its fallback value); the cache invalidation is what triggers the per-entry re-fire of `property_changed`.

Testing
Also adds a `TestHarness::set_default_properties` passthrough and a test (`set_default_properties_refires_property_changed` in masonry/src/tests/update.rs) that (1) seeds a cached `ContentColor` read, (2) swaps defaults, and (3) asserts `property_changed` re-fires for the cached type. The test fails against the fluke fix I made earlier (in issue #1819).

Disclosure: AI-assisted; my responsibility.

Relevant issues:

#1819: closes
#1765: related, leaves `background_color()`, system light/dark detection, and the per-widget override API to that issue.
#1786: related, leaves per-widget-type styling for later.

### DIFF
--- a/masonry/src/tests/update.rs
+++ b/masonry/src/tests/update.rs
@@ -805,3 +805,54 @@ fn status_flag_update_order() {
     assert!(!harness.get_widget(parent1_tag).ctx().is_focus_target());
     assert!(!harness.get_widget(parent1_tag).ctx().has_focus_target());
 }
+
+/// `RenderRoot::set_default_properties` should re-fire `property_changed` on
+/// every cached property across the tree, so a host can swap the default set at
+/// runtime and have widgets react.
+#[test]
+fn set_default_properties_refires_property_changed() {
+    use std::any::TypeId;
+    use std::cell::RefCell;
+    use std::rc::Rc;
+    use std::sync::Arc;
+
+    use crate::peniko::color::AlphaColor;
+    use crate::properties::ContentColor;
+
+    let changed: Rc<RefCell<Vec<TypeId>>> = Rc::new(RefCell::new(Vec::new()));
+
+    let widget = {
+        let changed = changed.clone();
+        ModularWidget::new(())
+            // Read `ContentColor` during paint so it lands in the property cache.
+            .paint_fn(|_, ctx, props, _painter| {
+                let cache = ctx.property_cache();
+                let _ = props.get::<ContentColor>(cache);
+            })
+            .property_change_fn(move |_, _ctx, property_type| {
+                changed.borrow_mut().push(property_type);
+            })
+    };
+
+    // Seed a default `ContentColor` for this widget type and run a frame so the
+    // widget caches it.
+    let mut defaults = test_property_set();
+    defaults.insert::<ModularWidget<()>, ContentColor>(ContentColor::new(AlphaColor::WHITE));
+    let mut harness = TestHarness::create(defaults, NewWidget::new(widget));
+    let _ = harness.redraw();
+    changed.borrow_mut().clear();
+
+    // Swap in a different default and run another frame; the update-props pass
+    // should re-fire `property_changed` for every cached property, including
+    // `ContentColor`.
+    let mut new_defaults = test_property_set();
+    new_defaults.insert::<ModularWidget<()>, ContentColor>(ContentColor::new(AlphaColor::BLACK));
+    harness.set_default_properties(Arc::new(new_defaults));
+    let _ = harness.redraw();
+
+    assert!(
+        changed.borrow().contains(&TypeId::of::<ContentColor>()),
+        "expected property_changed to re-fire for ContentColor after the default swap, got {:?}",
+        changed.borrow(),
+    );
+}
--- a/masonry_core/src/app/render_root.rs
+++ b/masonry_core/src/app/render_root.rs
@@ -413,6 +413,42 @@ impl RenderRoot {
         &mut self.property_arena
     }
 
+    /// Replace the tree-wide default properties at runtime.
+    ///
+    /// The default property set is normally fixed at construction. This lets
+    /// a host swap it mid-session, e.g. a light/dark theme toggle that
+    /// re-colors widgets relying on default `ContentColor` / `Background`.
+    ///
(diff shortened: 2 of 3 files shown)


masonry_core/src/app/render_root.rs
```diff
@@ -413,6 +413,47 @@ impl RenderRoot {
         &mut self.property_arena
     }
 
+    /// Replace the tree-wide default properties at runtime.
+    ///
+    /// The default property set is normally fixed at construction. This lets
+    /// a host swap it mid-session, e.g. a light/dark theme toggle that
+    /// re-colors widgets relying on default `ContentColor` / `Background`.
+    ///
```
mark-ik, line 421: The following comment block can be excised if it's too verbose.

masonry/src/tests/update.rs
```diff
@@ -805,3 +805,56 @@ fn status_flag_update_order() {
     assert!(!harness.get_widget(parent1_tag).ctx().is_focus_target());
     assert!(!harness.get_widget(parent1_tag).ctx().has_focus_target());
 }
+
+// PROPERTIES
```
PoignardAzur: Mh. I just realized we're not using marker comments in this file. Oh well, problem for later.

PoignardAzur:
> The test fails against the fluke fix I made earlier (in issue https://github.com/linebender/xilem/issues/1819).

Nice.

PoignardAzur (approved):

masonry_core/src/app/render_root.rs
PoignardAzur: Yes, I would recommend removing it. Property invalidation is an implementation detail, we shouldn't talk about it in detail in the public API doc.

I'd replace the whole paragraph with something like:

> This invalidates the computed properties of the entire widget tree, and calls [`Widget::property_changed`] for every property previously resolved by each widget.

The latter comment above `fn invalidate_props_all_in()` is enough to explain what happens.

masonry_core/src/app/render_root.rs
```diff
@@ -413,6 +413,47 @@ impl RenderRoot {
...
+            state.property_cache.invalidated = true;
+
+            let id = state.id;
+            recurse_on_children(id, widget, children, |node| {
+                invalidate_props_all_in(node);
+            });
+        }
+        let root_node = self.widget_arena.get_node_mut(self.root_id());
+        invalidate_props_all_in(root_node);
+    }
```
PoignardAzur, line 450: On second thought, I think this should include a call to `run_rewrite_passes()`.

PoignardAzur (changes requested):

mark-ik:
Ok, comment block truncated, and `self.run_rewrite_passes();` is in! Thank you! I'll save closing #1819 for when the next PR (Xilem) merges.

PoignardAzur (approved):
Thanks a lot for the PR!

mark-ik:
Thank you so much for your help! I'll start on the follow up Xilem PR now.

PR #1822, opened 2026-05-31
xilem: WindowView::with_default_properties for runtime theme swap

### LINKED ISSUE #1819: Runtime theme switching: swap DefaultProperties without a restart

mark-ik:
Scope
One file: `xilem/src/window_view.rs`.

Background
[`RenderRoot::set_default_properties`] (added in #1821) lets a host swap the tree-wide default property set at runtime, but it's only reachable from code that holds a `RenderRoot`. Xilem apps go through `WindowView`, so they can't reach the new method at the moment.

Fix
Adds `WindowView::with_default_properties(Arc<DefaultProperties>)`. The new set gets pushed to the render root in `rebuild` only when the `Arc` identity changes (`Arc::ptr_eq`), so steady-state frames don't re-apply. `None` (the default) leaves the startup set untouched.

Testing
No tests in this PR. Xilem doesn't have masonry's `TestHarness`, and the propagation invariant is already covered by the test that landed with #1821. Anecdotally, when applied to my fork in my app, a theme toggle correctly recolors ContentColor-defaulted widgets without restart. I have screenshots or screen recordings available if desired.

Disclosure: AI-assisted; still my responsibility.

Closes #1819.

Relevant issues:
- #1765: related, still leaves `background_color()`, system light/dark detection, and the per-widget override API to that issue.
- #1786: related, leaves per-widget-type styling for later.

### DIFF
--- a/xilem/src/window_view.rs
+++ b/xilem/src/window_view.rs
@@ -1,7 +1,9 @@
 // Copyright 2025 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use masonry::{theme::BACKGROUND_COLOR, util::debug_panic};
+use std::sync::Arc;
+
+use masonry::{core::DefaultProperties, theme::BACKGROUND_COLOR, util::debug_panic};
 use masonry_winit::app::{NewWindow, Window, WindowId};
 
 use crate::core::{MessageCtx, Mut, View, ViewElement, ViewMarker};
@@ -16,6 +18,8 @@ pub struct WindowView<State: 'static> {
     pub(crate) masonry_root: MasonryRoot<State>,
     /// The base color of the window.
     pub(crate) base_color: Option<Color>,
+    /// Tree-wide default properties, applied on `Arc` identity change.
+    pub(crate) default_properties: Option<Arc<DefaultProperties>>,
 }
 
 pub(crate) type WindowViewState = <Box<AnyWidgetView<(), ()>> as View<(), (), ViewCtx>>::ViewState;
@@ -37,6 +41,7 @@ pub fn window<V: WidgetView<State>, State: 'static>(
         options: WindowOptions::new(title),
         masonry_root: MasonryRoot::new(root_view),
         base_color: None,
+        default_properties: None,
     }
 }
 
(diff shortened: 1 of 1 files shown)


xilem/src/window_view.rs
```diff
@@ -110,6 +124,15 @@ impl<State> View<State, (), ViewCtx> for WindowView<State> {
...
         }
 
+        if let Some(props) = &self.default_properties
+            && prev
+                .default_properties
+                .as_ref()
+                .is_none_or(|p| !Arc::ptr_eq(p, props))
+        {
+            window.render_root().set_default_properties(props.clone());
+        }
```
PoignardAzur, line 134: I think this is fine for now, but I'd add a TODO comment to mention the case where default_properties go from `Some` to `None`.

PoignardAzur (approved):
LGTM!

PR #1824, opened 2026-06-12
Make `TextArea` measurement side-effect free

0k-dot-computer:
Hi, thanks for all of the work on `xilem` and the linebender projects in general. I've been using it for some time and have accumulated quite a few patches which I'd like to start contributing upstream if useful to the project.

I originally came up with this one when auditing my own widget set for correct definition of `measure` and ended up removing the hack in `measure` of masonry's `TextArea` along the way (from #1560):
```
// TODO: Remove this hack
  and do efficient side-effect free measurement with no
  alignment
```

`TextArea`'s`measure` worked by laying out `PlainEditor` with the new width, relayout, and then resetting the width, which means selection is re-resolved and text is shaped for a transient state. I demonstrated the issue with a new test that fails on `main`.

My approach for this attempt at a fix is to extract `Label`'s `TextLayoutCache` into a shared module and to re-use it from `TextArea`. The approach requires manually invalidating the cache, but avoiding that seems to require changes to the parley public API, so I was hesitant to go there without some initial feedback.

I directed Claude Code to make the changes and went through a few rounds of feedback and review with it before opening this PR. I also used Claude to help me understand more about how masonry works both in general and in relation to this change.

### DIFF
--- a/masonry/src/widgets/label.rs
+++ b/masonry/src/widgets/label.rs
@@ -6,7 +6,6 @@ use std::mem::Discriminant;
 
 use accesskit::{Node, Role};
 use include_doc_path::include_doc_path;
-use smallvec::SmallVec;
 use tracing::{Span, trace_span};
 
 use crate::core::{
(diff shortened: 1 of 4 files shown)


PoignardAzur:
Thanks for the PR! I haven't been available lately, but I'll take a look soon.

PR #1826, opened 2026-06-19
feat: emit TextAction::Cancelled on Escape key press and expose on_escape callback for TextInput

phamsongnewl:
### Summary
- Add `TextAction::Cancelled` variant to `TextAction` enum, emitted when the Escape key is pressed in a `TextArea`.
- Add `on_escape` callback builder to `TextInput` (xilem), mirroring the existing `on_enter` pattern.
- Handle `TextAction::Cancelled` in `TextInput`'s message dispatch to invoke the registered callback.

### Changes
**`masonry/src/widgets/text_area.rs`**
- New `TextAction::Cancelled` variant with doc comment.
- `Key::Named(NamedKey::Escape)` handling that submits `TextAction::Cancelled` via `ctx.submit_action`.
**`xilem_masonry/src/view/text_input.rs`**
- New `on_escape: Option<Callback<State, Action>>` field.
- `TextInput::on_escape(F)` builder method (matches `on_enter` signature).
- `TextAction::Cancelled` handling in `View::message` — calls the callback or returns `Stale`.

### Motivation
Escape is a universal cancel/dismiss gesture in desktop UIs. `TextArea` previously had no way to surface it to the view layer, forcing users to work around it at the event level.


xilem_masonry/src/view/text_input.rs
```diff
@@ -198,6 +200,15 @@ impl<State: 'static, Action: 'static> TextInput<State, Action> {
         self
     }
 
+    /// Set a callback that will be run when the user presses Escape <kbd>Esc</kbd>.
+    pub fn on_escape<F>(mut self, on_escape: F) -> Self
```
PoignardAzur: You should probably call this `on_cancel` for consistency with the Masonry action.

PoignardAzur (approved):
LGTM.

Thanks for the PR!

xilem_masonry/src/view/text_input.rs
songpham: thanks, I replaced `on_escape`  to `on_cancel`

Issue #1341, opened 2025-08-15
Most text widgets are missing tests


PoignardAzur:
The following tests have either no tests or very few tests:

- Prose
- TextInput
- TextArea
- VariableLabel

We should add tests to some or all of them (keeping in mind that both Prose and TextInput are thin wrappers around TextArea), probably copying from the Label text.

We should especially test various interaction scenarios (pointer events, inputting test, etc).

We should probably design a richer IME emulation in TestHarness as well, to help write some of those tests.

PR #1828, opened 2026-06-20
Add tests for VariableLabel's AnimatedF32 (#1341)

arimu1:
## What

`masonry/src/widgets/variable_label.rs` had no tests (the test module was a `// TODO - Add tests` stub).

`VariableLabel` is a non-editable widget, so the pointer/IME interaction scenarios mentioned in #1341 do not apply to it. Its distinguishing feature is the animated font weight, driven by the `AnimatedF32` helper in the same module — which was entirely untested. These tests cover that state machine:

- a `stable` value does not move and reports completion immediately;
- `move_to` with a zero or negative duration snaps straight to the target;
- `move_to` interpolates linearly over time, in both increasing and decreasing directions;
- `advance` past the target clamps to the target (no overshoot) and stays put afterwards;
- `AnimationStatus::is_completed`.

All deterministic unit tests — no rendering/snapshot dependency.

## Notes

Partially addresses #1341 (the `VariableLabel` item). The editable text widgets (`TextArea`, `TextInput`, `Prose`) and richer IME-emulation tests are left for follow-up.

`cargo test -p masonry`, `cargo fmt --check`, and `cargo clippy -p masonry` all pass locally.

### DIFF
--- a/masonry/src/widgets/variable_label.rs
+++ b/masonry/src/widgets/variable_label.rs
@@ -284,5 +284,88 @@ impl Widget for VariableLabel {
 // --- MARK: TESTS
 #[cfg(test)]
 mod tests {
-    // TODO - Add tests
+    use super::*;
+
+    #[test]
+    fn stable_value_does_not_animate() {
+        let mut value = AnimatedF32::stable(5.0);
+        assert_eq!(value.value(), 5.0);
+        assert_eq!(value.target(), 5.0);
+
+        // Advancing a stable value leaves it unchanged and reports completion.
+        let status = value.advance(100.0);
+        assert!(status.is_completed());
+        assert_eq!(value.value(), 5.0);
+    }
+
+    #[test]
+    fn move_to_zero_millis_jumps_immediately() {
+        let mut value = AnimatedF32::stable(0.0);
+        value.move_to(10.0, 0.0);
+        assert_eq!(value.target(), 10.0);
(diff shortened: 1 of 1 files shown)


DJMcNab:
Have you read https://linebender.org/wiki/llm-policy/? Apologies that this isn't more clearly linked; that's on the to-do list. I believe this pr to be almost entirely generated by an LLM, which is why I bring it up.

As this type isn't public API (I think anyway, although `unreachable_pub/unnameable_types` don't seem to be triggering), I don't think it needs extensive tests. Additionally, most of these tests aren't actually guarding against a realistic scenario, from a quick glance.

I'm glad that you're interested in supporting Xilem's development, but unfortunately this contribution does not meaningfully do so. If you would like to continue helping Xilem's development, I would be glad to help mentor that through review of a PR with more human effort. Please ping me if you do end up making one.

arimu1:
Thanks for the careful review, and apologies for the friction — you're right that this was AI-assisted, and I should have read and disclosed against the LLM policy up front. Your point also stands on the merits: `AnimatedF32` isn't public API and these tests don't guard a realistic scenario, so the added coverage isn't really pulling its weight.

I'm going to close this rather than have you spend more time on it. I appreciate the offer to mentor a more human-driven contribution — I'll take you up on that with something narrower and genuinely useful next time. Thanks again.

PR #1829, opened 2026-06-21
masonry: add hover and press highlight to Button

emnrdl:
Button::pre_paint now overlays a translucent highlight ring when the button is hovered, pressed, or focused, so every button gives pointer feedback regardless of its background and border colors. This mirrors the hover/focus indicator already drawn by Checkbox.


PoignardAzur:
Sorry, I've been busy, hoping to review this soon!

PoignardAzur (changes requested):
I might be missing something, but I think this is the wrong direction. This year we've started to move away from custom status-checking code in widgets, towards using conditional properties instead. This seems like a move in the opposite direction.

PR #1830, opened 2026-06-25
Adds initial Linux-specific window attributes

crutchcorn:
This PR adds the ability to pass a value to Wayland's `app_id` and X11's `WM_CLASS` through a platform-specific `WindowOptions`.

This is needed for my compositor project that uses Wayland's `app_id` to remove server-side chrome information for a shell.

Happy to edit however y'all see fit.

### DIFF
--- a/xilem/src/window_options.rs
+++ b/xilem/src/window_options.rs
@@ -710,7 +710,76 @@ pub use macos::*;
 #[cfg(windows)]
 pub use windows::*;
 
-#[cfg(not(any(target_os = "windows", target_os = "macos")))]
+#[cfg(target_os = "linux")]
+mod linux {
+    use winit::platform::wayland::WindowAttributesExtWayland;
+    use winit::platform::x11::WindowAttributesExtX11;
+    use winit::window::{Window, WindowAttributes};
+
+    #[derive(Debug, Clone, Default)]
+    pub(crate) struct PlatformSpecificInitialWindowAttrs {
+        /// `(general, instance)` identifier applied as the Wayland `app_id` and
+        /// the X11 `WM_CLASS`.
+        pub(crate) name: Option<(String, String)>,
+    }
+
+    #[derive(Debug, Clone, Default)]
+    pub(crate) struct PlatformSpecificReactiveWindowAttrs {}
+
+    impl PlatformSpecificInitialWindowAttrs {
+        pub(crate) fn build(&self, attrs: WindowAttributes) -> WindowAttributes {
+            let Some((general, instance)) = &self.name else {
+                return attrs;
+            };
+            // winit only knows which backend it will use at runtime, so the
+            // same identifier is applied to both: the Wayland `app_id` and the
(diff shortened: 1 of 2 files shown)


xilem/src/window_options.rs
```diff
@@ -710,7 +710,76 @@ pub use macos::*;
...
+        #[inline]
+        fn with_name(mut self, general: impl Into<String>, instance: impl Into<String>) -> Self {
+            self.initial.platform_specific.name = Some((general.into(), instance.into()));
+            self
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+pub use linux::*;
```
PoignardAzur: Small nits, but the other two `pub use platform::*;` should be grouped with this one.

(Really, they should all be grouped at the top or bottom.)

xilem/src/window_options.rs
```diff
@@ -710,7 +710,76 @@ pub use macos::*;
...
+mod linux {
+    use winit::platform::wayland::WindowAttributesExtWayland;
+    use winit::platform::x11::WindowAttributesExtX11;
+    use winit::window::{Window, WindowAttributes};
+
+    #[derive(Debug, Clone, Default)]
+    pub(crate) struct PlatformSpecificInitialWindowAttrs {
+        /// `(general, instance)` identifier applied as the Wayland `app_id` and
+        /// the X11 `WM_CLASS`.
+        pub(crate) name: Option<(String, String)>,
```
PoignardAzur, line 717: Feels like `app_identifier` would be a better name throughout the file? `name` seems a little vague. But I don't know X11/Wayland, maybe it's the expected name.

PoignardAzur (approved):
To be honest, I don't know much about X11/Wayland, but this follows the conventions established by the other platform options, so I'll assume it's correct.

xilem/src/window_options.rs
crutchcorn: `name` follows the convention of winit, so I didn't want to break convention here. Happy to change if you prefer though!

PoignardAzur: It's fine then.

xilem/src/window_options.rs
crutchcorn: Grouped them all at the bottom

PoignardAzur:
Perfect!

ricvelozo:
> This PR adds the ability to pass a value to Wayland's `app_id` and X11's `WM_CLASS` through a platform-specific `WindowOptions`.
> 
> This is needed for my compositor project that uses Wayland's `app_id` to remove server-side chrome information for a shell.
> 
> Happy to edit however y'all see fit.

@crutchcorn Just a question: what if it uses just an app ID, for both Wayland & X11, instead of two Strings? GTK does this.

crutchcorn:
@ricvelozo this is supported. The second argument is optional and noops on Wayland, even. This is how Winit has things structured, so I mirrored it.

PR #1831, opened 2026-06-27
Apply default font to fix number rendering in fontconfig environments

crutchcorn:
In https://github.com/linebender/xilem/pull/1479, labels were set to use `SystemUi`, but checkboxes and text inputs were not updated at that time.

This PR implements this fix in the other two places it was missing, to fix rendering problems with numbers in specific environments:

image

> These numbers are right next to the ascii text. This is what it looks like with this patch applied:
> 
> image

BTW, per:

> Can't reproduce the linked issues

It seems that this issue might be related to Linux-specific `fontconfig` environments, as [I was able to reproduce these issues in a Wayland compositor/shell I'm building using Xilem](https://github.com/crutchcorn/hearthspace/issues/2)

Let me know if there's any additional changes I can apply to help this land.

### DIFF
--- a/xilem_masonry/src/view/checkbox.rs
+++ b/xilem_masonry/src/view/checkbox.rs
@@ -5,7 +5,7 @@ use std::marker::PhantomData;
 
 use masonry::core::{ArcStr, NewWidget};
 use masonry::parley::style::FontWeight;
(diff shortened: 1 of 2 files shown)


PoignardAzur (approved):
Thanks for the PR!

PR #1832, opened 2026-07-15
Fix TextInput placeholder invisible to assistive technologies

Stasshe:
# Fix `TextInput` placeholder invisible to assistive technologies

Fixes #1701

## Problem

`TextInput`'s AccessKit representation exposed the placeholder text twice:

1. As a visual `Label` child node (used only for painting).
2. Via `accesskit::Node::set_placeholder`, called on the *parent* `GenericContainer` node instead of the `TextArea` node it actually describes.

Assistive technologies could not associate the placeholder with its text input, and screen readers announced the redundant label as a separate element.

## Fix

- `TextArea` gained a `placeholder` field and `TextArea::set_placeholder`, so the placeholder is set on the `TextInput`'s inner AccessKit `TextInput`-role node (the `TextArea`), not its container.
- `TextInput::accessibility` no longer calls `node.set_placeholder` on the container node.
- Added `NewWidget::accessibility_hidden(bool)` (backed by a new `accessibility_hidden` flag on `WidgetOptions`/`WidgetState`), which omits a widget and its whole subtree from the AccessKit tree without touching layout, painting, or event handling.
- `TextInput`'s visual placeholder `Label` is now created with `.accessibility_hidden(true)`, so it's no longer published to the accessibility tree at all.
- `run_accessibility_pass` skips building nodes under a hidden subtree, filters hidden children out of `set_children`, drops the root widget's node when hidden, and redirects `focus` away from a hidden widget to the window node.

## Testing

- `masonry::tests::accessibility::accessibility_hidden_omits_subtree` — a hidden widget and its descendants are absent from the tree, and its parent has zero exposed children.
- `masonry::widgets::text_input::tests::placeholder_is_exposed_on_text_area_accessibility_node` — the placeholder is set on the `TextArea` node (not the root `TextInput` node), the visual placeholder `Label` has no accessibility node, and the root exposes exactly one child. Updated after `set_placeholder` is called to confirm live updates propagate.
- `cargo test -p masonry --lib accessibility` passes.

### DIFF
--- a/masonry/src/widgets/text_area.rs
+++ b/masonry/src/widgets/text_area.rs
@@ -9,7 +9,7 @@ use tracing::{Span, trace_span};
 
 use crate::core::keyboard::{Key, KeyState, NamedKey};
 use crate::core::{
-    AccessCtx, AccessEvent, BrushIndex, ChildrenIds, CursorIcon, EventCtx, Ime, LayoutCtx,
+    AccessCtx, AccessEvent, ArcStr, BrushIndex, ChildrenIds, CursorIcon, EventCtx, Ime, LayoutCtx,
     MeasureCtx, PaintCtx, PointerButton, PointerButtonEvent, PointerEvent, PointerUpdate,
     PropertiesMut, PropertiesRef, QueryCtx, RegisterCtx, StyleProperty, TextEvent, Update,
     UpdateCtx, Widget, WidgetId, WidgetMut, render_text, set_accesskit_brush_properties,
@@ -45,9 +45,10 @@ use crate::{TextAlign, theme};
 // TODO: RichTextInput 👀
 // TODO: Support for links - https://github.com/linebender/xilem/issues/360
 pub struct TextArea<const USER_EDITABLE: bool> {
-    // TODO: Placeholder text?
     /// The underlying `PlainEditor`, which provides a high-level interface for us to dispatch into.
     editor: PlainEditor<BrushIndex>,
+    /// Placeholder text exposed to accessibility APIs by an editable text area.
+    placeholder: ArcStr,
     /// The generation of `editor` which we have rendered.
     ///
     /// TODO: Split into rendered and layout generation. This will make the `edited` mechanism in [`on_text_event`](Widget::on_text_event).
@@ -118,6 +119,7 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
         editor.set_text(text);
         Self {
             editor,
+            placeholder: "".into(),
             rendered_generation: Generation::default(),
             word_wrap: true,
             last_max_advance: None,
@@ -369,6 +371,11 @@ impl<const EDITABLE: bool> TextArea<EDITABLE> {
     /// Configures how this text area handles the user pressing Enter <kbd>↵</kbd>.
     pub fn set_insert_newline(this: &mut WidgetMut<'_, Self>, insert_newline: InsertNewline) {
         this.widget.insert_newline = insert_newline;
+        this.ctx.request_accessibility_update();
+    }
+
+    pub(super) fn set_placeholder(this: &mut WidgetMut<'_, Self>, placeholder: impl Into<ArcStr>) {
+        this.widget.placeholder = placeholder.into();
         this.ctx.request_accessibility_update();
     }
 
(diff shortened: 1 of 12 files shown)


PoignardAzur (commented):
Thanks for the PR!

I think this is a valuable improvement, but I'm a little uncomfortable adding yet another status flag to Masonry just for one specific use-case.

I think it would be feasible and simpler to hardcode a "is a placeholder" flag into the Label widget, and then just mark the widget's access node as hidden when that flag is sent.

Thoughts?

Stasshe:
@PoignardAzur 

Thanks for the review! The concern about adding a flag to core (`WidgetOptions`/`WidgetState`) for what looks like a single use-case is fair.

Ideally, `TextArea` itself would render the placeholder the way an HTML `<input placeholder>` does — no separate `Label`, no duplication at all. But `TextArea`/`PlainEditor` doesn't natively support placeholder rendering right now, and fixing that means touching text layout, which is beyond this PR's scope. So this PR works within the current structure (`TextArea` + a visual `Label`) and hides the `Label` from the accessibility tree instead.

Given that, the question is whether the hiding mechanism should be `Label`-specific or general. I looked for the same pattern elsewhere and found two more instances:

- `ProgressBar` (`progress_bar.rs:243-258`): sets `set_numeric_value()` on its own node while also exposing a `Label` child showing the same value — the value gets announced twice.
- `StepInput` (`step_input.rs:1444-1462`): same issue with `set_value()` and its display `Label`.

A `Label`-only flag wouldn't fix either of those. I think `accessibility_hidden` is needed as a general mechanism, similar to how `aria-hidden` is used on the web for visual-only/duplicate content in general.

I'd like to keep this PR scoped to `TextInput`, and follow up separately with native placeholder rendering for `TextArea`, plus fixes for `ProgressBar`/`StepInput` using the same mechanism.

This is just my take, though — if there's context or design intent I'm missing and you'd still prefer the dedicated-flag approach, I'm happy to send that PR instead.

PoignardAzur:
> * `ProgressBar` (`progress_bar.rs:243-258`): sets `set_numeric_value()` on its own node while also exposing a `Label` child showing the same value — the value gets announced twice.
> 
> * `StepInput` (`step_input.rs:1444-1462`): same issue with `set_value()` and its display `Label`.

Fair enough. Although if you add a "hide access node" flag to Label, these widgets could also use it, I think.

In general, I'm just worried about the amount of complexity that Masonry is accruing, especially given that we're in a period where maintainer bandwidth is low. Solutions with a low blast radius are better.

Stasshe:
@PoignardAzur 

That makes sense. I was treating the more general mechanism as the cleaner design, but the concrete reuse cases I found all involve hiding a `Label`. They justify a `Label`-level flag, not a core-wide mechanism for hiding arbitrary widgets and subtrees.

The core-wide version also introduced semantics for hidden descendants, focus, and the root node that aren't needed for the cases we have today. Given the maintenance cost and the current maintainer bandwidth, I agree that this is too large a blast radius.

I've updated the implementation accordingly: the flag now lives on `Label` and marks its AccessKit node as hidden, following the existing approach used by `Image` and `Svg`. The changes to `WidgetOptions`, `WidgetState`, and the accessibility pass have been removed.

Thanks for explaining the broader project context.

masonry/src/widgets/text_area.rs
```diff
@@ -1063,6 +1070,9 @@ impl<const EDITABLE: bool> Widget for TextArea<EDITABLE> {
         if !EDITABLE {
             node.set_read_only();
         }
+        if !self.placeholder.is_empty() {
+            node.set_placeholder(self.placeholder.to_string());
+        }
```
PoignardAzur, line 1075: Is there a reason this part is in TextArea instead of TextInput?

PoignardAzur (approved):
LGTM!

Sorry for the review delay.

masonry/src/widgets/text_area.rs
Stasshe: node.set_placeholder() can only be called inside a widget's own accessibility() method, since the node argument is that widget's own AccessKit node. TextInput has no way to reach into its child TextArea's node directly, so the actual placeholder-setting has to happen inside TextArea's own accessibility().

TextInput just forwards the text through TextArea::set_placeholder (pub(super)). TextArea owns the value and applies it to its node; TextInput is responsible for calling it and keeping it in sync.

Stasshe:
No problem!
I reply the review comment, and plz check it.

masonry/src/widgets/text_area.rs
PoignardAzur: Right, and the TextArea node is the one with the text content. I think it would still be possible to put the placeholder on TextInput, but for now this will do.

Stasshe:
Thanks a lot for your kind review!!!!!

PR #1833, opened 2026-07-29
Support `smol` as async runtime backend

HigherOrderLogic:
As title.

Currently the `tokio` and `smol` backend are mutually exclusive and cannot be turned on at the same time. This can change, however I dont see any value in doing so.


PoignardAzur:
I think we'll explore a different approach, for the reasons outlined in [#xilem > fully feature gating async](https://xi.zulipchat.com/#narrow/channel/354396-xilem/topic/fully.20feature.20gating.20async/with/614370757).

PR #1834, opened 2026-08-03
tree_arena: make the crate no_std + alloc

goyox86:
Hi there XIlem team! 

I've been experimenting with the Xilem architecture for a framework for embedded devices. `tree_arena` works really well in embedded devices too. The only thing is that my crate is `no_std + alloc` and in the `tree_arena` there are few `std` imports, that make impossible for me just to pull it as a dependency. This would be awesome so I don't have to vendor xilem in my repo / CI with this patch.

Please let me know if you want to actually see the embedded GUI library and I'll add some of you to my private repo. I am just no ready to share, it will be of course open source but want to validate a bunch of stuff before I release it.

Thanks for Xilem!

Changes:

- `lib.rs`: add `#![no_std]` and `extern crate alloc;` (the latter placed after the crate-level `//!` doc block so the module-doc position stays valid).
- `tree_arena_safe.rs`: add `use alloc::vec::Vec;` so the `Vec<NodeId>` return types still resolve without the std prelude. Replace `std::hint::black_box(())` in the hidden test helper with `core::hint::black_box(())` — same semantics, available on no_std.
- `tree_arena_unsafe.rs`: swap `use std::cell::UnsafeCell;` → `use core::cell::UnsafeCell;`, add `use alloc::{boxed::Box, vec::Vec};` for the `Box<UnsafeCell<…>>` and `Vec<NodeId>` fields, and the two `std::hint::black_box` → `core::hint::black_box` swaps in the `realloc_inner_storage` test helper.

Disclaimer:

I used some LLM agents to drive this change but is faily simple. I hope that is ok.

### DIFF
--- a/tree_arena/Cargo.toml
+++ b/tree_arena/Cargo.toml
@@ -16,11 +16,14 @@ default-target = "x86_64-unknown-linux-gnu"
 targets = []
 
 [features]
+default = ["std", "safe_tree"]
+# Enable future features which require the standard library.
+# This feature is provided for forwards compatibility only, and current behaviour is the same whether or not it is enabled.
+std = []
 # This crate contains two implementations of a tree for use in masonry, one safe and the other unsafe.
 # The safe tree is known to work, and serves as the baseline implementation and is used by default.
 # The unsafe tree leverages a hashmap as an arena and is designed for higher performance: it leverages unsafe code to achieve this.
 # The unsafe tree is not yet fully tested, and is not used by default.
-default = ["safe_tree"]
 safe_tree = []
 
 [dependencies]
--- a/tree_arena/README.md
+++ b/tree_arena/README.md
@@ -34,6 +34,17 @@ The safe tree is the priority. This means:
 * If both versions are at feature parity, [Masonry][] can switch on the unsafe version for best performance.
 * Otherwise, [Masonry][] uses the safe version.
 
+## Features
+
+The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
+
+- `std` (enabled by default): Enable future features which require the standard library.
+  This feature is provided for forwards compatibility only, and current behaviour is the same whether or not it is enabled.
+- `safe_tree` (enabled by default): Use the safe tree implementation instead of the unsafe one.
+
+This crate is `no_std` compatible; to use it without the standard library, disable default features.
+Note that this also disables `safe_tree`, so re-enable it explicitly if you want the safe implementation.
+
 ## Architecture
 
 ### Safe Tree
--- a/tree_arena/src/lib.rs
+++ b/tree_arena/src/lib.rs
@@ -1,8 +1,6 @@
 // Copyright 2024 the Xilem Authors
 // SPDX-License-Identifier: Apache-2.0
 
-#![no_std]
-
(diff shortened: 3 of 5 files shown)


DJMcNab (approved):
I'd be happy to land this as-is, but for awareness, our standard practise for no_std crates is laid out in something like Linebender Resource Handle - https://github.com/linebender/raw_resource_handle/blob/28076cda8767d09b27298392236d7ac8e8be9984/src/lib.rs#L14-L38.

Essentially, the pattern that we want to enforce is that we have `std` enabled by default. This allows us to have future flexibility about adding a `std` feature flag by default, without making it a silent breaking change in the future.

(Also, our usual practise is to stick crate-level attributes after the doc; but also this crate is missing some of the lint set metadata, so I'm not surprised you didn't know that)

If you're willing to implement those, that would be great. The formatting CI check is mandatory.

goyox86:
> I'd be happy to land this as-is, but for awareness, our standard practise for no_std crates is laid out in something like Linebender Resource Handle - https://github.com/linebender/raw_resource_handle/blob/28076cda8767d09b27298392236d7ac8e8be9984/src/lib.rs#L14-L38.
> 
> Essentially, the pattern that we want to enforce is that we have `std` enabled by default. This allows us to have future flexibility about adding a `std` feature flag by default, without making it a silent breaking change in the future.
> 
> (Also, our usual practise is to stick crate-level attributes after the doc; but also this crate is missing some of the lint set metadata, so I'm not surprised you didn't know that)
> 
> If you're willing to implement those, that would be great. The formatting CI check is mandatory.

Hey @DJMcNab !

I attempted to implement your suggestions in 34e5e2c, plus the formatting fix.

- **`std` enabled by default.** Followed the `raw_resource_handle` pattern: `default = ["std", "safe_tree"]` with an empty `std = []`, and `#[cfg(feature = "std")] extern crate std as _;` so enabling it on a platform without `std` fails loudly rather than silently. Added a `# Features` section to the crate docs covering both flags. `masonry_core` takes `tree_arena` with default features, so it picks `std` up implicitly and nothing changes for it.

- Fixed the `#![no_std]` failure. 

- **Lint set.** Added the `LINEBENDER LINT SET - lib.rs - v3` block, matching the copies in `xilem_core` and `masonry_core`. The `Cargo.toml` half (v8) was already being inherited through `[lints] workspace = true`. I also added `#![cfg_attr(docsrs, feature(doc_cfg))]`, since it sits inside the range you linked and every other crate in the workspace carries it. 

- `README.md` is regenerated with `cargo rdme` (1.4.8, matching CI), since the doc comment changed.


One thing I have not changed: `safe_tree` is also a default feature, so `default-features = false` quietly drops `no_std` users onto the unsafe tree. That predates this PR, so I have only documented the interaction in the new `# Features` section — happy to separate the two if you would rather `std` were the only thing that `default-features = false` turned off.

DJMcNab (approved):
Thanks!

For future reference, I'd advise you to be careful about how you use LLMs to generate PR comments - they will generally be much more useful to a reviewer if they have high information density (concretely, the only part of your comment which was actually informing about something novel was the part about `no-default-features` causing the unsafe tree to be used). It can be helpful to think about where a reviewer needs to put special attention - that is, the parts about which you're uncertain.

PR #1839, opened 2026-08-11
Remove screenshot tolerance from tests

### LINKED ISSUE #1759: Remove screenshot tolerance from tests
With the switch to vello_cpu, tests should be much more deterministic than when they ran on various GPUs (possibly completely deterministic?).

Because of that, the `TestHarness::screenshot_tolerance` parameter might be obsolete. We should try removing it, then running the tests on as many platforms as possible (CI, laptops, macbooks, possibly WebAssembly) and see it it's needed.

Una539:
## Summary

Remove screenshot tolerance from screenshot tests and use exact pixel comparison.

Fixes #1759

## Motivation

After switching to vello_cpu, screenshot tests appear deterministic enough on my machine to remove the tolerance parameter.

I tested locally by setting tolerance to 0. Existing screenshots were regenerated using:

MASONRY_TEST_BLESS=1 cargo test -p masonry

After regeneration, running the test suite twice with zero tolerance passed.

## Changes

- Remove screenshot_tolerance from TestHarness
- Make screenshot comparison exact
- Update screenshot references
- Update documentation


PoignardAzur:
Thanks for the PR!

This looks good at first glance, but I'm leaving for vacation today, so I won't have time to review this. Ping me in a week if I haven't followed up on this.

Una539:
Hi @PoignardAzur , just a quick ping as you suggested since it's been a little over a week. I'm still fairly new to contributing to open-source projects, so please let me know if there's anything else I should do on my side. Thanks!

PoignardAzur (approved):
> I'm still fairly new to contributing to open-source projects

And thus, apprentice, you have learned the forbidden secret of "Open source maintainers have an attention span of two hours and need constant reminders to work on something".

Anyway, everything looks perfect, and I've confirmed with a quick grep that every instance of the word "tolerance" has been removed from the repo, which, uh, shouldn't be taken out of context.

Thanks!

PoignardAzur:
(CI failed, I'm not sure why when reading the logs. I've re-ran the failed jobs, in case it's a spurious failure.)

PR #1841, opened 2026-08-15
fix(masonry/virtual_scroll): Fix unintended regression from #1796

zheylmun:
Laying out stashed children is an error.
Re-add the guard for stashed children in virtual_scroll::layout to prevent panic.

### DIFF
--- a/masonry/src/widgets/virtual_scroll.rs
+++ b/masonry/src/widgets/virtual_scroll.rs
@@ -717,6 +717,21 @@ impl Widget for VirtualScroll {
         // Calculate the sizes of all children
         self.virtual_list.model_mut().clear();
         for (idx, child) in &mut self.items {
+            if !self.active_range.contains(idx) {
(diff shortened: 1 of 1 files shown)


PoignardAzur:
At a glance, I'm a bit confused why stashing the children is necessary at all. "A child scrolled out of view" is explicitly what stashing is *not* for.

(This kind of complication is why I'm considering removing VirtualScroll.)

zheylmun:
Trying to lay out the children that aren't mounted causes a panic at the moment.  I agree this approach a clunky  fix, but it's restoring the behavior that was removed at the beginning of July.  I'm not familiar enough with the guts to really weigh in on the "right" fix. I simply tracked down the regression.  Happy enough to keep this on a fork if you are considering a more permanent solution.

PoignardAzur:
Fair enough. Will review this in a few days.

PR #1842, opened 2026-08-22
feat(masonry): property stack mutations

tonymushah:
Hello Xilem and Masonry Authors,

_A little bit of context_,
I am currently building a GUI framework called [`velona`](https://codeberg.org/tonymushah/velona) built on top of [`leptos/reactive_graph`](https://docs.rs/reactive_graph/), `masonry`, and `winit`. 
And I want it to have some component scope styling system. 
I initially though of using `DefaultProperties` and `reactively` changing it but that didn't sound like practical solution.
Thankfully, `PropertyStack`s exists but it does not have any API for property changes.

And that's what this pull request is trying to solve.

In a nutshell,

This PR add some new methods to the `RenderRoot`:
- ~~`edit_property_stack`: allow the "edit" of a property stack with its given id via the `PropertyStackMut` struct.
This struct will then "track" all "modification" done in the property stack (like adding selectors, inserting new props, etc) and propagates the changes to the widgets that linked to the given property stack.~~

- `replace_property_stack`: replace a property stack with a new one. (See https://github.com/linebender/xilem/pull/1842#issuecomment-5407041102 for more details)

- `remove_property_stack`: remove the property stack in the `PropertyStackArena`, and then propagates the changes to the widgets that is linked to the property stack that has been removed.

- `has_property_stack`: checks if the property stack is in the arena.

- `insert_property_stack`: inserts a property stack in the arena. 

I already made a [PR](https://codeberg.org/tonymushah/velona/pulls/106) to test if this work on my framework repo.

I did like to point out that most of the property update propagation is mostly a copy cat of #1821.

Disclosure: No AI have been used for this PR.

Relevant Issues: 
#1786


masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1017,146 @@ impl RenderRoot {
...
+    /// # Panics
+    ///
+    /// Panics if there is no property stack with the given id in the [`PropertyArena`].
+    pub fn edit_property_stack<E, O>(&mut self, property_stack_id: PropertyStackId, edit_fn: E) -> O
+    where
+        E: FnOnce(&mut PropertyStackMut<'_>) -> O,
+    {
+        let Some(property_stack) = self.property_arena.arena.get_mut(&property_stack_id) else {
+            panic!("property stack not found");
+        };
```
HigherOrderLogic: This should be shorter.

```suggestion
        let property_stack = self.property_arena.arena.get_mut(&property_stack_id).expect("property stack not found");
```

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1017,146 @@ impl RenderRoot {
...
+
+        let selector_changes = props_stack_mut.selector_changes;
+
+        // We don't continue any further
+        // if there is no changes inside done inside the property stack.
+        if selector_changes.is_empty() {
+            return out;
+        }
+
+        let _ = props_stack_mut;
```
HigherOrderLogic: I dont think this line does anything?

masonry_core/src/core/property_stack_mut.rs
```diff
@@ -0,0 +1,414 @@
...
+    /// # Panics
+    ///
+    /// Panics if the index is out of bounds.
+    pub fn edit_property_set<E, O>(&mut self, index: usize, edit_fn: E) -> O
+    where
+        E: FnOnce(&mut SelectorPropertiesMut<'_>) -> O,
+    {
+        let Some((selector, set)) = self.property_stack.stack.get_mut(index) else {
+            panic!("The index is out of bound?");
+        };
```
HigherOrderLogic:
```suggestion
        let (selector, set) = self.property_stack.stack.get_mut(index).expect("The index is out of bound?");
```

masonry_core/src/core/property_stack_mut.rs
```diff
@@ -0,0 +1,414 @@
...
+        let mut set_mut = SelectorPropertiesMut {
+            has_property_changed: false,
+            stack: set,
+        };
+        let res = edit_fn(&mut set_mut);
+        if set_mut.has_property_changed {
+            let selector = selector.clone();
+            let _ = set;
```
HigherOrderLogic: Same as above, this line seems to be doing nothing?

masonry_core/src/core/property_stack_mut.rs
```diff
@@ -0,0 +1,414 @@
...
+        self.property_stack
+            .stack
+            .iter()
+            .enumerate()
+            .filter_map(|(index, (selector_in, _))| {
+                if selector == selector_in {
+                    Some(index)
+                } else {
+                    None
+                }
```
HigherOrderLogic: Nit.

```suggestion
                (selector == selector_in).then_some(index)
```

masonry_core/src/core/selector_properties_mut.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Sets property `P` to given value. Returns the previous value if `P` was already set.
+    pub fn insert<P: Property>(&mut self, value: P) -> Option<P> {
+        let old_value = self.stack.insert(value);
+        self.has_property_changed = true;
+        old_value
+    }
+    /// Removes property `P`. Returns the previous value if `P` was set.
+    pub fn remove<P: Property>(&mut self) -> Option<P> {
+        let old_value = self.stack.remove();
+        self.has_property_changed = true;
```
HigherOrderLogic: This should be a little bit more efficient.

```suggestion
        self.has_property_changed = old_value.is_some();
```

masonry_core/src/app/render_root.rs
tonymushah: Hmm... Yeah. 
Removing the 
```
let _ = props_stack_mut;
```
is fine.
I initially though that if the compiler would get mad if I borrow the `PropertyStack` mutable reference much longer. 😅

masonry_core/src/core/selector_properties_mut.rs
tonymushah: Ah... I get it now.
We only mark the `SelectorPropertiesMut` as changed if the property have been really removed.

masonry_core/src/core/property_stack_mut.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+    /// Remove the latest inserted property set for the given [`Selector`] (aka `pop`).
+    pub fn pop_selector_property_set(&mut self, selector: &Selector) {
+        let maybe_index = self.property_stack.stack.iter().enumerate().rev().find_map(
+            |(index, (selector_in, _))| {
+                if selector_in == selector {
+                    Some(index)
+                } else {
+                    None
+                }
```
HigherOrderLogic:
```suggestion
                (selector_in == selector).then_some(index)
```

masonry_core/src/core/property_stack_mut.rs
```diff
@@ -0,0 +1,410 @@
...
+            .stack
+            .iter()
+            .enumerate()
+            .rev()
+            .find_map(|(index, (selector_in, _))| {
+                if selector_in == selector {
+                    Some(index)
+                } else {
+                    None
+                }
```
HigherOrderLogic:
```suggestion
                (selector_in == selector).then_some(index)
```

PoignardAzur:
Hi, thanks for the PR! Super happy to hear someone is building on top of Masonry.

I'm probably not gonna have time to review this PR in-depth this week, but some feedback after a short skim:

I agree we want the ability to add/remove/edit property stacks, but I don't think finer-grained editing is useful. I think it's fine if the only edit allowed is "you replace an entire property stack with a new stack, and if that re-builds half the widgets in the tree, too bad".

The idea being that changing property stacks should be relatively rare, and styling changes should instead happen by changing a widget's classes and pseudo-classes. (For the same reason you don't reload your CSS stylesheet every time you want to animate something on your page.)

Let me know if I missed something!

tonymushah:
Thank you for the response @PoignardAzur (and sorry for early response, i happened to wake up early).

I understand the CSS analogy here where a masonry property stack represents a CSS stylesheet that should be loaded once.

But, the fine-grain edit is intentional.
It might be not useful for most developer out there 
but this one specifically targets the developers who want to build highly customizable UI from user input.
(ex: user-defined themes where themes can be edited live from the UI).

Yeah, they could change property value from `WidgetMut` directly but what if they want more? like Customizable hover, active, focus? Possible, but more complex to setup.

> I think it's fine if the only edit allowed is "you replace an entire property stack with a new stack, and if that re-builds half the widgets in the tree, too bad"

Yeah, sure but that would be inefficient.

Since we are in CSS comparison here, 
CSS allows fine-grained property changes with [`var`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/var).

Edit:
I initially though of implementing some sort of "CSS variables" equivalent in masonry but that seems to be out of my technical reach.

PoignardAzur:
>  but this one specifically targets the developers who want to build highly customizable UI from user input.
> (ex: user-defined themes where themes can be edited live from the UI).

I'm fine reloading most of the widget tree in that case.

> Yeah, sure but that would be inefficient.

My philosophy for Masonry is "be very willing to do niche things inefficiently rather than add complex code to support them".

Replacing a single font stack could be a done in a single method, no need for the `PropertyStackMut` and `SelectorPropertiesMut` types. True, it's not efficient if you dynamically need to change the stylesheet very often, but that just isn't a very common use-case compared to e.g. switching between light and dark mode, or changing an item on hover. Also, reloading the entire widget tree isn't *that* expensive.

tonymushah:
> Replacing a single font stack could be a done in a single method, no need for the PropertyStackMut and SelectorPropertiesMut types. True, it's not efficient if you dynamically need to change the stylesheet very often, but that just isn't a very common use-case compared to e.g. switching between light and dark mode, or changing an item on hover. Also, reloading the entire widget tree isn't that expensive.

Alright then, I will try replacing the "edit_property_stack" to "replace_property_stack" later and see if that fits on my end. 
(Maybe in a few hours)
I will notify you once I am done.

tonymushah:
@PoignardAzur 
I removed the `edit` with its unnecessary "mutable" struct and replaced it with `replace`.

Everything is also fine on my end. (It is somewhat faster to replace the property stack instead of editing it).

But I had to make some sacrifices though:
- `PropertyStack` now implement `Clone`
- some of the `PropertyStackMut` methods have been moved to `PropertyStack`, but without the tracking system.

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1016,117 @@ impl RenderRoot {
...
+                invalidate_properties_resolution(node, property_stack_id);
+            });
+        }
+        let root_node = self.widget_arena.get_node_mut(self.root_id());
+        invalidate_properties_resolution(root_node, property_stack_id);
+
+        self.run_rewrite_passes();
+    }
+    /// Checks if a property stack with the given id is in the property arena.
+    pub fn has_property_stack(&self, property_stack_id: PropertyStackId) -> bool {
```
PoignardAzur, line 1039: You need to include whitespace between your functions.

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1016,117 @@ impl RenderRoot {
...
+    /// This also invalidate the computed properties of any widget in the entire tree that is linked to that property stack,
+    /// and calls [`Widget::property_changed`] for every property previously
+    /// resolved by each widget.
+    pub fn remove_property_stack(&mut self, property_stack_id: PropertyStackId) {
+        // Don't proceed any further if nothing have been removed
+        if self
+            .property_arena
+            .arena
+            .remove(&property_stack_id)
+            .is_none()
```
PoignardAzur: Small nit, but rustfmt collapses this paragraph to one line if you rename the parameter to `stack_id`, which I think reads better.

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1016,117 @@ impl RenderRoot {
...
+    pub fn has_property_stack(&self, property_stack_id: PropertyStackId) -> bool {
+        self.property_arena.arena.contains_key(&property_stack_id)
+    }
+    /// Remove a property stack with the given id in the property arena.
+    ///
+    /// This also invalidate the computed properties of any widget in the entire tree that is linked to that property stack,
+    /// and calls [`Widget::property_changed`] for every property previously
+    /// resolved by each widget.
+    pub fn remove_property_stack(&mut self, property_stack_id: PropertyStackId) {
+        // Don't proceed any further if nothing have been removed
```
PoignardAzur:
```suggestion
        // Don't proceed any further if nothing has been removed
```

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1016,117 @@ impl RenderRoot {
...
+            .is_none()
+        {
+            return;
+        }
+
+        // Only mark the node that is linked to this property stack for the update-properties pass: `need_update_props`
+        fn invalidate_properties_resolution(
+            node: ArenaMut<'_, WidgetArenaNode>,
+            property_stack_id: PropertyStackId,
+        ) {
```
PoignardAzur, line 1070: This function is almost byte-identical to the other `fn invalidate_properties_resolution` you wrote. You should factor them out and only keep one.

masonry_core/src/core/property_stack.rs
PoignardAzur, line 1: Overall, I think these methods are too specific to your desired use-case.

I think you should add the following general methods:

```rust
fn insert_layer(&mut self, index: usize, selector: Selector, properties: impl Into<PropertySet>);
fn remove_layer(&mut self, index: usize);
fn get_layers(&self) -> &[(Selector, PropertySet)];
```

User code can figure out the rest.

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1016,117 @@ impl RenderRoot {
...
+
+            // We tell the node state there some props update that needs to be done,
+            // since we can't tell in advance if its child has the property stack id.
+            // If not, the property changes will be not visible when we run the rewrite passes.
+            state.needs_update_props = true;
+
+            let is_linked_to_pstack = state
+                .property_stack_id
+                .as_ref()
+                .is_some_and(|id| property_stack_id == *id);
```
PoignardAzur:
```suggestion
            let is_linked_to_pstack = state.property_stack_id == Some(property_stack_id);
```

masonry_core/src/app/render_root.rs
```diff
@@ -1016,6 +1016,117 @@ impl RenderRoot {
...
+
+            let id = state.id;
+            recurse_on_children(id, widget, children, |node| {
+                invalidate_properties_resolution(node, property_stack_id);
+            });
+        }
+
+        let root_node = self.widget_arena.get_node_mut(self.root_id());
+
+        root_node.item.state.needs_update_props = true;
```
PoignardAzur: AFAICT, this line is redundant, since `invalidate_properties_resolution` immediately does the same.

PoignardAzur (commented):
Thanks for the large changes!

I think this will be good to merge after these comments are addressed.

masonry_core/src/app/render_root.rs
tonymushah: Is it fine if I extracted it in a new `RenderRoot` method?

masonry_core/src/app/render_root.rs
tonymushah: Sorry... 
I forgot to remove this one.

masonry_core/src/app/render_root.rs
PoignardAzur: Yes.

masonry_core/src/core/property_stack.rs
tonymushah, line 1: @PoignardAzur 
I feel like I am going to remove: 
- `get_selector_indexes`
- `get_last_selector_property_set_mut`
- `pop_selector_property_set`

Seems too redundant and not so useful since `get_layers`, `insert_layer`, `remove_layer` already exists.

PoignardAzur: I'd also recommend removing `has_selector` and `get_last_selector_index`.

tonymushah: Okay

tonymushah: Also, should we change the `push` to `push_layer`? 
(Mostly to make things clear with the other methods)

PoignardAzur: Fair enough.

tonymushah:
I guess, that is mostly it (right?).

(What a wild ride, well, I guess I should blame myself for being too selfish about this, 
although this is my real first time making a PR onto an open-source project so that should be expected.)

Anyway, `replace` was the "better" solution after all, not as efficient as the fine-grain one, 
but more useful (GUI frameworks would probably find this more useful in features like making property stack hot-reload from files, etc...).

Thanks for the review @PoignardAzur.

PoignardAzur (approved):
LGTM!

PoignardAzur:
> (What a wild ride, well, I guess I should blame myself for being too selfish about this,
although this is my real first time making a PR onto an open-source project so that should be expected.)

I think it went fine. Negotiating features is part of the contribution process.

PR #1843, opened 2026-08-23
Avoid panic when Windows clipboard read fails during paste

lvyaqiao:
## Summary

- Avoid panicking when reading Windows clipboard text fails during Ctrl+V.
- Log the clipboard error and keep the event loop running.
- Preserve normal text paste behavior.

## Reproduction

On Windows, an empty clipboard can make `copypasta 0.10.2` return error 1168.
Before this change, `masonry_winit` called `unwrap()` and the `to_do_list` example exited with code 101.

## Tests

- `cargo fmt --all -- --check`
- `cargo check -p masonry_winit --locked`
- `cargo build -p masonry --example to_do_list --locked`
- `cargo test -p masonry_winit --locked`

Related to #1766.


PoignardAzur (approved):
Thanks!

Issue #1836, opened 2026-08-07
Refactor `masonry_winit`


PoignardAzur:
The code in `masonry_winit` is pretty messy, in a way that I'm pretty sure is 95% accidental complexity.

Some things we could do to make the code more readable:

- Split off MasonryState
- Split off Window
- Remove `'a` lifetime from `MasonryState<'a>` and store static surfaces.

Niedzwiedzw:
hello @PoignardAzur, would it be fine if I gave it a shot? I've done a fair share of winit in the past

PoignardAzur:
Sure! I'm on vacation for a week, so don't expect quick feedback, but I'll happily accept help.

Overall, the coding style I'm trying to converge to is:
- Traits / types first, then impls.
- Small-ish files named after the most important traits/types in them.

PoignardAzur:
@Niedzwiedzw Do you still want to take this on?

HigherOrderLogic:
For the other two tasks, how do you plan on splitting those structs?

PoignardAzur:
Next step would be moving NewWindow, Window and WindowId to a new file.

After that, maybe merge MasonryState and MainState.

PR #1846, opened 2026-09-01
Merge `MasonryState` into `MainState`

HigherOrderLogic:
More progress on #1836.

Would love to get some feedback on the use of the optional `AppDriver`. I cant think of any better way to execute it.


PoignardAzur:
Looks good on first glance.

I'm not super clear what `MainState ::with_app_driver` is supposed to do, will need to look closer.

HigherOrderLogic:
> I'm not super clear what `MainState ::with_app_driver` is supposed to do, will need to look closer.

It's just a private helper to not repeating taking the app driver out of the `Option` wrapper.

masonry_winit/src/event_loop_runner.rs
```diff
@@ -381,15 +362,17 @@ impl Debug for MasonryState {
     }
 }
 
-impl MasonryState {
-    /// Creates the Masonry application's composition root.
+impl MainState {
+    /// Creates the Masonry application's main state.
```
PoignardAzur: I do like the sentence "composition root" better, for what it's worth.

PoignardAzur:
Sorry for the radio silence, I've been sick.

> Would love to get some feedback on the use of the optional AppDriver. I cant think of any better way to execute it.

I see that it's a way to get around the lifetime issues you get from merging the two types. If we have to add this kind of hack, I'd rather we keep the two types separate for now and do a deeper refactor later.

HigherOrderLogic:
I'll close the PR then.

PoignardAzur:
Yeah, sorry if that's unsatisfying.

PR #1847, opened 2026-09-01
xilem: Expose imaging backends as features

Nerixyz:
I found that vello_hybrid and skia both use less cpu and gpu when displaying animated images.

#1714 added the `imaging_vello`, `imaging_vello_hybrid` and `imaging_skia` features to masonry but weren't exposed to xilem/xilem_masonry users. This PR does that.

As with masonry, `imaging_vello` remains the default.


PoignardAzur (approved):
For the record, I really don't like feature flags as backend selectors, I don't like the "one crate with a dozen feature flags to represent all possible combinations" approach, I don't like the feature-flag threading it involves, and I don't like the fact that these flags are how Masonry currently handles rendering backends.

But since we're in for a penny, and I don't have an immediate alternative, we might as well do this.

LGTM.

